PR #9, opened 2019-09-09
Travis fixes

LinuxUser404:


LinuxUser404:
Ok, I fixed the travis build. There are some bonuses:
1) Travis build is simple now.
2) Enabled tests
3) Moved kamu-core into submodules.
4) Adjusted the docs to use the new command to clone the repository.

.gitignore
```diff
@@ -1 +1,4 @@
 target/
+.metals
+project/.bloop/
+.bloop/
```
sergiimk: If these are the folders that your IDE creates I don't think this is a good place for them.

General rule I use for `.gitignore` is that all things specific to developer's environment should not go into the project's `.gitignore` but into your global `.gitignore` instead. Otherwise we will end up with a crazy mix of ignores because everyone uses different tools, and will lose track of where certain things came from.

Example:
```
$ git config -l --global
core.excludesfile=/home/smikhtoniuk/.gitignore
...
```
```
cat /home/smikhtoniuk/.gitignore
.idea/
```

build.sbt
```diff
@@ -43,7 +43,8 @@ lazy val kamuCli = (project in file("."))
       "sqlline" % "sqlline" % "1.8.0",
       // NOTE: Using kamu-specific Hive version with some bugfixes
       // remove `kamu.X` part if you want a simpler build
-      ("org.spark-project.hive" % "hive-jdbc" % "1.2.1.spark2.kamu.1")
+      // ("org.spark-project.hive" % "hive-jdbc" % "1.2.1.spark2.kamu.1")
+      ("org.spark-project.hive" % "hive-jdbc" % "1.2.1.spark2")
```
sergiimk: Is this intentional? Wouldn't want to accidentally make a release with a wrong dependency... We need to research how to specify a version dynamically.

build.sbt
```diff
@@ -1,13 +1,13 @@
 import sbtassembly.AssemblyPlugin.defaultUniversalScript
 
-lazy val kamuCoreManifests = RootProject(file("../kamu-core-manifests"))
+lazy val kamuCoreManifests = RootProject(file("kamu-core-manifests"))
```
sergiimk: I think this needs a bit more reorganization. The structure with submodules I imagined (and seen in few other projects) is:
```
/kamu-cli
    src
    build.sbt
/kamu-core-manifests
    src
    build.sbt
/kamu-core-ingest-polling
    src
    build.sbt
/kamu-core-transform-streaming
    src
    build.sbt
/build.sbt
```
In other words we should put `kamu-cli` source code into subfolder with its own `build.sbt` and then have a `build.sbt` at project root that ties everything together.

docs/developer_guide.md
```diff
@@ -46,12 +46,10 @@ Note: You'll need to have a GPG key configured to sign the artifacts.
...
 
-Clone following projects into the same folder:
-
-* [kamu-data/kamu-core-manifests](https://github.com/kamu-data/kamu-core-manifests)
-* [kamu-data/kamu-core-ingest-polling](https://github.com/kamu-data/kamu-core-ingest-polling)
-* [kamu-data/kamu-core-transform-streaming](https://github.com/kamu-data/kamu-core-transform-streaming)
-* [kamu-data/kamu-cli](https://github.com/kamu-data/kamu-cli)
+Clone the repository with submodules:
+```sh
+git clone --recurse-submodules https://github.com/kamu-data/kamu-cli
```
sergiimk: This is great :+1:

.gitignore
LinuxUser404: Got it, thanks.

build.sbt
LinuxUser404: This would look like multiple packages within a single repository. There are repositories like that:
https://github.com/plredmond/GPipe-GLFW
https://github.com/portnov/libssh2-hs
But they don't use submodules. I would rather put submodules in a separate folder like here:
https://github.com/haskell/haskell-ide-engine (submodules)
https://gitlab.haskell.org/ghc/ghc (libraries)

build.sbt
LinuxUser404: Please clarify. Dynamically as in 'it is not bound to a particular version', or 'there is a flag we pass to the build tool to choose between pre-specified dependencies'?.

LinuxUser404: Either way I don't know how to do that just yet, so I just created the issue:
https://github.com/kamu-data/kamu-cli/issues/10

sergiimk: Let's revert this and use `-Dhive.jdbc.version=upstream` in travis file instead.

build.sbt
sergiimk: I like the idea of `libraries` folder. I was aiming for a standard multi-project layout of java/maven projects, but indeed none of them that I've seen used submodules, so having them alongside the other code might be confusing.

build.sbt
LinuxUser404: Done

build.sbt
LinuxUser404: Done

PR #12, opened 2019-09-11
Added dataset dependency resolution

LinuxUser404:
`kamu pull` is able to resolve dependencies between datasets now.

### DIFF
--- a/src/main/scala/dev/kamu/cli/commands/PullCommand.scala
+++ b/src/main/scala/dev/kamu/cli/commands/PullCommand.scala
@@ -2,6 +2,7 @@ package dev.kamu.cli.commands
 
 import dev.kamu.cli.external.SparkRunner
 import dev.kamu.cli.{MetadataRepository, RepositoryVolumeMap}
+import dev.kamu.cli.utility.DependencyGraph
 import dev.kamu.core.manifests.{Dataset, DatasetID}
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.log4j.LogManager
@@ -14,21 +15,27 @@ class PullCommand(
   repositoryVolumeMap: RepositoryVolumeMap,
   metadataRepository: MetadataRepository,
   sparkRunner: SparkRunner,
-  datasetIDs: Seq[String],
-  all: Boolean
+  ids: Seq[String],
+  all: Boolean,
+  recursive: Boolean
 ) extends Command {
   private val logger = LogManager.getLogger(getClass.getName)
 
   def run(): Unit = {
-    val datasets =
-      if (all)
-        metadataRepository.getAllDatasets()
-      else
-        datasetIDs
-          .map(DatasetID)
-          .map(metadataRepository.getDataset)
+    val datasetIDs = {
+      if (all) metadataRepository.getAllDatasetIDs()
+      else ids.map(DatasetID)
+    }
+    val plan = metadataRepository.getDatasetsInDependencyOrder(
+      datasetIDs,
+      recursive || all // All implies recursive, which is more efficient
+    )
 
-    val numUpdated = datasets
+    logger.debug(s"Pulling datasets in following order:")
+    plan.foreach(d => logger.debug(s"  ${d.id.toString}"))
+
+    // Use Spark to pull each item in the plan
+    val numUpdated = plan
       .map(
         ds =>
           ds.kind match {
--- a/src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
+++ b/src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
@@ -1,0 +1,25 @@
+package dev.kamu.cli.utility
+
+class DependencyGraph[A](dependsOn: A => List[A]) {
+  // TODO: Resolve should report encountered loops instead of not terminating
+  def resolve(goal: List[A]): Seq[A] = {
+    type State = (Seq[A], Set[A])
+    def add(x: A)(s: State): State = (s._1 :+ x, s._2 + x)
+    def isQueued(s: State)(x: A): Boolean = s._2.contains(x)
+    val emptyState = (Seq.empty[A], Set.empty[A])
+
+    @scala.annotation.tailrec
+    def queue(s: State, stack: List[A]): State = {
+      if (stack.nonEmpty) {
+        if (isQueued(s)(stack.head)) queue(s, stack.tail)
+        else {
+          val deps = dependsOn(stack.head).filterNot(isQueued(s))
+          if (deps.isEmpty) queue(add(stack.head)(s), stack.tail)
+          else queue(s, deps ++ stack)
+        }
+      } else s
+    }
+    queue(emptyState, goal)._1
+  }
+
+}
--- a/src/test/scala/dev/kamu/cli/utility/DependencyGraphSpec.scala
+++ b/src/test/scala/dev/kamu/cli/utility/DependencyGraphSpec.scala
@@ -1,0 +1,103 @@
+package dev.kamu.cli.utility
+
+import org.scalatest.{Matchers, FunSuite}
+
+class DependencyGraphSpec extends FunSuite with Matchers {
+  test("Empty graph or isolated vertices") {
+    val deps = List.empty[(Int, Int)]
+    def dependsOn(v: Int) = deps.filter(d => d._1 == v).map(d => d._2)
+    val g = new DependencyGraph(dependsOn)
+    g.resolve(Nil) should be(Nil)
+    g.resolve(List(0)) should be(List(0))
+    g.resolve(List(1)) should be(List(1))
+    g.resolve(List(0, 1)) should be(List(0, 1))
+    g.resolve(List(1, 0)) should be(List(1, 0))
+  }
+
+  test("Single dep") {
+    val deps = List((0, 1))
+    def dependsOn(v: Int) = deps.filter(d => d._1 == v).map(d => d._2)
+    val g = new DependencyGraph(dependsOn)
+    g.resolve(Nil) should be(Nil)
+    g.resolve(List(0)) should be(List(1, 0))
+    g.resolve(List(1)) should be(List(1))
+    g.resolve(List(0, 1)) should be(List(1, 0))
+    g.resolve(List(1, 0)) should be(List(1, 0))
+  }
+
+  test("Single dep reverse") {
+    val deps = List((1, 0))
+    def dependsOn(v: Int) = deps.filter(d => d._1 == v).map(d => d._2)
+    val g = new DependencyGraph(dependsOn)
+    g.resolve(Nil) should be(Nil)
+    g.resolve(List(0)) should be(List(0))
+    g.resolve(List(1)) should be(List(0, 1))
+    g.resolve(List(0, 1)) should be(List(0, 1))
+    g.resolve(List(1, 0)) should be(List(0, 1))
+  }
+
+  test("A and B depends on C") {
+    val deps = List((0, 2), (1, 2))
+    def dependsOn(v: Int) = deps.filter(d => d._1 == v).map(d => d._2)
+    val g = new DependencyGraph(dependsOn)
+    g.resolve(List(0)) should be(List(2, 0))
+    g.resolve(List(1)) should be(List(2, 1))
+    g.resolve(List(2)) should be(List(2))
+    g.resolve(List(0, 1)) should be(List(2, 0, 1))
+    g.resolve(List(1, 0)) should be(List(2, 1, 0))
+    g.resolve(List(0, 1, 2)) should be(List(2, 0, 1))
+    g.resolve(List(1, 0, 2)) should be(List(2, 1, 0))
+    g.resolve(List(0, 2, 1)) should be(List(2, 0, 1))
+    g.resolve(List(1, 2, 0)) should be(List(2, 1, 0))
+    g.resolve(List(2, 0, 1)) should be(List(2, 0, 1))
+    g.resolve(List(2, 1, 0)) should be(List(2, 1, 0))
+  }
+
+  test("A depends on B and C") {
+    val deps = List((0, 1), (0, 2))
+    def dependsOn(v: Int) = deps.filter(d => d._1 == v).map(d => d._2)
+    val g = new DependencyGraph(dependsOn)
+    g.resolve(List(0)) should be(List(1, 2, 0))
+    g.resolve(List(1)) should be(List(1))
+    g.resolve(List(2)) should be(List(2))
+    g.resolve(List(0, 1)) should be(List(1, 2, 0))
+    g.resolve(List(1, 0)) should be(List(1, 2, 0))
+    g.resolve(List(0, 1, 2)) should be(List(1, 2, 0))
+    g.resolve(List(1, 0, 2)) should be(List(1, 2, 0))
+    g.resolve(List(0, 2, 1)) should be(List(1, 2, 0))
+    g.resolve(List(1, 2, 0)) should be(List(1, 2, 0))
+    g.resolve(List(2, 0, 1)) should be(List(2, 1, 0))
+    g.resolve(List(2, 1, 0)) should be(List(2, 1, 0))
+  }
+
+  test("Rhombus") {
+    val deps = List((0, 1), (0, 2), (1, 3), (2, 3))
+    def dependsOn(v: Int) = deps.filter(d => d._1 == v).map(d => d._2)
+    val g = new DependencyGraph(dependsOn)
+    g.resolve(List(0)) should be(List(3, 1, 2, 0))
+    g.resolve(List(1)) should be(List(3, 1))
+    g.resolve(List(2)) should be(List(3, 2))
+    g.resolve(List(3)) should be(List(3))
+    g.resolve(List(0, 1)) should be(List(3, 1, 2, 0))
+    g.resolve(List(1, 0)) should be(List(3, 1, 2, 0))
+    g.resolve(List(2, 0)) should be(List(3, 2, 1, 0))
+    g.resolve(List(0, 1, 2)) should be(List(3, 1, 2, 0))
+    g.resolve(List(0, 2, 3)) should be(List(3, 1, 2, 0))
+    g.resolve(List(0, 3, 1)) should be(List(3, 1, 2, 0))
+    g.resolve(List(0, 1, 3)) should be(List(3, 1, 2, 0))
+    g.resolve(List(3, 2, 0)) should be(List(3, 2, 1, 0))
+    g.resolve(List(2, 3, 0)) should be(List(3, 2, 1, 0))
+    g.resolve(List(1, 2, 0)) should be(List(3, 1, 2, 0))
+    g.resolve(List(2, 0, 1)) should be(List(3, 2, 1, 0))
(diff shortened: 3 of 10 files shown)


src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
-          .map(DatasetID)
-          .map(metadataRepository.getDataset)
-
-    val numUpdated = datasets
+    val s = metadataRepository.getAllDatasetIDs()
+    val datasets: IndexedSeq[DatasetID] = s.toIndexedSeq
+    val datasetIntID = s.zipWithIndex.toMap
+    val idToInt: DatasetID => Int = d => datasetIntID(d)
+    val intToId: Int => DatasetID = d => datasets(d)
+    val idToDataset: DatasetID => Dataset = d => metadataRepository.getDataset(d)
```
sergiimk: Please install Scalafmt plugin for IntelliJ and configure it to format code each time you save a file to make sure our code is consistently formatted.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
-      else
-        datasetIDs
-          .map(DatasetID)
-          .map(metadataRepository.getDataset)
-
-    val numUpdated = datasets
+    val s = metadataRepository.getAllDatasetIDs()
+    val datasets: IndexedSeq[DatasetID] = s.toIndexedSeq
+    val datasetIntID = s.zipWithIndex.toMap
+    val idToInt: DatasetID => Int = d => datasetIntID(d)
```
sergiimk: can be written more concisely as:
`val idToInt = (id: DatasetID) => datasetIntID(d)`
or even
`val idToInt = datasetIntID.apply _`

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
+      if (0 == toResolve.length) {
+        if (!isScheduled(x)) {
+          plan = plan :+ x
+          scheduled.update(x, true)
+        }
+        toSchedule = xs
+      } else toSchedule = toResolve.foldRight(toSchedule)(_ :: _)
+    }
+
+    logger.info(s"This is what I am about to pull in order:")
```
sergiimk: Currently everything that is logged at INFO and above is displayed to the user as output in stderr, so be mindful of this. When user pulls datasets they probably only care that this happened and trust us to figure out correct order of things, so I think this should be logged at DEBUG.

Also I prefer a "Doing BLAH" style of logging (imperative style??) as it's least verbose, e.g. "Pulling datasets in following order: ..."

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
+    var plan: Seq[Int] = Nil
+    def isScheduled(d: Int): Boolean = scheduled(d)
+
+    if (all) toSchedule = datasets.indices.toList.reverse
+    else toSchedule = datasetIDs.map(DatasetID).toList.map(idToInt)
+
+    logger.info(s"Your pull request:")
+    for (d <- toSchedule) logger.info(s"  ${intToId(d).toString}")
+
+    while (Nil != toSchedule) {
```
sergiimk: This is a whole lot of code for one function, so we should modularize it - extract / break it into smaller cohesive functions.

Few things to think about when doing this:
* Maintain a consistent level of granularity in functions (e.g. mixing low-level sorting algorithms with high level logic like spawning docker containers is no bueno)
* Is this class the right place for this logic at all? I'm sure we'll need it in other commands in future

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
+    val datasetIntID = s.zipWithIndex.toMap
+    val idToInt: DatasetID => Int = d => datasetIntID(d)
+    val intToId: Int => DatasetID = d => datasets(d)
+    val idToDataset: DatasetID => Dataset = d => metadataRepository.getDataset(d)
+
+    val scheduled: Array[Boolean] = new Array(datasets.length)
+    for (i <- scheduled.indices) scheduled.update(i, false)
+    var toSchedule: List[Int] = Nil
+    var plan: Seq[Int] = Nil
+    def isScheduled(d: Int): Boolean = scheduled(d)
```
sergiimk: better not to mix different styles of function declaration in one function, pick either `val x = ...` or `def x() = ...`

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
+
+    if (all) toSchedule = datasets.indices.toList.reverse
```
sergiimk: what's the significance of reversing the indices here? `getAllDatasetIDs` returns datasets in arbitrary order (order of files on the file system) so reversing it ... only produces another collection with arbitrary order

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +20,44 @@ class PullCommand(
...
+    else toSchedule = datasetIDs.map(DatasetID).toList.map(idToInt)
+
+    logger.info(s"Your pull request:")
+    for (d <- toSchedule) logger.info(s"  ${intToId(d).toString}")
+
+    while (Nil != toSchedule) {
+      val x: Int = toSchedule.head
+      val xs: List[Int] = toSchedule.tail
+      val toResolve: Seq[Int] = idToDataset(intToId(x))
```
sergiimk: `idToDataset` here is a pretty expensive function (loads and parses yaml) I'm having hard time understanding how many times it's being called here ...

src/test/scala/dev/kamu/cli/utility/DependencyGraphSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+        val g = new DependencyGraph(2, Nil)
+        g.resolve(Nil) should be(Nil)
+    }
+
+    test("A and B depends on C") {
+        val g = new DependencyGraph(3, List((0, 2), (1, 2)))
+        g.resolve(List(0)) should be(List(2, 0))
+        g.resolve(List(1)) should be(List(2, 1))
+        g.resolve(List(2)) should be(List(2))
+        g.resolve(List(0, 1)) should be(List(2, 0, 1))
```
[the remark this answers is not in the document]
LinuxUser404: Got it. I believe that's the reason why one of tests is failing at the moment.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
```diff
@@ -0,0 +1,42 @@
...
+            resolved(x) = true
+          }
+        }
+        else {
+          toSchedule.pushAll(toResolve)
+        }
+      }
+    }
+    plan
+  }
```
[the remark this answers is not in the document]
LinuxUser404: It is simple indeed, but neither efficient nor correct.

This is not a sorting problem, thus sort might produce incorrect result. Lets say `A` depends on`B` and `C` depends on nothing. Imagine the user requested to pull `A,C,B`, none of the neighbors are connected, so the sorting algorithm might come to a conclusion the sequence `A,C,B` is already sorted. It just looks like `A=B=C`. It is not guaranteed to fail, since it depends on the exact implementation of the algorithm, yet it is possible to come up with an example for any predefined efficient sort algorithm. Only inefficient(at least quadratic in time) sorts, like bubble sort, would not suffer from this problem.

Recursive pull with dependency graph resolution is linear in time and space(compared to `n log^2 n` here). Not sure if linear time is possible with non-recursive reordering pull as in recursive one, but `n log n` is definitely possible.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -20,15 +21,38 @@ class PullCommand(
...
-      if (all)
-        metadataRepository.getAllDatasets()
-      else
-        datasetIDs
-          .map(DatasetID)
-          .map(metadataRepository.getDataset)
-
-    val numUpdated = datasets
+    // Helper functions to abstract from Dataset and DatasetID
+    val datasets: IndexedSeq[Dataset] = metadataRepository.getAllDatasets().toIndexedSeq
```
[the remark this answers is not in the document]
LinuxUser404: Ok, I think I know how to that.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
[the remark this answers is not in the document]
sergiimk: Yep, you're right it doesn't handle the disjoint case :(

Now that you simplified your algorithm a bit I can see what it's doing. You're starting at each input node and doing a traversal of its sub-graph while adding nodes you haven't seen before to the plan in depth-first order. This approach makes sense, but I still think we can simplify the implementation.

* We already have the `dependsOn` relationship captured in the `Dataset` itself so transforming the entire graph into a list of `(child, parent)` edges (as you currently do in `PullCommand`) only to later reconstruct it into `child -> list of parents` form seems redundant and this step can be eliminated.

* You are using a mutable stack to implement depth-first traversal. I think it's a lot more natural to implement it with recursion.

* I think you can avoid re-mapping datasetIDs into Ints and using indexes everywhere. The only two inputs this algorithm needs are: a sequence of DatasetIDs to resolve / order, and a `dependsOn` function that given a DatasetID returns list of its dependencies. And for keeping track of resolved nodes you could use a `Set`. You can still make the algorithm work on any graph-like objects if you want using generics. This approach also will lead you to a solution that loads the definitions of the datasets only for those datasets that matter.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -15,20 +16,35 @@ class PullCommand(
...
+          .getDataset(_)
+          .dependsOn
+          .toList
+          .filter(datasetIDs.map(DatasetID).contains)
+      }
+    }
+
+    val plan = {
+      val depGraph = new DependencyGraph[DatasetID](dependsOn)
+      depGraph.resolve(datasetIDs.map(DatasetID).toList)
```
sergiimk: Let's put this extra logic into a method in `MetadataRepository` e.g.:
```
def getDatasetsInDependencyOrder(ids: Seq[DatasetID], recursive: Boolean): List[Dataset]
```

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -15,20 +16,35 @@ class PullCommand(
...
+    }
+
+    logger.debug(s"Pulling datasets in following order:")
+    plan.foreach(d => logger.debug(s"  ${d.toString}"))
+
+    // Use Spark to pull each item in the plan
+    val numUpdated = plan
+      .map(metadataRepository.getDataset)
```
sergiimk: Ugh, kinda sucks that we'll be loading them the second time, perhaps if per my other suggestion we isolate the whole resolve step into a method we could cache the `Dataset`s as `dependsOn` loads the definitions.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
```diff
@@ -0,0 +1,25 @@
...
+  def resolve(goal: List[A]): Seq[A] = {
+    type State = (Seq[A], Set[A]) // (dataset queue, queued datasets)
+    def add(x: A)(s: State): State = (s._1 :+ x, s._2 + x)
+    def isQueued(s: State)(x: A): Boolean = s._2.contains(x)
+    val emptyState = (Seq.empty[A], Set.empty[A])
+
+    @scala.annotation.tailrec
+    def queue(s: State, stack: List[A]): State = {
+      if (stack.nonEmpty) {
+        val deps = dependsOn(stack.head).reverse.filterNot(isQueued(s))
```
sergiimk: what's the point of reversing the `dependsOn` here? all dependencies are equivalent, there's no meaningful order between them...

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
```diff
@@ -0,0 +1,25 @@
...
+        if (deps.isEmpty)
+          queue(add(stack.head)(s), stack.tail.dropWhile(isQueued(s)))
```
sergiimk: this `dropWhile` confused me for a while, wouldn't it be more clear to check `if(!isQueued(stack.head))` as part of the outer `if` block and skip the item?

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
```diff
@@ -0,0 +1,25 @@
...
+        else
+          queue(s, deps.reverse.foldRight(stack)(_ :: _).dropWhile(isQueued(s)))
```
sergiimk: same here, reverse seems not necessary, and foldRight ... can't you just concatenate the lists?

src/test/scala/dev/kamu/cli/utility/DependencyGraphSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+    g.resolve(Nil) should be(Nil)
+    g.resolve(List(0)) should be(List(0))
+    g.resolve(List(1)) should be(List(1))
+    g.resolve(List(0, 1)) should be(List(0, 1))
+    g.resolve(List(1, 0)) should be(List(1, 0))
+  }
+
+  test("Self-dep") {
+    // TODO: test case
+    pending // requires circular dependency resolution support
```
sergiimk: let's not merge with `pending`s - either implement the feature or remove the test

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -15,20 +16,35 @@ class PullCommand(
...
-      if (all)
-        metadataRepository.getAllDatasets()
-      else
-        datasetIDs
-          .map(DatasetID)
-          .map(metadataRepository.getDataset)
-
-    val numUpdated = datasets
+    val dependsOn: DatasetID => List[DatasetID] = {
+      if (all || recursive) {
```
sergiimk: this condition doesn't change, so instead of having it in the function it can be move outside of it

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -15,20 +16,35 @@ class PullCommand(
...
+        metadataRepository.getDataset(_).dependsOn.toList
+      } else {
+        metadataRepository
+          .getDataset(_)
+          .dependsOn
+          .toList
+          .filter(datasetIDs.map(DatasetID).contains)
```
sergiimk: why not map all strings to DatasetID at the beginning of this function instead of doing it in two places

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
```diff
@@ -15,20 +16,35 @@ class PullCommand(
...
-        datasetIDs
-          .map(DatasetID)
-          .map(metadataRepository.getDataset)
-
-    val numUpdated = datasets
+    val dependsOn: DatasetID => List[DatasetID] = {
+      if (all || recursive) {
+        metadataRepository.getDataset(_).dependsOn.toList
+      } else {
+        metadataRepository
```
sergiimk: Hm, I don't think this is a correct implementation of the non-recursive mode. You should add some tests for it.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
LinuxUser404: There is meaningful order for for the graph, so it produces vertices in post-order order, which is easy to test. Yet reserve was indeed unnecessary here since it gets reversed again later.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
LinuxUser404: Yeah, that foldRight is just concatenate. I guess I am just too used to folds.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
LinuxUser404: Added if statement, removed dropWhile.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
```diff
@@ -0,0 +1,25 @@
+package dev.kamu.cli.utility
+
+class DependencyGraph[A](dependsOn: A => List[A]) {
+  // TODO: Object creation should fail if the graph contains loops and report the loop as an error
+
+  def resolve(goal: List[A]): Seq[A] = {
+    type State = (Seq[A], Set[A]) // (dataset queue, queued datasets)
```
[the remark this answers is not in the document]
LinuxUser404: Got it.

src/main/scala/dev/kamu/cli/utility/DependencyGraph.scala
```diff
@@ -0,0 +1,25 @@
+package dev.kamu.cli.utility
+
+class DependencyGraph[A](dependsOn: A => List[A]) {
+  // TODO: Object creation should fail if the graph contains loops and report the loop as an error
```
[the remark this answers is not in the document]
LinuxUser404: Fixed it.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
LinuxUser404: Sure. Done

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
LinuxUser404: Done.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
LinuxUser404: Not all of them are loaded 2 times. Not sure if dependsOn needs cache. If it does I think it should be somewhere in MetadataRepository.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
LinuxUser404: It seems to me as I assign a function to dependsOn depending on the condition as I would assign a function pointer in other languages. I'll have to check that. Moved the condition outside for now.

src/main/scala/dev/kamu/cli/commands/PullCommand.scala
LinuxUser404: Not sure how to test that. An example would be useful.

src/test/scala/dev/kamu/cli/utility/DependencyGraphSpec.scala
LinuxUser404: removed the test

PR #22, opened 2019-09-17
'kamu add' support for remote sources

LinuxUser404:


.travis.yml
```diff
@@ -8,4 +11,7 @@ cache:
     - $HOME/.m2 # Maven repository
 script:
   - sbt -Dhive.jdbc.version=upstream assembly
+  - cd ./testsuit-extras/test-http-server
+  - python3 -m http.server 9000 & # start the server for http related tests
```
sergiimk: As the number of tricky things we need to test grows the approach of setting up the environment *before* running all tests will quickly become unsustainable, so I'm a big fan of tests that fully take care of their own setup.

In your case I think it would be better to:
- use `DatasetFactory` to create mock dataset
- dump it into a temporary yaml file
- spawn a docker container with an http server and mount this file into it (we already have a nice docker process wrapper for this) (you should be able to find a good 3rd party image for this task so we don't have to maintain our own)
- wait for it to start by trying to connect to it (we already have this for Livy startup, so can be generalized and reused)
- execute the rest of the test and clean up

LinuxUser404: Sounds a bit over complicated(third-party image, docker, mounting files to it, not to mention potential security risks from somebody-else's images) compared to single call to python's standard library.

sergiimk: I wouldn't call it a "single call" - this is something that every developer will have to do every time they run tests, and something that will need to be documented. `python3` is also not a portable way to launch python and will not work in my environment for example, so it really brings a lot of unnecessary complexity. I'd rather hide this complexity from people within a test.

You can still use python's http server for this, but starting it via docker from one of the [official slim images](https://hub.docker.com/_/python) is simpler than requiring people to install python in a certain way. Don't think there's any practical security concerns with those images - at work we use them both for local dev and for prod deployments.

So yes, it's more work to do within a test but it's more productive in the long run, it scales well to testing different configurations, and we already have all necessary building blocks to make it simple to implement.

src/main/scala/dev/kamu/cli/MetadataRepository.scala
```diff
@@ -62,13 +63,36 @@ class MetadataRepository(
...
       inputStream.close()
     }
   }
 
+  def loadDatasetFromURI(uriStr: String): Dataset = {
+    val uri = new URI(uriStr)
+    uri.getScheme match {
+      case "https" => loadDatasetFromURL(uri.toURL)
+      case "http"  => loadDatasetFromURL(uri.toURL)
+      case "ftp"   => loadDatasetFromURL(uri.toURL)
```
sergiimk: Reading `Source.fromURL` docs I doubt FTP source will work. In ingest-polling app I'm using a 3rd party FTP client, but here if it's not supported I wouldn't bother and just returned an error.

src/main/scala/dev/kamu/cli/commands/AddCommand.scala
```diff
@@ -5,20 +5,20 @@ import dev.kamu.cli.{
...
 }
-import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.hadoop.fs.FileSystem
 import org.apache.log4j.LogManager
 
 class AddCommand(
   fileSystem: FileSystem,
   metadataRepository: MetadataRepository,
-  manifests: Seq[Path]
+  manifests: Seq[String]
```
sergiimk: should we use the `URI` type here, that's what I'm using throughout the code to represent some resource (local or on the web)

src/main/scala/dev/kamu/cli/external/DockerProcessBuilder.scala
```diff
@@ -32,7 +32,7 @@ class DockerProcessBuilder(
     new DockerProcess(
       id,
       dockerClient,
-      runArgs.containerName.get,
+      runArgs.containerName.get, // TODO: containerName is optional... None.get exception possible
```
sergiimk: Yep, I have a ticket for this

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +58,56 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
       actual3 shouldEqual Seq(dsC.id, dsB.id, dsA.id)
     }
   }
+
+  test(raw"'kamu add' from HTTP") {
+    withEmptyRepo { kamu =>
+      val testDir = getRandomDir()
+      val serverPort = 8080
+      val hostPort = 9000
```
sergiimk: just fyi, to exclude the possibility of port collision you could expose port without assigning it a number and then use `DockerProcess.getHostPort` method to see which port on the host it was mapped to.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +58,56 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+      val testServerName = s"kamu_test_http_server"
+      val testHttpServerArgs = DockerRunArgs(
+        image = "python:3.7-slim",
+        args = List("python3", "-m", "http.server", s"${serverPort}"),
+        exposePortMap = Map(hostPort -> serverPort),
+        volumeMap = Map(testDir -> new Path("/mnt/kamu")),
+        containerName = Some(testServerName),
+        detached = true
+      )
+      val testHttpServer = new DockerClient(FileSystem.get(new Configuration()))
```
sergiimk: `KamuTestBase` should already have `fileSystem` member, so you don't need to create it anew

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +58,56 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+
+      val testHttpServerProc = new DockerProcessBuilder(
+        "http",
+        testHttpServer,
+        testHttpServerArgs
+      ).run()
+      val expected = DatasetFactory.newRootDataset()
+      kamu.metadataRepository.saveDataset(expected, path)
+      testHttpServerProc.join()
+      Thread.sleep(1000) // wait for the port to become ready, 'waitForHostPort' did work :-(
```
sergiimk: Hm, I'd like to investigate this a bit further. I'm generally very much against sleeping in tests - you always either sleep to long and make tests unnecessary long, or too little and then your tests break randomly.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +58,56 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+      val actual =
+        kamu.metadataRepository.loadDatasetFromURI(
+          s"http://localhost:${hostPort}/mnt/kamu/test-dataset.yaml"
+        )
+      testHttpServer.stop(testServerName, time = 1)
```
sergiimk: might need try/finally to ensure we stop the container even in case of exceptions

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +58,56 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
       actual3 shouldEqual Seq(dsC.id, dsB.id, dsA.id)
     }
   }
+
+  test(raw"'kamu add' from HTTP") {
+    withEmptyRepo { kamu =>
+      val testDir = getRandomDir()
```
sergiimk: `withEmptyRepo` already creates a temporary directory which will be deleted after the tests (`kamu.config.repositoryRoot`), so you can any files in it and not bother deleting them. Currently if your test fails your directory will remain in `/tmp`.

src/main/scala/dev/kamu/cli/external/DockerClient.scala
```diff
@@ -107,12 +110,22 @@ class DockerClient(fileSystem: FileSystem) {
...
+      .exitValue()
+  }
+
   def pull(image: String): Unit = {
     prepare(Seq("docker", "pull", image)).!
   }
 
   def inspectHostPort(container: String, port: Int): Option[Int] = {
-    val format = "--format={{ (index (index .NetworkSettings.Ports \"" + port + "/tcp\") 0).HostPort }}"
+    val format = "--format=\'{{ (index (index .NetworkSettings.Ports \"" + port + "/tcp\") 0).HostPort }}\'"
```
sergiimk: This is the change that causes tests to fail.

Quotes are necessary to run such command in the shell because shell has to split the input string into process arguments, but here the list of arguments is passed directly into the system call and extra quotes **become** part of an argument.

If your goal was to make this log a command that can be easily copy-pasted into a shell then argument quoting/escaping should be only added for the log output.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
sergiimk: that problem that I pointed out might actually be the reason why `waitForHostPort` didn't work for you

LinuxUser404: `waitForHostPort` would wait till it becomes available, but seem to rise the exception regardless...

src/main/scala/dev/kamu/cli/commands/AddCommand.scala
```diff
@@ -3,22 +3,32 @@ package dev.kamu.cli.commands
...
-      metadataRepository.loadDatasetFromFile(manifestPath)
+    val sources = manifests.map(manifestStr => {
+      logger.debug(s"Loading dataset from: $manifestStr")
+      try {
+        val manifestURI = new URI(manifestStr);
+        if (null == manifestURI.getScheme)
+          metadataRepository.loadDatasetFromFile(new Path(manifestStr))
+        else metadataRepository.loadDatasetFromURI(manifestURI)
+      } catch {
+        case _: URISyntaxException =>
```
sergiimk: I think this complexity can be entirely avoided if we assume that upper level logic should do the necessary escaping/encoding of file paths that contain non-URL-friendly characters, e.g. if there is a space it will be converted to `%20`. Then this class can then accept a `Seq[URI]` as a parameter and avoid this exception handling.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +61,63 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
       actual3 shouldEqual Seq(dsC.id, dsB.id, dsA.id)
     }
   }
+
+  test(raw"'kamu add' from HTTP") {
+    withEmptyRepo { kamu =>
+      val expected = DatasetFactory.newRootDataset()
+
+      // create a temporary directory with the dataset to host
+      val serverDir = getRandomDir()
```
sergiimk: this directory will not be cleaned up if test fails, you could simply create files in the directory already set up by `withEmptyRepo` and let this method clean them up for you instead

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +61,63 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+      )
+      val testHttpServer = new DockerClient(fileSystem)
+      val testHttpServerProc = new DockerProcessBuilder(
+        "http",
+        testHttpServer,
+        testHttpServerArgs
+      ).run()
+      testHttpServerProc.join()
+      val hostPort =
+        testHttpServerProc.waitForHostPort(serverPort, Duration("500 ms"))
```
sergiimk: no need for string parsing here, see [this](https://www.scala-lang.org/api/2.9.3/scala/concurrent/duration/Duration.html):
```
import scala.concurrent.duration._

val duration = 500 millis
```

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +61,63 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+
+      // pull the dataset from the server
+      val actual =
+        kamu.metadataRepository.loadDatasetFromURI(
+          new URI(s"http://localhost:${hostPort}/test-dataset.yaml")
+        )
+
+      // stop the server
+      testHttpServer.kill(testServerName)
```
sergiimk: probably should be in a `finally` block not to "leak" the container

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +61,63 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+      fileSystem.delete(serverDir, true)
+
+      actual shouldEqual expected
+    }
+  }
+
+  test(raw"'kamu add' from file") {
+    withEmptyRepo { kamu =>
+      val expected = DatasetFactory.newRootDataset()
+      val testDir = getRandomDir()
```
sergiimk: same here about the temp dir

src/main/scala/dev/kamu/cli/commands/AddCommand.scala
LinuxUser404: I am not sure how and where to add that upper level logic, since the argument parsing is done by another library, which seem not to handle URI. This is what I get if I remove that logic:
```
kamu add ~/.local/share/wineprefixes/steam/dosdevices/c\:/Program\ Files\ \(x86\)/ca.vancouver.data.local-area-boundaries.yaml 
Exception in thread "main" java.net.URISyntaxException: Illegal character in path at index 64: /home/nick/.local/share/wineprefixes/steam/dosdevices/c:/Program Files (x86)/ca.vancouver.data.local-area-boundaries.yaml
```
If I keep it it works just as expected:
```
kamu add ~/.local/share/wineprefixes/steam/dosdevices/c\:/Program\ Files\ \(x86\)/ca.vancouver.data.local-area-boundaries.yaml 
[INFO] Added 1 datasets
```

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
LinuxUser404: got it

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
LinuxUser404: Not sure how to get that directory path. `withEmptyRepo` does not change working directory path, so a relative path points outside of the temporary dir.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
LinuxUser404: got it

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
LinuxUser404: Oh,I see the comment above, which means `KamuTestAdapter.config` should be public. Got it.

src/main/scala/dev/kamu/cli/commands/AddCommand.scala
LinuxUser404: Looks like the previous implementation also fails with the above example. There is a bug in either the library or scala or both.
```
kamu add ~/.local/share/wineprefixes/steam/dosdevices/c\:/Program\ Files\ \(x86\)/ca.vancouver.data.local-area-boundaries.yaml 
[ERROR] Excess arguments provided: '/home/nick/.local/share/wineprefixes/steam/dosdevices/c:/Program Files (x86)/ca.vancouver.data.local-area-boundaries.yaml'
```

LinuxUser404: Oh, I've been missing the converters, which were producing the issue. Should be fixed in the next commit.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +61,72 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
   }
+
+  test(raw"'kamu add' from HTTP URI") {
+    withEmptyRepo { kamu =>
+      val expected = DatasetFactory.newRootDataset()
+      // create a temporary directory with the dataset to host
+      val serverDir =
+        new Path(
+          kamu.config.repositoryRoot,
+          // raw"test dir with: Colon and Space" // bug in scala?
```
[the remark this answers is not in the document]
LinuxUser404: Well, I'd like to uncomment it, but it leads to a bug I don't know how to fix.

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
```diff
@@ -53,4 +61,72 @@ class MetadataRepositorySpec extends FunSuite with Matchers with KamuTestBase {
...
+          raw"server"
```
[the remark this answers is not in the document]
LinuxUser404: Yeah, you're right. Seeing `raw` allows me to think about the string as just a sequence of characters(I guess I like it a bit too much :-) ). Yet this is not the default in scala, hence the keyword. I'll keep an eye on that and try to use the keywords only when needed.

src/main/scala/dev/kamu/cli/commands/AddCommand.scala
```diff
@@ -3,41 +3,47 @@ package dev.kamu.cli.commands
...
+              logger.warn(e.getMessage + " - skipping")
+              false
+            case e: MissingReferenceException =>
+              logger.warn(e.getMessage + " - skipping")
+              false
+          }
+        })
+        .count(added => added)
+      logger.info(s"Added $numAdded datasets")
+    } catch {
```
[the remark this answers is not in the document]
LinuxUser404: got it

src/test/scala/dev/kamu/cli/MetadataRepositorySpec.scala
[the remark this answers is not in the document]
LinuxUser404: Found a workaround.

PR #54, opened 2021-07-26
Fix sql init file lifetime

sergiimk:
FYI @coldeasy, hopefully this fixes your issue.


coldeasy:
That fixed it, thanks @sergiimk

PR #79, opened 2022-07-14
Improve paginator performance on large block counts

sergiimk:
This PR adds two workarounds for poor pagination performance on datasets with high block counts.

1) Individual writes into `minus::Pager` seem to have high overhead, so we format the entire block using a temporary buffer and only write to `Pager` once per block.

This already improves performance significanlty (e.g. `kamu log` on a 1000-block dataset takes about 3 seconds to render).

2) On top of that to limit performance degradation we introduce `--limit` parameter to the `log` command that defaults to 500 blocks.

Note:
* `kamu log x > file` is blazing fast even with 1K blocks, so its all overhead is in the `minus` pager library
* `minus` library has support for "dynamic paging" where a separate thread can push more data progressively but it doesn't work as I expected - it doesn't backpressure the producing thread and seems to make the problem only worse

@zaychenko-sergei please review

### DIFF
--- a/kamu-cli/src/commands/log_command.rs
+++ b/kamu-cli/src/commands/log_command.rs
@@ -15,7 +15,7 @@ use opendatafabric::*;
 
 use chrono::prelude::*;
 use console::style;
-use futures::TryStreamExt;
+use futures::{StreamExt, TryStreamExt};
 use opendatafabric::serde::yaml::YamlMetadataBlockSerializer;
 use opendatafabric::serde::MetadataBlockSerializer;
 use opendatafabric::MetadataBlock;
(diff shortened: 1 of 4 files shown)


kamu-cli/src/commands/log_command.rs
```diff
@@ -132,21 +135,34 @@ trait MetadataRenderer {
...
         &'a self,
         output: &mut impl Write,
-        mut blocks: DynMetadataStream<'b>,
+        blocks: DynMetadataStream<'b>,
     ) -> Result<(), CLIError> {
+        let mut blocks = blocks.take(self.limit);
+
+        // TODO: We buffer output per block as writing directly to minus::Pager seems to
+        // have a lot of overhead. In future we should improve its async paging support.
+        let mut buf = Vec::new();
```
sergiimk, line 158: I probably could wrap the `output` with some sort of fixed-size buffered writer, but I didn't bother, thinking we'll revisit this anyway and implement proper async pagination.

PR #82, opened 2022-07-21
Added ODF ReadStep extension for Parquet format

zaychenko-sergei:


opendatafabric/src/serde/flatbuffers/proxies_generated.rs
```diff
@@ -9,8 +9,8 @@
 
 // automatically generated by the FlatBuffers compiler, do not modify
 
-use std::cmp::Ordering;
-use std::mem;
+use core::cmp::Ordering;
```
sergiimk, line 12: Can you check which version of Flatbuffers compiler you are using? I we use different ones we'll be producing too much diffs.

sergiimk (commented):
Looks as expected so far, but we'll need to bump the Spark engine image version once we release it.

opendatafabric/src/serde/flatbuffers/proxies_generated.rs
sergiimk: Mine is:
```bash
$ flatc --version
flatc version 2.0.0

$ yay -Q | grep flat
flatbuffers-git 2358-1
```
I moved to latest git version after encountering some issues in Rust codegen, but can switch back to a latest stable if it is indeed stable :)

zaychenko-sergei: I'm using the latest version available on today, 2.0.6

kamu-core/tests/tests/engine/test_ingest_engine.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright Kamu Data, Inc. and contributors. All rights reserved.
+// Copyright Kamu Data, Inc. &&and contributors. All rights reserved.
```
sergiimk, line 1: accidental?

sergiimk (approved):
One small thing, otherwise LGTM!

PR #94, opened 2022-09-05
Display progress on push/pull

sergiimk:
Ticket: https://www.notion.so/Display-sync-progress-8ee519cf84b1453e842be047080b5c96


kamu-core/src/domain/repos/metadata_chain_comparator.rs
```diff
@@ -23,12 +24,35 @@ impl MetadataChainComparator {
...
         rhs_chain: &dyn MetadataChain,
         rhs_head: Option<&Multihash>,
+        listener: &dyn CompareChainsListener,
     ) -> Result<CompareChainsResult, CompareChainsError> {
         // When source and destination point to the same block, chains are equal, no further scanning required
         if Some(&lhs_head) == rhs_head.as_ref() {
             return Ok(CompareChainsResult::Equal);
         }
 
+        let lhs_chain = MetadataChainWithStats::new(
```
sergiimk, line 34: Not too happy with this... I suspect it's not the last time we need some introspection into what's happening in the metadata chain, so we might support some listener interface directly in `MetadataChain` trait in the future.

opendatafabric/src/lib.rs
```diff
@@ -8,6 +8,8 @@
 // by the Apache License, Version 2.0.
 
 #![feature(backtrace)]
+#![feature(provide_any)]
+#![feature(error_generic_member_access)]
```
sergiimk, line 12: This is needed by `thiserror` crate to work on nightly rustc - they are starting some big refactoring of the `Error` trait.

sergiimk:
Will merge this as I think it will significantly improve the UX of the demo environment when our datasets grow larger over time, but will happily follow up on any comments / suggestions.

PR #116, opened 2023-01-17
Implementing Smart Transfer Protocol

zaychenko-sergei:
- Support Simple Transfer Protocol operations via api-server


kamu-cli/src/explore/api_server.rs
```diff
@@ -81,3 +139,157 @@ async fn graphql_playground() -> impl axum::response::IntoResponse {
         async_graphql::http::GraphQLPlaygroundConfig::new("/graphql"),
     ))
 }
+
+async fn dataset_refs_handler(
```
sergiimk, line 143: What do you think about moving all this out of `kamu-cli` module into `kamu-adapter-http`? In the spirit of hexagonal architecture we will then have a self-contained module implementing simple & smart HTTP-based ODF protocols which can be included here into overall routing table alongside the `kamu-adapter-gql`.

kamu-core/src/domain/repos/dataset.rs
```diff
@@ -26,6 +27,9 @@ pub trait Dataset: Send + Sync {
     fn as_checkpoint_repo(&self) -> &dyn ObjectRepository;
     fn as_cache_repo(&self) -> &dyn NamedObjectRepository;
 
+    fn supports_smart_protocol(&self) -> bool;
```
sergiimk, line 30: I was surprised to see this method here - protocols are an "adapter" concern, so I think we should avoid them leaking into the core domain model.

I was thinking that `http(s)://` will always mean simple transfer and `odf+http(s)://` will indicate smart transfer - so based on the URL sheme we pick the desired protocol and there will be no need to ask dataset what protocol it supports.

kamu-core/src/infra/sync_service_impl.rs
```diff
@@ -241,20 +243,46 @@ impl SyncServiceImpl {
...
                 validation,
                 opts.trust_source.unwrap_or(src_is_local),
                 opts.force,
                 listener,
-            )
-            .await
-        {
+            ).await
+
+        } else if dst_dataset.supports_smart_protocol() {
```
sergiimk, line 260: Perhaps `sync_generic` should not exist any more - it's not really "generic" if it just branches based on protocol and the operation. 

In the method where we analyze the dataset URL we probably should pick the specific protocol implementation to use (function pointer / strategy object).

kamu-cli/src/explore/api_server.rs
zaychenko-sergei: Added `kamu-adapter-http`, where we will host the protocol implementation

kamu-adapter-http/src/ws_client_smart_transfer_protocol.rs
```diff
@@ -0,0 +1,95 @@
...
+use kamu::domain::{Dataset, SyncError, SyncResult, SyncListener};
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+
+#[component(pub)]
+pub struct WsSmartTransferProtocolClient {}
+
+#[async_trait::async_trait]
+impl SmartTransferProtocolClient for WsSmartTransferProtocolClient {
```
sergiimk, line 30: name of the file doesn't correspond to to the name of the class - should it be `ws_smart_transfer_protocol_client.rs`?

kamu-adapter-http/src/ws_client_smart_transfer_protocol.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+    
+        unimplemented!("Not supported yet")
+
+    }
+
+    async fn push_protocol_client_flow(
+        &self,
+        src: & dyn Dataset,
+        dst: & dyn Dataset,
```
sergiimk, line 67: should we point to a remote dataset purely by URL, not by supplying `dyn Dataset`?

zaychenko-sergei: Yes, reflected this

kamu-adapter-http/src/ws_client_smart_transfer_protocol.rs
zaychenko-sergei: Renamed

kamu-core/src/domain/repos/dataset.rs
zaychenko-sergei: Resolved

kamu-core/src/infra/sync_service_impl.rs
zaychenko-sergei: I'm analyzing URLs to select an algorithm, but I've left `sync_generic` untouched - it's a default case, when ODF refs are not detected

kamu-core/src/infra/repos/object_repository_in_memory.rs
```diff
@@ -59,6 +72,10 @@ impl ObjectRepository for ObjectRepositoryInMemory {
         unimplemented!()
     }
 
+    async fn get_download_url(&self, _prefix_url: &Url, _hash: &Multihash) -> Result<(Url, Option<DateTime<Utc>>), GetError> {
+        unimplemented!()
```
sergiimk, line 76: I used `unimplemented!()` above for methods that can be implemented but there was simply to use case for doing it back then. In this case `panic!()` is probably more appropriate as it cannot ever be implemented.

kamu-core/src/infra/repos/metadata_chain_impl.rs
```diff
@@ -156,34 +188,48 @@ where
...
-                        BlockMalformedError {
-                            hash: hash.clone(),
-                            source: e.into(),
-                        }
-                        .int_err(),
-                    ));
-                }
-            },
+    async fn get_block_download_url(&self, prefix_url: &Url, hash: &Multihash) -> Result<(Url, Option<DateTime<Utc>>), GetBlockError> {
+        match self.obj_repo.get_download_url(prefix_url, hash).await {
```
sergiimk, line 226: Can you explain `prefix_url` here? I'm confused why it's needed as object repos should know exactly where data is.

kamu-core/src/infra/repos/object_repository_s3.rs
```diff
@@ -187,6 +192,29 @@ where
...
+        let options = PreSignedRequestOption {
+            expires_in: std::time::Duration::from_secs(validity_period_seconds),
+        };
+        
+        let presigned_url = get_object_request.get_presigned_url(&(Region::default()), &credentials, &options);
+        match Url::parse(presigned_url.as_str()) {
+            Ok(url) => Ok((url, Some(Utc::now() + chrono::Duration::seconds(validity_period_seconds as i64)))),
+            Err(e) => Err(GetError::Internal(e.int_err()))
+        }
+    }
```
sergiimk, line 216: This feels misplaced to me in the `ObjectRepository` interface... but I need to think about this more to propose an alternative.

kamu-core/src/infra/utils/smart_transfer_protocol.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use std::sync::Arc;
+use url::Url;
+
+use crate::domain::{Dataset, SyncResult, SyncError, SyncListener};
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait SmartTransferProtocolClient: Sync + Send {
```
sergiimk, line 18: We could probably even put this into `opendatafabric` crate as it's a part of the spec.

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
```diff
@@ -0,0 +1,280 @@
...
+    messages::*, 
+    ws_common::{WriteMessageError, ReadMessageError, self}, 
+    dataset_protocol_helper::*
+};
+
+
+/////////////////////////////////////////////////////////////////////////////////
+
+
+async fn axum_ws_read_generic_payload<TMessagePayload: DeserializeOwned>(
```
sergiimk, line 28: for a module-private function this seems like a very verbose name :)

sergiimk (changes requested):

kamu-adapter-http/src/messages/smart_transfer_protocol_messages.rs
```diff
@@ -0,0 +1,198 @@
...
+
+
+/// Initial dataset pull request message
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct DatasetPullRequest {
+    pub begin_after: Option<Multihash>,
+    pub stop_at: Option<Multihash>,
+}
+
+
```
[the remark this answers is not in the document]
zaychenko-sergei: Right, somehow it was disabled originally since Day 1 of my work.
I've ran formatting all over the project.
Btw, maybe we should think of installing it as a pre-commit hook.

kamu-adapter-http/src/messages/smart_transfer_protocol_messages.rs
```diff
@@ -0,0 +1,198 @@
...
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct DatasetPushRequest {
+    pub current_head: Multihash,
+    pub size_estimation: TransferSizeEstimation,
+}
+
+
+/// Response to initial dataset push request
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct DatasetPushRequestAccepted {}
```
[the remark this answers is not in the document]
zaychenko-sergei: The idea in the protocol was that for logical errors we push `DatasetError` message.

```
/// Error message
#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
pub struct DatasetError {
    pub error_details: ErrorDetails,
}

/// Error message details
#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
pub struct ErrorDetails {
    pub error_code: String,
    pub description: String,
}
```

kamu-adapter-http/src/http_server_dataset_router.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use std::sync::Arc;
+
+use kamu::domain::LocalDatasetRepository;
+
+use crate::http_server_constants::*;
+use crate::http_server_simple_transfer_protocol::*;
+
+
+pub fn create_dataset_routes(local_repo: Arc<dyn LocalDatasetRepository>) -> axum::Router {
```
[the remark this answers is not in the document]
zaychenko-sergei: This covers simple transfer protocol URLs too. I renamed to `create_dataset_transfer_protocol_routes`

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
```diff
@@ -0,0 +1,162 @@
...
+};
+use std::{
+    collections::HashMap, 
+    sync::Arc,
+    str::FromStr,
+};
+
+pub async fn dataset_refs_handler(
+    local_dataset_repository: axum::extract::Extension<Arc<dyn LocalDatasetRepository>>,
+    axum::extract::Path(params): axum::extract::Path<HashMap<String, String>>
```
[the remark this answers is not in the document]
zaychenko-sergei: Originally I thought of binding an optional `accountName` parameter, and it wasn't clear how to express this with simple `Path`. We don't have it anyway for now, can think of this later. Converted to simple path matching.

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
```diff
@@ -0,0 +1,162 @@
...
+    )
+}
+
+async fn resolve_dataset(
+    local_dataset_repository: axum::extract::Extension<Arc<dyn LocalDatasetRepository>>,
+    params: &HashMap<String, String>
+) -> Result<Arc<dyn Dataset>, GetDatasetError> {
+    
+    // TODO: support 'accountName' parameter
+    let dataset_name_param = params.get(PARAMETER_DATASET_NAME).unwrap();
```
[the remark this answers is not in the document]
zaychenko-sergei: What do you mean by "the routes don't mention the datasetName"? We have a nested router.
```
            .nest(
                format!("/:{}", kamu_adapter_http::PARAMETER_DATASET_NAME).as_str(),
                kamu_adapter_http::create_dataset_transfer_protocol_routes(local_repo),
            )
```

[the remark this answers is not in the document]
zaychenko-sergei: If I understood the idea correctly, you suggest to split out the lower-level part of the route handlers from the operations implementation logic completely. So that current handler, i.e., `dataset_blocks_handler`, receives a dataset reference as an argument, constructed block hash as an argument, and does not even know of Axum existence.

With regards to the account name, that would be another row in the router, likely preceeding the current, and I assumed it would reuse the nested router, so it is isolated in the `create_dataset_transfer_protocol_routes`.

What you've described about "externalizing" sounds a lot like middleware layer that propagates custom state to the final handlers. Does [this](https://docs.rs/axum/latest/axum/middleware/index.html#passing-state-from-middleware-to-handlers) look similar?

kamu-adapter-http/src/ws_common.rs
```diff
@@ -0,0 +1,91 @@
...
+        #[from]
+        #[backtrace]
+        BoxedError,
+    ),
+}
+
+/////////////////////////////////////////////////////////////////////////////////
+
+pub fn parse_payload<TMessagePayload: DeserializeOwned>(
+    raw_message: String
```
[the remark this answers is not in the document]
zaychenko-sergei: Corrected

kamu-core/src/domain/repos/metadata_chain.rs
```diff
@@ -26,6 +30,15 @@ pub trait MetadataChain: Send + Sync {
...
 
+    /// Returns size of the specified block in bytes
+    async fn get_block_size(&self, hash: &Multihash) -> Result<u64, GetBlockError>;
+
+    /// Returns raw bytes associated with the block
+    async fn get_block_bytes(&self, hash: &Multihash) -> Result<Bytes, GetBlockError>;
+
+    /// Returns block download URL
+    async fn get_block_download_url(&self, prefix_url: &Url, hash: &Multihash) -> Result<(Url, Option<DateTime<Utc>>), GetBlockError> ;
+
```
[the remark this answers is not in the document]
zaychenko-sergei: Right, I did not know `as_object_repo` exists.

kamu-core/src/infra/repos/metadata_chain_impl.rs
zaychenko-sergei: With S3 repository, the output URL depends solely on the external storage system + bucket key.
With local repository, the repository can only tell the relative URL part, but it has no idea on the API server context (host+port), nor the dataset context (author?+name).

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
zaychenko-sergei: Simplified here, both read & write, and similar in Tungstenite part.

kamu-core/src/infra/repos/object_repository_s3.rs
zaychenko-sergei: It feels the same way too, and this is related to `prefix_url` as well, which is ignored here, as you can see, and only used in local FS. I thought of a Visitor pattern, but the implementation depends on the internal fields a lot.

kamu-core/src/infra/repos/object_repository_in_memory.rs
zaychenko-sergei: Makes sense

kamu-core/src/infra/repos/dataset_factory_impl.rs
```diff
@@ -42,6 +42,7 @@ impl DatasetFactoryImpl {
 
     pub fn get_local_fs(layout: DatasetLayout) -> DatasetImplLocalFS {
         DatasetImpl::new(
+            Url::parse(format!("file://{}", layout.root_dir.to_str().unwrap()).as_str()).unwrap(),
```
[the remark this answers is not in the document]
zaychenko-sergei: `from_directory_path` matched even better

kamu-core/src/domain/repos/metadata_chain.rs
```diff
@@ -225,6 +247,20 @@ pub enum GetBlockError {
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Error, Debug)]
+pub enum ConstructBlockFromBytesError {
```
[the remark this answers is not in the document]
zaychenko-sergei: Not anymore. It used to in one of the intermediate versions.
Moved to impl and made module-private.

kamu-core/src/domain/repos/metadata_chain.rs
```diff
@@ -334,6 +370,17 @@ impl From<super::reference_repository::SetRefError> for AppendError {
     }
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Error, Debug)]
+pub enum AppendFromBytesError {
+    #[error(transparent)]
+    Append(#[from] AppendError),
```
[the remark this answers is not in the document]
zaychenko-sergei: This looks a bit wider, as the same flow is activated in `get_block` implementation. The `construct_block_from_bytes` is just an extracted sub-method, that was originally used in `get_block`.

kamu-adapter-http/src/messages/smart_transfer_protocol_messages.rs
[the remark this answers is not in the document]
sergiimk: Ok, I'll wait for push implementation to see how it looks like. 

What I was mostly wondering about is how will you read from the socket a message that can be either `*Accepted` or  `*Error`, whether its by:
- trying to to parse `*Accepted` and falling back to other types on parse failures
- or by creating an enum type that combines successes and errors

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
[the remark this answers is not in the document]
sergiimk: Yes, to be specific my suggestion is to write `kamu-adapter-http` in such a way that its routes can be mounted under any URL structure like:
- `/:datasetName/...` - single-tenant, by name
- `/:datasetID/...` - single-tenant, by ID
- `/:accountName/:datasetName/...` - multi-tenant, by name
- `/...` - or even a server that serves only one specific dataset

Therefore the logic that determines the identity of a dataset that STP handlers will operate on has to be defined at the place of routes **nesting**.

I'm not familiar enough with Axum's services/layers/middleware structure to say for sure, but what you found looks very close how I imagined it.

kamu-core/src/infra/repos/object_repository_s3.rs
sergiimk: Some poorly structured thoughts:

Imagine we're doing a "pull" via `https://my-odf-server.com/:accountName/:datasetName/pull` - when transferring data I see following possible modes of operation:
- **Proxied** - data is served through simple transfer protocol
  - URL is determined as `reverse_url(/blocks/:hash)`
  - e.g. `https://my-odf-server.com/:accountName/:datasetName/blocks/:hash`
- **Delegated** - downloads/uploads go directly to storage
  - e.g. `https://s3.us-west-2...aws.com/datasetsBucket/:datasetID/blocks/:hash`
- **In-Band** - data is streamed via same websocket (we're not implementing this yet, but useful to keep in mind)

Some ODF servers can only choose to support proxied mode, some storage back-ends may not support delegation, some clients may prefer only proxied/in-band mode (e.g. some firewall concerns).

Now how this all relates to `ObjectRepository` - should it be the one issuing these URLs or do we need a completely different interface for this...

The platform-specific implementation of the `ObjectRepository` can be pretty complex:
- It may utilize some cache / index to read blocks really fast
  - e.g. Redis / Elastic, or holding last 100 blocks of datasets in memory...
- Writes and cache misses will have to use a URL of the **internal** storage, e.g. `s3://datasets.kamu.dev/:datasetID/blocks/`
- Which in most cases will be different from public storage URLs for downloads, e.g. `https://some-cdn-host.com/bucketname/:datasetID/blocks/:hash`


Putting everything together I think:
- I don't see a reason to move `get_download_url()` out of `ObjectRepository` yet
- I think `prefix_url` param should be removed though
- Instead we should be configuring `ObjectRepository` with both
  - internal URL to use for normal reads/writes
  - external URL for issuing download/upload URLs to the clients


I suggest to change the signature to:
```rust
trait ObjectRepository {
  async fn get_download_url(&self, hash: &Multihash, opts: &DownloadOpts)
    -> Result<GetDownloadUrlResult, GetDownloadUrlError>;
}

struct DownloadOpts {
  expiration: Option<TimeDelta>,
}

struct GetDownloadUrlResult {
  url,
  expires_at,
}

enum GetDownloadUrlError {
  NotSupported,
  ...
}
```

This way if a specific storage back-end does not support delegated downloads/uploads - the "smart" handler can fall back onto proxied / "simple" protocol. Note that this will require some sort of URL reversal `reverse_url("relative/url")` - which Axum may have support for.

kamu-core/src/infra/repos/metadata_chain_impl.rs
sergiimk: Suggested a solution below - will resolve this comment.

kamu-core/src/infra/repos/object_repository_s3.rs
zaychenko-sergei: Changed the signature of `get_download_url` as described above. 
Eliminated `prefix_url` argument from all operations.
Shifted resolution of the base URL to dataset resolver - passing external base URL as an argument.
Local FS repository optionally can store a base URL, which is used to reply on download links

kamu-core/src/domain/repos/dataset.rs
```diff
@@ -26,6 +27,8 @@ pub trait Dataset: Send + Sync {
     fn as_checkpoint_repo(&self) -> &dyn ObjectRepository;
     fn as_cache_repo(&self) -> &dyn NamedObjectRepository;
 
+    fn base_url(&self) -> &Url;
```
sergiimk, line 30: Looks like this method is not used any more - probably best to remove it to avoid anyone making assumptions about the specific layout of a dataset under such URL (e.g. `dataset.base_url().join("blocks/abcd")`)

kamu-core/src/domain/repos/local_dataset_repository.rs
```diff
@@ -51,6 +52,12 @@ pub trait LocalDatasetRepository: DatasetRegistry + Sync + Send {
         dataset_ref: &DatasetRefLocal,
     ) -> Result<Arc<dyn Dataset>, GetDatasetError>;
 
+    async fn get_dataset_with_external_url(
```
sergiimk, line 55: As discussed on Slack - my suggestion is to remove this in favor of having existing `get_dataset()` method return datasets with external URLs pre-configured, which can be achieved by having a S3-specific implementation of the `LocalDatasetRepository`.

kamu-core/src/domain/repos/metadata_chain.rs
```diff
@@ -59,6 +61,14 @@ pub trait MetadataChain: Send + Sync {
         opts: AppendOpts<'a>,
     ) -> Result<Multihash, AppendError>;
 
+    /// Loads the block from bytes and appends it to the chain
+    async fn append_block_from_bytes<'a>(
+        &'a self,
+        hash: &Multihash,
+        block_bytes: Bytes,
```
sergiimk, line 68: What is the advantage of passing `Bytes` here instead of a slice `&[u8]`?

kamu-core/src/infra/repos/object_repository_in_memory.rs
```diff
@@ -56,7 +67,15 @@ impl ObjectRepository for ObjectRepositoryInMemory {
...
-        unimplemented!()
+        panic!("get_stream not allowed for in-memory repository");
+    }
+
+    async fn get_download_url(
+        &self,
+        _hash: &Multihash,
+        _opts: DownloadOpts,
+    ) -> Result<GetDownloadUrlResult, GetDownloadUrlError> {
+        panic!("get_download_url not allowed for in-memory repository");
```
sergiimk, line 78: Why not return `Err(GetDownloadUrlError::NotSupported)` instead of panic?

kamu-core/src/infra/repos/object_repository_local_fs.rs
```diff
@@ -38,13 +40,14 @@ where
     D: Send + Sync,
     D: digest::Digest,
 {
-    pub fn new<P>(root: P) -> Self
+    pub fn new<P>(root: P, external_base_url: Option<Url>) -> Self
```
sergiimk, line 43: If we put `LocalDatasetRepositoryS3` in charge of S3 external URLs then base URL will not appear anywhere but in `ObjectRepositoryS3`. All others would simply return `Err(GetDownloadUrlError::NotSupported)` making protocol to fall back to proxying traffic.

sergiimk (changes requested):

kamu-core/src/domain/repos/metadata_chain.rs
zaychenko-sergei: I did not invent this. This extends the existing code path in `get_block` about deserializing block content from disk

kamu-core/src/infra/repos/object_repository_in_memory.rs
zaychenko-sergei: In-memory repository is only used by tests and panics in many if the overrides.

kamu-core/src/infra/repos/object_repository_local_fs.rs
zaychenko-sergei: I do not understand what do you mean by "fall back to proxying traffic". We have already started a websocket-based communication loop that has different rules. What I read in this comment looks like something that we did not define in the smart transfer protocol description at all..

zaychenko-sergei: Who would inject this new dataset repository into the catalog? We don't have anything else on today except our CLI app. What you are describing here, is that "platform" would have it's own catalog version, and that would inject different dataset repository. So, if I still use CLI app as the ground, this means these download strategies I tried to implement will never even enable themselves.

kamu-cli/src/explore/api_server.rs
```diff
@@ -27,6 +35,14 @@ impl APIServer {
...
                 axum::routing::get(graphql_playground).post(graphql_handler),
             )
+            .nest(
+                format!("/:{}", kamu_adapter_http::PARAMETER_DATASET_NAME).as_str(),
+                kamu_adapter_http::create_dataset_transfer_protocol_routes()
+                    .layer(axum::middleware::from_fn(
+                        kamu_adapter_http::resolve_dataset_by_name,
+                    ))
+                    .layer(axum::extract::Extension(local_dataset_resolver)),
+            )
```
sergiimk, line 45: This is great - almost how I imagined it. 

I wonder if we can further simplify it to something like this:
```rust
.nest(
    "/:datasetName",
    kamu_adapter_http::create_dataset_transfer_protocol_routes()
        .layer(kamu_adapter_http::LocalDatasetResolver(
            |request| odf::DatasetName::try_parse(request.path_params.get("datasetName"))?.as_local_ref()),
        ),
)
````
i.e. have just one `DatasetResolverLocal` middleware that we can parametrize with a lambda that knows how to extract a `DatasetRefLocal` from request parts.

File: kamu-core/src/domain/repos/local_dataset_repository.rs (hunk 22 above)
Comment (zaychenko-sergei) [reply kamu-core/src/domain/repos/local_dataset_repository.rs]: Got rid of this method

File: kamu-core/src/infra/repos/dataset_repository_s3.rs [hunk 27]
```diff
@@ -0,0 +1,473 @@
...
+
+    async fn eliminate_bucket_items(
+        &self,
+        dataset_name: &DatasetName,
+    ) -> Result<(), InternalError> {
+        let dataset_key_prefix = self.s3_context.get_key(dataset_name.as_str());
+        let list_response = self
+            .s3_context
+            .client
+            .list_objects(ListObjectsRequest {
```
sergiimk, line 65: Looks like this is deprecated in favor of `list_objects_v2`: https://rusoto.github.io/rusoto/rusoto_s3/trait.S3.html#tymethod.list_objects

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+                prefix: Some(dataset_key_prefix),
+                ..ListObjectsRequest::default()
+            })
+            .await
+            .int_err()?;
+
+        if let Some(contents) = list_response.contents {
+            self.s3_context
+                .client
+                .delete_objects(DeleteObjectsRequest {
```
sergiimk, line 76: `list_objects` only returns up to 1000 items
likewise `delete_objects` can accept up to 1000 keys
If you don't implement "pagination" then at lease an assertion is in order to detect when we exceeded the limit.

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+        );
+        Url::parse(dataset_url_string.as_str()).unwrap()
+    }
+
+    fn get_dataset_impl(&self, dataset_name: &DatasetName) -> Result<impl Dataset, InternalError> {
+        let dataset_url = self.get_s3_bucket_path(dataset_name.as_str());
+        DatasetFactoryImpl::get_s3(dataset_url)
+    }
+
+    async fn eliminate_bucket_items(
```
sergiimk, line 57: Very creative name - I had to read code to understand what it does :)
How about `delete_s3_objects_by_prefix`?

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+            Ok(s) => Ok(s),
+            Err(GetSummaryError::EmptyDataset) => unreachable!(),
+            Err(GetSummaryError::Access(e)) => Err(e.int_err().into()),
+            Err(GetSummaryError::Internal(e)) => Err(CreateDatasetError::Internal(e)),
+        }?;
+
+        let handle = DatasetHandle::new(summary.id, dataset_name.clone());
+
+        match self
+            .move_bucket_items_on_dataset_rename(staging_name, dataset_name.as_str())
```
sergiimk, line 113: Moving in S3 is expensive, so since it doesn't give us atomicity mb we shouldn't even bother, and better solve the concurrency problem on the database level.

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+                if resp {
+                    return Err(MoveBucketItemsOnRenameError::NameCollision(
+                        NameCollisionError {
+                            name: DatasetName::from_str(new_dataset_name).unwrap(),
+                        },
+                    ));
+                }
+            }
+            Err(e) => return Err(MoveBucketItemsOnRenameError::Internal(e)),
+        }
```
sergiimk, line 159: I think this can be simplified to:
```rust
if self.bucket_path_exists(target_key_prefix.as_str()).await? {
  return Err(...);
}
```

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+            })
+            .await;
+
+        match listing {
+            Ok(resp) => Ok(resp.contents.is_some()),
+            Err(e) => Err(e.int_err().into()),
+        }
+    }
+
+    async fn move_bucket_items_on_dataset_rename(
```
sergiimk, line 142: I feel like there's too much inconsistency between functions taking dataset names as parameters and those that take S3 paths / prefixes.

I would recommend creating a utility module for S3 with list / recursive delete / recursive move functionality and then using it here. Such module can probably unify some code with `ObjectStoreS3` too.

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+                }
+            }
+            Err(e) => return Err(MoveBucketItemsOnRenameError::Internal(e)),
+        }
+
+        let old_key_prefix = self.s3_context.get_key(old_dataset_name);
+        let list_response = self
+            .s3_context
+            .client
+            .list_objects(ListObjectsRequest {
```
sergiimk, line 165: Same problem with 1,000 item limit

kamu-core/src/domain/dataset_repository_utils.rs
```diff
@@ -0,0 +1,213 @@
...
+};
+
+use crate::domain::{
+    AppendOpts, CreateDatasetFromSnapshotError, CreateDatasetResult, DatasetRepository,
+    GetDatasetError, InvalidSnapshotError, MissingInputsError, ResultIntoInternal,
+};
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub async fn create_dataset_from_snapshot(
```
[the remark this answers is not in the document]
sergiimk: Functions like `get_downstream_dependencies_impl` for example make more sense as "utils".

kamu-core/src/infra/repos/dataset_repository_s3.rs
```diff
@@ -0,0 +1,473 @@
...
+    ) -> DatasetHandleStream<'s> {
+        Box::pin(get_downstream_dependencies_impl(self, dataset_ref))
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+// DatasetS3BuilderImpl
+/////////////////////////////////////////////////////////////////////////////////////////
+
+struct DatasetS3BuilderImpl {
```
sergiimk, line 397: mb `DatasetBuilderS3`?

kamu-core/src/infra/sync_service_impl.rs
```diff
@@ -268,6 +275,73 @@ impl SyncServiceImpl {
...
+        force: bool,
+        listener: Arc<dyn SyncListener + 'static>,
+    ) -> Result<SyncResult, SyncError> {
+        info!("Starting sync using Simple Transfer Protocol");
+        SimpleTransferProtocol
+            .sync(
+                src,
+                src_ref,
+                dst,
+                _dst_ref,
```
sergiimk, line 336: underscore should be removed since variable is in use?

sergiimk (changes requested):

kamu-core/src/infra/sync_service_impl.rs
zaychenko-sergei: oops

kamu-core/src/infra/repos/dataset_repository_s3.rs
zaychenko-sergei: Renamed as suggested. Similarly, renamed the builder in the local FS version to `DatasetBuilderLocalFs`

kamu-core/src/domain/repos/dataset_repository.rs
```diff
@@ -103,7 +90,7 @@ pub trait DatasetBuilder: Send + Sync {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[async_trait]
-pub trait LocalDatasetRepositoryExt: LocalDatasetRepository {
+pub trait LocalDatasetRepositoryExt: DatasetRepository {
```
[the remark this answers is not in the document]
zaychenko-sergei: Fine

kamu-cli/src/commands/add_command.rs
```diff
@@ -193,10 +193,8 @@ impl Command for AddCommand {
             }
         };
 
-        let mut add_results = self
-            .local_repo
-            .create_datasets_from_snapshots(snapshots)
-            .await;
+        let mut add_results =
+            create_datasets_from_snapshots(self.local_repo.as_ref(), snapshots).await;
```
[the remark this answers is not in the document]
zaychenko-sergei: The idea did not work well, generating weird lifetime errors, as we discussed.

kamu-core/src/domain/dataset_repository_utils.rs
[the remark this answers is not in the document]
zaychenko-sergei: Other functions moved to extension

kamu-core/src/infra/repos/dataset_repository_s3.rs
zaychenko-sergei: Replaced

kamu-core/src/infra/repos/dataset_repository_s3.rs
zaychenko-sergei: Agreed

kamu-core/src/infra/repos/dataset_repository_s3.rs
zaychenko-sergei: Addressed

sergiimk (changes requested):

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
```diff
@@ -0,0 +1,229 @@
...
+            return Err(axum::http::StatusCode::INTERNAL_SERVER_ERROR);
+        }
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////
+
+pub async fn dataset_blocks_handler(
+    dataset: Extension<Arc<dyn Dataset>>,
+    axum::extract::Path(params): axum::extract::Path<HashMap<String, String>>,
```
sergiimk, line 53: See my PR where I extract dataset id/name using a serde-derived struct - this should be much more efficient and cleaner than to access params via hashmap.

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
```diff
@@ -0,0 +1,229 @@
...
+                Err(axum::http::StatusCode::INTERNAL_SERVER_ERROR)
+            }
+        },
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////
+
+fn resolve_base_api_url(api_host: axum::headers::Host) -> Url {
+    let mut base_url_str = String::from("http://");
```
sergiimk, line 218: Note that hard-coded HTTP will fail once we deploy on AWS.

sergiimk: The easiest fix would be to configure `base_url` from the outside.

Alternatively we could try writing an extractor similar to `axum::extract::Host` to guess the scheme e.g. from `X-Forwarded-Proto` and local port (80 / 443).

sergiimk (changes requested):

kamu-core/src/infra/sync_service_impl.rs
```diff
@@ -241,33 +246,112 @@ impl SyncServiceImpl {
...
+    async fn sync_smart_pull_transfer_protocol<'a>(
+        &'a self,
+        odf_src: &DatasetRefAny,
+        dst: &DatasetRefAny,
+        opts: SyncOptions,
+        listener: Arc<dyn SyncListener>,
+    ) -> Result<SyncResult, SyncError> {
+        let odf_src_remote_ref = odf_src.as_remote_ref().unwrap();
+        let odf_src_url = self.resolve_remote_dataset_url(&odf_src_remote_ref).await?;
+        let http_src_url = Url::parse(&(odf_src_url.as_str())[4..]).unwrap(); // odf+http, odf+https - cut odf+
```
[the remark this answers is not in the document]
zaychenko-sergei: This wont' work, see the comment in `set_scheme`:

```
    /// Change this URL’s scheme.
    ///
    /// Do nothing and return `Err` under the following circumstances:
    ///
    /// * If the new scheme is not in `[a-zA-Z][a-zA-Z0-9+.-]+`
    /// * If this URL is cannot-be-a-base and the new scheme is one of
    ///   `http`, `https`, `ws`, `wss` or `ftp`
    /// * If either the old or new scheme is `http`, `https`, `ws`,
    ///   `wss` or `ftp` and the other is not one of these
    /// * If the new scheme is `file` and this URL includes credentials
    ///   or has a non-null port
    /// * If this URL's scheme is `file` and its host is empty or null
```

In particular, 
 > If either the old or new scheme is `http`, `https`, `ws`, `wss` or `ftp` and the other is not one of these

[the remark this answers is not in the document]
zaychenko-sergei: I've written the intent more explicitly:

```
let http_src_url = Url::parse(&(odf_src_url.as_str())["odf+".len()..]).unwrap(); // odf+http, odf+https - cut odf+
```

kamu-core/src/infra/utils/s3_context.rs
```diff
@@ -0,0 +1,330 @@
...
+                    bucket: self.bucket.clone(),
+                    prefix: Some(key_prefix.clone()),
+                    max_keys: Some(S3Context::MAX_LISTED_OBJECTS),
+                    ..ListObjectsV2Request::default()
+                })
+                .await
+                .int_err()?;
+
+            if let Some(contents) = list_response.contents {
+                has_next_page = (contents.len() as i64) == S3Context::MAX_LISTED_OBJECTS;
```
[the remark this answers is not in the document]
zaychenko-sergei: Right, improved the criteria

kamu-core/src/infra/utils/s3_context.rs
```diff
@@ -0,0 +1,330 @@
...
+                self.client
+                    .delete_objects(DeleteObjectsRequest {
+                        bucket: self.bucket.clone(),
+                        delete: Delete {
+                            objects: contents
+                                .iter()
+                                .map(|obj| ObjectIdentifier {
+                                    key: obj.key.clone().unwrap(),
```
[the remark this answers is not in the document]
zaychenko-sergei: Improved here and in 1 more location. Still, not enough Rust standard library knowledge, I need more time to read and practice.

kamu-adapter-http/tests/tests/test_smart_pull.rs
```diff
@@ -0,0 +1,302 @@
...
+        )
+        .await
+        .unwrap();
+
+    let client_harness = client_side_harness();
+
+    let foo_odf_url = server_harness.dataset_url("foo");
+    let foo_dataset_ref = DatasetRefRemote::from(&foo_odf_url);
+
+    let api_server_handle = async { server_harness.api_server_run().await };
```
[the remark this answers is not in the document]
zaychenko-sergei: Makes sense

kamu-adapter-http/tests/tests/test_smart_pull.rs
```diff
@@ -0,0 +1,302 @@
...
+    let client_handle = async {
+        let pull_result = client_harness
+            .pull_dataset(DatasetRefAny::from(foo_dataset_ref))
+            .await
+            .get(0)
+            .unwrap()
+            .result
+            .as_ref()
+            .unwrap()
+            .clone();
```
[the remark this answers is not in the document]
zaychenko-sergei: The problem here is that `PullResponse` is non-cloneable, as `PullError` isn't.
In one of the tests I need to access an error, not just `PullResult`.
I've decided to write another client helper method for normal cases, and only use the multi-version for error part.

kamu-adapter-http/tests/tests/test_smart_pull.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        assert_eq!(
```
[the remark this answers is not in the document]
zaychenko-sergei: It seems that temp files in Minio server only exist in the container memory, and are never flushed to the disk temp directory during the test. At least I don't see any files, the bucket directory is always empty.

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
zaychenko-sergei: image

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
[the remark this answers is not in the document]
zaychenko-sergei: I assume this is solved by your middleware branch now

kamu-cli/src/explore/api_server.rs
zaychenko-sergei: Solved in your branch

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
zaychenko-sergei: Used. It really works with the named fields in structs, unlike tuple syntax, which expects all prefixes to be listed.

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
zaychenko-sergei: Added `KAMU_PROTOCOL_SCHEME` env var

CHANGELOG.md
```diff
@@ -1,66 +1,106 @@
...
+
+- Smart Transfer Protocol: implemented Pull flow based on web-sockets
+- S3-based dataset repository implementation
+
+### Changed
+
+- Improved status codes and error handling in API server's routes for Simple Transfer Protocol
+
 ## [0.115.0] - 2023-03-19
+
```
sergiimk, line 20: I prefer the compact format (and changed changelog formats accordingly in few other projects), so if you're not proposing a new style - let's revert these whitespace changes.

CHANGELOG.md
```diff
@@ -1,66 +1,106 @@
...
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.116.0] - 2023-03-28
+
+### Added
+
+- Smart Transfer Protocol: implemented Pull flow based on web-sockets
```
sergiimk, line 12: Would be nice to give a short example, e.g. "if you run `kamu system api-server` you can pull from it as ..."

kamu-adapter-http/tests/tests/test_smart_pull.rs
```diff
@@ -0,0 +1,272 @@
...
+        // TODO: try expecting better error message
+        assert!(pull_responses[0].result.is_err());
+    };
+
+    await_client_server_flow!(api_server_handle, client_handle)
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+fn copy_folder_recursively(src: &Path, dst: &Path) -> io::Result<()> {
```
sergiimk, line 258: Consider replacing with [`fs_extra::dir::copy()`](https://docs.rs/fs_extra/latest/fs_extra/dir/fn.copy.html) - we already depend on `fs_extra` crate in other places.

sergiimk (approved):
Last few comments - other than that looks ready to go! :rocket:

kamu-adapter-http/tests/tests/test_smart_pull.rs
zaychenko-sergei: Utilized

CHANGELOG.md
zaychenko-sergei: Added

CHANGELOG.md
zaychenko-sergei: Restored

kamu-adapter-http/tests/tests/test_smart_pull.rs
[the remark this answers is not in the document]
zaychenko-sergei: Solution found

PR #123, opened 2023-02-08
Handling situations when dataset schema is not yet defined

zaychenko-sergei:
CLI:
- kamu inspect schema prints a warning
- kamu tail prints an error

GraphQL adapter:
- optional schema access taken into account
- empty tail when there is no schema

### DIFF
--- a/kamu-adapter-graphql/src/scalars/dataset.rs
+++ b/kamu-adapter-graphql/src/scalars/dataset.rs
@@ -178,6 +178,14 @@ pub struct DataBatch {
 }
 
 impl DataBatch {
+    pub fn empty(format: DataBatchFormat) -> DataBatch {
+        DataBatch {
+            format,
+            content: String::from("{}"),
+            num_records: 0
+        }
+    }
+
     pub fn from_records(
         record_batches: &Vec<datafusion::arrow::record_batch::RecordBatch>,
         format: DataBatchFormat,
@@ -228,7 +236,7 @@ impl DataBatch {
 
 #[derive(SimpleObject)]
 pub struct DataQueryResultSuccess {
-    pub schema: DataSchema,
+    pub schema: Option<DataSchema>,
     pub data: DataBatch,
     pub limit: u64,
 }
@@ -252,7 +260,7 @@ pub enum DataQueryResult {
 }
 
 impl DataQueryResult {
-    pub fn success(schema: DataSchema, data: DataBatch, limit: u64) -> DataQueryResult {
+    pub fn success(schema: Option<DataSchema>, data: DataBatch, limit: u64) -> DataQueryResult {
         DataQueryResult::Success(DataQueryResultSuccess {
             schema,
             data,
@@ -273,6 +281,7 @@ impl DataQueryResult {
             error_kind: DataQueryResultErrorKind::InternalError,
         })
     }
+    
 }
 
 impl From<QueryError> for DataQueryResult {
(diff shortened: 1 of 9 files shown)


kamu-adapter-graphql/src/scalars/dataset.rs
```diff
@@ -178,6 +178,14 @@ pub struct DataBatch {
 }
 
 impl DataBatch {
+    pub fn empty(format: DataBatchFormat) -> DataBatch {
+        DataBatch {
+            format,
+            content: String::from("{}"),
```
sergiimk, line 184: This is incorrect since `format` can be many things other than `JSON` and `{}` will be invalid for CSV, JsonLD and other supported formats. See `from_records` below as I had to add some code to handle this already.

kamu-adapter-graphql/src/scalars/dataset.rs
```diff
@@ -252,7 +260,7 @@ pub enum DataQueryResult {
 }
 
 impl DataQueryResult {
-    pub fn success(schema: DataSchema, data: DataBatch, limit: u64) -> DataQueryResult {
+    pub fn success(schema: Option<DataSchema>, data: DataBatch, limit: u64) -> DataQueryResult {
```
sergiimk, line 263: Should we also introduce an `empty()` factory function, so that `success()` continues to take value not `Option`? This way it won't be possible to create a result with data but without a schema.

kamu-cli/src/commands/inspect_schema_command.rs
```diff
@@ -104,26 +104,41 @@ impl InspectSchemaCommand {
...
                 e @ QueryError::DataFusionError(_) => CLIError::failure(e),
                 e @ QueryError::Internal(_) => CLIError::critical(e),
             })?;
 
-        match self.output_format.as_ref().map(|s| s.as_str()) {
-            None | Some("ddl") => self.print_schema_ddl(&schema),
-            Some("parquet") => self.print_schema_parquet(&schema)?,
-            Some("json") => self.print_schema_json(&schema)?,
-            _ => unreachable!(),
+        match schema_or_none {
```
sergiimk, line 132: For future: I see other codebases often use `maybe_<x>` naming e.g. `maybe_schema` for optional values, mb we should adopt it too.

kamu-core/src/domain/query_service.rs
```diff
@@ -75,6 +82,16 @@ pub enum QueryError {
     ),
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Error, Clone, PartialEq, Eq, Debug)]
+#[error("Dataset schema is not available yet: {dataset_ref}")]
```
sergiimk, line 88: Nitpick: "not yet available" I think is more grammatically correct, but I don't have a proof :)

sergiimk (approved):
Pointed out one issue - rest looks good to merge.
Let's add a CHANGELOG entry too.

kamu-adapter-graphql/src/scalars/dataset.rs
zaychenko-sergei: Changed string value depending on the format.  assume `JsonXXX` are still `Json`

kamu-cli/src/commands/inspect_schema_command.rs
zaychenko-sergei: Maybe :-)

kamu-core/src/domain/query_service.rs
zaychenko-sergei: Let it be

PR #129, opened 2023-02-19
GQL dataset creation and append APIs

sergiimk:
Adds dataset creation APIs:

```gql
"""
Creates a new empty dataset
"""
createEmpty(accountId: AccountID!, datasetKind: DatasetKind!, datasetName: DatasetName!): CreateDatasetResult!
"""
Creates a new dataset from provided DatasetSnapshot manifest
"""
createFromSnapshot(accountId: AccountID!, datasetSnapshot: String!, datasetSnapshotFormat: MetadataManifestFormat!): CreateDatasetFromSnapshotResult!
```

Also adds a convenience method to get raw metadata block (e.g. in YAML format):
```gql
"""
Returns a metadata block corresponding to the specified hash and encoded in desired format
"""
blockByHashRaw(hash: Multihash!, blockFormat: MetadataManifestFormat!): String
```

And looking ahead - a new method for appending metadata events to a chain:
```gql
"""
Commits new event to the metadata chain
"""
commitEvent(metadataEvent: String!, metadataEventFormat: MetadataManifestFormat!): CommitResult!
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -220,7 +220,7 @@ dependencies = [
  "arrow-cast",
  "arrow-data",
  "arrow-schema",
- "flatbuffers 22.9.29",
+ "flatbuffers",
 ]
 
 [[package]]
@@ -454,9 +454,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "axum"
-version = "0.6.6"
+version = "0.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e246206a63c9830e118d12c894f56a82033da1a2361f5544deeee3df85c99d9"
+checksum = "2fb79c228270dcf2426e74864cabc94babb5dbab01a4314e702d2f16540e1591"
 dependencies = [
  "async-trait",
  "axum-core",
@@ -806,7 +806,7 @@ checksum = "f3ad85c1f65dc7b37604eb0e89748faf0b9653065f2a8ef69f96a687ec1e9279"
 
 [[package]]
 name = "container-runtime"
-version = "0.111.0"
+version = "0.112.0"
 dependencies = [
  "dill",
  "regex",
@@ -1515,16 +1515,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "flatbuffers"
-version = "23.1.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "77f5399c2c9c50ae9418e522842ad362f61ee48b346ac106807bd355a8a7c619"
-dependencies = [
- "bitflags",
- "rustc_version",
-]
-
-[[package]]
 name = "flate2"
 version = "1.0.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1861,9 +1851,9 @@ dependencies = [
 
 [[package]]
 name = "http"
-version = "0.2.8"
+version = "0.2.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75f43d41e26995c17e71ee126451dd3941010b0514a81a9d11f3b341debc2399"
+checksum = "bd6effc99afb63425aff9b05836f029929e345a6148a14b7ecd5ab67af944482"
 dependencies = [
  "bytes",
  "fnv",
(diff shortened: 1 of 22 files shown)


kamu-adapter-graphql/src/queries/datasets.rs
```diff
@@ -103,6 +105,206 @@ impl Datasets {
...
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// CreateDatasetResult
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Interface, Debug, Clone)]
+#[graphql(field(name = "message", type = "String"))]
+pub enum CreateDatasetResult {
```
sergiimk, line 235: Note the new approach to error handling that I'm suggesting:
* Result is an interface with single field `message` - this allows for a "catch-all" error handling code
* Success and individual errors can be unpacked as `... on CreateDatasetResultSuccess { dataset { id } }` so that detailed fields can be accessed
* This can be extended with an error code if we need it, but `_typename` should already serve as a unique error identifier

What do you think?

kamu-adapter-graphql/src/queries/metadata_chain.rs
```diff
@@ -77,6 +78,29 @@ impl MetadataChain {
         Ok(block.map(|b| MetadataBlockExtended::new(hash, b, Account::mock())))
     }
 
+    /// Returns a metadata block corresponding to the specified hash and encoded in desired format
+    async fn block_by_hash_raw(
```
sergiimk, line 82: I suggest using this method for the YAML view - this is the only way we will display ALL fields of the metadata block, unlike when converting GQL object to YAML, since latter will only contain fields that platform is aware of (i.e. queries via GQL).

kamu-adapter-graphql/src/queries/datasets.rs
```diff
@@ -103,6 +105,206 @@ impl Datasets {
...
+            .append(
+                odf::MetadataBlock {
+                    system_time: Utc::now(),
+                    prev_block_hash: None,
+                    sequence_number: 0,
+                    event: odf::MetadataEvent::Seed(odf::Seed {
+                        dataset_id,
+                        dataset_kind: dataset_kind.into(),
+                    }),
+                },
```
zaychenko-sergei, line 139: I prefer if this low-level code would be encapsulated behind a core service. API plugin should work at higher abstraction level.

kamu-adapter-graphql/src/queries/datasets.rs
zaychenko-sergei: UI only needs to display an error. Most of the time it would not need any complex polymorphic behavior.

kamu-adapter-graphql/src/queries/metadata_chain.rs
zaychenko-sergei: Why calling it "raw"? It looks like "export as" to me.

zaychenko-sergei (changes requested):

Cargo.lock
```diff
@@ -220,7 +220,7 @@ dependencies = [
  "arrow-cast",
  "arrow-data",
  "arrow-schema",
- "flatbuffers 22.9.29",
+ "flatbuffers",
```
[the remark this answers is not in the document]
sergiimk: The problem was that we directly depend on `flatbuffers` and also indirectly via `arrow`. Because we were specifying our dependency as `flatbuffers = "*"` - we were pulling latest version v23 while arrow still depends on v22.

I changed our dependency to `flatbuffers = "22"` in order to avoid compiling both versions and use same version as arrow.

There are cargo plugins that allow you to whitelist which crates can be duplicated and which can't - we probably should start using them to ensure we don't compile with duplicated versions of some major crates.

kamu-adapter-graphql/src/queries/datasets.rs
```diff
@@ -103,6 +105,206 @@ impl Datasets {
         let account = Account::mock();
         self.by_account_impl(ctx, account, page, per_page).await
     }
+
+    // TODO: Multitenancy
```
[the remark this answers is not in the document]
sergiimk: accountID is currently ignored, so we'll need to migrate this code to a multitenancy-aware repository once we have it

kamu-adapter-graphql/src/queries/datasets.rs
sergiimk: Yeah, was bothering me too - I found a simple solution.

kamu-adapter-graphql/src/queries/metadata_chain.rs
sergiimk: "Raw" as in returning a serialized data as opposed to returning a GQL object. But I agree it's not a great name.

Renamed to `block_by_hash_encoded` as "export" didn't sound right to me and I also wanted this method to have the same name prefix as the one returning a GQL object.

PR #151, opened 2023-03-26
Introduce DatasetResolver middleware

sergiimk:
This PR:
- Introduces `DatasetResolverLayer` middleware parameterizable with a lambda that extracts dataset identity from any part of the request
- Adds tests for different mount points and resolution methods
- Un-hardcodes related parts of the protocol


kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
```diff
@@ -162,15 +162,10 @@ pub async fn dataset_push_ws_upgrade_handler(
 pub async fn dataset_pull_ws_upgrade_handler(
     ws: axum::extract::ws::WebSocketUpgrade,
     dataset: Extension<Arc<dyn Dataset>>,
-    axum::TypedHeader(api_host): axum::TypedHeader<axum::headers::Host>,
-    axum::extract::Path(params): axum::extract::Path<HashMap<String, String>>,
+    host: axum::extract::Host,
```
sergiimk, line 165: Notice the use of different `Host` extractor - the one you were using did not respect the "X-Forwarded-Host" or other types of proxy headers.

kamu-adapter-http/src/http_server_simple_transfer_protocol.rs
```diff
@@ -183,47 +178,20 @@ pub async fn dataset_pull_ws_upgrade_handler(
...
-    if let Some(port) = api_host.port() {
-        base_url_str += ":";
-        base_url_str += &port.to_string();
+fn get_base_dataset_url(
+    host: axum::extract::Host,
+    uri: axum::extract::OriginalUri,
+    depth: usize,
+) -> Url {
+    // TODO: HTTP is hardcoded
+    let url = Url::parse(&format!("http://{}", host.0)).unwrap();
```
sergiimk, line 187: Note that hard-coded HTTP will fail once we deploy on AWS.

kamu-adapter-http/tests/harness/server_side_harness.rs
```diff
@@ -136,25 +137,27 @@ struct TestAPIServer {
...
-                kamu_adapter_http::create_dataset_transfer_protocol_routes()
-                    .layer(axum::middleware::from_fn(
-                        kamu_adapter_http::resolve_dataset_by_name,
+                "/:dataset_name",
+                kamu_adapter_http::smart_transfer_protocol_routes()
+                    .layer(kamu_adapter_http::DatasetResolverLayer::new(
+                        |Path(p): Path<DatasetByName>| p.dataset_name.as_local_ref(),
                     ))
-                    .layer(axum::extract::Extension(local_dataset_resolver)),
+                    .layer(axum::extract::Extension(catalog)),
```
sergiimk, line 154: Note that I'm only passing `Catalog` now as an extension. 

This is consistent with GQL adapter. 

It allows to only construct repo instances when they are needed and gives objects back the control over their own lifetimes, so in future we could have flexibility of scoping instances per task / thread / singleton.

PR #155, opened 2023-03-29
Smart push protocol

zaychenko-sergei:


kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
```diff
@@ -318,11 +350,251 @@ pub async fn dataset_pull_ws_handler(
...
+
+            if let Err(e) = response_result_objects {
+                Err(SmartProtocolPushServerError::PushObjectResponseWriteFailed(
+                    e,
+                ))
+            } else {
+                Ok(request.is_truncated)
+            }
+        }
+        Err(e) => Err(SmartProtocolPushServerError::PushObjectRequestReadFailed(e)),
```
sergiimk, line 493: Lots of nested `match` blocks are kinda hard to read, so I would recommend using `map_err(...)?` when you only match for error handling and do early returns:

```rust
let maybe_push_objects_transfer_request =
  read_payload::<DatasetPushObjectsTransferRequest>(socket)
  .await
  .map_err(...)?;
```

kamu-adapter-http/src/http_server_middleware.rs
```diff
@@ -132,25 +140,41 @@ where
...
-                    tracing::error!("Could not get dataset: {:?}", err);
-                    return Ok(Response::builder()
-                        .status(StatusCode::INTERNAL_SERVER_ERROR)
-                        .body(Default::default())
-                        .unwrap());
-                }
-            };
+            if Self::is_allowed_to_create_dataset(&request) {
+                let dataset_builder = match dataset_repo.get_or_create_dataset(&dataset_ref).await {
+                    Ok(dsb) => Arc::new(dsb),
```
sergiimk, line 145: Does putting a value into axum's extensions requires it to be `Clone`?
It's very unusual to see `Arc<Box<T>>` - it's a shared pointer that holds a unique pointer to T.

I guess my suggestion above would avoid this altogether.

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
```diff
@@ -318,11 +350,251 @@ pub async fn dataset_pull_ws_handler(
...
+    mut socket: axum::extract::ws::WebSocket,
+    dataset_builder: Arc<Box<dyn DatasetBuilder>>,
+    dataset_url: Url,
 ) {
-    unimplemented!("Push smart protocol flow to be implemented")
+    let dataset = dataset_builder.as_dataset();
+    let push_request = match handle_push_request_initiation(&mut socket, dataset).await {
+        Ok(push_request) => push_request,
+        Err(e) => {
+            discard_dataset_building_on_error(dataset_builder.as_ref().as_ref(), e).await;
```
sergiimk, line 563: Can we have a wrapper function that calls `discard_dataset_building_on_error` based on whether inner function succeeds or fails. Inner function will be able to use `?` operator for error handling, and avoid repeating this discard call in every place an error may originate from.

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -72,6 +69,45 @@ enum SmartProtocolPullClientError {
 
 /////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Error, Debug)]
+enum SmartProtocolPushClientError {
```
sergiimk, line 73: Same comment as for the other `SmartProtocolPushServerError` - hard to imagine how clients would handle this error.

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -20,14 +20,11 @@ use url::Url;
...
-    Dataset, ErrorIntoInternal, GetRefError, InternalError, InvalidIntervalError,
-    ResultIntoInternal, SyncError, SyncListener, SyncResult,
-};
-use kamu::{domain::BlockRef, infra::utils::smart_transfer_protocol::SmartTransferProtocolClient};
+use kamu::domain::*;
+use kamu::infra::utils::smart_transfer_protocol::SmartTransferProtocolClient;
 
 use crate::{
-    dataset_protocol_helper::{dataset_import_object_file, dataset_import_pulled_metadata},
+    dataset_protocol_helper::*,
```
sergiimk, line 27: As an idea - we could change crate structure to clearly separate by protocol type and client vs server.

E.g. users would do `use kamu_adapter_http::smart_transfer_protocol::client::*` and within submodules we won't have to use verbose names like `ws_smart_transfer_protocol_tungstenite_client` and `SmartProtocolPullClientError` - protocol and client server differences would be in the "namespace" rather than types.

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -208,6 +244,186 @@ impl WsSmartTransferProtocolClient {
...
+
+        let metadata_batch = match prepare_dataset_metadata_batch(
+            src_dataset.as_metadata_chain(),
+            src_head,
+            dst_head,
+        )
+        .await
+        {
+            Ok(metadata_batch) => metadata_batch,
+            Err(e) => return Err(SmartProtocolPushClientError::Internal(e)),
```
sergiimk, line 312: Can be simplified as:
```rust
let metadata_batch = prepare_dataset_metadata_batch(
            src_dataset.as_metadata_chain(),
            src_head,
            dst_head,
        )
        .await
        .int_err()?
```

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -208,6 +244,186 @@ impl WsSmartTransferProtocolClient {
...
+        );
+
+        let write_result_metadata = write_payload(
+            socket,
+            DatasetPushMetadataRequest {
+                new_blocks: metadata_batch,
+            },
+        )
+        .await;
+        if let Err(e) = write_result_metadata {
```
sergiimk, line 328: Same here

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -332,11 +548,119 @@ impl SmartTransferProtocolClient for WsSmartTransferProtocolClient {
...
+            .get_ref(&BlockRef::Head)
+            .await
+            .int_err()?;
+
+        let size_estimation =
+            match prepare_dataset_transfer_estimaton(src.as_metadata_chain(), &src_head, dst_head)
+                .await
+            {
+                Ok(size_estimation) => size_estimation,
+                Err(e) => return Err(SyncError::Internal(e.int_err())),
```
sergiimk, line 569: same here

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -332,11 +548,119 @@ impl SmartTransferProtocolClient for WsSmartTransferProtocolClient {
...
+                tracing::debug!("Push process aborted with error: {}", e);
+                return Err(SyncError::Internal(e.int_err()));
+            }
+        };
+
+        use futures::stream::{StreamExt, TryStreamExt};
+        futures::stream::iter(push_objects_response.object_transfer_strategies)
+            .map(Ok)
+            .try_for_each_concurrent(
+                /* limit */ 4, // TODO: external configuration?
```
sergiimk, line 637: Yes, let's make it a struct field to be configured when creating the client.

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
```diff
@@ -332,11 +548,119 @@ impl SmartTransferProtocolClient for WsSmartTransferProtocolClient {
...
+                tracing::debug!("Push process aborted with error: {}", e);
+                return Err(SyncError::Internal(e.int_err()));
+            }
+        };
+
+        use tokio_tungstenite::tungstenite::protocol::{frame::coding::CloseCode, CloseFrame};
+        ws_stream
+            .close(Some(CloseFrame {
+                code: CloseCode::Normal,
+                reason: Cow::Borrowed("Client push flow succeeded"),
```
sergiimk, line 654: I think you should be able to write this as:
```rust
reason: "Client push flow succeeded".into(),
```

kamu-adapter-http/tests/tests/test_smart_push.rs
```diff
@@ -0,0 +1,256 @@
...
+use opendatafabric::*;
+
+use crate::harness::{
+    await_client_server_flow, copy_folder_recursively, ClientSideHarness, ServerSideHarness,
+};
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[test_log::test(tokio::test)]
+#[cfg_attr(feature = "skip_docker_tests", ignore)]
```
sergiimk, line 25: Just a thought: 

We probably could run the majority of these tests dockerless, using in-memory storage and having server issue proxied (simple transfer protocol) URLs for object download/upload.

Then we could have just a couple of tests that do use docker for specifically testing the delegated / non-proxied URLs with Minio.

sergiimk (changes requested):
Left some suggestions

kamu-adapter-http/src/dataset_protocol_helper.rs
```diff
@@ -240,76 +249,130 @@ async fn load_dataset_blocks(
...
+                    target_references.push(ObjectFileReference {
+                        object_type: ObjectType::Checkpoint,
+                        physical_hash: checkpoint.physical_hash.clone(),
+                        size: checkpoint.size,
+                    });
                 }
             }
-            _ => (),
         }
+        _ => (),
```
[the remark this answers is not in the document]
zaychenko-sergei: Expanded

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
```diff
@@ -318,11 +350,251 @@ pub async fn dataset_pull_ws_handler(
...
+    match maybe_push_request {
+        Ok(push_request) => {
+            tracing::debug!(
+                "Push client sent a push request: currentHead={:?} sizeEstimation={:?}",
+                push_request
+                    .current_head
+                    .as_ref()
+                    .map(|ba| ba.to_string())
+                    .ok_or("None"),
+                push_request.size_estimation
```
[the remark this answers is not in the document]
zaychenko-sergei: Renamed everywhere

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
```diff
@@ -318,11 +350,251 @@ pub async fn dataset_pull_ws_handler(
...
+            tracing::debug!(
+                "Obtained object batch with {} objects of type {:?}, media type {:?}, encoding {:?}, bytes in compressed blocks {}",
+                push_metadata_request.new_blocks.objects_count,
+                push_metadata_request.new_blocks.object_type,
+                push_metadata_request.new_blocks.media_type,
+                push_metadata_request.new_blocks.encoding,
+                push_metadata_request.new_blocks.payload.len()
+            );
+
+            assert!(
```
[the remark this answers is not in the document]
zaychenko-sergei: Agree

kamu-adapter-http/src/http_server_middleware.rs
```diff
@@ -132,25 +140,41 @@ where
...
-                }
-                Err(err) => {
-                    tracing::error!("Could not get dataset: {:?}", err);
-                    return Ok(Response::builder()
-                        .status(StatusCode::INTERNAL_SERVER_ERROR)
-                        .body(Default::default())
-                        .unwrap());
-                }
-            };
+            if Self::is_allowed_to_create_dataset(&request) {
```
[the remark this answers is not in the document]
zaychenko-sergei: I've addressed this, but differently. Optional dataset is not solving anything, I need a builder, not a dataset. So, propagating a reference instead, and creating a builder in the handler already.

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
zaychenko-sergei: Love this idiom. I've simplified a lot of code with `map_err`

kamu-adapter-http/src/http_server_middleware.rs
```diff
@@ -73,6 +73,12 @@ pub struct DatasetResolverMiddleware<Svc, IdExt, Extractor> {
     layer: DatasetResolverLayer<IdExt, Extractor>,
 }
 
+impl<Svc, IdExt, Extractor> DatasetResolverMiddleware<Svc, IdExt, Extractor> {
+    fn is_dataset_optional(request: &Request<Body>) -> bool {
+        "/push" == request.uri().path()
```
sergiimk, line 78: Will this fail to match under a different mount point, e.g. `/:dataset_id`? Or is axum giving us request URI relative to the mount?

kamu-adapter-http/src/http_server_middleware.rs
zaychenko-sergei: I think I had a weird error about Sync trait when passing the Box to lambda within the WS upgrade callback. But it seems it is solved now in this new form. Hard to tell.

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
zaychenko-sergei: Further simplified a few places

kamu-adapter-http/src/http_server_middleware.rs
zaychenko-sergei: As you can see, we don't have dataset id here as well. We'll test of course when it's time.

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
```diff
@@ -87,6 +88,42 @@ enum SmartProtocolPullServerError {
 
 /////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Error, Debug)]
+enum SmartProtocolPushServerError {
```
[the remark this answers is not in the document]
zaychenko-sergei: Implemented struct-based errors with phase, also shortened name. Ouch, this was a tough change

kamu-adapter-http/src/ws_smart_transfer_protocol_axum_server.rs
zaychenko-sergei: Done

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
zaychenko-sergei: Partially reorganized the structure of modules, and closed a lot of redundand public accesses.
The users now have a convenient type alias for the client implementation, and invoke like this:

`kamu_adapter_http::SmartTransferProtocolClientImpl`

kamu-adapter-http/src/ws_smart_transfer_protocol_tungstenite_client.rs
zaychenko-sergei: Externalized `ObjectTransferOptions` struct. By default, it is taking the value of [available parallelism](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html) on the target platform.

sergiimk (approved):
Looks ready to merge. :+1:

Once we test and stabilize this I would like to eventually:
- Move message types and main traits into `opendatafabric` - making them an official part of the spec
- Ensure messages serialization format is stable. Currently we don't actually know / check / care about what the serialized JSON looks like, because we have `serde` on both ends, but we would need to define the format and keep it stable once it's part of ODF.
  - Possibly use JSONSchema-first approach, just like with metadata
- Separate smart protocol client into its own crate

kamu-adapter-http/src/lib.rs
```diff
@@ -10,21 +10,15 @@
...
-mod http_server_simple_transfer_protocol;
-pub use http_server_simple_transfer_protocol::*;
-
-mod dataset_protocol_helper;
+mod simple_protocol;
+mod smart_protocol;
 mod ws_common;
+
+pub type SmartTransferProtocolClientImpl =
+    smart_protocol::ws_tungstenite_client::WsSmartTransferProtocolClient;
```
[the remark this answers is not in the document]
zaychenko-sergei: Ok, makes sense

kamu-adapter-http/src/smart_protocol/messages.rs
```diff
@@ -74,17 +76,30 @@ pub enum ObjectPullStrategy {
...
 pub struct DatasetPushRequestAccepted {}
 
+// Unsuccesful response to initial dataset push request
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub enum DatasetPushRequestError {
+    Internal(DatasetInternalError),
+    InvalidHead {
+        suggested_head: Option<Multihash>,
+        actual_head: Option<Multihash>,
+    },
```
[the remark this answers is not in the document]
zaychenko-sergei: Okay. Modified here, and in a similar case in pull flow.

kamu-core/src/domain/repos/metadata_chain.rs
```diff
@@ -394,6 +394,15 @@ pub struct InvalidIntervalError {
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Error, Debug)]
+#[error("Invalid head, expected {suggested_head:?}, actually {actual_head:?}")]
+pub struct InvalidHeadError {
```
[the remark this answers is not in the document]
zaychenko-sergei: Okay, I understood `RefCASError`, the only thing unclear is what is `CAS` standing for, so maybe let's avoid unclear abbreviations.

zaychenko-sergei:
> Looks ready to merge. 👍
> 
> Once we test and stabilize this I would like to eventually:
> 
> * Move message types and main traits into `opendatafabric` - making them an official part of the spec
> * Ensure messages serialization format is stable. Currently we don't actually know / check / care about what the serialized JSON looks like, because we have `serde` on both ends, but we would need to define the format and keep it stable once it's part of ODF.
>   
>   * Possibly use JSONSchema-first approach, just like with metadata
> * Separate smart protocol client into its own crate



> Looks ready to merge. 👍
> 
> Once we test and stabilize this I would like to eventually:
> 
> * Move message types and main traits into `opendatafabric` - making them an official part of the spec
> * Ensure messages serialization format is stable. Currently we don't actually know / check / care about what the serialized JSON looks like, because we have `serde` on both ends, but we would need to define the format and keep it stable once it's part of ODF.
>   
>   * Possibly use JSONSchema-first approach, just like with metadata
> * Separate smart protocol client into its own crate

Initially we assumed we will try to implement code generation from [Async API definition of the protocol](https://github.com/open-data-fabric/open-data-fabric/blob/master/protocols/smart-transfer-protocol.asyncapi.yaml). However, matching tools could not be found, so we will likely have to implement the generation script on our own.

PR #157, opened 2023-04-02
Chore/migrate to aws sdk

sergiimk:
Goodbye `rusoto`!

Most API changes are trivial. The only annoying thing is that construction of S3 Client object is async which has to spread into a bunch of functions.

### DIFF
--- a/kamu-core/src/domain/repos/object_repository.rs
+++ b/kamu-core/src/domain/repos/object_repository.rs
@@ -70,8 +70,8 @@ pub trait ObjectRepository: Send + Sync {
 
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct InsertResult {
+    /// Hash of the inserted object
     pub hash: Multihash,
-    pub already_existed: bool,
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////
--- a/kamu-core/src/infra/engine/engine_odf.rs
+++ b/kamu-core/src/infra/engine/engine_odf.rs
@@ -216,7 +216,8 @@ impl RunInfo {
         let reader = std::io::BufReader::new(file);
         for line in reader.lines() {
             let line = line?;
-            let obj = json::parse(&line).unwrap_or(json::Null);
+            let obj: serde_json::Value =
+                serde_json::from_str(&line).unwrap_or(serde_json::Value::Null);
             let process = &obj["process"];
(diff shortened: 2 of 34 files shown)


kamu-core/src/domain/repos/object_repository.rs
```diff
@@ -70,8 +70,8 @@ pub trait ObjectRepository: Send + Sync {
 
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct InsertResult {
+    /// Hash of the inserted object
     pub hash: Multihash,
-    pub already_existed: bool,
```
sergiimk, line 74: S3 does not support an atomic "insert if does not exist" operation, and I thought that checking with HEAD on every object write just to fill this flag would be too expensive (adds a whole extra round-trip).

So I decided to get rid of the flag altogether. In a situation where likelihood of large object already existing in the target repo - a higher level logic can do the pre-insert checks.

kamu-core/src/infra/engine/engine_odf.rs
```diff
@@ -216,7 +216,8 @@ impl RunInfo {
         let reader = std::io::BufReader::new(file);
         for line in reader.lines() {
             let line = line?;
-            let obj = json::parse(&line).unwrap_or(json::Null);
+            let obj: serde_json::Value =
```
sergiimk, line 219: Also got rid of unmaintained `json` crate!

kamu-core/src/infra/repos/object_repository_s3.rs
```diff
@@ -70,14 +65,12 @@ where
...
-            Ok(_) => {
-                return Ok(true);
-            }
-            // TODO: This error type doesn't work
-            // See: https://github.com/rusoto/rusoto/issues/716
-            Err(RusotoError::Service(HeadObjectError::NoSuchKey(_))) => Ok(false),
-            Err(e @ RusotoError::Credentials(_)) => Err(AccessError::Unauthorized(e.into()).into()),
-            Err(_) => Ok(false), // return Err(e.into()),
+            Ok(_) => Ok(true),
+            Err(err) => match err.into_service_error() {
```
sergiimk, line 69: this error handling style is pretty horrible, but SDK maintainers are thinking about improving it

zaychenko-sergei (approved):

PR #161, opened 2023-04-11
Smart protocol should work both with S3 and local workspace servers

zaychenko-sergei:
Unified tests for pull/push.
Added AddData generation to S3 tests, and solved an upload problem.
Fixed URL issue for simple protocol redirects.
Fixed incorrect file collectioning for PUSH flow
Trying to build PUT queries to upload objects.


kamu-adapter-http/src/simple_protocol/handlers.rs
```diff
@@ -154,6 +164,81 @@ pub async fn dataset_checkpoints_handler(
...
+
+    let temp_data_folder_suffix = format!(
+        "{}/{}/{}",
+        staging_name,
+        internal_folder_name,
+        get_staging_name()
+    );
+
+    let temp_upload_path = workspace_layout.datasets_dir.join(temp_data_folder_suffix);
+    upload_file_from_axum_to_disk(&temp_upload_path, body_stream)
```
sergiimk, line 222: Can't we stream object in via `ObjectRepository` interface, so this is generic over whatever storage implementation we chose (lfs, in-memory, proxied S3)?

kamu-adapter-http/src/smart_protocol/dataset_helper.rs
```diff
@@ -487,10 +508,17 @@ pub async fn prepare_push_object_transfer_strategy(
...
             Err(error) => match error {
-                GetTransferUrlError::NotSupported => Ok(TransferUrl {
-                    url: get_simple_transfer_protocol_url(object_file_ref, dataset_url),
-                    expires_at: None,
-                }),
+                GetTransferUrlError::NotSupported => {
+                    let mut simple_transfer_protocl_url =
+                        get_simple_transfer_protocol_url(object_file_ref, dataset_url);
+                    let staging_name_query_part = format!("staging_name={}", dataset_staging_name);
+                    simple_transfer_protocl_url.set_query(Some(staging_name_query_part.as_str()));
```
sergiimk, line 515: Why do we need to carry over staging name between this phase and the PUT request?

How I see it:
- Client does "push"
- Server sees that the data object A is missing and sends `/data/A` URL
- Client sends `PUT /data/A` and streams the body
- Server (via `ObjectRepositoryFs`) streams into a temporary staging file, and only when stream completes and hash sum matches does atomic move `staging-xyz -> A`

sergiimk (commented):
Love the new test structure, but let's discuss the staging approach more.

kamu-adapter-http/src/smart_protocol/dataset_helper.rs
sergiimk: Discussed on Slack: key problem is that from perspective of PUT reuqest the target dataset does not yet exist as `DatasetBulder` is not finalized in another thread/process yet.

kamu-adapter-http/src/simple_protocol/handlers.rs
zaychenko-sergei: We can't. We do not have any way to obtain the correct `ObjectRepository` in the 2ndary call that has no access to the `DatasetBuilder`

kamu-adapter-http/src/smart_protocol/dataset_helper.rs
zaychenko-sergei: Resolved

kamu-adapter-http/src/simple_protocol/handlers.rs
zaychenko-sergei: Resolved

PR #164, opened 2023-05-01
Improved recovery of smart protocol operations in case partial file uploads/downloads succeeded earlier

zaychenko-sergei:
Corrected aborted push writes / pull reads recovery for new & updated datasets in the smart protocol.
Extended unit test coverage for smart protocol mimicing related errors.
Patch release + minor dependency updates.


kamu-adapter-http/src/smart_protocol/dataset_helper.rs
```diff
@@ -597,6 +597,14 @@ pub async fn dataset_export_object_file(
     dataset: &dyn Dataset,
     object_transfer_strategy: &PushObjectTransferStrategy,
 ) -> Result<(), SyncError> {
+    if object_transfer_strategy.push_strategy == ObjectPushStrategy::SkipUpload {
+        tracing::debug!(
+            "Skipping upload of {:?} '{}'",
```
sergiimk, line 602: Minor: Rather than formatting everything into a message consider named arguments. This makes logs more structured and will simplify searching for logs once we have ElasticSearch or a similar log aggregator.

Example: https://github.com/kamu-data/kamu-cli/blob/e77c330727b8a6af3621620066f816ebb626ce1d/kamu-core/src/infra/repos/dataset_factory_impl.rs#L140

sergiimk (approved):

kamu-adapter-http/src/smart_protocol/dataset_helper.rs
zaychenko-sergei: Restructured this idiomatic way where it felt applicable

PR #167, opened 2023-05-11
Fix tracing of async functions

sergiimk:
In this PR:
* I removed the use of explicit `Span` creation in favor of `#[instrument(...)]` macro that works correctly with `Future`s (suspend/resume)
* I unified the use of tracing to always use qualified path e.g. (`tracing::debug!(...)` instead of `use tracing; debug!(...)`) so tracing code was easier to find/replace


sergiimk:
Note: We might need a similar change in API server.

zaychenko-sergei (approved):

PR #171, opened 2023-05-14
Move HTTP ingest to reqwest and make FTP optional

sergiimk:
I eliminated the use of `curl` for anything but FTP access. I was on the edge of dropping FTP support entirely, but for now kept it as an optional feature.

This feature will only be enabled for testing in Linux build and in release, so it should speed up dev builds.

Not linking `curl` and `openssl` in Windows build might even stop if from failing - this is to be seen.


kamu-core/Cargo.toml
```diff
@@ -34,8 +36,8 @@ serde_yaml = "0.9"
 # Ingest
 # TODO: Using curl brings a lot of overhead including compiling and linking openssl
 # We should replace it with reqwest + a separate FTP client or drop FTP support in favor of container-based ingest.
-curl = { version = "0.4", features = ["http2", "protocol-ftp", "static-curl", "static-ssl"] }
-curl-sys = "0.4"
+curl = { optional = true, version = "0.4", features = ["protocol-ftp", "static-curl", "static-ssl"] }
```
sergiimk, line 39: Some ideas for future of FTP support:
- We could link to curl dynamically - i.e. you get FTP support only if you have `libcurl` on your system
  - This needs good testing to ensure `kamu` doesn't just stop working if `libcurl` is not present
- We could try native rust FTP clients, but last time I looked I didn't find anything mature enough
- We could make people who need FTP use a containerized ingest, and even provide a pre-built image

PR #175, opened 2023-05-16
Resilience against smart protocol push timeout

zaychenko-sergei:
Implemented an upload progress exchange extension in the smart push flow.


kamu-adapter-http/src/smart_protocol/messages.rs
```diff
@@ -147,7 +147,24 @@ pub enum ObjectPushStrategy {
     HttpUpload,
 }
 
-/// Push stage 3: complete handshake indication
+/// Push stage 3: object upload progress
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+pub struct DatasetPushObjectsUploadInProgress {
```
sergiimk, line 152: I don't really like using this as PING substitute.
- On a poor connection a single large enough object might upload slow enough for connection to be cut even before there's any change in progress to report;
- I highly doubt that any more "sophisticated" payload will have higher chances to prevent idle connection drop than sending a single byte of data like ping.

How I imagine ideal keep-alive is:
- Server may at any point send PING message to the client
- Client has to reply with PONG as soon as possible

This way:
- Server is in charge of ping frequency (it can be adjusted centrally to always be below idle timeout of an LB)
- It's a much simpler rule to follow for client-side protocol implementors - it doesn't complicate the state machine as much as upload progress does.

kamu-adapter-http/src/smart_protocol/ws_axum_server.rs
```diff
@@ -384,8 +384,10 @@ async fn try_handle_push_objects_request(
             PushServerError::ReadFailed(PushReadError::new(e, PushPhase::ObjectsRequest))
         })?;
 
+    let objects_count = request.object_files.len();
+
     tracing::debug!(
-        objects_count = % request.object_files.len(),
+        objects_count = % objects_count,
```
sergiimk, line 390: nitpick: can be simplified to just `%object_count`

kamu-adapter-http/src/smart_protocol/messages.rs
sergiimk: I see now that you have `min_upload_progress_delay_sec` that would solve "single large object" problem, but I still like the idea of server being in charge of the ping frequency better.

kamu-adapter-http/src/smart_protocol/ws_tungstenite_client.rs
```diff
@@ -320,6 +322,110 @@ impl WsSmartTransferProtocolClient {
...
 
+    async fn export_group_of_object_files(
+        &self,
+        socket: &mut TungsteniteStream,
+        push_objects_response: DatasetPushObjectsTransferResponse,
+        src: Arc<dyn Dataset>,
+        transfer_options: ObjectTransferOptions,
+    ) -> Result<(), SyncError> {
+        let total_files_count = push_objects_response.object_transfer_strategies.len();
+        let uploaded_files_counter = Arc::new(AtomicI32::new(0));
```
sergiimk, line 333: Nice use of atomics!

kamu-adapter-http/src/smart_protocol/ws_tungstenite_client.rs
```diff
@@ -320,6 +322,110 @@ impl WsSmartTransferProtocolClient {
...
+        let cloned_counters: Vec<_> = iter::repeat_with(|| uploaded_files_counter.clone())
+            .take(total_files_count)
+            .collect();
```
sergiimk, line 336: Took me a while to understand what `cloned_counters` is...

Can this initialization and `zip` later on be replaced with something like:
```rust
let task_data: Vec<_> = push_objects_response.object_transfer_strategies
  .into_iter(|s| (s, uploaded_files_counter.clone())
  .collect();

futures::stream::iter(task_data)
  .try_for_each_concurrent(|s, c| ...)
  .await
```

kamu-adapter-http/src/smart_protocol/ws_tungstenite_client.rs
```diff
@@ -320,6 +322,110 @@ impl WsSmartTransferProtocolClient {
...
+                },
+            )
+            .await
+        });
+
+        while !export_task.is_finished() {
+            tokio::time::sleep(std::time::Duration::from_secs(
+                transfer_options.min_upload_progress_delay_sec,
+            ))
+            .await;
```
sergiimk, line 361: A better approach that lets the process terminate faster could be something like:
```rust
loop {
  tokio::select! {
    _ = time::sleep(timeout) => ...,
    export_task => ...
  }
}
```

There is also [tokio::time::interval](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) that might have less overhead than creating `sleep` task every time.

sergiimk (changes requested):
Suggested a different approach, lmk what you think.

kamu-adapter-http/src/smart_protocol/ws_tungstenite_client.rs
zaychenko-sergei: I agree, same sense, but slightly cleaner.

kamu-adapter-http/src/smart_protocol/messages.rs
zaychenko-sergei: Addressed

PR #176, opened 2023-05-22
Feature/Async container runtime

sergiimk:
In this PR:
- Major improvements to `container-runtime` crate, including fully `async` interface
- Unifying container process management and graceful shutdown procedure
- This removes a lot of tech debt like signal handling, lots of `tokio::spawn_blocking()` etc.
- Also sneaked in a few improvements:
  - Converted most of `FetchService` to async (only FTP is left)
  - Added progress reporting to containerized fetch
  - Fixed issue with progress bars / spinners not refreshing during pull/push


utils/container-runtime/src/container.rs
```diff
@@ -0,0 +1,443 @@
...
+        if let Some(_) = self.child.id() {
+            tracing::warn!(
+                container_name = %self.container_name,
+                "Container was not terminated - cleaning up synchronously"
+            );
+
+            let _ = self.child.terminate_blocking(self.terminate_timeout);
+        }
+    }
+}
```
sergiimk, line 443: One of key places in PR. We are hoping that use will call `container.terminate().await?` for graceful shutdown, but if they forget to do so we will still clean up resources but in a blocking way.

This may stall the event loop in some circumstances, but will prevent leaking resources.

Hopefully this limitation will be removed once [AsyncDrop](https://rust-lang.github.io/async-fundamentals-initiative/roadmap/async_drop.html) problem is addressed in the language.

kamu-cli/src/commands/pull_command.rs
```diff
@@ -355,30 +344,15 @@ impl Command for PullCommand {
...
-            if self.finished.load(Ordering::SeqCst) {
-                break;
-            }
-            std::thread::sleep(std::time::Duration::from_millis(100));
-        }
-    }
-
-    fn finish(&self) {
-        self.finished.store(true, Ordering::SeqCst);
-    }
```
sergiimk, line 381: Looks like after some major upgrade of `indicatif` this code became meaningless and actually didn't refresh the progress bars. This is why I noticed that our UI became less "responsive". 

I replaced this with `enable_steady_tick()` calls.

zaychenko-sergei (approved):

PR #184, opened 2023-05-26
Features/Datafusion engine

sergiimk:


opendatafabric/src/dtos/dtos_extra.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct DatasetVocabularyResolved<'a> {
+    /// Name of the system time column.
+    pub system_time_column: Cow<'a, str>,
+    /// Name of the event time column.
+    pub event_time_column: Cow<'a, str>,
+    /// Name of the offset column.
+    pub offset_column: Cow<'a, str>,
+}
+
+pub type DatasetVocabularyResolvedOwned = DatasetVocabularyResolved<'static>;
```
sergiimk, line 64: This is a really cool trick I picked up from `datafusion` - it allows to have a single type that can contain both borrowed references with limited lifetime and owned objects (with `'static`) lifetime.

utils/container-runtime/src/runtime.rs
```diff
@@ -316,7 +316,7 @@ impl ContainerRuntime {
         }
     }
 
-    #[tracing::instrument(level = "debug", skip_all, fields(container_name))]
+    #[tracing::instrument(level = "debug", skip_all, fields(%container_name))]
```
sergiimk, line 319: No idea why, but without `%` these values are not showing up in traces ... this is a really unexpected behavior and I will look into it some day.

utils/container-runtime/tests/tests/test_setup.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use container_runtime::ContainerRuntime;
+
+use super::test_container::TEST_IMAGE;
+
+// Not really a test - used by CI to separate pulling of test images
+// into its own phase
+#[cfg_attr(feature = "skip_docker_tests", ignore)]
+#[test_log::test(tokio::test)]
+async fn test_setup_pull_images() {
```
sergiimk, line 18: This *might* help to stop my container test from failing - it appears to be some race condition in signal handling in `podman` itself that allows the controlling attached process of `podman` to exit while container is still running. Last time I saw image pull happening in parallel (by another test) so I want to make sure that we pull all images before test execution starts, as a separate CI step.

kamu-cli/src/app.rs
```diff
@@ -219,6 +219,16 @@ fn load_config(workspace_layout: &WorkspaceLayout, catalog: &mut CatalogBuilder)
...
+        datafusion_image: config
+            .engine
+            .as_ref()
+            .unwrap()
+            .images
+            .as_ref()
+            .unwrap()
+            .datafusion
+            .clone()
+            .unwrap(),
```
sergiimk, line 231: Yeah... we really should improve our config service to avoid this crazy chain. Need to do a survey of config crates to see if there's something suitable.

PR #185, opened 2023-05-26
Hotfix: AWS S3 object store needs explicit credentials

zaychenko-sergei:


kamu-core/Cargo.toml
```diff
@@ -56,6 +56,7 @@ sha3 = "0.10"
 aws-config = { version = "0.55" }
 aws-sdk-s3 = { version = "0.28" }
 aws-smithy-http = { version = "0.55", features = ["rt-tokio"] }
+aws-credential-types = "*"
```
sergiimk, line 59: Please specify a version - I'm making `cargo deny` rules more strict to reject the wildcard dependencies.

kamu-core/tests/tests/test_query_service_impl.rs
```diff
@@ -138,8 +140,8 @@ async fn create_catalog_with_s3_workspace(s3: &S3) -> dill::Catalog {
         .add_value(infra::ObjectStoreBuilderLocalFs::new())
         .bind::<dyn ObjectStoreBuilder, infra::ObjectStoreBuilderLocalFs>()
         .add_value(infra::ObjectStoreBuilderS3::new(
-            s3_context.bucket,
-            endpoint.unwrap(),
+            s3_context,
+            s3_credentials,
```
sergiimk, line 144: Do we need to pass credentials separately if they can already be obtained from `S3Context`?

sergiimk (approved):

kamu-core/Cargo.toml
zaychenko-sergei: Done

kamu-core/tests/tests/test_query_service_impl.rs
zaychenko-sergei: Yes, because it's async operation, and the context around is synchronous

PR #192, opened 2023-06-11
Bug/tracing grapqhl data queries

zaychenko-sergei:
Added more tracing instrumentation to catch QueryService issues on the platform


src/adapter/graphql/src/queries/datasets/dataset.rs
```diff
@@ -53,23 +53,27 @@ impl Dataset {
     }
 
     /// Unique identifier of the dataset
+    #[tracing::instrument(level = "info", skip_all)]
```
sergiimk, line 56: These will generate per-property enter/exit events at `INFO` level ... this looks super spammy to me.
What's the purpose of these logs if we don't log neither parameters nor return values?

Would enabling debug logging on GQL library achieve better effect?

Would really want these to be reverted once debugging is done.

zaychenko-sergei: For now I've added a trace around any GraphQL public entry, as we have no clue what is failing

PR #193, opened 2023-06-11
Feature/Task system

sergiimk:
In this PR:
- I extracted `enum-variants` crate for convenient type conversions
- Added new `event-sourcing` crate that defines core types and interfaces for aggregates and event stores
- Implemented `Task` aggregate and separated `TaskService` into `TaskScheduler` (maintains the queue, controls concurrency, resources etc.) and `TaskExecutor` (pulls tasks from the queue and runs them

You will also see a lot of changes related to removing `?Send` from service interfaces. When I started using `async_trait` I barely knew what I was doing and using `?Send` allowed me to make compiler happy. But we actually want our core services to efficiently use all cores and this means having tasks migrate from one thread to another (i.e. futures need to be `Send`), thus I refactored those old interfaces accordingly.


src/adapter/graphql/tests/tests/test_tasks.rs
```diff
@@ -8,15 +8,35 @@
...
+    impl TaskScheduler for TaskScheduler {
+        async fn get_task(&self, task_id: &TaskID) -> Result<TaskState, GetTaskError>;
+        fn list_tasks_by_dataset<'a>(&'a self, dataset_id: &DatasetID) -> TaskStateStream<'a>;
+        async fn create_task(&self, plan: LogicalPlan) -> Result<TaskState, CreateTaskError>;
+        async fn cancel_task(&self, task_id: &TaskID) -> Result<TaskState, CancelTaskError>;
+        async fn take(&self) -> Result<TaskID, TakeTaskError>;
+        async fn try_take(&self) -> Result<Option<TaskID>, TakeTaskError>;
+    }
+}
+
```
sergiimk, line 27: @zaychenko-sergei Note that I started using `mockall` library here. It allows be to test GQL interface without linking any concrete implementations - not even "in-memory". I think we should eventually move to this model in all adapters, so that we are not linking them to impl crates even in test builds.

src/app/cli/src/explore/api_server.rs
```diff
@@ -63,15 +69,21 @@ impl APIServer {
...
         self.server.local_addr()
     }
 
-    pub async fn run(self) -> Result<(), hyper::Error> {
-        self.server.await
+    pub async fn run(self) -> Result<(), InternalError> {
+        tokio::select! {
+            res = self.server => { res.int_err() },
+            res = self.task_executor.run() => { res.int_err() },
+        }
```
sergiimk, line 86: Notice: `kamu system api-server` now spawns an HTTP server and a task executor loop.

src/app/cli/tests/tests/test_di_graph.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[test_log::test(tokio::test)]
+async fn test_di_graph_validates() {
+    let mut catalog_builder = kamu_cli::configure_catalog();
+    kamu_cli::register_config_in_catalog(&kamu_cli::CLIConfig::default(), &mut catalog_builder);
+
+    // TODO: We should ensure this test covers parameters requested by commands and
+    // types needed for GQL/HTTP adapter that are currently being constructed
+    // manually
+    let validate_result = catalog_builder.validate().ignore::<WorkspaceLayout>();
```
sergiimk, line 22: Notice: `dill` now has a feature that will statically validate the dependency graph to avoid most runtime errors.

Because GraphQL and HTTP adapters are using `Catalog` manually to create objects - they are still exposed to runtime errors. We will need to find a way to create a `dill-axum` and `dill-async-graphql` adapters to be able to statically analyze the dependencies of their endpoints and include them into this validation.

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+use crate::entities::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug)]
+pub struct Task {
+    state: TaskState,
+    pending_events: Vec<TaskEvent>,
+    last_synced_event: Option<EventID>,
+}
```
sergiimk, line 23: Our first event-sourced aggregate

src/domain/task-system/src/entities/logical_plan.rs
```diff
@@ -7,26 +7,57 @@
...
 /// Represents logical steps needed to carry out a task
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum LogicalPlan {
     /// Perform an update on a dataset like update from polling source or a
     /// derivative transform
     UpdateDataset(UpdateDataset),
+    /// A task that can be used for testing the scheduling system
+    Probe(Probe),
+}
```
sergiimk, line 25: These are the only supported task types for now.

src/domain/task-system/src/entities/task_event.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// New task entered the queue
+    Created(TaskCreated),
+    /// Task execution had started
+    Running(TaskRunning),
+    /// Cancellation of task was requested (this is not immediate and task may
+    /// still finish with a different outcome than cancelled)
+    Cancelled(TaskCancelled),
+    /// Task has reached a final outcome
+    Finished(TaskFinished),
+}
```
sergiimk, line 29: Domain events underpinning the `Task` aggregate.

src/domain/task-system/src/entities/task_state.rs
```diff
@@ -7,16 +7,28 @@
...
 pub struct TaskState {
+    /// Unique and stable identitfier of this task
     pub task_id: TaskID,
+    /// Life-cycle status of a task
     pub status: TaskStatus,
+    /// Whether the task was ordered to be cancelled
+    pub cancellation_requested: bool,
+    /// Execution plan of the task
     pub logical_plan: LogicalPlan,
-}
```
sergiimk, line 20: A state / projection of task events maintained by `Task` aggregate

src/domain/task-system/src/services/task_scheduler.rs
```diff
@@ -16,16 +17,26 @@ use crate::*;
...
+
+    /// Attempts to cancel the given task
+    async fn cancel_task(&self, task_id: &TaskID) -> Result<TaskState, CancelTaskError>;
+
+    /// Blocks until the next task is available for execution and takes it out
+    /// of the queue (called by [TaskExecutor])
+    async fn take(&self) -> Result<TaskID, TakeTaskError>;
+
+    /// A non-blocking version of [TaskScheduler::take()]
+    async fn try_take(&self) -> Result<Option<TaskID>, TakeTaskError>;
```
sergiimk, line 39: Scheduler will maintain queues of tasks and decide when a task can run based on capacity. The executor(s) will be stealing tasks from the queues through these methods.

src/utils/event-sourcing/src/aggregate.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use super::errors::*;
+use crate::EventID;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+/// Common set of operations for an event-sourced aggregate
+#[async_trait::async_trait]
+#[allow(drop_bounds)]
+pub trait Aggregate
```
sergiimk, line 18: The heart of event sourcing - I'm using this trait to enforce same structure on all aggregates.

There exist multiple event-sourcing libraries for Rust, but I decided to start with my own implementation to get a feel for how it can look like in Rust and understand key requirements.

I will then use experience from writing this to evaluate these libraries to hopefully switch to something off-the-shelf.

resources/schema.gql
```diff
@@ -790,12 +826,29 @@ type Tasks {
...
+type TasksMutations {
+	"""
+	Requests cancellation of the specified task
+	"""
+	cancelTask(taskId: TaskID!): Task!
+	"""
+	Schedules a task to update the specified dataset by performing polling
+	ingest or a derivative transformation
+	"""
+	crateUpdateDatasetTask(datasetId: DatasetID!): Task!
```
zaychenko-sergei, line 846: create ?

resources/schema.gql
```diff
@@ -790,12 +826,29 @@ type Tasks {
...
+	"""
+	Schedules a task to update the specified dataset by performing polling
+	ingest or a derivative transformation
+	"""
+	crateProbeTask(datasetId: DatasetID, busyTimeMs: Int, endWithOutcome: TaskOutcome): Task!
```
zaychenko-sergei, line 851: same here

src/adapter/graphql/tests/tests/test_tasks.rs
zaychenko-sergei: I see, pretty typical mocking approach. Asserting expected interactions.

src/app/cli/src/commands/pull_command.rs
```diff
@@ -165,35 +162,32 @@ impl PullCommand {
...
-                .into_iter(),
-                PullOptions {
+                }],
+                PullMultiOptions {
                     ingest_options: IngestOptions {
                         fetch_uncacheable: self.fetch_uncacheable,
                         exhaust_sources: true,
                     },
-                    ..PullOptions::default()
+                    ..Default::default()
```
zaychenko-sergei, line 175: Similar dilemma

src/domain/task-system/src/entities/logical_plan.rs
```diff
@@ -7,26 +7,57 @@
...
 
 /// Represents logical steps needed to carry out a task
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum LogicalPlan {
     /// Perform an update on a dataset like update from polling source or a
     /// derivative transform
     UpdateDataset(UpdateDataset),
+    /// A task that can be used for testing the scheduling system
+    Probe(Probe),
```
zaychenko-sergei, line 24: Do you mean health checks?

src/domain/task-system/src/entities/task_event.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TaskCreated {
+    pub event_time: DateTime<Utc>,
+    pub task_id: TaskID,
```
zaychenko-sergei, line 36: These 2 fields repeat in each event.
I understand that Rust has no inheritance for structs, but maybe we can have a different design:

```
struct TaskEvent {
    pub event_time: DateTime<Utc>,
    pub task_id: TaskID,
    pub body: TaskEventBody,
}

pub enum TaskEventBody {
    /// New task entered the queue
    Created(TaskCreated),
    /// Task execution had started
    Running(TaskRunning),
    /// Cancellation of task was requested (this is not immediate and task may
    /// still finish with a different outcome than cancelled)
    Cancelled(TaskCancelled),
    /// Task has reached a final outcome
    Finished(TaskFinished),
}

pub struct TaskCreated {
   pub logical_plan: LogicalPlan,
}

pub struct TaskRunning {}

pub struct TaskCancelled {}

pub struct TaskFinished {
   pub outcome: TaskOutcome,
}


```

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+                status: TaskStatus::Queued,
+                cancellation_requested: false,
+                logical_plan: genesis.logical_plan.clone(),
+                created_at: genesis.event_time.clone(),
+                ran_at: None,
+                cancellation_requested_at: None,
+                finished_at: None,
+            },
+            pending_events: vec![genesis.into()],
+            last_synced_event: None,
```
zaychenko-sergei, line 47: Isn't first "TaskCreated"already synchronized to the state ?

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+    /// Transition task to a `Running` state
+    pub fn run(&mut self) -> Result<(), IllegalSequenceError<Self>> {
+        self.apply(TaskRunning {
```
zaychenko-sergei, line 53: No state check? I guess it should not be running or finished to allow `run`

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        self.apply(TaskCancelled {
+            event_time: Utc::now(),
+            task_id: self.task_id,
+        })
+    }
+
+    /// Transition task to a `Finished` state with the specified outcome
+    pub fn finish(&mut self, outcome: TaskOutcome) -> Result<(), IllegalSequenceError<Self>> {
+        self.apply(TaskFinished {
```
zaychenko-sergei, line 82: Similar, no state check?

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+            TaskStatus::Running if !self.cancellation_requested => true,
+            _ => false,
+        }
+    }
+
+    /// Set cancellation flag (if not already set)
+    pub fn cancel(&mut self) -> Result<(), IllegalSequenceError<Self>> {
+        if self.cancellation_requested {
+            return Ok(());
+        }
```
zaychenko-sergei, line 72: Why is this check needed? I guess if we have multiple cancellations in the queue of events, well, the 2nd one should just silently ignore

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+    fn update_state(&mut self, event: TaskEvent) -> Result<(), IllegalSequenceError<Self>> {
+        assert_eq!(self.state.task_id, event.task_id());
+
+        // Check if state transition is legal
+        match (self.state.status, &event) {
+            (TaskStatus::Queued, TaskEvent::Running(_)) => {}
+            (TaskStatus::Queued | TaskStatus::Running, TaskEvent::Cancelled(_)) => {}
+            (TaskStatus::Queued | TaskStatus::Running, TaskEvent::Finished(_)) => {}
+            (_, _) => return Err(IllegalSequenceError::new(self.state.clone(), event).into()),
+        }
```
zaychenko-sergei, line 105: Are Queued->Queued, Running->Running, Finished->Finished, Cancelled->Cancelled possible?
Probably these should be just ignored to satisfy idempotence property.

src/domain/task-system/src/aggregates/task.rs
```diff
@@ -0,0 +1,276 @@
...
+        &self.task_id
+    }
+
+    fn from_genesis_event(
+        event_id: EventID,
+        event: TaskEvent,
+    ) -> Result<Self, IllegalGenesisError<Self>> {
+        if !event.is_variant::<TaskCreated>() {
+            return Err(IllegalGenesisError { event });
+        }
```
zaychenko-sergei, line 155: Why is this a dynamic check instead of better static type?

resources/schema.gql
sergiimk: :man_facepalming:

src/adapter/http/tests/harness/client_side_harness.rs
```diff
@@ -101,17 +101,15 @@ impl ClientSideHarness {
...
-                &mut vec![dataset_ref].into_iter(),
-                PullOptions {
+                vec![dataset_ref],
+                PullMultiOptions {
                     sync_options: SyncOptions {
                         create_if_not_exists: true,
                         ..SyncOptions::default()
                     },
-                    ..PullOptions::default()
+                    ..Default::default()
```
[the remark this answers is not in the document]
sergiimk: Both are equivalent (accessing same method through type name vs through the trait) so I sometimes write `Type::default()` to be explicit, but the next day would write `Default::default()` to avoid name repetition. I'll revert back to type name, but yeah we need some convention :)

src/app/cli/src/app.rs
```diff
@@ -183,19 +184,32 @@ pub fn configure_catalog() -> CatalogBuilder {
     b.add::<kamu_adapter_http::SmartTransferProtocolClientWs>();
     b.bind::<dyn SmartTransferProtocolClient, kamu_adapter_http::SmartTransferProtocolClientWs>();
 
+    b.add::<kamu_task_system_inmem::TaskSchedulerInMemory>();
```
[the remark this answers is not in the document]
sergiimk: Yes, I will integrate and test the task schedule in `api-server` right after I merge this PR

src/adapter/http/tests/harness/client_side_harness.rs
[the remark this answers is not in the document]
sergiimk: The convention I would propose is:

Use explicit type name when passing default parameters to functions, e.g.:

```rust
pull_svc.pull(dataset, PullOptions::default())
```
This ensures that when method's second argument is refactored to a different type - client code will need to be updated and reconsidered (as opposed to passing `Default::default()` that might just work for new parameter type too.

Use `Default::default()` in struct initializers, e.g.:

```rust
pull_svc.pull(dataset, PullOptions {
  create_if_exists: true,
  ..Default::default()
})
```
This is to avoid type name repetition == less diff during renaming refactorings.

src/domain/task-system/src/entities/task_event.rs
sergiimk: Yeah, this is how I started - with `task_id` and `event_time` in a separate type ... but then moved to the current design.

The reasoning is that this enum represents all events of a "Task Scheduling" domain, and there might be events that are in this domain but not directly linked to a task. 

For example in a distributed system we might want to track availability of task executors and might have `ExecutorOnline` `ExecutorOffline` events that only have `executor_id`. When executor commits to a task such event might have both `executor_id` AND `task_id`.

This series of blog posts (while a bit annoying in its style) explains this approach: https://sara.event-thinking.io/2023/04/kill-aggregate-chapter-1-I-am-here-to-kill-the-aggregate.html

sergiimk: Here's a rust framework that specifically builds on the "Kill the Aggregate" idea of this blog post series https://github.com/disintegrate-es/disintegrate (one of the libraries we can evaluate later).

The idea of "events are just facts" and what we call "aggregates" are just projections reconstructed from different subsets of domain event streams really resonates with me and how I was already thinking about projections in Kamu.

src/domain/task-system/src/aggregates/task.rs
sergiimk: `last_synced_event` tracks last event that was persisted in the EventStore. Here we are creating a completely new task, so the `TaskCreated` should be considered a pending update.

src/domain/task-system/src/aggregates/task.rs
sergiimk: While it's legal to cancel already cancelled task I though it's better to save some work of writing a redundant event.

src/domain/task-system/src/aggregates/task.rs
sergiimk: `Queued->Queued` is impossible as this is the initial state.

For the rest ... not sure. I think I would rather be strict and not allow them in the aggregate.

There is a separate question of conflict resolution, e.g. if two users concurrently cancel the same task should we silently deduplicate the event on save? My instincts is that this adds too much complexity, and I would rather have optimistic concurrency errors and retry the transaction a few times than have an overly smart deduplication logic at the event store level.

src/domain/task-system/src/aggregates/task.rs
sergiimk: I guess I could express this function as:
```rust
fn from_genesis_event<E>(
        event_id: EventID,
        event: E,
    ) -> Result<Self, IllegalGenesisError<Self>>
    where
        E: VariantOf<Self::Event>;
```

This would allow `Task::from_genesis_event` accept only `TaskCreated`.

BUT

This also limits aggregate (projection) to only initialize from a single type of event. I feel like in some cases it may be useful to allow multiple (or even all) event types to be able to define an initial state of projection.

src/infra/task-system-inmem/src/task_event_store_inmem.rs
```diff
@@ -0,0 +1,162 @@
...
+#[derive(Default)]
+struct State {
+    events: Vec<TaskEvent>,
+    tasks_by_dataset: HashMap<DatasetID, Vec<TaskID>>,
+    last_task_id: Option<TaskID>,
+}
+
+impl State {
+    fn next_task_id(&mut self) -> TaskID {
+        let new_task_id = if let Some(last_task_id) = self.last_task_id {
```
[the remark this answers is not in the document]
sergiimk: This accepts `&mut self` - i.e. an exclusive non-shared reference. This forces the rest of the code to use `State::next_task_id` only when holding a lock.

src/domain/task-system/src/entities/task_event.rs
zaychenko-sergei: Nobody cancelled DRY principle. 
In a OOP language I would have had:
 - Event (with just event time)
    - ExecutorEvent (with executor_id)
       - subclasses
    - TaskEvent (with task_id)
       - subclasses

It's weird that we cannot express such a simple typical thing easily with Rust.

src/infra/task-system-inmem/src/task_event_store_inmem.rs
```diff
@@ -0,0 +1,162 @@
...
+use opendatafabric::DatasetID;
+
+pub struct TaskEventStoreInMemory {
+    state: Arc<Mutex<State>>,
+}
+
+#[derive(Default)]
+struct State {
+    events: Vec<TaskEvent>,
+    tasks_by_dataset: HashMap<DatasetID, Vec<TaskID>>,
```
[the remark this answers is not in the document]
sergiimk: Not all events will have an associated DatasetID, so I decided to keep a single `Vec<TaskEvent>` as the core of event store. Everything else is optimization - when implementing a DB-backed event store can chose whether we use an association table `[DatasetID, TaskID]` or de-normalize the events etc.

src/infra/task-system-inmem/src/task_executor_inmem.rs
```diff
@@ -0,0 +1,89 @@
...
+                    dataset_id: _,
+                    busy_time,
+                    end_with_outcome,
+                }) => {
+                    if let Some(busy_time) = busy_time {
+                        tokio::time::sleep(busy_time.clone()).await;
+                    }
+                    end_with_outcome.unwrap_or(TaskOutcome::Success)
+                }
+            };
```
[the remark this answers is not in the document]
sergiimk: In future I will likely split it into different executor types that can only process certain kinds of events. This was not essential for Phase 1 - my goal here is to provide a sane GQL API on top of which task Web UI can be prototyped in parallel while we refine the task system.

src/app/cli/tests/tests/test_di_graph.rs
sergiimk: For every registered `Builder` it will check that all of its dependencies can be satisfied. See: https://github.com/sergiimk/dill-rs/blob/master/dill/tests/tests/test_validation.rs

src/utils/event-sourcing/src/aggregate.rs
sergiimk: Yes you are right. I was going to try to refactor this into `Aggregate<Projection>` where `Projection` specifies the state transition logic and `Aggregate` type implements all the event handling. I'd prefer to do this in a separate PR though, after I finish API server integration. Don't want to be stuck perfecting the event sourcing "framework" without progressing to the actual goal which is to have dataset updates running in the platform backend.

PR #196, opened 2023-06-15
Add knownEngines GQL api

sergiimk:
New `knownEngines` GQL API for displaying a list of recommended engines in Web UI

Example:
image


resources/schema.gql
```diff
@@ -578,7 +603,9 @@ type Query {
 }
 
 enum QueryDialect {
-	DATA_FUSION
+	SQL_SPARK
+	SQL_FLINK
+	SQL_DATA_FUSION
```
sergiimk, line 608: Note a small breaking change here. @dmitriy-borzenko will need to update the `{ data { query(queryDialect: ...) {} } }` parameter accordingly.

Motivation is to be start preparing for non-SQL dialects, e.g. `SPARQL`.

zaychenko-sergei (approved):

PR #199, opened 2023-06-26
Feature/multitenant workspace

zaychenko-sergei:


src/domain/core/src/repos/dataset_repository.rs
```diff
@@ -433,6 +461,8 @@ pub enum GetDatasetError {
     #[error(transparent)]
     NotFound(#[from] DatasetNotFoundError),
     #[error(transparent)]
+    MultiTenantRefUnexpected(#[from] MultiTenantRefUnexpectedError),
```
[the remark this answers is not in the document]
sergiimk: Looking further in the PR I feel even strongly about this - `MultiTenantRefUnexpected` ended up being passed through in almost every other error type, but I really don't think it deserves the same "fundamental error class" status as `Unauthorized` or `Internal`.

src/infra/core/src/pull_service_impl.rs
```diff
@@ -129,9 +132,14 @@ impl PullServiceImpl {
                     DatasetRefRemote::Alias(alias)
                     | DatasetRefRemote::Handle(DatasetHandleRemote { alias, .. }),
                 ) => DatasetAlias::new(alias.account_name.clone(), alias.dataset_name.clone()),
-                Some(DatasetRefRemote::Url(url)) => {
-                    DatasetAlias::new(None, self.infer_local_name_from_url(url)?)
-                }
+                Some(DatasetRefRemote::Url(url)) => DatasetAlias::new(
+                    if self.dataset_repo.is_multi_tenant() {
```
sergiimk, line 136: Do we really need to branch here based on the "tenancy"? I thought the single-tenant repo will function by using certain hard-coded account name (e.g. `kamu`) for the current user. So that all service code can always work in a multi-tenant mode but this account name will simply be ignored when forming a dataset fs path.

It would be nice to have all services just assume we are always multi-tenant and not make any special cases.

src/infra/core/src/pull_service_impl.rs
```diff
@@ -257,9 +265,14 @@ impl PullServiceImpl {
             }
         }
 
-        // No luck - now have to search through aliases
+        // No luck - now have to search through aliases (of current user)
         use tokio_stream::StreamExt;
-        let mut datasets = self.local_repo.get_all_datasets();
+        let mut datasets = if self.dataset_repo.is_multi_tenant() {
```
sergiimk, line 270: Same here, if `current_account.account_name` is `kamu` in single-tenant mode then two branches should return same result?

src/infra/core/src/pull_service_impl.rs
```diff
@@ -431,6 +444,16 @@ impl PullServiceImpl {
...
             .collect())
     }
+
+    fn classify_dataset_reference(
+        &self,
+        r: &DatasetRefAny,
+    ) -> Result<DatasetRef, DatasetRefRemote> {
+        // Single-tenant workspace => treat all repo-like references as repos.
+        // Multi-tenant workspace => treat all repo-like references as accounts, use
+        // repo:// for repos
```
sergiimk, line 454: Does `repo//name` format actually work? I didn't see any changes in the reference parsing code.

src/infra/core/src/pull_service_impl.rs
```diff
@@ -495,20 +517,17 @@ impl PullService for PullServiceImpl {
...
-                        ingest_from: None,
-                    },
-                }
+            .map(|r| match self.classify_dataset_reference(&r) {
+                Ok(local_ref) => PullRequest {
+                    local_ref: Some(local_ref),
+                    remote_ref: None,
+                    ingest_from: None,
+                },
+                Err(remote_ref) => PullRequest {
```
sergiimk, line 526: Would it make sense to extract this code into `PullRequest::from_any_ref()` to avoid duplication?

src/infra/core/src/workspace_layout.rs
```diff
@@ -106,3 +117,59 @@ impl std::fmt::Display for WorkspaceVersion {
         write!(f, "{}", value)
     }
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+const WORKSPACE_CONFIG_VERSION: i32 = 1;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WorkspaceConfig {
```
sergiimk, line 126: Why did you decide to use a separate config file instead of the existing config?

src/app/cli/src/commands/verify_command.rs
```diff
@@ -46,7 +46,7 @@ impl VerifyCommand {
         I: Iterator<Item = DatasetRef>,
     {
         Self {
-            local_repo,
+            dataset_repo,
```
sergiimk, line 49: In future whenever you think we need to mass-rename something let's do it on `master` branch as a quick isolated change - this would make this PR a lot easier to read.

src/domain/core/src/entities/current_account_config.rs
```diff
@@ -0,0 +1,36 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use opendatafabric::AccountName;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone)]
+pub struct CurrentAccountConfig {
```
sergiimk, line 15: When we integrate authorization library this will likely be called `Subject`, perhaps we should call it that from the beginning?

src/domain/core/src/entities/current_account_config.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub account_name: AccountName,
+    pub specified_explicitly: bool,
```
sergiimk, line 17: Is `specified_explicitly` property very CLI-focused? If subject making the call is determined from HTTP `Authorization` header - is it implicit or explicit? I feel like the explicitness shouldn't matter... at least on `domain` level.

sergiimk (changes requested):
Phew, this is a big one!

It looks great overall :+1: 

My main suggestions are:
- Simplify error handling
- Consider moving tenancy into existing `.kamuconfig`
- Support multi-tenancy in `SetTransform` input aliases

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -398,17 +400,27 @@ impl DatasetRefAny {
...
                 Err(DatasetRefRemote::Alias(DatasetAliasRemote {
-                    repo_name: RepoName::from_inner_unchecked(repo_name),
-                    account_name: Some(AccountName::from_inner_unchecked(account_name)),
+                    repo_name,
+                    account_name: Some(account_name),
                     dataset_name,
                 }))
             }
-            Self::Alias(Some(prefix), None, dataset_name) => {
+            Self::RemoteAlias(repo_name, None, dataset_name) => {
```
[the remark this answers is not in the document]
zaychenko-sergei: Right, simplified

src/domain/core/src/repos/dataset_repository.rs
```diff
@@ -22,6 +22,10 @@ use crate::*;
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+pub const DEFAULT_DATASET_OWNER_NAME: &str = "kamu";
```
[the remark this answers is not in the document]
zaychenko-sergei: There's one place in `list` command:

```
            if show_owners {
                owner.push(String::from(match &hdl.alias.account_name {
                    Some(name) => name.as_str(),
                    None => DEFAULT_DATASET_OWNER_NAME,
                }));
            }
```

Conceptually, we might think of displaying it in the "Owner" field even in single-tenant mode, but I decided to omit the column instead.

Anyway, we have to inject something for `CurrentAccountConfig` in the core domain services, in particular, in local fs dataset repository and into pull services, where the account name is used in multi-tenant ref resolutions. And tests utilize this identifier.

src/infra/core/src/pull_service_impl.rs
zaychenko-sergei: This is required when 2 users have datasets with identical names, and we are pulling via alternative user (`-a` switch) without specifying account name in the reference:

  `kamu -a wasya pull alberta.case-details`

src/infra/core/src/pull_service_impl.rs
zaychenko-sergei: Right now a single-tenant dataset repository would panic if we query it by account:

```
    fn get_account_datasets<'s>(&'s self, _account_name: AccountName) -> DatasetHandleStream<'s> {
        panic!("Single-tenant dataset repository queried by account");
    }
```

What you basically suggest could work if we compared the account name with `DEFAULT_DATASET_OWNER_NAME` here and redirect to `get_all_datasets`. Still panicking if name does not match default.

src/infra/core/src/pull_service_impl.rs
zaychenko-sergei: Never tried. We need to extend tests to check

src/infra/core/src/pull_service_impl.rs
zaychenko-sergei: Good point, extracted.

src/domain/core/src/entities/current_account_config.rs
zaychenko-sergei: It does not matter at domain level. It affects the behavior of 2 CLI commands: `list` and `add`.

src/infra/core/src/workspace_layout.rs
zaychenko-sergei: As it's not a responsibility of the CLI layer. `WorkspaceLayout` is not a truly CLI concept on today, as it lives in the `infra` part. I cannot inject CLI-specific abstractions in the `infra` part.

zaychenko-sergei: Ideally, it should be living in a new local-fs adapter, that is not referenced by the tests directly.

src/adapter/graphql/src/queries/datasets/datasets.rs
```diff
@@ -173,8 +173,8 @@ impl Datasets {
             .await
     }
 
-    // TODO: Multitenancy
-    // TODO: Multitenant resolution for derivative dataset inputs (should it only
+    // TODO: Multi-tenancy
+    // TODO: Multi-tenant resolution for derivative dataset inputs (should it only
     // work by ID?)
```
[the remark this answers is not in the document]
zaychenko-sergei: I haven't touched the GraphQL parts at all here yet, that's a separate subject.
The problem with `TransformInput` for the current stage is rather here - basically multi-tentant ref in a snapshot for derived dataset transformation:

```
async fn resolve_transform_inputs<T>(
    repo: &T,
    dataset_name: &DatasetName,
    inputs: &mut Vec<TransformInput>,
) -> Result<(), CreateDatasetFromSnapshotError>
where
    T: DatasetRepository,
    T: ?Sized,
    {...}
```

src/domain/core/src/repos/dataset_repository.rs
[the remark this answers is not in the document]
sergiimk: I agree that it's a usage mistake and it would make sense to report it in methods like `resolve_dataset_ref()`, to let caller handle it.

But when some high-level call like `sync_dataset()` crashes with `MultiTenantRefUnexpected` - it might as well crash with `InternalError` because caller will have no means to handle this error - such error should've been prevented earlier at input validation.

So overall I think:
- It's a scary amount of changes in error handling for this small error
- I would be OK to see internal error on tenancy issues in CLI as they shouldn't happen during normal usage
- If we really want to display `MultiTenantRefUnexpected` as user error - we could use RTTI to detect it even when it's wrapped into `InternalError`
- But the best solution would probably be to prevent it from happening by validating user input early (but I'm not sure extra core is worth it in this specific case)

src/domain/core/src/repos/dataset_repository.rs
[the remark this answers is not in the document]
sergiimk: While it's convenient for tests no doubt - my issue it that there is no such concept as "Default Owner" in our domain language, so it looks odd in this module. 

I would probably prefer even hard-coding `"kamu"` in every test to having this constant in domain layer... just to keep our domain language pure. Or at the very least renaming it to `TEST_OWNER_NAME` or `TEST_SUBJECT`.

As for use in CLI layer - I thought you were using OS account name in `CurrentAccountConfig`.

src/infra/core/src/pull_service_impl.rs
sergiimk: Yes, that's exactly what I had in mind. Making single-tenant repo work:
- with aliases without `<account>`
- and aliases where account matches the one in `CurrentAccountConfig`

src/infra/core/src/workspace_layout.rs
sergiimk: Take a look at how we configure everything else on infra layer (container-runtime, ipfs gateway, engines etc.). The configuration lives in `CLIConfig` which during initialization is parsed and deconstructed into infra-specific config objects.

Is there any reason why we cannot do the same here and store tenancy as something like:
```yaml
workspace:
  tenancy: "multi-tenant"
```
in our main config?

src/domain/core/src/entities/current_account_config.rs
sergiimk: Not a big deal for now - let's leave it until we integrate authorization

src/domain/core/src/entities/current_account_config.rs
sergiimk: Can we pass it as an extra argument to those commands and eliminate this field from domain layer?

src/adapter/graphql/src/queries/datasets/datasets.rs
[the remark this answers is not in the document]
zaychenko-sergei: I haven't touched GraphQL parts much, but the core transformations across accounts are now supported, and ODF was extended.

src/infra/core/src/pull_service_impl.rs
zaychenko-sergei: Improved

src/domain/core/src/entities/current_account_config.rs
zaychenko-sergei: I've separated a domain-level `CurrentAccountSubject` with account name only, and CLI-level `CurrentAccountIndication`, that looks similar to `RelatedAccountIndication`, and is not injected via DI, but passed as a value into dependent commands directly.

src/domain/core/src/entities/current_account_config.rs
zaychenko-sergei: I renamed, but I still don't understand the convention you are referring too.

src/infra/core/src/workspace_layout.rs
zaychenko-sergei: Two reasons:

1) I does not feel architecturally correct to me. There is no strict == between being in CLI mode and having a local FS workspace layout. 

2) Technically, I need to know if workspace is multi-tenant before DI catalog is started to get filled, as it has to be explicitly injected via builder arguments into several services. At least on today, reading `CLIConfig` happens at the very end of dependencies initialization. Changing the order of initialization feels unmotivated to me.

sergiimk: Re. `2.` I think it's a matter of time when we need to change the DI catalog composition based on the configuration. I don't see anything that prevents us now from reading the config first, and it seems more logical to me. Current order just kinda "happened" rather than was designed.

Re. `1.` You can say this about every config option: IPFS  != container runtime != engine configuration. But this doesn't mean we should put each in a different config file, does it? If you follow same separation of:
- there is a config DTO in `infra` layer (agnostic of where values actually come from)
- there is a config section in `CLIConfig`
- we read CLI config and convert it into infra DTO

... then CLI config is purely "presentation" layer concern - just like any CLI flag or an argument. In fact coupling core tenancy configuration with file format and serialization feels less architecturally sound to me.

It also:
- Complicates workspace structure
- And any improvements we make to `CLIConfig` (e.g. allowing to override config via ENV vars or CLI args) will not benefit the tenancy config...

zaychenko-sergei: Re 3. Config is modifiable via commands. Workspace multi-tenance property should be readonly.

src/domain/core/src/repos/dataset_repository.rs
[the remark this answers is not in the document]
zaychenko-sergei: > But the best solution would probably be to prevent it from happening by validating user input early (but I'm not sure extra core is worth it in this specific case)

That's a right comment. I've migrated the `MultiTenantRefUnexpectedError` error to CLI error and implemented validation at the level of command lines, where we expect `DatasetRef`, not `DatasetRefAny`.

src/infra/core/src/workspace_layout.rs
sergiimk: How about we continue with the current (separate file) solution, but modify `WorkspaceLayout::create()` not to write a config file for a single-tenant workspace, so that in prod cases this config was not present?

sergiimk (approved):
Made small suggestion re. config, but other than that - everything looks great and good to merge :+1: :rocket:

src/infra/core/src/workspace_layout.rs
zaychenko-sergei: Done.

```
        // Only save the workspace configuration if it is different from default
        let ws_config = WorkspaceConfig::new(multi_tenant);
        if ws_config != WorkspaceConfig::default() {
            ws_config.save_to(&ws.config_path).int_err()?;
        }
```

PR #200, opened 2023-07-01
Custom perfetto layer

sergiimk:
After some discussion with the author of `tracing-chrome` crate, and asking around in Perfetto dev Discord I decided to prototype my own tracer.

It works as follows:
* every time span is entered / exited it writes **synchronous** begin/end timelines - thus giving us a nice stacked per-thread flame graph for each time a coroutine was actively running on a CPU
* when span is created / closed it writes **async** begin/end timelines under unique timeline IDs - this gives us independent async timeline per every span in the program (similarly to how Jaeger displays spans)

It's not perfect but much better than what we had before.

Here's an example of pulling two datasets concurrently using old tracer:
image

and here is the new one:
image

### DIFF
--- a/src/infra/core/src/repos/dataset_impl.rs
+++ b/src/infra/core/src/repos/dataset_impl.rs
@@ -8,6 +8,7 @@
 // by the Apache License, Version 2.0.
 
 use std::path::Path;
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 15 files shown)


src/infra/core/src/repos/dataset_impl.rs
```diff
@@ -356,19 +357,29 @@ where
         let output_data = match data_interval {
             None => None,
             Some(data_interval) => {
-                let from_data_path = movable_data_file.unwrap();
+                let from_data_path = Arc::new(movable_data_file.unwrap().to_path_buf());
+
+                let path = from_data_path.clone();
+                let logical_hash = tokio::task::spawn_blocking(move || {
```
sergiimk, line 363: Just by looking at the trace I was able to easily spot that main thread is blocked for a very long time when hashing - so adding a fix that moves hashing to a background thread.

zaychenko-sergei: I was actually recently surprised those data hashing helper functions were synchronous, when I did `QueryService` refactoring.

PR #204, opened 2023-07-17
Feature/engine data proxy

sergiimk:
In this PR:
- Isolated Ingest/Transform services from FS paths
  - How data gets into the engine is now decided on the engine level
  - There will be several ways to pass data in and out of the engine in future (mounting files, direct remote storage links, via shared memory)
  - This is the layer I will work on next, to make ingest/transform work for S3 datasets (initially without touching the engines)
- These changes allowed to better formalize the commit procedure
- AND to finally move `WorkspaceLayout` to app layer
- Also we're no longer mounting entire workspace into engine containers:
  - most things are mounted on per-file basis (except ingest where we still need to read all data directory)
  - all inputs are mounted as Read-Only for safety


src/domain/core/src/entities/dataset.rs
```diff
@@ -148,3 +165,59 @@ pub enum CommitError {
         InternalError,
     ),
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////
+// Commit helpers
+/////////////////////////////////////////////////////////////////////////////////////////
+
+/// Replicates [AddData] event prior to hasing of data and checkpoint
```
zaychenko-sergei, line 173: prior to has**h**ing

src/domain/core/src/entities/dataset.rs
```diff
@@ -148,3 +165,59 @@ pub enum CommitError {
...
+    path: Option<PathBuf>,
+}
+
+impl OwnedFile {
+    pub fn new(path: impl Into<PathBuf>) -> Self {
+        let path = path.into();
+        assert!(path.exists());
+        Self { path: Some(path) }
+    }
+
```
zaychenko-sergei, line 207: I guess this should be isolated in `domain/core/src/utils` rather then staying in the `dataset.rs`

src/infra/core/src/engine/engine_io_strategy.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    async fn materialize_object(
+        &self,
+        repo: &dyn ObjectRepository,
+        hash: &Multihash,
+        container_in_dir: &Path,
+        volumes: &mut Vec<VolumeSpec>,
+    ) -> Result<PathBuf, InternalError> {
+        let url = repo.get_internal_url(hash).await;
+        let host_path = kamu_data_utils::data::local_url::into_local_path(url).int_err()?;
```
zaychenko-sergei, line 39: Should we use `CachedObject` here maybe?

src/domain/core/src/entities/dataset.rs
sergiimk: thanks, fixed

src/domain/core/src/entities/dataset.rs
sergiimk: done!

src/infra/core/src/engine/engine_io_strategy.rs
sergiimk: Yep, this will likely be a part of my next PR

PR #215, opened 2023-08-03
S3 repo id based structure

zaychenko-sergei:


src/infra/core/src/engine/engine_spark.rs
```diff
@@ -324,7 +324,7 @@ impl SparkEngine {
 
         {
             tracing::info!(?request, path = ?request_path, "Writing request");
-            let file = File::create(&request_path)?;
+            let file: File = File::create(&request_path)?;
```
sergiimk, line 327: Why was this necessary?

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -26,76 +25,105 @@ use crate::utils::s3_context::S3Context;
...
-            .await?
-        {
-            return Err(MoveBucketItemsOnRenameError::NameCollision(
-                NameCollisionError {
-                    alias: new_alias.clone(),
-                },
-            ));
+        dataset: &dyn Dataset,
+    ) -> Result<DatasetAlias, InternalError> {
+        match dataset.as_info_repo().get("alias").await {
```
sergiimk, line 66: why use `match` instead of `dataset.as_info_repo().get("alias").await.int_err()?` just like below?

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -26,76 +25,105 @@ use crate::utils::s3_context::S3Context;
...
+                    }
+
+                }
+            }
+        })
+    }
+
+    fn normalize_alias(&self, alias: &DatasetAlias) -> DatasetAlias {
+        if alias.is_multi_tenant() {
+            alias.clone()
```
sergiimk, line 117: Nitpick: Would it be better to accept `alias` param as owned to avoid excessive cloning?

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -123,94 +150,119 @@ impl DatasetRepository for DatasetRepositoryS3 {
...
-
-                let name = DatasetAlias::new(None, DatasetName::try_from(prefix).int_err()?);
-                let hdl = self.resolve_dataset_ref(&name.into()).await.int_err()?;
-                yield hdl;
-            }
-        })
+        self.stream_datasets_if(|_| true)
     }
 
     fn get_account_datasets<'s>(&'s self, account_name: AccountName) -> DatasetHandleStream<'s> {
```
sergiimk, line 189: Should we rename this method to `get_datasets_by_owner`? It's a bit ambiguous about the relationship it's checking:
- is it all datasets an account has access to
- all datasets it owns
- or something else.

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -123,94 +150,119 @@ impl DatasetRepository for DatasetRepositoryS3 {
...
         dataset_ref: &DatasetRef,
     ) -> Result<Arc<dyn Dataset>, GetDatasetError> {
-        let handle = self.resolve_dataset_ref(dataset_ref).await?;
-        let dataset = self.get_dataset_impl(&handle.alias).await?;
+        let dataset_handle = self.resolve_dataset_ref(dataset_ref).await?;
+        let dataset = self.get_dataset_impl(&dataset_handle.id).await?;
         Ok(Arc::new(dataset))
     }
 
+    // TODO: name collision check
```
sergiimk, line 212: It seems to be checking for name collisions below. TODO either lacking some details or can be removed.

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -220,45 +272,71 @@ impl DatasetRepository for DatasetRepositoryS3 {
...
         {
-            Ok(hash) => Ok(hash),
+            Ok(head) => head,
             Err(AppendError::RefCASFailed(_)) => {
-                // We are using head ref CAS to detect previous existence of a dataset
-                // as atomically as possible
-                Err(CreateDatasetError::NameCollision(NameCollisionError {
-                    alias: dataset_alias.clone(),
-                }))
+                unreachable!("The duplication protection should have triggered by now");
```
sergiimk, line 288: CAS error is still possible if same user is pushing same dataset in two concurrent processes at once - they would race to write the `head` ref file. We shouldn't be crashing the server in such situation.

But I guess it wouldn't be a "NameCollision" in this case, just an optimistic concurrency error...

sergiimk (approved):
Some minor comments, looks good otherwise :ship:

src/infra/core/src/engine/engine_spark.rs
zaychenko-sergei: It's enough to make 1 unexpected click in VSCode to insert a type, and visually it looks almost the same as an auto hint

src/infra/core/src/repos/dataset_repository_s3.rs
zaychenko-sergei: Simplified. This was copied & cut from a more complex match in Local FS

src/infra/core/src/repos/dataset_repository_s3.rs
zaychenko-sergei: No, that's not easy. I'm getting alias references from higher level APIs, like `create_dataset` and `resolve_dataset_ref`

src/infra/core/src/repos/dataset_repository_s3.rs
zaychenko-sergei: Right, clarification makes sense

src/infra/core/src/repos/dataset_repository_s3.rs
zaychenko-sergei: Resolved as internal error

src/infra/core/src/repos/dataset_repository_s3.rs
zaychenko-sergei: Killed. No longer relevant.

PR #225, opened 2023-08-27
DataFusion ingest

sergiimk:
Making our ingest 100X faster by switching from Spark to DataFusion!

Highlights:
* DF ingest is experimental - people have to opt-in to use it by specifying a `preprocess` step in their root dataset polling source that uses `engine: datafusion`
  *  If your source does not require a preprocess step - you'll have to use a no-op SQL like `select * from input`
  * All examples were converted to use DF - after the release I will update the demo and `contrib` repo
* You will see that new ingest is implemented as a completely separate code path in `IngestServiceImpl`
  *  Spark path is left almost untouched
  * There is some code duplication between `IngestServiceImpl` and legacy `IngestTask`, but a ton of code will disappear one Spark ingest is removed
* DF ingest is implemented as a separate crate
  * New `DataWriter` domain object encapsulates logic of adding data to a root dataset (merge strategies, adding system columns, validating schema, commit)
* DF ingest is not only faster - it can also work over S3 without us downloading the entire dataset locally
* Format updates based on open-data-fabric/open-data-fabric#58
* Some further work is needed to make DF ingest the default


src/adapter/http/tests/harness/client_side_harness.rs
```diff
@@ -48,6 +48,9 @@ impl ClientSideHarness {
         b.add::<auth::AlwaysHappyDatasetActionAuthorizer>()
             .bind::<dyn auth::DatasetActionAuthorizer, auth::AlwaysHappyDatasetActionAuthorizer>();
 
+        b.add::<SystemTimeSourceDefault>();
```
sergiimk, line 51: @zaychenko-sergei  Note that I started introducing this interface to have more control over time in tests. Just letting you know it exists.

src/infra/core/src/ingest_service_impl.rs
```diff
@@ -7,44 +7,685 @@
...
+
+        // TODO: Temporarily we continue to default to Spark ingest, but use new engine
+        // if preprocessing query specifies datafusion
+        let engine = polling_source_block
+            .event
+            .preprocess
+            .as_ref()
+            .map(|Transform::Sql(t)| t.engine.as_str());
+
+        if engine == Some("datafusion") {
```
sergiimk, line 122: This is where Spark and DF ingest code paths branch off

src/infra/core/src/ingest_service_impl.rs
```diff
@@ -95,30 +736,23 @@ impl IngestServiceImpl {
...
         }
     }
+}
 
-    async fn do_ingest(
+///////////////////////////////////////////////////////////////////////////////
+// Legacy Spark-based ingest path
+///////////////////////////////////////////////////////////////////////////////
+
+impl IngestServiceImpl {
```
sergiimk, line 745: In a separate impl block I put all the legacy stuff that will be deleted once we drop Spark ingest.

src/infra/ingest-datafusion/benches/ledger.rs
```diff
@@ -0,0 +1,189 @@
...
+        .unwrap();
+
+    let res = res.cache().await.unwrap();
+
+    assert_eq!(res.count().await.unwrap(), expected_rows);
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+fn bench(c: &mut Criterion) {
```
sergiimk, line 168: First few benchmarks!

sergiimk:
Apologies, but I'll rush the merge/release and will follow up on comments in separate PRs.

The tech demo I'm preparing for our partners depends on DF ingest and I need the release to be able to share it with them.

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
```
zaychenko-sergei, line 7: It seems you've made 0.140.0 already, right?

src/adapter/http/tests/harness/client_side_harness.rs
zaychenko-sergei: It's a totally valid practice. Accessing "current time" directly is considered untestable. In languages like Python nobody cares, as they allow "freezing" time via dynamic language capabilities, but in compiled languages like Rust that would be the only valid way to deal with "current time".

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -106,6 +106,34 @@ pub trait MetadataChainExt: MetadataChain {
         }
     }
 
+    /// Same as [Self::last_of_type_ref] defaulting to the head reference
+    async fn last_of_type<E: VariantOf<MetadataEvent>>(
```
zaychenko-sergei, line 110: Just a note: our UI would need to filter pages of N elements of type to implement filters properly. See https://github.com/kamu-data/kamu-web-ui/issues/122

src/domain/core/src/services/ingest/reader.rs
```diff
@@ -0,0 +1,92 @@
...
+/// defined in the [ReadStep].
+#[async_trait::async_trait]
+pub trait Reader: Send + Sync {
+    /// Returns schema that the output will be coerced into, if such schema is
+    /// defined in the [ReadStep].
+    async fn output_schema(
+        &self,
+        ctx: &SessionContext,
+        conf: &ReadStep,
+    ) -> Result<Option<Schema>, ReadError>;
```
zaychenko-sergei, line 30: Can we use this to improve UX with quick schema feedback?

src/domain/core/src/utils/time_source.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl SystemTimeSource for SystemTimeSourceDefault {
+    fn now(&self) -> DateTime<Utc> {
+        Utc::now()
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct SystemTimeSourceMock {
```
zaychenko-sergei, line 34: I prefer if we stick to [Gerard Meszaros](http://xunitpatterns.com/Test%20Double%20Patterns.html) classic terminilogy:
 - "mock" is something that tests behavioral impact (asserting calls, arguments, sequence)
 - "fake" is a testing-friendly alternative implementation
 - "stub"  is a test-specific object that feeds the desired indirect inputs
 
 From this POV, this is more of a "stub".

src/domain/opendatafabric/src/dtos/dtos_extra.rs
```diff
@@ -169,6 +174,25 @@ impl TransformSql {
...
+    pub fn schema(&self) -> Option<&Vec<String>> {
+        match self {
+            ReadStep::Csv(v) => v.schema.as_ref(),
+            ReadStep::Json(v) => v.schema.as_ref(),
+            ReadStep::JsonLines(v) => v.schema.as_ref(),
+            ReadStep::NdJson(v) => v.schema.as_ref(),
+            ReadStep::GeoJson(v) => v.schema.as_ref(),
+            ReadStep::NdGeoJson(v) => v.schema.as_ref(),
+            ReadStep::EsriShapefile(v) => v.schema.as_ref(),
+            ReadStep::Parquet(v) => v.schema.as_ref(),
```
zaychenko-sergei, line 191: Note, this code is shouting `schema` field is stored at a wrong hierarchy level

src/infra/core/src/ingest_service_impl.rs
zaychenko-sergei: Don't we have an enum for engines?

src/infra/core/src/ingest_service_impl.rs
```diff
@@ -7,44 +7,685 @@
...
+
+        let mut data_writer = DataWriterDataFusion::builder(args.dataset.clone(), ctx.clone())
+            .with_metadata_state_scanned()
+            .await?
+            .build()
+            .await?;
+
+        let mut iteration = 0;
+        let mut combined_result = None;
+        loop {
```
zaychenko-sergei, line 156: As I understand, an iteration per query?

src/infra/core/tests/tests/engine/test_engine_ingest.rs
```diff
@@ -85,87 +100,816 @@ async fn test_ingest_csv_with_engine() {
...
+        indoc!(
+            r#"
+            +--------+----------------------+----------------------+----------+------+------------+
+            | offset | system_time          | event_time           | observed | city | population |
+            +--------+----------------------+----------------------+----------+------+------------+
+            | 0      | 2050-01-01T12:00:00Z | 2050-01-01T12:00:00Z | I        | A    | 10001      |
+            | 1      | 2050-01-01T12:00:00Z | 2050-01-01T12:00:00Z | I        | B    | 20001      |
+            | 2      | 2050-01-01T12:00:00Z | 2050-01-01T12:00:00Z | I        | C    | 30001      |
+            +--------+----------------------+----------------------+----------+------+------------+
+            "#
```
zaychenko-sergei, line 233: Nice. I think I had a test somewhere where I did the `DataFrame` unpacking manually to compare 1 value.

src/infra/core/tests/tests/engine/test_engine_ingest.rs
```diff
@@ -174,21 +918,20 @@ struct IngestTestHarness {
...
         let cache_dir = temp_dir.path().join("cache");
         std::fs::create_dir(&run_info_dir).unwrap();
         std::fs::create_dir(&cache_dir).unwrap();
 
-        let dataset_action_authorizer = Arc::new(
-            mock_dataset_action_authorizer::MockDatasetActionAuthorizer::new()
-                .expect_check_write_dataset(DatasetAlias::new(None, dataset_name.clone()), 1),
-        );
+        let dataset_action_authorizer =
+            Arc::new(kamu_core::auth::AlwaysHappyDatasetActionAuthorizer::new());
```
zaychenko-sergei, line 934: What is the problem here? Why replaced on forgiving fake?

CHANGELOG.md
sergiimk: yes, I just didn't push version bump to this branch

src/domain/core/src/entities/metadata_chain.rs
sergiimk: Yeah, I think we need and API like `iter_blocks` that accepts a mask of event types, e.g.

```rust
metadata_chain.iter_blocks_ref(
  &BlockRef::Head, 
  MetadataEventType::SetPollingSource | MetadataEventType::SetLicense | MetadataEventType::SetInfo
)`
```
This would let `MetadataChain` implementations decide which speedups they can use to make traversal most efficient. The `last_of_type` can then be a helper that delegates to such function.

I was going to add such function first, but my PR was already growing out of control.

src/domain/core/src/services/ingest/reader.rs
sergiimk: Not this function specifically, but I definitely had the schema inference for UI in mind when writing these reader types. I think DataFusion ingest is a significant step in this direction.

src/domain/core/src/utils/time_source.rs
sergiimk: Thanks - will rename to `SystemTimeSourceStub` in a follow up PR

src/domain/opendatafabric/src/dtos/dtos_extra.rs
sergiimk: This is a "poor man's base class" :)

Looking at metadata structure I think `schema` property is well positioned:
```yaml
read:
  kind: csv
  schema: ...
```

I wouldn't want adding another level of nesting there just for field reuse.

src/infra/core/src/ingest_service_impl.rs
sergiimk: Some data sources (currently only `filesGlob`) will return `has_more: true` to say "call me again - I have more data", so the ingest will loop, committing each chunk, until it exhausts the source.

src/infra/core/tests/tests/engine/test_engine_ingest.rs
sergiimk: Thanks, I tried a dozen different approaches to writing tests for `DataFrame`s and this seems like the least verbose and flexible one I had.

sergiimk: I'm also using [pretty_assertions crate](https://docs.rs/pretty_assertions/latest/pretty_assertions/) making errors very easy to read.

src/infra/core/tests/tests/engine/test_engine_ingest.rs
sergiimk: Good spot. I struggled with it as I was creating tests that ingest multiple times. I have restored auth check in my follow up PR.

src/infra/core/src/ingest_service_impl.rs
sergiimk: Anyone can implement an engine, so we cannot enumerate them all, but I will introduce an enum for those engine we care about (we don't have one yet).

PR #230, opened 2023-08-30
Feature/UI rebac

zaychenko-sergei:


src/adapter/graphql/src/root.rs
```diff
@@ -91,6 +91,22 @@ impl Mutation {
...
+impl AccessToken {
+    pub fn new<S>(token: S) -> Self
+    where
+        S: Into<String>,
+    {
+        Self {
+            token: token.into(),
+        }
+    }
+}
```
sergiimk, line 108: Should this live in `scalars/` sub-module?

src/adapter/graphql/tests/tests/test_gql_metadata_chain.rs
```diff
@@ -78,9 +89,11 @@ async fn metadata_chain_append_event() {
                     }
                 }
                 "#
+                )
+                .replace("<id>", &create_result.dataset_handle.id.to_string())
+                .replace("<content>", &event_yaml.escape_default().to_string()),
             )
-            .replace("<id>", &create_result.dataset_handle.id.to_string())
-            .replace("<content>", &event_yaml.escape_default().to_string()),
+            .data(AccessToken::new(access_token)),
```
sergiimk, line 96: I hope all these changes can be simplified after we figure out how to pass catalog with token into GQL handler, coz this became quite messy.

src/app/cli/src/services/account_service.rs
```diff
@@ -101,8 +252,25 @@ impl CurrentAccountIndication {
     }
 
     pub fn as_current_account_subject(&self) -> CurrentAccountSubject {
```
sergiimk, line 254: Nitpick: In rust:
- `as_*()` usually means a zero overhead type conversion (e.g. `String` to `&str`)
- `into_*()` usually means conversion that consumes `self` and returns converted type
- `to_*()` creates a new instance of a target type while leaving original intact (e.g. `Path::to_path_buf()`)

src/app/cli/src/services/config_service.rs
```diff
@@ -43,6 +45,8 @@ pub struct CLIConfig {
     /// Data access and visualization configuration
     #[merge(strategy = merge_recursive)]
     pub frontend: Option<FrontendConfig>,
+    /// Users configuration
+    pub users: Option<UsersConfig>,
```
sergiimk, line 49: merge decorator might be needed in case we ever want to define common set of users for all workspaces within some environment.

src/domain/core/src/entities/current_account_subject.rs
```diff
@@ -9,28 +9,30 @@
...
 
-pub const TEST_ACCOUNT_NAME: &str = "kamu";
+pub const DEFAULT_ACCOUNT_NAME: &str = "kamu";
+pub const DEFAULT_AVATAR_URL: &str = "https://avatars.githubusercontent.com/u/50896974?s=200&v=4";
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
 pub struct CurrentAccountSubject {
-    pub account_name: AccountName,
+    pub account: auth::AccountInfo,
```
sergiimk, line 22: This feels very wrong to me. Auth subject should be the carrier of identity and nothing else. Definitely would not expect it to carry avatar URLs.

Why do you think it's necessary to have `AccountInfo` as part of the Subject? Wouldn't it be enough if where necessary we could resolve Subject into `AccountInfo` via `AuthenticationService`?

src/infra/core/src/auth/authentication_service_impl.rs
```diff
@@ -0,0 +1,214 @@
...
+    kamu_jwt_secret: String,
+    authentication_providers_by_method: HashMap<&'static str, Arc<dyn AuthenticationProvider>>,
+}
+
+#[component(pub)]
+impl AuthenticationServiceImpl {
+    pub fn new(authentication_providers: Vec<Arc<dyn AuthenticationProvider>>) -> Self {
+        let kamu_jwt_secret = match std::env::var("KAMU_JWT_SECRET") {
+            Ok(jwt_secret) => jwt_secret,
+            Err(_) => Self::random_jwt_secret(),
```
sergiimk, line 41: I'm usually scared of such fallbacks - one small typo and our backend is silently miss-configured.

Should we make it a required DI argument?

src/infra/core/src/auth/authentication_service_impl.rs
```diff
@@ -0,0 +1,214 @@
...
+        };
+
+        let mut service = Self {
+            authentication_providers_by_method: HashMap::new(),
```
sergiimk, line 45: Why not build the hash map first and create the type after?

src/infra/core/src/auth/authentication_service_impl.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    fn make_access_token(
+        &self,
+        subject: String,
+        login_method: &str,
+        provider_credentials_json: String,
+    ) -> Result<String, InternalError> {
+        let claims = KamuAccessTokenClaims {
+            iat: Utc::now().timestamp() as usize,
```
sergiimk, line 94: Let's use our new `SystemTimeSource`! We'll need it anyway once we start testing token expiration.

src/infra/core/src/auth/authentication_service_impl.rs
```diff
@@ -0,0 +1,214 @@
...
+            access_credentials: KamuAccessCredentials {
+                login_method: String::from(login_method),
+                provider_credentials_json,
+            },
+        };
+
+        encode(
+            &Header::new(KAMU_JWT_ALGORITHM),
+            &claims,
+            &EncodingKey::from_secret(self.kamu_jwt_secret.as_bytes()),
```
sergiimk, line 107: `from_secret` is probably expensive? Is it better to cache the `EncodingKey`?

src/infra/core/src/auth/authentication_service_impl.rs
```diff
@@ -0,0 +1,214 @@
...
+            exp: (Utc::now() + Duration::days(1)).timestamp() as usize,
+            iss: String::from(KAMU_JWT_ISSUER),
+            sub: subject,
+            access_credentials: KamuAccessCredentials {
+                login_method: String::from(login_method),
+                provider_credentials_json,
+            },
+        };
+
+        encode(
```
sergiimk, line 104: Does this create an encrypted + signed token or only signed?

src/infra/core/src/auth/authentication_service_impl.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn decode_access_token(
+        &self,
+        access_token: String,
+    ) -> Result<KamuAccessCredentials, AccessTokenError> {
+        let mut validation = Validation::new(KAMU_JWT_ALGORITHM);
+        validation.set_issuer(vec![KAMU_JWT_ISSUER].as_slice());
+
+        let token_data = decode::<KamuAccessTokenClaims>(
+            &access_token,
+            &DecodingKey::from_secret(self.kamu_jwt_secret.as_bytes()),
```
sergiimk, line 121: Same here re. `from_secret`

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -484,7 +483,7 @@ impl DatasetStorageStrategy for DatasetSingleTenantStorageStrategy {
     }
 
     fn get_datasets_by_owner<'s>(&'s self, account_name: AccountName) -> DatasetHandleStream<'s> {
-        if account_name == self.current_account_subject.account_name {
+        if account_name.eq(&self.current_account_subject.account.login) {
```
sergiimk, line 486: why this change?

src/adapter/graphql/src/mutations/auth_mut.rs
```diff
@@ -7,117 +7,110 @@
...
 }
 
-#[derive(SimpleObject, Debug, Clone, Deserialize)]
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(SimpleObject, Debug, Clone)]
 pub(crate) struct AccountInfo {
-    login: String,
+    login: AccountName,
     name: String,
```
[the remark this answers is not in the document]
sergiimk: Term `login` makes sense for our password-based auth provider, but then I would expect provider to map it into `account_name`.

sergiimk (changes requested):

src/adapter/graphql/src/root.rs
zaychenko-sergei: Why? It's not going outside API

src/infra/core/src/auth/authentication_service_impl.rs
zaychenko-sergei: Having a coordinated JWT secret would be necessary for scaling API server instances.
It could be randomized at Kubernetes level, however, a different secret would mean dropping existing sessions after restart.
On the other hand, secret has to be periodially randomized, so some rotation rules are needed.

src/infra/core/src/auth/authentication_service_impl.rs
zaychenko-sergei: Agree, no reason

src/app/cli/src/services/account_service.rs
zaychenko-sergei: `to_*()` is the right match here

src/app/cli/src/services/config_service.rs
zaychenko-sergei: Missed

src/infra/core/src/auth/authentication_service_impl.rs
zaychenko-sergei: `encode` here sends for Base64 conversion. Then it's signed. But it's not encrypted - there is no sensitive information there anyway.

src/infra/core/src/repos/dataset_repository_local_fs.rs
zaychenko-sergei: Wasn't sure on the idiomatic way. Corrected here and in one more place

src/infra/core/src/auth/authentication_service_impl.rs
zaychenko-sergei: Not much, but copies the secret internally. I think I've followed the style of the examples, but yeah, avoiding a copy is not a bad thing too.

```
    pub fn from_secret(secret: &[u8]) -> Self {
        EncodingKey { family: AlgorithmFamily::Hmac, content: secret.to_vec() }
    }
````

File: src/infra/core/src/auth/authentication_service_impl.rs (hunk 8 above)
Comment (zaychenko-sergei) [reply src/infra/core/src/auth/authentication_service_impl.rs:94]: Utilized

File: src/infra/core/src/auth/authentication_service_impl.rs (hunk 6 above)
Comment (zaychenko-sergei) [reply src/infra/core/src/auth/authentication_service_impl.rs:41]: An open question is also whether to externalize the issuer and signing algorithm properties.

File: src/adapter/graphql/src/mutations/auth_mut.rs (hunk 13 above)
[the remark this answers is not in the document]
Comment (zaychenko-sergei) [reply src/adapter/graphql/src/mutations/auth_mut.rs:102]: `login` -> `account_name`
`name` -> 'display_name`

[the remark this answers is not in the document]
Comment (zaychenko-sergei) [reply src/adapter/graphql/src/mutations/auth_mut.rs]: I've introduced `AccountDisplayName` as a `String` typedef temporarily, but in future it needs some constraints, in particular length (>0, <N)

File: src/domain/core/src/entities/current_account_subject.rs (hunk 5 above)
Comment (zaychenko-sergei) [reply src/domain/core/src/entities/current_account_subject.rs:22]: I agree, it was a hasty move, reverted

File: src/infra/core/tests/tests/ingest/test_fetch.rs [hunk 14]
```diff
@@ -636,6 +636,7 @@ async fn test_fetch_files_glob() {
 // Container
 ///////////////////////////////////////////////////////////////////////////////
 
+#[ignore]
```
sergiimk, line 639: Never seen this test fail randomly before - let's investigate.

src/adapter/graphql/src/mutations/dataset_metadata_mut.rs
```diff
@@ -58,11 +60,25 @@ impl DatasetMetadataMut {
...
 
     /// Updates or clears the dataset readme
+    #[graphql(guard = "LoggedInGuard::new()")]
+    // TODO: guard should include permission check
     async fn update_readme(
         &self,
         ctx: &Context<'_>,
         content: Option<String>,
     ) -> Result<UpdateReadmeResult> {
+        match check_dataset_write_access(ctx, &self.dataset_handle).await {
```
sergiimk, line 70: This is a lot of code ... would suck for this to spread in to every mutation.

Let's think (outside of this PR) how we can push these permission checks down to the dataset repo layer (e.g. one idea would be to separate `Dataset` interface into "read" and "write" parts).

src/adapter/graphql/src/root.rs
```diff
@@ -91,18 +94,90 @@ impl Mutation {
...
         .extension(Tracing)
         .extension(extensions::ApolloTracing)
         .finish()
 }
+
+/// Executes GraphQL query
+pub async fn execute_query(
+    schema: Schema,
+    base_catalog: dill::Catalog,
+    maybe_access_token: Option<AccessToken>,
```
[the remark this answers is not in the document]
zaychenko-sergei: Smart protocol would require a similar approach, as well as the login entry in the REST API too.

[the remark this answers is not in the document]
zaychenko-sergei: Implementing those Axum middlewares is easier said than done. Last time (dataset resolution) you did some magic generic code, which I don't understand.

src/adapter/graphql/src/mutations/auth_mut.rs
```diff
@@ -7,117 +7,110 @@
...
-            .await
-            .int_err()?;
-
-        Ok(account_info)
+impl From<kamu_core::auth::GetAccountInfoError> for GqlError {
+    fn from(value: kamu_core::auth::GetAccountInfoError) -> Self {
+        match value {
+            kamu_core::auth::GetAccountInfoError::AccessToken(e) => GqlError::Gql(
+                Error::new("Access token error")
+                    .extend_with(|_, eev| eev.set("token_error", e.to_string())),
```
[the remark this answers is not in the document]
zaychenko-sergei: I don't think I understand the proper approach here in general, and why GraphQL is not using standard HTTP statuses. Token error or missing token should be status 401 - Unauthorized, while lack of permission should return 403 - Forbidden. Instead, GraphQL suggests returning 200 always and transport errors as valid response.

I think I'm in a need to read something conceptual why it has to be done one way or another.

src/infra/core/tests/tests/ingest/test_fetch.rs
zaychenko-sergei: Not even randomly. It is failing all the time no my machine.

src/adapter/graphql/src/mutations/dataset_metadata_mut.rs
zaychenko-sergei: It's the same problem as we had with authorization checks at CLI commands level.

src/adapter/graphql/src/queries/datasets/datasets.rs
```diff
@@ -66,7 +63,12 @@ impl Datasets {
         let page = page.unwrap_or(0);
         let per_page = per_page.unwrap_or(Self::DEFAULT_PER_PAGE);
 
-        let mut all_datasets: Vec<_> = dataset_repo.get_all_datasets().try_collect().await?;
+        let account_name = account.name(ctx).await?;
```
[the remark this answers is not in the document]
zaychenko-sergei: I don't understand what `account.as_account_ref()` is supposed to return. We only have `AccountID` and `AccountName` primitives, and accessing those is an attribute access, not a type conversion.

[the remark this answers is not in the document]
zaychenko-sergei: `AccountID` topic in the core layer is out of scope, I have to work with what's available now, and not with what will be available some day

src/adapter/graphql/src/root.rs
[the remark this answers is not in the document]
zaychenko-sergei: Btw, relying on middleware would mean injecting subject in every GraphQL test instead of a token

src/adapter/graphql/src/mutations/auth_mut.rs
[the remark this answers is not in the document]
zaychenko-sergei: Removed `AccessError` variant from response type enums, it's a runtime GQL error instead.

src/adapter/graphql/tests/tests/test_gql_metadata_chain.rs
zaychenko-sergei: Simplified, still, we have to attach catalog to the schema post-factum

src/adapter/graphql/src/root.rs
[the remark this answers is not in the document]
zaychenko-sergei: Implemented Axum middleware and the rest of the suggested handling approach

src/adapter/graphql/src/mutations/auth_mut.rs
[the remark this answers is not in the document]
sergiimk: I think GQL just treats HTTP as transport protocol. I imagine one could potentially run GQL over SSH or GRPC or any other channel. Thus it decides to implement all error mechanisms on its own (poorly).

src/adapter/graphql/src/queries/datasets/dataset.rs
```diff
@@ -17,16 +17,16 @@ use crate::queries::*;
 
 #[derive(Debug, Clone)]
 pub struct Dataset {
-    account: Account,
+    owner_ref: AccountRef,
```
sergiimk, line 20: What's the motivation for this change of using `AccountRef` instead of `Account` everywhere?

It seems to break the whole "Graphness" of GQL:
- Previously you could potentially traverse nodes infinitely through connections like `dataset -> owner -> organizations[0] -> teams[0] -> datasets[0] -> owner ....`.
- But now `AccountRef` is a dead-end node with just an ID and Name, which is very REST-like.

In some comment previously I was suggesting to introduce `AccountRef`, but it was mostly for the ODF side of things - as a direct counterpart for `DatasetRef` that allows to refer to an account EITHER by its name or by ID, and then a `AccountHandle` which is a resolved reference that has BOTH name and ID. (this is of course out of scope of this PR).

src/adapter/http/src/middleware/current_account_resolver_layer.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone)]
+pub struct CurrentAccountResolverMiddleware<Svc> {
+    inner: Svc,
+}
+
+impl<Svc> CurrentAccountResolverMiddleware<Svc> {
```
sergiimk, line 49: Would `AuthenticationMiddleware` be a better name for this?

src/adapter/http/src/middleware/current_account_resolver_layer.rs
```diff
@@ -0,0 +1,136 @@
...
+                match Self::current_account_subject(base_catalog, maybe_access_token).await {
+                    Ok(current_account_subject) => current_account_subject,
+                    Err(response) => return Ok(response),
+                };
+
+            let derived_catalog = dill::CatalogBuilder::new_chained(base_catalog)
+                .add_value(current_account_subject)
+                .build();
+
+            request.extensions_mut().insert(derived_catalog.clone());
```
sergiimk, line 131: Why do we need to clone here? Doesn't look like original instance is needed any more.

sergiimk (changes requested):

src/adapter/graphql/src/queries/datasets/datasets.rs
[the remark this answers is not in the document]
sergiimk: I imagined `account.as_account_ref()` would return an inner field of type `&odf::AccountName` for now (later to be replaced by `odf::AccountRef` that would mimic `DatasetRef` - a value that refers to an account by ID, by Name, or by both).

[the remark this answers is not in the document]
zaychenko-sergei: Looking forward, it's not so easy to introduce account IDs. In several places we only have account name in the alias within a dataset reference, and there is nowhere to take the account ID from - it must be stored with the name in all such places, like inputs of transform, i.e.

src/adapter/http/src/middleware/current_account_resolver_layer.rs
zaychenko-sergei: Eliminated

src/adapter/http/src/middleware/current_account_resolver_layer.rs
zaychenko-sergei: Renamed middleware and layer

src/adapter/graphql/src/queries/datasets/dataset.rs
zaychenko-sergei: If I only have  a reference (we usually only have Name to be exact), anything else like avatar URL, not even talking of tracing organizations, needs to interact with `AuthenticiationService` to resolve full account info. This is expensive to precompute (remember, we are going to GitHub API right now - in future, a database round trip), and most of the time these attributes would not be needed much. Computing full account info on each GraphQL field resolution is also expensive. This would need some sort of lazy load wrapper, if it's possible at all in Rust with the immutability rules...

zaychenko-sergei: Making an experiment with `tokio::sync::OnceCell` for lazy evaluation

PR #239, opened 2023-09-22
Feature/kamu login

zaychenko-sergei:


src/app/cli/src/services/odf_server/login_service.rs
```diff
@@ -0,0 +1,282 @@
...
+        if self.output_config.is_tty
+            && self.output_config.verbosity_level == 0
+            && !self.output_config.quiet
+        {
+            eprintln!(
+                "{}\n  {}",
+                s("HTTP server is listening on:").green().bold(),
+                s(&cli_web_server_url).bold(),
+            );
+            eprintln!("{}", s("Use Ctrl+C to stop the server").yellow());
```
sergiimk, line 124: 1) Can we keep all output in the command instead of the service?

2) Do we want to show user anything about the HTTP server running and how to stop it? I would expect a message like this instead:
```
Please open this URL in the browser to login https://....
This process will exit automatically after receiving the login result.
```

src/app/cli/src/services/odf_server/access_token_registry_service.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn write_access_tokens_registry(
+        &self,
+        scope: AccessTokenStoreScope,
+        registry: &OdfServerAccessTokenRegistry,
+    ) -> Result<(), InternalError>;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////
+
+const KAMU_TOKEN_STORE: &str = ".kamutokenstore";
```
sergiimk, line 229: Not a fan of this file name. Makes me wonder if tokens will be the only thing we store there... 

Git uses very generic `.git-credentials` - perhaps we can name it similarly?

sergiimk (approved):
Looks great! :+1: 
Just a few minor comments.

src/app/cli/src/services/odf_server/login_service.rs
zaychenko-sergei: 1. Moved reporting into command via a listener callback.
2. Improved as recommended:
image

src/app/cli/src/services/odf_server/access_token_registry_service.rs
zaychenko-sergei: I explicitly wanted to avoid using term "credentials", as this is not a correct terminology. By definition, "Credentials" is not about tokens, it's about data required before login, not what is obtained after:

> Credentials are a set of login or authentication data that verify a user's identity and grant them access to a particular system or service. Usually, credentials include a username and a password.

I.e. for AWS this would be valid, as it stores API keys, which is something used before login.

Looking at my own computer, I've found a few similar files that end with "rc", i.e. `.npmrc`, `.yarnrc`. This stands for "runtime configuration", although the term is debatable too.

Docker uses something close to what I've proposed: `.tokenseed`.

IPFS uses term `.keystore`.

PR #256, opened 2023-10-13
v0.145.2: caching GitHub account  data

zaychenko-sergei:


src/adapter/oauth/src/oauth_github.rs
```diff
@@ -19,12 +22,19 @@ const LOGIN_METHOD_GITHUB: &str = "oauth_github";
 
 ///////////////////////////////////////////////////////////////////////////////
 
-pub struct OAuthGithub;
+pub struct OAuthGithub {
+    cached_accounts_by_name: Mutex<HashMap<String, GithubAccountInfo>>,
```
sergiimk, line 26: I'd recommend to create a private struct called `State` with two hash maps, and then wrap it into a single mutex.

This way we can only lock once per update and be sure that we don't deadlock.

sergiimk (approved):
LGTM, just one small suggestion

src/adapter/oauth/src/oauth_github.rs
zaychenko-sergei: Addressed

PR #266, opened 2023-10-26
FlightSQL tweaks

sergiimk:
Rustc was choking when my async callback for creating DataFusion sessions per user got a bit complex, so the main change here was to replace a functor with a `SessionFactory` async trait interface.

I also used this opportunity to include python FlightSQL client code into examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.145.5] - 2023-10-26
+### Changed
+- FlightSQL interface tweaks to simplify integration into `api-server`
+
+## [0.145.4] - 2023-10-24
+### Fixed
+- `kamu ui` should run without specifying any auth secrets
+
 ## [0.145.3] - 2023-10-18
 ### Changed
 - Demanding required env vars to be set before API server / Web UI server startup
 - Custom error messages for non-valid session in GraphQL queries depending on the reason
 ### Fixed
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -4,12 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.145.5] - 2023-10-26
```
zaychenko-sergei, line 7: I guess it was more than qualifying for minor release, rather than a patch.

src/adapter/flight-sql/src/session_factory.rs
```diff
@@ -0,0 +1,29 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub type Token = String;
+
+#[async_trait::async_trait]
+#[allow(unused_variables)]
+pub trait SessionFactory: Send + Sync {
+    async fn authenticate(&self, username: &str, password: &str) -> Result<Token, Status> {
+        Err(Status::unauthenticated("Invalid credentials!"))
```
zaychenko-sergei, line 23: Shouldn't this be connected to our `AuthenticationService` ?

zaychenko-sergei (commented):
1) There is `DEFAULT_ACCOUNT_NAME` constant to use instead of hard-coded `kamu` user name.

2) I don't understand why we need a separate authentication mechanism here different than the implemented one. Will it even work in multi-tenant mode?

sergiimk:
@zaychenko-sergei  This is still highly experimental stuff so securing it was not a priority. It does work in multi-tenant mode because our `QueryService` does not yet enforce permissions, so I can query any dataset when not logged in.

PR #272, opened 2023-11-10
Feature / Push ingest

sergiimk:
# Added
- New `kamu ingest` command allows you to push data into a root dataset, examples:
  - `kamu ingest my.dataset data-2023-*.json` - to push from files
  - `echo '{"city": "Vancouver", "population": 675218}' | kamu ingest cities --stdin`
- New `/:dataset/data/ingest` REST endpoint also allows you to push data via API, example:
  - Run API server and get JWT token: `kamu ui --http-port 8080 --get-token`
  - Push data: `echo '{...}' | curl -v -X POST http://localhost:8080/freezer/data/ingest -H 'Authorization:  Bearer <token>'`
- The `kamu ui` command now supports `--get-token` flag to print out the access token upon server start that you can use to experiment with API


src/app/cli/src/commands/ingest_command.rs
```diff
@@ -0,0 +1,151 @@
...
+            .canonicalize()
+            .map_err(|e| CLIError::usage_error(format!("Ivalid path {}: {}", path, e)))?;
+        url::Url::from_file_path(p)
+            .map_err(|_| CLIError::usage_error(format!("Ivalid path {}", path)))
+    }
+}
+
+#[async_trait::async_trait(?Send)]
+impl Command for IngestCommand {
+    async fn run(&mut self) -> Result<(), CLIError> {
```
zaychenko-sergei, line 71: Minor: the method is quite long. Try splitting out validations and progress tracking inits.

src/app/cli/src/commands/pull_command.rs
```diff
@@ -412,7 +304,7 @@ enum ProgressStyle {
     Bar,
 }
 
-struct PrettyIngestProgress {
+pub(crate) struct PrettyIngestProgress {
```
zaychenko-sergei, line 307: Minor: move to a separate file, as it's getting reused

src/app/cli/src/commands/ui_command.rs
```diff
@@ -64,14 +70,23 @@ impl Command for UICommand {
...
+        .await;
 
         let web_server_url = format!("http://{}", web_server.local_addr());
         tracing::info!("HTTP server is listening on: {}", web_server_url);
 
+        if self.get_token {
+            tracing::info!(
+                token = %web_server.get_access_token(),
+                "Issued API server token"
+            );
```
zaychenko-sergei, line 87: I guess we need this for push ingest external client to be able to connect.
But how will it work in real life on the platform? Simply `kamu login` won't be fine for the IoT device case.

src/infra/core/src/ingest_service_impl.rs
```diff
@@ -909,85 +927,120 @@ impl IngestServiceImpl {
...
+            let mut buf = [0u8; 2048];
+            loop {
+                let read = data.read(&mut buf).await.int_err()?;
+                if read == 0 {
+                    break;
+                }
+                file.write_all(&buf[..read]).await.int_err()?;
+            }
+            file.flush().await.int_err()?;
+        }
```
zaychenko-sergei, line 1025: Yes, this thing looks extremely bad

zaychenko-sergei: We must set a size limit and throw an error in the API handler, if it is exceeded.
Right now this is an easy attack vector against our server - I don't think that container has more than a few gigs of free space in the cluster.

src/app/cli/src/commands/ingest_command.rs
sergiimk: Dome on my current branch

src/app/cli/src/commands/pull_command.rs
sergiimk: I separated push and poll on my current branch and removed this export

src/app/cli/src/commands/ui_command.rs
sergiimk: I think we will need API token issuing feature for IoT device case

PR #288, opened 2023-11-16
Push ingest refactoring

sergiimk:
In this PR we:
- Add tests for push ingest REST API
- Separate ingest into two services `PollingIngestService` and `PushIngestService`
- Remove `fetch_override` support from polling path
- Add support for multiple media types on push
  - e.g. a source can be defined to read NDJSON, but you now will be able to push CSV / JSON etc. into it and ingest will do best-effort conversion
  - CLI adds `--input-type` parameter for this
  - API endpoint will use `Content-Type` header
- Introduce `MediaType` and a `DataFormatRegistry` that can construct `ingest::Reader` based on it
- Refactor `kamu_ingest_datafusion::Reader*` API slightly to do all configuration upon construction, not at runtime

### DIFF
--- a/src/app/cli/src/commands/ingest_command.rs
+++ b/src/app/cli/src/commands/ingest_command.rs
@@ -22,6 +22,7 @@ use crate::OutputConfig;
 ///////////////////////////////////////////////////////////////////////////////
 
 pub struct IngestCommand {
+    data_format_reg: Arc<dyn DataFormatRegistry>,
     dataset_repo: Arc<dyn DatasetRepository>,
     push_ingest_svc: Arc<dyn PushIngestService>,
     output_config: Arc<OutputConfig>,
@@ -35,6 +36,7 @@ pub struct IngestCommand {
 
 impl IngestCommand {
     pub fn new<'s, I, S>(
+        data_format_reg: Arc<dyn DataFormatRegistry>,
         dataset_repo: Arc<dyn DatasetRepository>,
         push_ingest_svc: Arc<dyn PushIngestService>,
         output_config: Arc<OutputConfig>,
@@ -50,6 +52,7 @@ impl IngestCommand {
         S: Into<String>,
     {
         Self {
+            data_format_reg,
             dataset_repo,
             push_ingest_svc,
             output_config,
@@ -93,25 +96,23 @@ impl IngestCommand {
         Ok(())
     }
 
-    fn get_media_type(&self) -> Result<Option<&str>, CLIError> {
+    fn get_media_type(&self) -> Result<Option<MediaType>, CLIError> {
         let Some(short_name) = &self.input_format else {
             return Ok(None);
         };
 
-        // TODO: Keep in sync with CLI parser
-        match short_name.to_lowercase().as_str() {
-            "csv" => Ok(Some(IngestMediaTypes::CSV)),
-            "json" => Ok(Some(IngestMediaTypes::JSON)),
-            "ndjson" => Ok(Some(IngestMediaTypes::NDJSON)),
-            "geojson" => Ok(Some(IngestMediaTypes::GEOJSON)),
-            "ndgeojson" => Ok(Some(IngestMediaTypes::NDGEOJSON)),
-            "parquet" => Ok(Some(IngestMediaTypes::PARQUET)),
-            "esrishapefile" => Ok(Some(IngestMediaTypes::ESRISHAPEFILE)),
-            _ => Err(CLIError::usage_error(format!(
-                "Unsupported format {}",
-                short_name
-            ))),
+        let short_name_lower = short_name.to_lowercase();
+
+        for fmt in self.data_format_reg.list_formats() {
+            if fmt.short_name.to_lowercase() == short_name_lower {
+                return Ok(Some(fmt.media_type.to_owned()));
+            }
         }
+
+        Err(CLIError::usage_error(format!(
+            "Unsupported format {}",
+            short_name
+        )))
     }
 }
 
--- a/src/app/cli/src/commands/ingest_command.rs
+++ b/src/app/cli/src/commands/ingest_command.rs
@@ -30,10 +30,11 @@ pub struct IngestCommand {
     files_refs: Vec<String>,
     stdin: bool,
     recursive: bool,
+    input_format: Option<String>,
 }
 
 impl IngestCommand {
-    pub fn new<'s, I>(
+    pub fn new<'s, I, S>(
         dataset_repo: Arc<dyn DatasetRepository>,
         push_ingest_svc: Arc<dyn PushIngestService>,
         output_config: Arc<OutputConfig>,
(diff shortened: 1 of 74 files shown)


src/app/cli/src/commands/ingest_command.rs
```diff
@@ -149,3 +188,186 @@ impl Command for IngestCommand {
...
+            console::style(step_str).bold().dim(),
+            msg,
+            console::style(dataset).dim(),
+        )
+    }
+
+    fn message_for_stage(&self, stage: PushIngestStage) -> String {
+        let msg = match stage {
+            PushIngestStage::CheckSource => "Checking source for updates",
+            PushIngestStage::Fetch => "Downloading data",
```
zaychenko-sergei, line 253: Didn't expect to see such phases for Push flow

src/infra/core/src/ingest/data_format_registry_impl.rs
```diff
@@ -0,0 +1,149 @@
...
+            ReadStep::EsriShapefile(_) => Self::FMT_ESRI_SHAPEFILE,
+        }
+    }
+
+    async fn get_reader(
+        &self,
+        ctx: SessionContext,
+        conf: ReadStep,
+        temp_path: PathBuf,
+    ) -> Result<Arc<dyn Reader>, ReadError> {
```
zaychenko-sergei, line 90: Maybe the name `Reader` is a bit too general.

src/infra/core/src/ingest/push_ingest_service_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+        )
+    )]
+    async fn do_ingest_inner(
+        &self,
+        mut args: PushIngestArgs,
+    ) -> Result<PushIngestResult, PushIngestError> {
+        args.listener
+            .on_stage_progress(PushIngestStage::CheckSource, 0, TotalSteps::Exact(1));
+        args.listener
+            .on_stage_progress(PushIngestStage::Fetch, 0, TotalSteps::Exact(1));
```
zaychenko-sergei, line 147: Exactly what I was talking about - if we don't need it , why we still keep it.

src/infra/core/src/ingest/push_ingest_service_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+                let p = args
+                    .url
+                    .to_file_path()
+                    .map_err(|_| format!("Invalid file URL {}", args.url).int_err())?;
+
+                // TODO: In case of STDIN (/dev/fd/0) or other pipes and special device files
+                // we have to copy data into a temporary file, as DataFusion cannot read from
+                // them directly.
+                cfg_if::cfg_if! {
+                    if #[cfg(unix)] {
```
zaychenko-sergei, line 213: What happens on Windows?

src/infra/core/src/ingest/push_ingest_service_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+            let mut buf = [0u8; 2048];
+            loop {
+                let read = source.read(&mut buf)?;
+                if read == 0 {
+                    break;
+                }
+                target.write_all(&buf[..read])?;
+            }
+            Ok(())
+        })
```
zaychenko-sergei, line 286: A code like this is repeating from time to time in many places all over the project. Can we think of reusing these adaptations universally?

src/infra/core/src/ingest/push_ingest_service_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+        listener: Option<Arc<dyn PushIngestListener>>,
+    ) -> Result<PushIngestResult, PushIngestError> {
+        // Save stream to a file in cache
+        //
+        // TODO: Breaking all architecture layers here - need to extract cache into a
+        // service
+        let path = self
+            .cache_dir
+            .join(ingest_common::get_random_cache_key("push-ingest-"));
+        Self::copy_stream_to_file(data, &path).await.int_err()?;
```
zaychenko-sergei, line 344: So, technically speaking, it's not exactly cached, but just stored in the cache directory? Maybe there should be a different temp directory?

src/app/cli/src/commands/ingest_command.rs
sergiimk: I'm still not fully sure about the scope of responsibilities of `PushIngestService`:
- It could accept just an input `DataFrame` and only handle preprocessing/merge/commit
- It could (as it is now) also handle reading raw data, and in this case:
  - some data can be read directly into a `DataFrame` (if location and data format are natively supported by DataFusion)
  - some data has to be fetched/transformed before reading if the location or format are not natively supported by DataFusion (e.g. if in future we will have push data arriving via Kafka)

It's because of the latter I left the `CheckSource` and `Fetch` stages.

I will remove them for now, not to muddy the current state of things, until we get to the Kafka stage. Perhaps Kafka micro-batching will be handled by a completely different service.

src/infra/core/src/ingest/data_format_registry_impl.rs
sergiimk: I actually would like us to use modularization/namespacing more instead of long and complex type names, e.g.

Instead of: `kamu::DatasetIngestReader`
Use: `kamu::ingest::Reader`

I think this would create nice small sub-domain languages instead of really large and complex single domain language.

But I do see that without looking at imports name `Reader` can look super generic and confusing. Will try to come up with a nice middleground (but outside of this PR).

src/infra/core/src/ingest/push_ingest_service_impl.rs
sergiimk: As per other comment - will remove these stages

src/infra/core/src/ingest/push_ingest_service_impl.rs
sergiimk: Windows is an unsupported target - we just compile on it not to ruin our long-term chances for compatibility. If you run `kamu` on Windows today the non-working piping will probably be the smallest of issues there.

src/infra/core/src/ingest/push_ingest_service_impl.rs
```diff
@@ -0,0 +1,368 @@
...
+        let mut file = tokio::fs::File::create(target_path).await?;
+        let mut buf = [0u8; 2048];
+        loop {
+            let read = data.read(&mut buf).await?;
+            if read == 0 {
+                break;
+            }
+            file.write_all(&buf[..read]).await?;
+        }
+        file.flush().await?;
```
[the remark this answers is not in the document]
sergiimk: That's the thing - generalizing both to async stream types would significantly slow down the copy from file (or pipe) to file because of async overhead.

src/infra/core/src/ingest/push_ingest_service_impl.rs
sergiimk: I managed to replace the low level code with `std::io::copy` and `tokio::io::copy`.

src/infra/core/src/ingest/push_ingest_service_impl.rs
sergiimk: I removed the dependency on `cache_dir` and no longer abusing the cache.

I'm now passing the stream down into the body of the ingest flow, and similarly to `STDIN` the stream will be saved into a temporary `operation_dir` directory.

Issue #244, opened 2023-09-29
Persistent Accounts


sergiimk:
Tasks:
- [x] https://github.com/kamu-data/kamu-node/issues/52
- [x] https://github.com/kamu-data/kamu-node/issues/68
- [x] https://github.com/kamu-data/kamu-cli/issues/645
- [x] #540 
- [x] https://github.com/kamu-data/kamu-cli/issues/566

sergiimk:
@zaychenko-sergei can this epic be closed?

PR #292, opened 2023-11-18
Experiment/dataset update flow

sergiimk:


src/domain/dataset-update-flow/src/entities/update/update_trigger.rs
```diff
@@ -0,0 +1,52 @@
...
+// by the Apache License, Version 2.0.
+
+use opendatafabric::{AccountID, AccountName, DatasetID};
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum UpdateTrigger {
```
sergiimk, line 17: # General: Enum naming
I suggest using this naming scheme:
```rust
enum Enum {
    VariantA(EnumVariantA),
    VariantB(EnumVariantB),
}
```

This removes a lot of creativity from naming things and great for autocomplete.

Suggested renames:

`UpdateManualTrigger` -> `UpdateTriggerManual`
`UpdateAutoPollingTrigger` -> `UpdateTriggerAutoPolling`
`UpdatePushTrigger` -> `UpdateTriggerPush`
`TimeDeltaSchedule` -> `ScheduleTimeDelta`
`CronExpressionSchedule` -> `ScheduleCronExpression`
etc.

src/domain/dataset-update-flow/src/entities/schedule/schedule_type.rs
```diff
@@ -0,0 +1,45 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+/// Represents dataset update settings
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ScheduleType {
```
sergiimk, line 14: Rename to `Schedule`? As it's not just the type, but the schedule itself.

src/domain/dataset-update-flow/src/aggregates/schedule.rs
```diff
@@ -0,0 +1,69 @@
...
+use chrono::Utc;
+use event_sourcing::*;
+use opendatafabric::DatasetID;
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Aggregate, Debug)]
+pub struct UpdateSchedule(Aggregate<UpdateScheduleState, (dyn UpdateScheduleEventStore + 'static)>);
```
sergiimk, line 19: # Generic scheduling service?
I feel like we will have so much more uses for recurrent / cron-based actions:
- poll/push/deriv updates
- compactions?
- replication?
- archival / truncation?
- garbage collection?

Should we make this into a standalone super-generic service, e.g. one that can publish any **command to our domain bus** on a specified schedule?

Note: I am trying hard to resist the urge to over-generalize, but it would also suck to end up with half a dozen cron services for every different kind of upkeep actions.

src/domain/dataset-update-flow/src/aggregates/update.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct Update(Aggregate<UpdateState, (dyn UpdateEventStore + 'static)>);
+
+impl Update {
+    /// Creates a dataset update process
+    pub fn new(update_id: UpdateID, dataset_id: DatasetID, trigger: UpdateTrigger) -> Self {
+        Self(
+            Aggregate::new(
+                dataset_id.clone(),
+                UpdateInitiated {
+                    event_time: Utc::now(),
```
sergiimk, line 38: Would be nice to externalize time sources from the start

src/domain/dataset-update-flow/src/aggregates/update.rs
```diff
@@ -0,0 +1,134 @@
...
+use event_sourcing::*;
+use kamu_task_system::TaskID;
+use opendatafabric::DatasetID;
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Aggregate, Debug)]
+pub struct Update(Aggregate<UpdateState, (dyn UpdateEventStore + 'static)>);
```
sergiimk, line 29: # Updates and Triggers
`retrigger()` and `UpdateSecondaryTrigger` - can you explain the idea behind this? 

Can there be a tertiary trigger?

Is this for cases when e.g. an `AutoPolling` update was scheduled, and then user clicks "Refresh" button ... then `Manual` will become a secondary trigger?

In this case does it make sense to have a `Vec<UpdateTrigger>` in the state? e.g. in case when one derivative dataset accumulates 3 `InputDataset` triggers while being throttled?

I don't see any `UpdateTrigger`s in the `UpdateState` struct currently btw.

src/domain/dataset-update-flow/src/entities/update/update_event.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// Queued for time
+    Queued(UpdateQueued),
+    /// Secondary triger
+    SecondaryTrigger(UpdateSecondaryTrigger),
+    /// Scheduled/Rescheduled a task
+    TaskScheduled(UpdateTaskScheduled),
+    /// Succeeded task
+    TaskSucceeded(UpdateTaskSucceeded),
+    /// Failed task
+    TaskFailed(UpdateTaskFailed),
```
sergiimk, line 35: In the `Task` model I found it very nice to separate the fact that task **finished** from its **outcome** (success/failure/cancellation). I suggest doing the same here.

src/domain/dataset-update-flow/src/entities/update/update_event.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// Cancelled task
+    TaskCancelled(UpdateTaskCancelled),
```
sergiimk, line 37: Per the suggestion above - cancellation might be just an outcome of an associated task.

Also, can **update itself be cancelled** while it's still waiting for a trigger?
i.e. before an associated task is even scheduled?

src/domain/dataset-update-flow/src/entities/update/update_delay_reason.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use chrono::Duration;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum UpdateDelayReason {
+    Throttling(UpdateThrottlingDelayReason),
+    Batching(UpdateBatchingDelayReason),
+}
```
sergiimk, line 18: Not sure about this.

Would it be better to model this not as a delay reason but as an `UpdateStartCondition`?

i.e. dataset updates system continuously evaluates new incoming **triggers** and based on them decides what is the **start condition** of the next update (immediate, throttled, pending batch).

src/domain/dataset-update-flow/src/aggregates/update.rs
```diff
@@ -0,0 +1,134 @@
...
+                    dataset_id,
+                    trigger,
+                },
+            )
+            .unwrap(),
+        )
+    }
+
+    /// Postpone launching for now
+    pub fn postpone(
```
sergiimk, line 49: I'm having trouble imagining how this API function will be used.

In my mind there will be a service controlling dataset updates:
* It listens on event bus for evens like `DatasetUpdated`, `ScheduledCommand<DatasetUpdatePolling>`, etc.
* It translates thes events into `Updatetrigger`s
* Checks if there is already an `Updated` scheduled for a dataset or creates a new one
* It then makes a decision on when to run the update (based on all triggers) and updates the state of an `Update` aggregate

So I imagine something like `fn add_trigger(trigger)` and `fn set_start_condition(condition)`.

src/domain/dataset-update-flow/src/entities/update/update_trigger.rs
zaychenko-sergei: Accepted as a rule

src/domain/dataset-update-flow/src/entities/schedule/schedule_type.rs
zaychenko-sergei: This was one of the draft names as well. I have some doubts on whether having `Schedule` (value object) and `UpdateSchedule` (aggregate) is distintcitve enough.

src/domain/dataset-update-flow/src/entities/update/update_trigger.rs
zaychenko-sergei: Note: currently events in Task aggregate violate this rule.

src/domain/dataset-update-flow/src/aggregates/update.rs
zaychenko-sergei: Externalized here and in Tasks aggregate, but in the form of an argument.
The dependency on the time source service should be established in services creating domain objects.

src/domain/dataset-update-flow/src/entities/update/update_event.rs
zaychenko-sergei: Compacted

src/domain/dataset-update-flow/src/aggregates/schedule.rs
zaychenko-sergei: The specific things for now are:
 - weak link with dataset via ID
 - trigger type
 - delay type (wait condition)
 
The last 2 could be generalized, while the 1st thing might be solved via objects composition.
If there are multiple job kinds for the datasets, we can also introduce job type enumeration.

src/domain/dataset-update-flow/src/aggregates/update.rs
zaychenko-sergei: The idea is that event store of the `DatasetUpdate` may be used to draw provenance records directly. It does not change the state of the aggregate, as it's not getting rescheduled or something, but just forms the detailed history.

Secondary triggers could be:
 - refresh manually when auto polling triggered (like you suggested)
 - derived dataset re-triggered from one or more inputs as it's still waiting for scheduling

src/domain/dataset-update-flow/src/entities/update/update_event.rs
zaychenko-sergei: A good point, however, I've added an additional invariant - no task has been scheduled. Otherwise, cancellation in one domain will have to initiate cancellation in the other domain. Not impossible, but needs an EventBus already.

src/domain/dataset-update-flow/src/entities/update/update_delay_reason.rs
zaychenko-sergei: I originally saw this as a way to cover provenance history needs. Perhaps, modeling it as a state as well will be needed to implement the evaluations - this part is still to be defined, so I wan't not in a hurry to define it as state.

src/domain/dataset-update-flow/src/aggregates/update.rs
zaychenko-sergei: Accepted the suggested naming

src/domain/dataset-update-flow/src/aggregates/update.rs
zaychenko-sergei: I think I will add triggers and start conditions to the state model as soon as this will be required for scheduling decisions. This part is not prototyped yet, so for now wasn't required.

src/domain/core/src/entities/events.rs
```diff
@@ -0,0 +1,19 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use opendatafabric::DatasetID;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone)]
+pub struct DatasetEventRemoved {
```
sergiimk, line 15: 1) The method on repository is called `delete_dataset`, so let's use "Deleted" here for consistency?

2) If there will not be `DatasetEvent` enum, should we just name it `DatasetDeleted`? Perhaps having it in `events` module is enough to classify it as an event and we don't need a special naming scheme/prefix?

zaychenko-sergei: I've renamed as recommended, but the end result does not make much sense either:

```
    b.bind::<dyn AsyncEventHandler<TaskFinished>, ...;
    b.bind::<dyn AsyncEventHandler<DatasetDeleted>, ...;
    b.bind::<dyn AsyncEventHandler<UpdateScheduleEventModified>, ...;
```

My gut feeling is to have a uniform naming scheme for events, regardless of whether they belong to enum or not.

src/utils/event-bus/src/event_bus.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn sync_dispatch<TEvent: 'static + Clone>(&self, event: &TEvent) -> Result<(), InternalError> {
+        let sync_handlers = self
+            .catalog
+            .get::<dill::AllOf<dyn EventHandler<TEvent>>>()
+            .unwrap();
+
+        for sync_handler in sync_handlers {
+            sync_handler.handle(event.clone())?;
```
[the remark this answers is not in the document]
zaychenko-sergei: Agree. I was able to implement simple pass-by-reference for sync and async handlers. `Arc<TEvent>` utilized only for async closures, which is a harder case from lifetime PoV.

src/utils/event-sourcing/src/event_store.rs
```diff
@@ -16,6 +16,9 @@ use crate::{EventID, Projection};
 /// Common set of operations for an event store
 #[async_trait::async_trait]
 pub trait EventStore<Proj: Projection>: Send + Sync {
+    /// Returns all unique values of event queries
+    fn get_queries<'a>(&'a self) -> QueryStream<'a, Proj::Query>;
```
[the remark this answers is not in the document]
zaychenko-sergei: Agree. Shifted down to `UpdateScheduleEventStore` and removed the introduced argument for `save_events`

src/utils/event-bus/src/event_bus.rs
```diff
@@ -0,0 +1,182 @@
...
+            .catalog
+            .get::<dill::AllOf<dyn AsyncEventHandler<TEvent>>>()
+            .unwrap();
+
+        let async_handler_futures: Vec<_> = async_handlers
+            .iter()
+            .map(|handler| handler.handle(event.clone()))
+            .collect();
+
+        futures::future::try_join_all(async_handler_futures).await?;
```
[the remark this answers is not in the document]
zaychenko-sergei: This is interesting. We assume these events run after the aggregate transaction is complete, and we assume that falures of handlers should not affect one another. I've replaced on `join_all` and implemented reporting reduction. This is simple, but we will only report the first error, even if multiple errors happen.

```
            let results = futures::future::join_all(closure_handler_futures).await;
            results.into_iter().try_for_each(|res| res)?;
```

src/domain/core/src/entities/events.rs
sergiimk: What do you think about **requiring** that events of a certain sub-domain are in `enum`?

While we don't use even sourcing in `DatasetRepository` now - potentially we will soon (e.g. when dataset is renamed we would want old URLs in the platform to redirect to new name until old name is taken). So we will likely have a `DatasetRepositoryEvent` enum.

Also when I look at `AsyncEventHandler<UpdateScheduleEventModified>` - this seems very granular to me. Perhaps event handlers should always operate on enum level: `AsyncEventHandler<UpdateScheduleEvent>`?

While this will dispatch some events a handler might not be interested in, it had a significant added bonus: event handlers will have to `match` on event type and the compiler will force us to consider whether a newly-introduced event needs to be handled by certain logic or not.

src/utils/event-sourcing/src/event_store.rs
```diff
@@ -28,11 +28,7 @@ pub trait EventStore<Proj: Projection>: Send + Sync {
     /// The `query` argument must be the same as query passed when retrieving
     /// the events. It will be used prior to saving events to ensure that there
     /// were no concurrent updates that could've influenced this transaction.
-    async fn save_events(
-        &self,
-        query: &Proj::Query,
-        events: Vec<Proj::Event>,
-    ) -> Result<EventID, SaveEventsError>;
+    async fn save_events(&self, events: Vec<Proj::Event>) -> Result<EventID, SaveEventsError>;
```
sergiimk, line 31: What's the reason for this change?

`query` is actually very important here as it will play the key role in optimistic concurrency control (see comment above).

src/utils/event-sourcing/src/event_store.rs
```diff
@@ -44,6 +40,9 @@ pub type EventStream<'a, Event> = std::pin::Pin<
     Box<dyn tokio_stream::Stream<Item = Result<(EventID, Event), GetEventsError>> + Send + 'a>,
 >;
 
+pub type QueryStream<'a, Query> =
```
sergiimk, line 43: Is not used any more?

src/utils/event-sourcing/src/event_store.rs
```diff
@@ -64,6 +63,12 @@ pub enum GetEventsError {
     Internal(#[from] InternalError),
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum GetQueriesError {
```
sergiimk, line 67: Not used?

src/domain/dataset-update-flow/src/aggregates/update_schedule.rs
```diff
@@ -0,0 +1,93 @@
...
+                },
+            )
+            .unwrap(),
+        )
+    }
+
+    /// Pause flow
+    pub fn pause(
+        &mut self,
+        now: DateTime<Utc>,
```
sergiimk, line 40: It seems unfortunate that `now` time appears in every single method now. Should we make time source a filed in the `Aggregate`?

src/infra/dataset-update-flow-inmem/src/activity_time_wheel.rs
```diff
@@ -0,0 +1,194 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub(crate) struct ActivityTimeWheel {
+    time_slots: VecDeque<TimeSlot>,
+    activity_times: HashMap<u64, DateTime<Utc>>,
+}
+
+struct TimeSlot {
+    pub activation_time: DateTime<Utc>,
+    pub planned_activity_ids: HashSet<u64>,
```
sergiimk, line 25: How likely is it that one time slot will be shared by several updates? It seems that in our case with microsecond resolution timestamps they will never be shared unless two updates were triggered under one "transaction" that reused the same system time.

If so `planned_activity_ids` would allocate a lot more memory than actually needed.

I'm trying to understand if the complexity of `TimeSlot`s is really needed here or would a simple [min-heap](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html) be enough. Using heap would result in zero allocations (once we reach a certain "stable" array capacity).

src/domain/dataset-update-flow/src/entities/schedule/update_schedule_state.rs
```diff
@@ -0,0 +1,124 @@
...
+// by the Apache License, Version 2.0.
+
+use event_sourcing::*;
+use opendatafabric::DatasetID;
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+/// Represents the state of the task at specific point in time (projection)
```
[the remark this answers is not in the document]
zaychenko-sergei: Killed. One more reason not to write dumb comments.

src/domain/dataset-update-flow/src/entities/update/update_event.rs
```diff
@@ -0,0 +1,143 @@
...
+use chrono::{DateTime, Utc};
+use enum_variants::*;
+use kamu_task_system::{TaskID, TaskOutcome};
+use opendatafabric::{AccountID, AccountName, DatasetID};
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+/// All events that model life-cycle of a task
```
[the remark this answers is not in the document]
zaychenko-sergei: Killed

src/domain/dataset-update-flow/src/entities/update/update_trigger.rs
```diff
@@ -0,0 +1,52 @@
...
+    Manual(UpdateTriggerManual),
+    AutoPolling(UpdateTriggerAutoPolling),
+    Push(UpdateTriggerPush),
+    InputDataset(UpdateTriggerInputDataset),
```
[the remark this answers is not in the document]
zaychenko-sergei: I liked `InputUpdated`

src/domain/dataset-update-flow/src/entities/update/update_event.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum UpdateEvent {
+    /// Update initiated
+    Initiated(UpdateEventInitiated),
+    /// Start condition defined
+    StartConditionDefined(UpdateEventStartConditionDefined),
```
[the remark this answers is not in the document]
zaychenko-sergei: This is only used for derived dataset, to show why the launch is waiting in the History tab. I think it wasn't me, who suggested the term "start condition".

src/infra/dataset-update-flow-inmem/src/update_schedule_service_inmem.rs
```diff
@@ -0,0 +1,194 @@
...
+use event_bus::{AsyncEventHandler, EventBus};
+use futures::StreamExt;
+use kamu_core::events::DatasetDeleted;
+use kamu_core::SystemTimeSource;
+use kamu_dataset_update_flow::*;
+use opendatafabric::DatasetID;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct UpdateScheduleServiceInMemory {
```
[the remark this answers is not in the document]
zaychenko-sergei: Interesting observation. Yes, the service implementation is independent on where repository stores the data - in memory or elsewhere. I don't even think it's in the right project to place the service impl. But it's not clear which one would be the right one, and the same problem exists for `TaskExecutorInMemory`. Actually, for `TaskSchedulerInMemory` too, as the main queue could be also hidden behind a repository: `task_queue: VecDeque<TaskID>,`

src/utils/event-sourcing/src/event_store.rs
zaychenko-sergei: It was used, but I replaced on a more specialized `DatasetIDStream`

src/utils/event-sourcing/src/event_store.rs
zaychenko-sergei: Killed

src/infra/dataset-update-flow-inmem/src/activity_time_wheel.rs
zaychenko-sergei: About the memory, this could be later optimized with unions - single activity vs multiple activities, replace the object if more than 1 is registered.

src/domain/dataset-update-flow/src/aggregates/update_schedule.rs
zaychenko-sergei: It seems very weird to include service into aggregate.

src/utils/event-sourcing/src/event_store.rs
zaychenko-sergei: Restored. Nothing uses this, this causes confusion.

src/domain/dataset-update-flow/src/repos/update_schedule_event_store.rs
```diff
@@ -0,0 +1,23 @@
...
+use opendatafabric::DatasetID;
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait UpdateScheduleEventStore: EventStore<UpdateScheduleState> {
+    /// Returns all unique values of dataset IDs associated with schedules
+    fn list_all_dataset_ids<'a>(&'a self) -> QueryStream<'a, DatasetID>;
```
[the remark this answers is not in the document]
zaychenko-sergei: At the server startup, we must load all defined active schedules and queue their updates according to the settings.
This has to be applied to all existing schedules.

src/domain/dataset-update-flow/src/entities/update/update_trigger.rs
```diff
@@ -0,0 +1,52 @@
...
+pub struct UpdateTriggerPush {
+    // TODO: source (HTTP, MQTT, CMD, ...)
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpdateTriggerInputDataset {
+    pub input_dataset_id: DatasetID,
+    pub input_update_id: UpdateID,
```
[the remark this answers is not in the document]
zaychenko-sergei: 1) Why push itself can't become a task? The trigger might contain data such as temp file path.
2) Without task, the logic to trigger the subsequent dataset updates won't work, it would need a separate "completion endpoint", as currently we bind it to the task finishing.

src/domain/dataset-update-flow/src/entities/update/update_state.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// Represents the state of the task at specific point in time (projection)
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpdateState {
+    /// Unique update identifier
+    pub update_id: UpdateID,
+    /// Identifier of the related dataset
+    pub dataset_id: DatasetID,
+    /// Queued for time
+    pub queued_for: Option<DateTime<Utc>>,
```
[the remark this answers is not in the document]
zaychenko-sergei: I don't now that now. I will be thinking of that when implementing API for UI.

src/domain/dataset-update-flow/src/entities/update/update_state.rs
```diff
@@ -0,0 +1,178 @@
...
+use chrono::{DateTime, Utc};
+use event_sourcing::*;
+use kamu_task_system::{TaskID, TaskOutcome};
+use opendatafabric::DatasetID;
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+const MAX_UPDATE_TASKS: usize = 3;
```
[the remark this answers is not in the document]
zaychenko-sergei: Retries are not yet implemented. It can be done in different ways, and one of them is to have a domain-level constraint, as the state sees all previously schedule task IDs.

src/domain/dataset-update-flow/src/aggregates/update.rs
```diff
@@ -0,0 +1,147 @@
...
+        let event = UpdateEventStartConditionDefined {
+            event_time: now,
+            update_id: self.update_id.clone(),
+            start_condition,
+        };
+        self.apply(event)
+    }
+
+    /// Queue for time
+    pub fn queued_for_time(
```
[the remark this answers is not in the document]
zaychenko-sergei: Renamed to `activate_at`.

src/domain/dataset-update-flow/src/entities/schedule/update_schedule_state.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpdateScheduleState {
+    /// Identifier of the related dataset
+    pub dataset_id: DatasetID,
+    /// Relevance state
+    pub relevance: UpdateScheduleStateRelevance,
```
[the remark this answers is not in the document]
zaychenko-sergei: The only purpose of the `None` schedule is to somehow store schedules of deleted dataset. I could not locate any healthy pattern how to correctly model "deleted" state. Adding a boolean field `deleted` seems to be too dumb.

src/domain/dataset-update-flow/src/entities/schedule/update_schedule_event.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+/// All events that model life-cycle of a task
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum UpdateScheduleEvent {
+    Created(UpdateScheduleEventCreated),
+    Modified(UpdateScheduleEventModified),
+    DatasetRemoved(UpdateScheduleEventDatasetRemoved),
```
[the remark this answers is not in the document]
zaychenko-sergei: You can't remove a schedule. You can only pause it.

src/domain/dataset-update-flow/src/entities/schedule/schedule.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+use chrono::{DateTime, Utc};
+
+/// Represents dataset update settings
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Schedule {
+    /// No schedule
+    None,
```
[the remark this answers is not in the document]
zaychenko-sergei: Explained below, I don't know a good pattern to model deleted schedules.

src/domain/core/src/entities/events.rs
zaychenko-sergei: > What do you think about requiring that events of a certain sub-domain are in enum?

I don't see benefits. If we convert the `DatasetRepository` to ES (hard to imagine the cost of it), then this idea will be relevant.

>> Perhaps event handlers should always operate on enum level: AsyncEventHandler<UpdateScheduleEvent>?

Again, I don't see any benefits. Right now I'm very specific about the type of event I want to get, and it gives me perfect type checking. Introducing a new event in the same aggregate does not imply that old handlers would be interested in handling it as well.

src/domain/dataset-update-flow/src/entities/schedule/update_schedule_state.rs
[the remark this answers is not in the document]
zaychenko-sergei: Maybe like this:

```
struct UpdateScheduleState {
   dataset_id: DatasetID,
   schedule_state: UpdateScheduleState;
}

enum UpdateScheduleState {
   Deleted,
   Active(UpdateScheduleStateAlive),
}

struct UpdateScheduleStateAlive {
   paused: bool,
   schedule: Schedule,
}
```

and then a `Schedule` without `None` option.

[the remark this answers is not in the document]
sergiimk: As discussed in Slack, my suggestion is not to do any "deleted" state at all and simply pause the schedule upon dataset deletion.

It's important that dataset can be re-added after the deletion with the same ID, so we have to handle this gracefully.

src/domain/dataset-update-flow/src/aggregates/update_schedule.rs
sergiimk: I don't see it as a "service" in DDD sense ... just a tiny component behind an abstract interface. Dunno, I don't feel strongly about this.

src/infra/dataset-update-flow-inmem/src/activity_time_wheel.rs
sergiimk: > Just think of cron scheduling, and let's say 1000 datasets will specify they update every Friday at 8:00am

> I don't see connection to heaps. This is not a priority queue.

A sorted array of upcoming activations (in chronological order) gives us O(1) access to "next wake up time". If there happen to be 1000 datasets scheduled to update at the same time - we would just `pop_front` it a 1000 times. I saw min-heap as just a more efficient version of this.

I definitely how aggregating updates per time slots is useful in hardware simulations that we discussed, but in our case we probably don't gain anything from this batching.

src/domain/dataset-update-flow/src/entities/update/update_event.rs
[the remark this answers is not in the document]
sergiimk: The name is fine, I'm just asking if start condition can be specified during aggregate creation and appear as a field in the `Initiated` event instead of having its own event.

src/domain/dataset-update-flow/src/entities/update/update_trigger.rs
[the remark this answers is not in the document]
sergiimk: Yeah I guess push will need to be a task. Perhaps we will even run the hash consistency validation as part of it.

src/infra/dataset-update-flow-inmem/src/update_schedule_service_inmem.rs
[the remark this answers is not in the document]
sergiimk: Yeah, I arrived at the same conclusion ... let's put a pin in this and sort it out once we create a DB-backed implementation.

Cargo.toml
```diff
@@ -13,10 +14,12 @@ members = [
     "src/domain/opendatafabric",
     "src/domain/core",
     "src/domain/task-system",
+    "src/domain/dataset-update-flow",
```
sergiimk, line 17: I really dislike this name...

Let's think about our nomenclature:
- We agreed that what we currently call "task" is a low-level primitive for **doing work** which we will use for capacity management
- What you called "dataset update" here is a higher-level work that currently can be:
  - a polling / push ingest
  - a derivative transform
- But we will likely soon have other kinds of dataset upkeep *things to do*
  - Verification
  - Periodic compactions
  - Backups / replication / syncing cache with external sources
  - ...
- I think all/most of these will appear in the same UI as "dataset updates" and will be scheduled in similar ways
- Thus perhaps we need an overarching term for all of them

What do you think of having "tasks" and "jobs", and this package becoming a `job-system`?

src/domain/core/src/entities/events.rs
zaychenko-sergei: I've added an enum for dataset events, and unified naming in all 3 event domains.

zaychenko-sergei: I understand the match principle, that's very often used to safely match enums, however both strategies for events are valid, and on today I do not see cases of full aggregate handlers, but I clearly see services which pick 1 event from N aggregates.

src/domain/dataset-update-flow/src/entities/schedule/update_schedule_state.rs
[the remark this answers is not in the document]
zaychenko-sergei: Ended with the following structure:

```

pub struct UpdateScheduleState {
    pub dataset_id: DatasetID,
    pub schedule: Schedule,
    pub status: UpdateScheduleStatus,
}

pub enum UpdateScheduleStatus {
    Active,
    PausedTemporarily,
    StoppedPermanently,
}
```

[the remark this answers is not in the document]
zaychenko-sergei: Re-adding handled gracefully.

src/infra/dataset-update-flow-inmem/src/activity_time_wheel.rs
zaychenko-sergei: I understood what you mean with the heap-based priority queue. It has some extra benefit compared to ring-buffer - the complexity of inserting in the middle is O(log N), while currently we have to shift stuff.

Put it another way, a time wheel is a nice Abstract Data Type idea. But yes, we can replace slot-hashtables with a heap, while preserving more or less the same API. I'm not sure yet on cancellations easy, but let's try.

Cargo.toml
zaychenko-sergei: Let's clarify if these jobs are dataset-related only or maybe system-wide?

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
```diff
@@ -0,0 +1,733 @@
...
+            .unwrap()
+            .time_wheel
+            .activate_at(activation_time, flow_id)?;
+        Ok(())
+    }
+
+    async fn schedule_flow_task(&self, flow: &mut Flow) -> Result<(), InternalError> {
+        let logical_plan = match &flow.flow_key {
+            FlowKey::Dataset(flow_key) => match flow_key.flow_type {
+                DatasetFlowType::Update => LogicalPlan::UpdateDataset(UpdateDataset {
```
sergiimk, line 382: Should we have something like `Probe` for the flow system too, so we could test it manually without spawning expensive tasks?

Perhaps a `Probe` flow could run `Probe` tasks on a certain dataset and spawn `Probe` flows for all downstream, like a cascading update but the actual work will be a no-op?

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
```diff
@@ -0,0 +1,733 @@
...
+        loop {
+            // Do we have a timeslot scheduled?
+            let maybe_nearest_activation_time = {
+                let state = self.state.lock().unwrap();
+                state.time_wheel.nearest_activation_moment()
+            };
+
+            // Is it time to execute it yet?
+            if let Some(nearest_activation_time) = maybe_nearest_activation_time
+                && nearest_activation_time <= self.time_source.now()
```
sergiimk, line 438: I had no idea you can unpack something with `if let` and then compare it in a single statement!

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
```diff
@@ -0,0 +1,733 @@
...
+            })
+            .collect();
+
+        let results = futures::future::join_all(planned_task_futures).await;
+        results
+            .into_iter()
+            .filter(|res| res.is_err())
+            .map(|e| e.err().unwrap())
+            .for_each(|e: InternalError| {
+                tracing::error!("Scheduling flow failed: {:?}", e);
```
sergiimk, line 99: Don't format values into the message - use structured fields instead:

```rust
tracing::error!(error = ?e, "Scheduling flow failed");
```

sergiimk: We actually should have a standard of how to capture error in tracing correctly. Here's what I was doing in HTTP adapter:
```rust
tracing::error!(
                error = ?self.source,
                error_msg = %self.source,
                status_code = %self.status_code,
                "Internal API error",
);
```

This captures both the debug representation and the display message, but we'll likely want a backtrace too.

Cargo.toml
zaychenko-sergei: Implemented

src/domain/dataset-update-flow/src/entities/flow/flow_trigger.rs
```diff
@@ -0,0 +1,53 @@
...
+    Push(FlowTriggerPush),
+    InputDatasetFlow(FlowTriggerInputDatasetFlow),
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FlowTriggerManual {
+    pub initiator_account_id: AccountID,
+    pub initiator_account_name: AccountName,
```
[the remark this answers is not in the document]
zaychenko-sergei: We have to keep it for now, until we don't have impplemented `AccountID`.
Otherwise, UI will not be able to properly print the initiating accounts, task owners, ...

src/domain/dataset-update-flow/src/entities/shared/system_flow_type.rs
```diff
@@ -0,0 +1,23 @@
...
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum SystemFlowType {
+    GC,
+}
+
+impl SystemFlowType {
+    pub fn iterator() -> impl Iterator<Item = SystemFlowType> {
+        [Self::GC].iter().copied()
+    }
```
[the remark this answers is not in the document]
zaychenko-sergei: I think it's a comparable runtime cost, if not identical, but let it be.

src/domain/dataset-update-flow/src/services/flow/flow_service.rs
```diff
@@ -0,0 +1,174 @@
...
+    async fn request_manual_flow(
+        &self,
+        flow_key: FlowKey,
+        initiator_account_id: AccountID,
+        initiator_account_name: AccountName,
+    ) -> Result<FlowState, RequestFlowError>;
+
+    /// Returns states of flows of certian type associated with a given dataset
+    /// ordered by creation time from newest to oldest
+    fn list_specific_flows_by_dataset(
```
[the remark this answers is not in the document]
zaychenko-sergei: Ok

src/domain/dataset-update-flow/src/services/flow/flow_service.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[error(transparent)]
+    Internal(#[from] InternalError),
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum CancelFlowError {
+    #[error(transparent)]
+    NotFound(#[from] FlowNotFoundError),
+    #[error(transparent)]
+    Internal(#[from] InternalError),
```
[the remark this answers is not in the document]
zaychenko-sergei: It's not very clear where exactly the authorization checks would fit architecturally at the end, but likely, you are correct.

src/domain/dataset-update-flow/src/entities/flow/flow_trigger.rs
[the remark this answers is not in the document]
sergiimk: Let's create a ticket for `AccountID`s if we don't have one yet and mention removing this field innnnnn it.

src/infra/dataset-update-flow-inmem/src/dataset_flow_key.rs
```diff
@@ -0,0 +1,95 @@
...
+use kamu_dataset_update_flow::DatasetFlowType;
+use opendatafabric::DatasetID;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Eq, PartialEq, Hash, Clone)]
+pub(crate) struct OwnedDatasetFlowKey {
+    dataset_id: DatasetID,
+    flow_type: DatasetFlowType,
+}
```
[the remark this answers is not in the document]
zaychenko-sergei: Right, it's existed before switching to dynamic flow keys, and now became a duplicate.
Removed and simplified a couple uses.

src/infra/dataset-update-flow-inmem/src/dataset_flow_key.rs
```diff
@@ -0,0 +1,95 @@
...
+            flow_type,
+        }
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Copy, Clone, Hash, PartialEq, Eq)]
+pub(crate) struct BorrowedDatasetFlowKey<'a> {
+    dataset_id: &'a DatasetID,
```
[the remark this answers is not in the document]
zaychenko-sergei: I did not know it's a type that has no heap pointers, and, if so, why doesn't it implement `Copy` trait? It would simplify half of the codebase.

Still, it's a 48-byte structure, and copying it around seems to be thick in theory. Copying 12 bytes seems to be faster, but that is not guaranteed as well, as memory is read by pages anyway. At least, it won't fit into registers.

I would say "cheap copy" if it had `Rc<>` over a 32-bit buffer inside, but in our case such approach will require `Arc`, and that is unlikely cheaper than copying 48 bytes.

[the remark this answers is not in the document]
zaychenko-sergei: I'm going to keep this little optimization anyway. We need to have a working example of multi-field hashing :-)

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
zaychenko-sergei: This requires a feature flag:
`#![feature(let_chains)]`

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
zaychenko-sergei: I've corrected this one, but I haven't really started working yet on tracing in these new crates.

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
zaychenko-sergei: I would likely introducing something like this when testing.

src/infra/dataset-update-flow-inmem/src/services/dependency_graph_service_inmem.rs
```diff
@@ -0,0 +1,174 @@
...
+        Self {
+            state: Arc::new(Mutex::new(State::new())),
+        }
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+impl DependencyGraphService for DependencyGraphServiceInMemory {
```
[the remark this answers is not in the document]
zaychenko-sergei: Moved.

src/domain/dataset-update-flow/src/entities/flow/flow_start_condition.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FlowStartConditionThrottling {
+    pub interval: Duration,
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FlowStartConditionBatching {
+    pub available_records: usize,
```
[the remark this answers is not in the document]
zaychenko-sergei: This isn't used yet. I removed, but we will have to have some source of truth in this regard.
Following the example with the IoT device, we still have to create an every-second secondary trigger event and must show each in the History view.

Currently this trigger is an empty struct:

```
pub struct FlowTriggerPush {
    // TODO: source (HTTP, MQTT, CMD, ...)
}
```

similarly, we don't store yet the number of records added for the flow outcome, and can't use it yet for the derived dataset throttling:

```

pub struct FlowTriggerInputDatasetFlow {
    pub input_dataset_id: DatasetID,
    pub input_flow_type: DatasetFlowType,
    pub input_flow_id: FlowID,
}
```

Perhaps both these structures should have information about the offset change. Sum of these numbers might be the criteria we are looking for.

src/domain/dataset-update-flow/src/entities/flow/flow_trigger.rs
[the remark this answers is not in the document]
zaychenko-sergei: We have an [epic ](https://github.com/kamu-data/kamu-cli/issues/244)

src/domain/dataset-update-flow/src/entities/update/update_event.rs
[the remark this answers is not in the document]
zaychenko-sergei: The experiment failed, we have a single aggregate finally.

src/domain/dataset-update-flow/src/entities/flow/flow_start_condition.rs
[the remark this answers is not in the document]
sergiimk: Hm, I was imaging it differently:
- Kafka controller will know about batching configuration
- When Kafka queue becomes non-empty - it will trigger the flow (to show users that batch is pending)
- While more records are flowing - it will NOT touch the flow system, until the configured batch size is reached to finally trigger the flow

This way secondary trigger events will appear only for other types of trigger, e.g. "manual".

While the flow is already in the state of waiting for a batch - I don't think showing more data arriving as secondary triggers is useful, and can result in too many updates.

src/domain/flow-system/src/entities/shared/dataset_flow_type.rs
```diff
@@ -0,0 +1,24 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq)]
+pub enum DatasetFlowType {
+    Update,
```
sergiimk, line 14: Would it make sense to break this flow into `Ingest` and `ExecuteQuery`? I feel like `Update` is too generic and it would be very nice for these flow types to correspond to the names of operations the engines will be executing.

sergiimk: Even if all implementation will be share for these types of flow, having an ability to query flow system e.g. to see how much compute time we're spending on ingest vs. derivative updates could be nice to have.

src/domain/flow-system/src/services/flow/flow_service.rs
```diff
@@ -0,0 +1,174 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait FlowService: Sync + Send {
+    /// Runs the update main loop
+    async fn run(&self) -> Result<(), InternalError>;
+
+    /// Creates a new manual flow request
+    async fn request_manual_flow(
```
sergiimk, line 26: "flow request" sounds confusing as no "requests" are seen anywhere else in this sub-domain's terminology.

Perhaps a better name would be `create_or_trigger_manual_flow`?

src/domain/flow-system/src/repos/flow/flow_event_store.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait FlowEventStore: EventStore<FlowState> {
+    /// Generates new unique flow identifier
+    fn new_flow_id(&self) -> FlowID;
+
+    /// Returns the last dataset flow of certain type
+    fn get_last_specific_dataset_flow(
```
sergiimk, line 23: I see you have indulged me in replacing the word "specific" in the FlowSystem trait with `of_type`. Should we do the same here for consistency?

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -0,0 +1,316 @@
...
+    async fn ensure_datasets_initially_scanned_with(
+        &self,
+        state: &mut State,
+        repository: &dyn DependencyGraphRepository,
+    ) -> Result<(), InternalError> {
+        if state.initially_scanned {
+            return Ok(());
+        }
+
+        tracing::debug!("Started initializing dependencies graph lazily");
```
sergiimk, line 109: Instead of logging started / finished events consider separating initialization into its own function with a `#[tracing::instrument()]` macro. It will log the end event even if function terminates with error and also measure the wall clock time for us to know how long this span took.

sergiimk (commented):
Some minor suggestions, but overall I'm really happy with the state of this PR :+1:

src/domain/dataset-update-flow/src/entities/flow/flow_start_condition.rs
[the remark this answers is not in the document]
zaychenko-sergei: Assuming this suggestion, without the visibility of Kafka's queue state, let's say, showing something like  `accumulated 17,245/25,000 records`, the user will not understand what is the flow waiting for.

src/domain/flow-system/src/entities/shared/dataset_flow_type.rs
```diff
@@ -0,0 +1,24 @@
...
+    Compacting,
```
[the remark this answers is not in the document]
zaychenko-sergei: Renamed

src/domain/flow-system/src/services/flow/flow_service.rs
zaychenko-sergei: `trigger_manual_flow`

src/domain/flow-system/src/repos/flow/flow_event_store.rs
zaychenko-sergei: My bad, should have renamed both parts.

src/infra/core/src/dependency_graph_service_inmem.rs
zaychenko-sergei: Done. Although, I've left a final trace with different text, which displays node/edge count in debug mode.

src/domain/flow-system/src/entities/shared/dataset_flow_type.rs
zaychenko-sergei: I've separated at flow level, but it stays as "Update" in Tasks level.

src/infra/dataset-update-flow-inmem/src/services/flow/flow_service_inmem.rs
zaychenko-sergei: There is no mechanism to pass any parameters to the flow yet. This would be necessary for `Probe` idea to make sense.

PR #364, opened 2023-12-13
Experiment/dependencies graph

zaychenko-sergei:


src/adapter/graphql/src/mutations/metadata_chain_mut.rs
```diff
@@ -80,6 +83,35 @@ impl MetadataChainMut {
...
+
+        let mut new_upstream_ids: Vec<opendatafabric::DatasetID> = vec![];
+
+        if let opendatafabric::MetadataEvent::SetTransform(transform) = &event {
+            for new_input in transform.inputs.iter() {
+                if let Some(id) = &new_input.id {
+                    new_upstream_ids.push(id.clone());
+                } else if let Some(dataset_ref_any) = &new_input.dataset_ref {
+                    let maybe_hdl = dataset_repo
+                        .try_resolve_dataset_ref_any(dataset_ref_any)
```
sergiimk, line 97: This actually points out a big problem - we currently allow to commit `SetTransform` event with unresolved IDs. We are missing validation in MetadataChain to consider events with aliases but without IDs invalid

Created: https://github.com/kamu-data/kamu-cli/issues/366

sergiimk: I think in the API we don't need the convenience of auto-resolution of aliases. Unlike when creating from YAML snapshots, users will never be writing these events by hand.

I suggest to consider ID to be always present and error out otherwise, not to attempt an alias resolution.

src/adapter/graphql/src/mutations/metadata_chain_mut.rs
```diff
@@ -103,6 +135,18 @@ impl MetadataChainMut {
             Err(e @ domain::CommitError::Internal(_)) => return Err(e.int_err().into()),
         };
 
+        // TODO: encapsulate this inside dataset/chain
```
sergiimk, line 138: Since `dataset.commit_event()` (unlike `chain.append()`) actually advances the reference - we know that block is the new `head` and could encapsulate the event sending logic in that function (make event bus an optional parameter).

It's still hacky, but wouldn't it be less duplication and a step closer to what we were discussing as an "end state", with dataset controlling all these things?

src/domain/core/src/services/dependency_graph_service.rs
```diff
@@ -7,75 +7,111 @@
...
+pub struct DependencyGraphServiceInitializer {
+    dataset_repo: Arc<dyn DatasetRepository>,
+}
+
+impl DependencyGraphServiceInitializer {
+    pub fn new(dataset_repo: Arc<dyn DatasetRepository>) -> Self {
+        Self { dataset_repo }
+    }
+
+    pub fn browse_dependencies_of_all_datasets(&self) -> DatasetDependenciesIDStream {
```
sergiimk, line 56: `list_` would be better than `browse_`

again, trying to limit our function naming vocabulary to a base minimum

src/domain/core/src/services/dependency_graph_service.rs
```diff
@@ -7,75 +7,111 @@
...
+                    .get_dataset(&dataset_handle.as_local_ref())
+                    .await
+                    .int_err()?
+                    .get_summary(GetSummaryOpts::default())
+                    .await
+                    .int_err()?;
+
+                let mut upstream_dataset_ids = Vec::new();
+                for transform_input in summary.dependencies.iter() {
+                    if let Some(input_id) = &transform_input.id {
```
sergiimk, line 75: Strange, why would summary have an ID as optional. We never let create a transform whose aliases are unresolvable.

src/domain/core/src/services/dependency_graph_service.rs
```diff
@@ -7,75 +7,111 @@
...
+                    .dataset_repo
+                    .get_dataset(&dataset_handle.as_local_ref())
+                    .await
+                    .int_err()?
+                    .get_summary(GetSummaryOpts::default())
+                    .await
+                    .int_err()?;
+
+                let mut upstream_dataset_ids = Vec::new();
+                for transform_input in summary.dependencies.iter() {
```
sergiimk, line 74: mb replace with `summary.dependencies.filter_map(...).collect()`?

not only for functional style, but cloning IDs seems unnecessary

sergiimk: not to allocate another Vec we could also change the interface to return a stream of (ID, ID) pairs, instead of (ID, Vec<ID>)

src/domain/core/src/services/dependency_graph_service.rs
sergiimk: Summary should probably have nothing but an ID - will add this to backwards incompatible changes list.

src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
```diff
@@ -190,16 +190,51 @@ pub async fn decode_metadata_batch(
...
     for (hash, block) in metadata {
         tracing::debug!(sequence_numer = %block.sequence_number, hash = %hash, "Appending block");
+
+        if let opendatafabric::MetadataEvent::SetTransform(transform) = &block.event {
+            // Collect only the latest upstream dataset IDs
+            new_upstream_ids.clear();
+            for new_input in transform.inputs.iter() {
+                if let Some(id) = &new_input.id {
+                    new_upstream_ids.push(id.clone());
+                } else if let Some(dataset_ref_any) = &new_input.dataset_ref {
```
sergiimk, line 218: Same here - if ID is not set let's consider block invalid

src/domain/core/src/services/dependency_graph_service.rs
```diff
@@ -7,75 +7,111 @@
...
-        dataset_downstream_id: &DatasetID,
-    ) -> Result<(), RemoveDependencyError>;
+        dataset_id: &DatasetID,
+    ) -> Result<DatasetIDStream, GetUpstreamDependenciesError>;
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[dill::component(pub)]
+pub struct DependencyGraphServiceInitializer {
```
sergiimk, line 47: I quite like this separation!

Perhaps this is actually not even an "initializer" but a read-only version of `DependencyGraphRepository`. Like you said before, we might store the graph in a database or some cache and load it in memory upon startup.

And eventually we will start writing back to this cache too.

src/adapter/graphql/src/mutations/metadata_chain_mut.rs
zaychenko-sergei: Ok, simplified

src/adapter/graphql/src/mutations/metadata_chain_mut.rs
zaychenko-sergei: Reworked. Wasn't easy, as this triggers injection of `EventBus` into `DatasetFactory` and it's methods.

src/domain/core/src/services/dependency_graph_service.rs
zaychenko-sergei: Renamed

src/domain/core/src/services/dependency_graph_service.rs
zaychenko-sergei: I replaced to (ID, ID) interface, and I'm not cloning the transform input's id, yet in this form we are cloning the dataset ID, so not much of a win except a temporary `Vec`.

src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
zaychenko-sergei: Done

src/domain/core/src/services/dependency_graph_service.rs
zaychenko-sergei: Made this concept of `DependencyGraphRepository` explicit - a dedicated trait + current struct became in-memory impl.

PR #373, opened 2023-12-15
Use KAMU_WORKSPACE envvar to find a workspace

s373r:
Useful for test and debug activities -- to be able to run a debugger with the custom workspace path (e.g. from an IDE)

### DIFF
--- a/src/app/cli/src/services/workspace/workspace_service.rs
+++ b/src/app/cli/src/services/workspace/workspace_service.rs
@@ -7,12 +7,17 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use kamu::domain::*;
 
 use crate::{WorkspaceConfig, WorkspaceLayout, WorkspaceVersion};
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+const KAMU_WORKSPACE_DIR_NAME: &str = ".kamu";
+const ENV_VAR_KAMU_WORKSPACE: &str = "KAMU_WORKSPACE";
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
@@ -42,17 +47,29 @@ impl WorkspaceService {
         }
     }
 
+    fn try_read_workspace_from_env(cwd: &PathBuf) -> Option<WorkspaceLayout> {
+        std::env::var_os(ENV_VAR_KAMU_WORKSPACE)
+            .map(|workspace| cwd.join(workspace).join(KAMU_WORKSPACE_DIR_NAME))
+            .map(WorkspaceLayout::new)
+    }
+
     pub fn find_workspace() -> WorkspaceLayout {
         let cwd = Path::new(".").canonicalize().unwrap();
+
+        if let Some(ws) = Self::try_read_workspace_from_env(&cwd) {
+            return ws;
+        }
+
         if let Some(ws) = Self::find_workspace_rec(&cwd) {
             ws
         } else {
-            WorkspaceLayout::new(cwd.join(".kamu"))
+            WorkspaceLayout::new(cwd.join(KAMU_WORKSPACE_DIR_NAME))
         }
     }
 
     fn find_workspace_rec(p: &Path) -> Option<WorkspaceLayout> {
-        let root_dir = p.join(".kamu");
+        let root_dir = p.join(KAMU_WORKSPACE_DIR_NAME);
+
         if root_dir.exists() {
             Some(WorkspaceLayout::new(root_dir))
         } else if let Some(parent) = p.parent() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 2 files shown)


src/app/cli/src/services/workspace/workspace_service.rs
```diff
@@ -42,17 +47,34 @@ impl WorkspaceService {
         }
     }
 
+    fn try_read_workspace_from_env(cwd: &PathBuf) -> Option<WorkspaceLayout> {
+        let Some(raw_workspace) = std::env::var_os(ENV_VAR_KAMU_WORKSPACE) else {
+            return None;
+        };
+
+        let workspace = Path::new(&raw_workspace).to_path_buf();
+        let workspace_full_path = cwd.join(workspace).join(KAMU_WORKSPACE_DIR_NAME);
```
sergiimk, line 56: Why do we need to join to `cwd` here? It seems that it will have no effect:
- if env var is a relative path - path will be unchanged
- if env var is an absolute path - it will override `cwd`

Can this be re-written something like:
```rust
std:env::var_os(...).map(Path::new).map(WorkspaceLayout::new)
```

sergiimk (commented):
I like the idea, but I think code can be simplified a bit

src/app/cli/src/services/workspace/workspace_service.rs
s373r: @sergiimk 

> Why do we need to join to cwd here? It seems that it will have no effect

While looking at the source code, I thought we needed to have absolute paths in `WorkspaceLayout` -- in this case it seemed appropriate.

I came to this conclusion based on the following lines:
```rust
    pub fn find_workspace() -> WorkspaceLayout {
        let cwd = Path::new(".").canonicalize().unwrap(); // ---> returns a absolute path
        // ...
```

Which led to absolute paths in `WorkspaceLayout`, like:
```
$ cd
$ kamu ls

[src/app/cli/src/main.rs:13] &workspace_layout = WorkspaceLayout {
    root_dir: "/home/s373r/.kamu",
    datasets_dir: "/home/s373r/.kamu/datasets",
    repos_dir: "/home/s373r/.kamu/repos",
    cache_dir: "/home/s373r/.kamu/cache",
    run_info_dir: "/home/s373r/.kamu/run",
    version_path: "/home/s373r/.kamu/version",
    config_path: "/home/s373r/.kamu/workspace.config",
}
```

In general, yes, you are right -- we can work with relative paths as well, it doesn't create a problem:
```
➜  ~ KAMU_WORKSPACE=./src/work/kamu-data/kamu-contrib/net.rocketpool/ kamu ls
[src/app/cli/src/main.rs:13] &workspace_layout = WorkspaceLayout {
    root_dir: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu",
    datasets_dir: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu/datasets",
    repos_dir: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu/repos",
    cache_dir: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu/cache",
    run_info_dir: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu/run",
    version_path: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu/version",
    config_path: "./src/work/kamu-data/kamu-contrib/net.rocketpool/.kamu/workspace.config",
}
┌───────────────────────────────┬──────┬────────┬─────────┬──────┐
│             Name              │ Kind │ Pulled │ Records │ Size │
├───────────────────────────────┼──────┼────────┼─────────┼──────┤
│ net.rocketpool.reth.mint-burn │ Root │   -    │       - │    - │
└───────────────────────────────┴──────┴────────┴─────────┴──────┘
```

Just to double-check, Since I don't know (yet) many of the nuances, I'd like to ask for confirmation that we can allow relative `WorkspaceLayout`

sergiimk: Ah, I see. That's a really good point - some parts of our code might actually depend on this path being absolute, and we should probably assert this in `WorkspaceLayout::new`.

I still find it awkward for this function to receive `cwd` as a parameter. Perhaps what we need here is the [absolute](https://doc.rust-lang.org/std/path/fn.absolute.html) function?

s373r: > Ah, I see. That's a really good point - some parts of our code might
> actually depend on this path being absolute, and we should probably assert
> this in WorkspaceLayout::new.

I'll create a ticket later (probably can be one of my next tasks) -- what do you think regarding that?

> I still find it awkward for this function to receive cwd as a
> parameter. Perhaps what we need here is the absolute function?

the [absolute](https://doc.rust-lang.org/std/path/fn.absolute.html) function is depended on [current_dir](https://doc.rust-lang.org/std/env/fn.current_dir.html) function, which about:

> [Platform-specific behavior](https://doc.rust-lang.org/std/env/fn.current_dir.html#platform-specific-behavior)
> This function [currently](https://doc.rust-lang.org/std/io/index.html#platform-specific-behavior)
> corresponds to the getcwd function on Unix and the
> GetCurrentDirectoryW function on Windows.

so, since we have already done this call and have a ref to `cwd`, I'd like to propose just use it

s373r: Btw, I updated `try_read_workspace_from_env()` function to use the fp approach like you proposed before

s373r: @sergiimk , please review

sergiimk (approved):
LGTM :+1:

Issue #302, opened 2023-11-28
Support input limits and `has_more` loop in containerized ingest


sergiimk:
### Problem
When using containerized fetch sources ([docs](https://docs.kamu.dev/cli/ingest/ingest-examples/#using-ingest-scripts), [example dataset](https://github.com/kamu-data/kamu-contrib/tree/master/net.rocketpool)) we currently wait for the container to fully exhaust the source. 

This is bad for several reasons:
- all data ends up in one (potentially very big) data slice
- any error during ingest of the streaming source means we have to restart from scratch

Last one is especially problematic for blockchain sources that require lots of requests to the node. If a single request fails due to some transient network blip - the entire progress is lost.

### Proposal
We propose to introduce a configurable "soft limit" that `kamu` will be passing to the container to indicate how many rows it should read. As container streams data and reaches this limit - it can choose to exit and provide "I have more" indication back to `kamu`. `kamu` can then continue polling the source until it's exhausted. If any errors occurs during this process - only the progress of last iteration will be lost.

**Warning:** some high-frequency sources **may never reach the exhausted state** - they may always have some new data. We need some protection against this either on `kamu` or container level.

### Progress

- [x] Implementation & tests
  - [x] `MO1`https://github.com/kamu-data/kamu-cli/pull/377 
  - [x] Follow-up PR: https://github.com/kamu-data/kamu-cli/pull/393
  - [x] Fix found bugs
    - [x] `MO3` https://github.com/kamu-data/kamu-cli/issues/402
    - [x] `MO4` https://github.com/kamu-data/kamu-cli/issues/404
- [x] Update documentation
  - [x] `MO2` Application interface description: https://github.com/kamu-data/kamu-docs/pull/11
  - [x] An example how to terminate an infinity-like data sequence (`net.rocketpool`) (done by https://github.com/kamu-data/kamu-contrib/pull/13)
- [ ] Related activities
  - [x] `MO5` Introduce `0.2.0` (https://github.com/kamu-data/kamu-contrib/pull/13)
  - [x] `MO6` Use `0.2.0` (https://github.com/kamu-data/kamu-contrib/pull/14)
  - [x] `MO7` Update `kamu-cli/examples/reth-vs-snp500` (and others) (https://github.com/kamu-data/kamu-cli/pull/411)
  - [ ] Rebuild affected datasets
  
Merge order: `MO1`, `MO2`, etc

sergiimk:
[...]

PR #377, opened 2023-12-18
Support input limits and `has_more` loop in containerized ingest

s373r:
Related issue: https://github.com/kamu-data/kamu-cli/issues/302

### DIFF
--- a/src/infra/core/tests/tests/ingest/test_fetch.rs
+++ b/src/infra/core/tests/tests/ingest/test_fetch.rs
@@ -707,17 +707,9 @@ async fn test_fetch_container_batch_size_default() {
         ]),
         env: None,
     });
-    let listener = Arc::new(TestListener::new());
 
     let res = fetch_svc
-        .fetch(
-            "1",
-            &fetch_step,
-            None,
-            &target_path,
-            &Utc::now(),
-            Some(listener.clone()),
-        )
+        .fetch("1", &fetch_step, None, &target_path, &Utc::now(), None)
         .await
         .unwrap();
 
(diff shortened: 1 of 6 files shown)


src/infra/core/tests/tests/ingest/test_fetch.rs
```diff
@@ -637,22 +590,10 @@ async fn test_fetch_container_ok() {
...
     let fetch_step = FetchStep::Container(FetchStepContainer {
-        image: crate::utils::HttpServer::IMAGE.to_owned(),
-        command: Some(vec!["/bin/bash".to_owned()]),
-        args: Some(vec![
-            "-c".to_owned(),
-            "printf \"city,population\nA,1000\nB,2000\nC,3000\n\"".to_owned(),
-        ]),
+        image: BUSYBOX.to_owned(),
+        command: Some(vec!["printf".to_owned()]),
+        args: Some(vec!["city,population\nA,1000\nB,2000\nC,3000\n".to_owned()]),
```
sergiimk, line 596: Should we use `CSV_BATCH_OUTPUT` constant here too, now that you separated data from the command?

src/infra/core/tests/tests/ingest/test_fetch.rs
```diff
@@ -686,6 +630,387 @@ async fn test_fetch_container_ok() {
...
+const HAS_MORE_TESTER_SCRIPT: &str = indoc! {r#"
+    #!/usr/bin/env sh
+
+    set -euo pipefail
+
+    if [[ -n "${DEBUG:-}" ]]; then
+      set -x
+    fi
+
+    ROWS_COUNT=100
```
sergiimk, line 980: I found it hard to understand the tests without closely reading this script, and it's mainly because it **hardcodes** the row count at 100.

I feel like it would be a bit simpler to understand if `ROW_COUNT` was passed as a parameter, so that in tests you could clearly see how `ROW_COUNT` compares to `ODF_BATCH_SIZE`.

src/infra/core/tests/tests/ingest/test_fetch.rs
```diff
@@ -686,6 +630,387 @@ async fn test_fetch_container_ok() {
...
+            CSV_BATCH_OUTPUT
+        );
+
+        match res_2 {
+            FetchResult::Updated(x) => x.source_state,
+            _ => unreachable!(),
+        }
+    };
+    let prev_source_state = {
+        let fetch_step_3 = FetchStep::Container(FetchStepContainer {
```
sergiimk, line 888: To me it seems like two iterations (one expecting `has_more==true`  another `false`) is enough here. If 3rd and 4th iterations are somehow special please put a comment above them explaining what is being checked.

src/infra/core/tests/tests/ingest/test_fetch.rs
s373r: A good catch! Yes, can be replaced -- Done

src/infra/core/tests/tests/ingest/test_fetch.rs
s373r: It definitely makes sense -- Done

src/infra/core/tests/tests/ingest/test_fetch.rs
s373r: Added comments to each step, I hope it looks like more clear -- **Q:** What do you think?

sergiimk: I think 2nd iteration doesn't test any new boundary conditions that are not covered elsewhere and could be removed to make the test shorter, but I'm fine with it either way.

zaychenko-sergei (approved):
Plz sync with the latest master

PR #381, opened 2023-12-20
Add currentPushSources GQL endpoint

### LINKED ISSUE #372: Expose current push sources in GQL API
Similarly to:
```
metadata { current_source }
metadata { current_transform }
```
we should provide fast access to all active push sources by:
- Renaming the `current_source` to `current_polling_source`
- Introducing new array query for `current_push_sources`

sergiimk:
closes #372


zaychenko-sergei (approved):

zaychenko-sergei:
The solution is not taking disabling and updates into account yet, but it's fine at this stage.

Issue #360, opened 2023-12-13
Parallel downloads/uploads in Simple Transfer Protocol


sergiimk:
Simple Transfer Protocol is currently fully sequential, downloading/uploading one object at a time.

When downloading, iterating the metadata blocks is sequential (see open-data-fabric/open-data-fabric#22), but we could improve parallelism of data/checkpoint downloads after they are discovered from metadata.

When uploading - all objects can be uploaded with some parallelism.

The degree of parallelism should be configurable.

PR #382, opened 2023-12-20
Add parallel downloads/uploads in Simple Transfer Protocol

rmn-boiko:
Related issue #360


src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -247,130 +410,41 @@ impl SimpleTransferProtocol {
...
-                stats.dst.checkpoints_written += 1;
-                stats.src.bytes_read += checkpoint.size as usize;
-                stats.dst.bytes_written += checkpoint.size as usize;
-                listener.on_status(SyncStage::TransferData, &stats);
-            }
-        }
+        stream::iter(block_download_tasks)
+            .for_each_concurrent(
+                SimpleProtocolTransferOptions::default().max_parallel_transfers,
+                |future| async move { future.await.unwrap() },
```
sergiimk, line 445: unwrap will cause the whole process to panic if a single file download fails, let's:
- add a test that highlights this
- find a way to abort transfer upon first error
- surface the error in result

src/infra/core/Cargo.toml
```diff
@@ -99,22 +99,26 @@ tracing = "0.1"
...
-# Http file server
-http = "0.2"
-hyper = "0.14"
-tower = "0.4"
-tower-http = { version = "0.4", features = ["fs", "trace"] }
-axum = "0.6"
+[[bench]]
+name = "parallel_simple_transfer_protocol"
+harness = false
+path = "tests/benches/parallel_simple_transfer_protocol.rs"
```
s373r, line 124: please add end blank line

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -19,13 +20,40 @@ use crate::*;
...
+impl Default for SimpleProtocolTransferOptions {
+    fn default() -> Self {
+        let max_parallel_transfers =
+            match std::env::var(ENV_VAR_SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS) {
+                Ok(string_value) => string_value
+                    .parse::<usize>()
+                    .unwrap_or(DEFAULT_SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS),
+                _ => DEFAULT_SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS,
+            };
+        // Use number of allowed cuncurrent file transfers. Default value is 10
```
s373r, line 50: I guess this fragment is self-documented, so I propose to remove it

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +240,143 @@ impl SimpleTransferProtocol {
...
+            .await
+        {
+            Ok(s) => Ok(s),
+            Err(GetError::NotFound(e)) => Err(CorruptedSourceError {
+                message: "Source data file is missing".to_owned(),
+                source: Some(e.into()),
+            }
+            .into()),
+            Err(GetError::Access(e)) => Err(SyncError::Access(e)),
+            Err(GetError::Internal(e)) => Err(SyncError::Internal(e)),
```
s373r, line 266: Proposition: extract error mapping logic into `impl From<GetError> for SyncError`

For example, like `impl From<ContainsError> for GetError`

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +240,143 @@ impl SimpleTransferProtocol {
...
+                    "Data file hash declared by the source didn't match ",
+                    "the computed - this may be an indication of hashing ",
+                    "algorithm mismatch or an attempted tampering",
+                )
+                .to_owned(),
+                source: Some(e.into()),
+            }
+            .into()),
+            Err(InsertError::Access(e)) => Err(SyncError::Access(e)),
+            Err(InsertError::Internal(e)) => Err(SyncError::Internal(e)),
```
s373r, line 298: Proposition: extract error mapping logic into `impl From<InsertError> for SyncError`

For example, like `impl From<ContainsError> for GetError`

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +240,143 @@ impl SimpleTransferProtocol {
...
+        stats.src.data_slices_read += 1;
+        stats.dst.data_slices_written += 1;
+        stats.src.bytes_read += data_slice.size as usize;
+        stats.dst.bytes_written += data_slice.size as usize;
+        listener.on_status(SyncStage::TransferData, &stats);
+
+        Ok(())
+    }
+
+    async fn downlaod_block_checkpoints<'a>(
```
s373r, line 310: a typo: downl`ao`d => download

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +240,143 @@ impl SimpleTransferProtocol {
...
+            .await
+        {
+            Ok(s) => Ok(s),
+            Err(GetError::NotFound(e)) => Err(CorruptedSourceError {
+                message: "Source checkpoint file is missing".to_owned(),
+                source: Some(e.into()),
+            }
+            .into()),
+            Err(GetError::Access(e)) => Err(SyncError::Access(e)),
+            Err(GetError::Internal(e)) => Err(SyncError::Internal(e)),
```
s373r, line 333: Proposition: extract error mapping logic into `impl From<GetError> for SyncError`

For example, like `impl From<ContainsError> for GetError`

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +240,143 @@ impl SimpleTransferProtocol {
...
+                    "Checkpoint file hash declared by the source didn't ",
+                    "match the computed - this may be an indication of hashing ",
+                    "algorithm mismatch or an attempted tampering",
+                )
+                .to_owned(),
+                source: Some(e.into()),
+            }
+            .into()),
+            Err(InsertError::Access(e)) => Err(SyncError::Access(e)),
+            Err(InsertError::Internal(e)) => Err(SyncError::Internal(e)),
```
s373r, line 368: Proposition: extract error mapping logic into `impl From<InsertError> for SyncError`

For example, like impl `From<ContainsError> for GetError`

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +240,143 @@ impl SimpleTransferProtocol {
         }
     }
 
+    async fn downlaod_block_data<'a>(
```
s373r, line 243: a typo: downl`ao`d => download

s373r (changes requested):

s373r (commented):
LGTM

s373r (approved):

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -214,6 +239,91 @@ impl SimpleTransferProtocol {
...
+        stats.src.data_slices_read += 1;
+        stats.dst.data_slices_written += 1;
+        stats.src.bytes_read += data_slice.size as usize;
+        stats.dst.bytes_written += data_slice.size as usize;
+        listener.on_status(SyncStage::TransferData, &stats);
+
+        Ok(())
+    }
+
+    async fn download_block_checkpoints<'a>(
```
zaychenko-sergei, line 283: It's downloading a single checkpoint, why calling method in plural?

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -19,13 +20,39 @@ use crate::*;
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+const ENV_VAR_SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS: &str =
+    "SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS";
+const DEFAULT_SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS: usize = 10;
```
zaychenko-sergei, line 25: What is the motivation for exactly 10 ?

src/domain/core/src/services/sync_service.rs
```diff
@@ -197,6 +197,39 @@ pub enum SyncError {
...
+                )
+                .to_owned(),
+                source: Some(e.into()),
+            }
+            .into(),
+            InsertError::Access(e) => SyncError::Access(e),
+            InsertError::Internal(e) => SyncError::Internal(e),
+        }
+    }
+}
```
zaychenko-sergei, line 231: The same text of error is applied both the checkpoints and to data blocks.
Instead of this, handle errors locally at the point of current `?` operator.

src/infra/core/src/testing/dataset_test_helper.rs
```diff
@@ -74,6 +78,91 @@ impl DatasetTestHelper {
             })
             .collect()
     }
+
+    pub async fn append_random_data(
+        dataset_repo: &dyn DatasetRepository,
+        dataset_ref: impl Into<DatasetRef>,
+        is_large_file: bool,
```
zaychenko-sergei, line 85: Pass a byte size instead of this boolean, since this is not a choice of algorithm.
The only difference between "small" and "large", is the size of the byte buffer.
Maybe you could declare 2 typical sizes as constants, saying "LARGE_SIZE", "SMALL_SIZE".

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -247,130 +357,42 @@ impl SimpleTransferProtocol {
...
+                        src,
+                        dst,
+                        data_slice,
+                        trust_source_hashes,
+                        listener.clone(),
+                        stats.clone(),
+                    ))
+                    // Each function return unique future
+                    // cast future to next type to allow storing them in vector
+                        as Pin<Box<dyn Future<Output = Result<(), SyncError>> + Send>>)
```
zaychenko-sergei, line 374: This might be done easier if vectors for data block futures and checkpoint futures would be separate.
I don't think it will hurt any parallelism on a realistically large dataset.

src/infra/core/src/utils/simple_transfer_protocol.rs
```diff
@@ -247,130 +357,42 @@ impl SimpleTransferProtocol {
...
+                }
+                if let Some(checkpoint) = block_stream.event.output_checkpoint {
+                    block_download_tasks.push(Box::pin(self.download_block_checkpoints(
+                        src,
+                        dst,
+                        checkpoint,
+                        trust_source_hashes,
+                        listener.clone(),
+                        stats.clone(),
```
zaychenko-sergei, line 383: I don't understand this cloning of `stats` - here and above.
How does the stats from N futures gets merged back?

In a similar place in Smart protocol, we are using atomic counters.

src/infra/core/tests/benches/parallel_simple_transfer_protocol.rs
```diff
@@ -0,0 +1,205 @@
...
+            &pull_repo_url.as_any_ref(),
+            &dataset_alias.as_any_ref(),
+            SyncOptions::default(),
+            None,
+        )
+        .await;
+
+    // Generate additional 70 blocks in dataset to make sure next iteration will be
+    // the same as previos
+    for _ in 1..70 {
```
zaychenko-sergei, line 120: Number 70 needs to be an algorithm argument. 
It's randomly written over different test routines, and if we change the value in once place, it's easy to forget to change it in another.

src/infra/core/tests/benches/parallel_simple_transfer_protocol.rs
```diff
@@ -0,0 +1,205 @@
...
+    let dataset_alias = Arc::new(DatasetAlias::new(None, DatasetName::new_unchecked("foo")));
+    rt.spawn(server.run());
+
+    (dataset_alias, pull_repo_url, push_repo_url)
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+fn bench_with_1_parallel(c: &mut Criterion) {
+    std::env::set_var("SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS", "1");
```
zaychenko-sergei, line 145: Why not using your constant `ENV_VAR_SIMPLE_PROTOCOL_MAX_PARALLEL_TRANSFERS` ?

src/infra/core/src/utils/simple_transfer_protocol.rs
rmn-boiko: It was my idea in the first place. I was using try_join to join both data/checkpoint futures and after controlling their amount in the `try_for_each_concurrently`. I may try to implement it and see how its going

src/infra/core/src/utils/simple_transfer_protocol.rs
rmn-boiko: It is debatable, but for me, it is a good number for parallel requests that will not spam any server.

src/infra/core/src/utils/simple_transfer_protocol.rs
rmn-boiko: I've tried to do it in such a way. Here are brief results: When we have two separate vectors for data and checkpoint, we have to run two separate streams, which hurts a lot and performance decreases decently.

If we try to make the way I was thinking about(with try_join inside each iteration) we have 2 issues that somehow block us: 
1. Not each iteration we have both processes and due to this we have to write not smooth if statements where we will return try_join result or just a single future
2. In stream try_for_each_concurrent we have to write a complex error handler to handle errors correctly

zaychenko-sergei: All right, let it be. At least, we experimented.

zaychenko-sergei (approved):

PR #384, opened 2023-12-24
Extract and improve multiformats

sergiimk:
**Breaking changes** - upon review this will be merge into a staging `feature/breaking-changes` branch, not master.

In this PR:
- new crate `utils/multiformats` for all generic [multiformats](https://multiformats.io) types
  - includes lots of improvements for consistent parsing and formatting interface
  - avoids most allocations via use of stack-based buffers when encoding to binary or multibase
- `CID` type was removed and `DatasetID` is now based on `DidOdf`, which in turn uses `multiformats::DidKey`
- As per pending ODF RFC-012:
  - `Multihash::as_multibase()` will by default use `base16` encoding
  - while `DidKey::as_multibase()` continues to use `base56btc` encoding, `DidOdf` will use `base16` too

Some test will fail as changes affect the coordinator <> engine interface. I will update all the engines as the last step.


src/utils/multiformats/src/stack_string.rs
```diff
@@ -0,0 +1,69 @@
...
+// by the Apache License, Version 2.0.
+
+// A stack-allocated string to avoid allocations
+pub struct StackString<const S: usize> {
+    buf: [u8; S],
+    len: usize,
+}
+
+impl<const S: usize> StackString<S> {
+    pub fn new(buf: [u8; S], len: usize) -> Self {
```
s373r, line 17: just for my understanding:
**Q:** do we always have a stack allocated buffer here? How is it checked? Is that granted by type declaration (array decl)?

sergiimk: The idea is that higher level code that uses `StackString` has to know the max buffer size of the string its producing and pass it as const generic param.

Not fully sure what you mean by "How is it checked?". By accepting `buf` as array `[u8; S]` and not as slice `&[u8]` we always know the exact size at compile time, i.e. it's impossible to pass `[0; 10]` to a `StackString<11>`.

sergiimk: Note that this type is likely temporary. If we like the ergonomics of it and will want to extend the use of stack-allocated strings to other parts (like dataset names, aliases, references) - we will probably switch to [compact_str](https://lib.rs/crates/compact_str).

s373r: > By accepting `buf` as array `[u8; S]` and not as slice `&[u8]` we always know the exact size at compile time, i.e. it's impossible to pass `[0; 10]` to a `StackString<11>`.

yep, thanks for the clarification -- I got the idea right

zaychenko-sergei:
There's likely some UI impact, as new identifiers are noticeably longer. We will need to double check UI, mostly block history pages, and confirm the styling is acceptable - otherwise, CSS corrections of columns or maybe different representation will be necessary:

image

PR #386, opened 2023-12-26
API to read and write dataset flow configurations

zaychenko-sergei:


resources/schema.gql
```diff
@@ -286,6 +294,25 @@ type DatasetEdge {
...
+	Returns defined configuration for a flow of specified type
+	"""
+	flowConfig(datasetFlowType: DatasetFlowType!): FlowConfiguration
+}
+
+type DatasetFlowConfigsMut {
+	setConfigTimeDelta(datasetFlowType: DatasetFlowType!, paused: Boolean!, every: TimeDeltaInput!): FlowConfiguration!
+	setConfigCronExpression(datasetFlowType: DatasetFlowType!, paused: Boolean!, cronExpression: String!): FlowConfiguration!
+	setConfigBatching(datasetFlowType: DatasetFlowType!, paused: Boolean!, throttlingPeriod: TimeDeltaInput, minimalDataBatch: Int): FlowConfiguration!
+}
```
sergiimk, line 308: This seems to be begging for a `union` too, but on the input side.

I found this [closed discussion](https://github.com/graphql/graphql-spec/issues/488) about adding `inputUnion` support.

They recommend [`@oneOf`](https://github.com/graphql/graphql-spec/pull/825) as an alternative - have you checked if `async-graphql` supports it?

resources/schema.gql
```diff
@@ -1056,6 +1100,23 @@ type TemporalTable {
...
 
+type TimeDelta {
+	every: Int!
+	unit: TimeUnit!
+}
+
+input TimeDeltaInput {
+	every: Int!
+	unit: TimeUnit!
+}
```
sergiimk, line 1111: Ugh, this makes me not want to use GQL for any write operations :(

resources/schema.gql
```diff
@@ -233,6 +237,10 @@ type Dataset {
 	"""
 	metadata: DatasetMetadata!
 	"""
+	Access to the flow configurations of this dataset
+	"""
+	flowConfigs: DatasetFlowConfigs!
```
sergiimk, line 242: Would it be better to add some extra nesting here to have a whole group dedicated to flows?

I imagine we will have more than just config management, but also accessing history of flows, triggering etc. so I would prefer all of those to be grouped together instead of adding more stuff to `Dataset` node.

sergiimk: If you agree - let's make sure that mutations also reside in such a group.

src/adapter/graphql/src/queries/datasets/dataset_flow_configs.rs
```diff
@@ -0,0 +1,198 @@
...
+            .int_err()?;
+
+        Ok(maybe_flow_config.map(|flow_config| flow_config.into()))
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Enum, Copy, Clone, Eq, PartialEq)]
+#[graphql(remote = "kamu_flow_system::DatasetFlowType")]
```
sergiimk, line 71: cool

src/adapter/graphql/src/queries/datasets/dataset_flow_configs.rs
```diff
@@ -0,0 +1,198 @@
...
+        Self {
+            cron_expression: value.cron_expression,
+        }
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////
+
+#[derive(SimpleObject, Debug, Clone, PartialEq, Eq)]
+pub struct TimeDelta {
```
sergiimk, line 143: Would it make sense to create a separate file for these simple objects / enums in our `scalars` directory?

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
```diff
@@ -0,0 +1,730 @@
...
+
+    let create_result = harness.create_root_dataset().await;
+
+    let request_code = indoc!(
+        r#"
+        {
+            datasets {
+                byId (datasetId: "<id>") {
+                    flowConfigs {
+                        flowConfig (datasetFlowType: "INGEST") {
```
sergiimk, line 38: Perhaps a structure like this:
```
datasets {
  byId(...) {
    flows {
      configs {
        byType(...) {
```
would be more consistent with our current API style

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
```diff
@@ -0,0 +1,730 @@
...
+                }
+            }
+            "#
+        )
+        .replace("<id>", &id.to_string())
+        .replace("<dataset_flow_type>", dataset_flow_type)
+        .replace("<paused>", if paused { "true" } else { "false" })
+        .replace("<every>", &throttling_period.0.to_string())
+        .replace("<unit>", throttling_period.1)
+        .replace("<min_data_batch>", &minimial_data_batch.to_string())
```
sergiimk, line 726: reminds me of `jinja` :)

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
```diff
@@ -0,0 +1,730 @@
...
+        )
+        .await;
+
+    assert!(res.is_err());
+    assert_eq!(
+        res.errors
+            .into_iter()
+            .map(|e| e.message)
+            .collect::<Vec<_>>(),
+        vec!["Expected Root dataset kind".to_string()]
```
sergiimk, line 450: Do we want to represent these errors as an enum result type, or you'd rather keep them as "internal errors" for now?

sergiimk (changes requested):

resources/schema.gql
zaychenko-sergei: Yes, it makes sense, it will likely grow in this direction.

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
zaychenko-sergei: It basically is, but as you may have noticed, it greately reduced the size of the tests. Each of these templates are repeating at least 3 times.

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
zaychenko-sergei: Restructured, although it restored another level with configs.

src/adapter/graphql/src/queries/datasets/dataset_flow_configs.rs
zaychenko-sergei: Created for output part, but it's not very clear where to put `InputObject` models.

resources/schema.gql
zaychenko-sergei: At least we reused `TimeUnit` primitive. But yes, a struct cannot be a `SimpleObject` or `InputObject` at the same time.

resources/schema.gql
zaychenko-sergei: I tried, and surprisingly it worked. My main concern was more of TypeScript side.
It looks weird enough both in schema:

```
input ScheduleInput @oneOf {
	timeDelta: TimeDeltaInput
	cronExpression: String
}
```

and in TypeScript:

```
export type ScheduleInput =
    | { cronExpression: Scalars["String"]; timeDelta?: never }
    | { cronExpression?: never; timeDelta: TimeDeltaInput };
```

but it works fine:

```
const schedule_1: ScheduleInput = { timeDelta: { every: 1, unit: TimeUnit.Weeks } };
const schedule_2: ScheduleInput = { cronExpression: "* */2 * * *" };
```

and event type checking is fine - you can't write both or none.

zaychenko-sergei: So, with this weapon, I decided to have 2 instead of 3 methods - I still prefer explicit separation for root and derived datasets, as these would be differently created forms in UI:

```
type DatasetFlowConfigsMut {
	setConfigSchedule(datasetFlowType: DatasetFlowType!, paused: Boolean!, schedule: ScheduleInput!): FlowConfiguration!
	setConfigBatching(datasetFlowType: DatasetFlowType!, paused: Boolean!, throttlingPeriod: TimeDeltaInput, minimalDataBatch: Int): FlowConfiguration!
}
```

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
zaychenko-sergei: Implemented similarly to `RenameResult`

sergiimk (approved):
Looks good to me! :+1:

PR #387, opened 2023-12-26
Update to ODF schema changes

sergiimk:
This PR reflects changes proposed in open-data-fabric/open-data-fabric#71, most notably:
- New enum tagging scheme in YAML per ODF RFC-013
- New offset and block range propagation scheme per ODF RFC-014
  - This significantly simplified logic in some places!
- A much simpler structure of `TransformInput` containing `dataset_ref` (ID) and an `alias` (string for queries only)
- Some other renaming and format changes
  - Most notably a switch to `u64` sequence numbers

### DIFF
--- a/src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
+++ b/src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
@@ -68,8 +68,8 @@ pub async fn prepare_dataset_transfer_estimate(
 
     let mut data_objects_count: u32 = 0;
     let mut checkpoint_objects_count: u32 = 0;
-    let mut bytes_in_data_objects: i64 = 0;
-    let mut bytes_in_checkpoint_objects: i64 = 0;
+    let mut bytes_in_data_objects: u64 = 0;
+    let mut bytes_in_checkpoint_objects: u64 = 0;
 
     while let Some((hash, block)) = block_stream.try_next().await? {
         blocks_count += 1;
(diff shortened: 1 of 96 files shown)


sergiimk:
I recommend reviewing individual commits if you want to separate the core changes from the noise of updating the tests.

src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
```diff
@@ -220,15 +220,12 @@ pub async fn dataset_append_metadata(
...
                 } else {
-                    return Err(AppendError::InvalidBlock(
-                        AppendValidationError::InvalidEvent(InvalidEventError::new(
-                            block.event,
-                            "Transform input with unresolved ID",
-                        )),
-                    ));
+                    // Input references must be resolved to IDs here, but we
+                    // ignore the errors and let the metadata chain reject this
+                    // event
```
s373r, line 228: Previously we had:
```rust
                    return Err(AppendError::InvalidBlock(
                        AppendValidationError::InvalidEvent(InvalidEventError::new(
                            block.event,
                            "Transform input with unresolved ID",
                        )),
                    ));
```

Q: Could you please clarify: how will the event be rejected?

src/infra/core/src/repos/dataset_impl.rs
```diff
@@ -414,16 +421,13 @@ where
...
                 } else {
-                    return Err(CommitError::MetadataAppendError(AppendError::InvalidBlock(
-                        AppendValidationError::InvalidEvent(InvalidEventError::new(
-                            event,
-                            "Transform input with unresolved ID",
-                        )),
-                    )));
+                    // Normally all references must be resolved to IDs already.
+                    // We continue here while expecting MetadataChain to reject
+                    // this event.
```
s373r, line 430: Previously we had:
```rust
return Err(CommitError::MetadataAppendError(AppendError::InvalidBlock(
    AppendValidationError::InvalidEvent(InvalidEventError::new(
        event,
        "Transform input with unresolved ID",
    )),
)));
```
**Q:** Could you please clarify: how will the event be rejected?

src/infra/core/tests/tests/repos/test_metadata_chain_impl.rs
```diff
@@ -209,18 +209,25 @@ async fn test_append_prev_block_sequence_integrity_broken() {
...
-    let block_too_low =
-        MetadataFactory::metadata_block(MetadataFactory::add_data().interval(10, 19).build())
-            .prev(&hash, 1 /* should be 2 */)
-            .build();
+    let block_too_low = MetadataFactory::metadata_block(
+        MetadataFactory::add_data()
+            .new_offset_interval(10, 19)
+            .build(),
+    )
+    .prev(&hash, 1 /* should be 2 */)
```
s373r, line 229: please add a comment w/ context why it should be

src/infra/core/tests/tests/repos/test_metadata_chain_impl.rs
```diff
@@ -234,10 +241,13 @@ async fn test_append_prev_block_sequence_integrity_broken() {
...
-    let block_too_high =
-        MetadataFactory::metadata_block(MetadataFactory::add_data().interval(10, 19).build())
-            .prev(&hash, 3 /* should be 2 */)
-            .build();
+    let block_too_high = MetadataFactory::metadata_block(
+        MetadataFactory::add_data()
+            .new_offset_interval(10, 19)
+            .build(),
+    )
+    .prev(&hash, 3 /* should be 2 */)
```
s373r, line 249: please add a comment w/ context why it should be

s373r (commented):
Added several questions and minor comments

src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
sergiimk: See `metadata_chain_impl.rs` - it contains a bunch of validation logic and is our "ground source of truth" on what is considered a valid metadata.

Ideally we want to make states like this unrepresentable in the type system and I already have a ticket for that:
https://github.com/open-data-fabric/open-data-fabric/issues/73

src/infra/core/tests/tests/repos/test_metadata_chain_impl.rs
sergiimk: I think the code here is fairly self-documenting:
- the test name is about block sequence integrity
- right below we assert on sequence integrity error

PR #388, opened 2023-12-26
Remove set watermark

sergiimk:
Per the updated ODF RFC-014 (see https://github.com/open-data-fabric/open-data-fabric/pull/71) this PR proposes to remove `SetWatermark` event, replacing it with "no-data" variant of `AddData`.

Note that "no-data" variants of `AddData` already exist on `master` - they are generated when the data source returns no data but bumps the watermark or the source state during the pull.

As you (hopefully) can see from code - not having and extra event to represent this greatly simplifies things.


s373r (approved):

zaychenko-sergei (commented):
I see that a lot of tests are still failing, so this change must be considered a draft.

In particular, I don't see a commit in GraphQL schema. Tried to generate it manually, and noticed a huge number of changes. This will break UI completely. It's really hard to track those, but we must carefully write down the required breaking UI changes as tickets for the FE team.

sergiimk:
Merged into the `feature/breaking-changes` staging branch

PR #393, opened 2023-12-27
tests::ingest::test_fetch::*: fix race condition during parallel execution

s373r:
- Context: OperationID is used for container name generation. For parallel test running, we had several failed attempts to create a container with the same name.


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Fixed
+- Fixed race condition during tests parallel execution (`tests::ingest::test_fetch::*` test group)
```
sergiimk, line 9: Normally I wouldn't mention non customer facing changes like these in the changelog.

Issue #402, opened 2023-12-28
Containerized ingestion: iterative reading fail on the second iteration (DataFusion engine)


s373r:
## Specifications
- `kamu-cli` version: latest

## Steps to Reproduce the Problem
1. Use the following dataset spec (https://github.com/kamu-data/kamu-contrib/pull/13):
```yaml
---
kind: DatasetSnapshot
version: 1
content:
  name: net.rocketpool.reth.mint-burn
  kind: root
  metadata:
    - kind: setPollingSource
      fetch:
        kind: container
        image: "ghcr.io/kamu-data/fetch-net.rocketpool.reth.mint-burn:0.2.0"
        env:
          - name: ETH_NODE_PROVIDER_URL
          - name: BLOCK_BATCH_SIZE
            value: "10000"
          - name: ODF_BATCH_SIZE
            value: "100000"
      read:
        kind: ndJson
      preprocess:
        kind: sql
        engine: datafusion
        query: |
          SELECT
            'rETH' as token_symbol,
            "eventName" as event_name,
            cast("amount" as double) / pow(10.0, 18) as amount,
            cast("ethAmount" as double) / pow(10.0, 18) as eth_amount,
            "blockNumber" as block_number,
            "blockHash" as block_hash,
            to_timestamp_seconds("eventTime") as event_time,
            "transactionIndex" as transaction_index,
            "transactionHash" as transaction_hash,
            "logIndex" as log_index
          FROM input
      merge:
        kind: ledger
        primaryKey:
          - transaction_hash
    - kind: setVocab
      eventTimeColumn: block_time
```
2. Try to pull the dataset:
```shell
rm -rf .kamu \ 
    && kamu init \
    && kamu add . -r \
    && ETH_NODE_PROVIDER_URL=<INSERT_YOUR_VALUE> kamu pull --all
```

## Expected Behavior
Execute without errors

## Actual Behavior
Trimmed output:
```
[...]

PR #403, opened 2023-12-28
PollingIngestServiceImpl::ingest_loop(): fix iterative processing

s373r:
Fixes: #402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Fixed
-- Fixed race condition during tests parallel execution (`tests::ingest::test_fetch::*` test group)
-- Fixed missed `flush()`'s for `tokio::fs:file`'s
+- Fixed async file flushing issues that could result in a race condition when using containerized ingest
 - `kamu pull`: fixed containerized iterative ingestion
 
 ## [0.150.0] - 2023-12-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed race condition during tests parallel execution (`tests::ingest::test_fetch::*` test group)
 - Fixed missed `flush()`'s for `tokio::fs:file`'s
+- `kamu pull`: fixed containerized iterative ingestion
 
 ## [0.150.0] - 2023-12-27
 ### Added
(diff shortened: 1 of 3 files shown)


src/infra/core/src/ingest/polling_ingest_service_impl.rs
```diff
@@ -140,17 +140,17 @@ impl PollingIngestServiceImpl {
...
-            .int_err()?
-            .build();
-
         let mut iteration = 0;
         let mut combined_result = None;
+
         loop {
+            let ctx = ingest_common::new_session_context(self.object_store_registry.clone());
+            let mut data_writer = DataWriterDataFusion::builder(args.dataset.clone(), ctx.clone())
+                .with_metadata_state_scanned(None)
```
sergiimk, line 149: During the initialization `DataWriter` needs to restore the state of processing from metadata. This requires finding a lot of events like polling source (for merge strategy), vocabulary (for system column names), etc. - very often this will result in a full-scan of the metadata chain.

Reusing writer across iterations allows it to incrementally update its internal state, minimizing scanning - I'd hate to lose this optimization.

Please hold off on this change and let me take a closer took at Datafusion API to see if there is a better way. All we really need is to restore the context state to a point before any views were registered in it for processing or clear the state entirely.

sergiimk (changes requested):

src/infra/core/src/ingest/polling_ingest_service_impl.rs
s373r: Done, proposed another solution approach

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Fixed
 - Fixed race condition during tests parallel execution (`tests::ingest::test_fetch::*` test group)
```
sergiimk, line 9: Can you please also remove this line?

sergiimk: Also I think this fix warrants a release

src/infra/ingest-datafusion/src/writer.rs
```diff
@@ -89,6 +91,10 @@ impl DataWriterDataFusion {
         self.meta.source_event.as_ref()
     }
 
+    pub fn set_session_context(&mut self, ctx: SessionContext) {
+        self.ctx = ctx;
+    }
```
sergiimk, line 96: Not ideal, but this is the best way I see to fix this for now.
I have some ideas for refactoring the writer type soon, so we'll see if this extra complexity goes away with it.

sergiimk (approved):
:+1:

CHANGELOG.md
s373r: Will do in the next PR

s373r: Just for record, release PR: https://github.com/kamu-data/kamu-cli/pull/410

Issue #376, opened 2023-12-15
Simplify obtaining auth token in `kamu system api-server`


sergiimk:
For testing purposes it's often useful to run `kamu system api-server` and issue GQL requests and mutations.

Since we added auth, testing mutations became harder as they require auth token.

This ticket is to either add a `--get-token` flag (similar to one in `kamu ui`) to generate and print out a token. This token can then be used in request header filed of the playground:
![image](https://github.com/kamu-data/kamu-cli/assets/204914/72e52451-672c-4540-a615-c7f3c65ce03c)

PR #408, opened 2023-12-29
Add get-token flag to api-server start

rmn-boiko:
Related issue https://github.com/kamu-data/kamu-cli/issues/376

### DIFF
--- a/src/app/cli/src/cli_commands.rs
+++ b/src/app/cli/src/cli_commands.rs
@@ -437,12 +437,23 @@ pub fn get_command(
             Some(("api-server", server_matches)) => match server_matches.subcommand() {
                 None => {
                     let workspace_svc = cli_catalog.get_one::<WorkspaceService>()?;
+
+                    let current_account_subject = cli_catalog.get_one::<CurrentAccountSubject>()?;
+                    let current_account_name = match current_account_subject.as_ref() {
+                        CurrentAccountSubject::Logged(l) => l.account_name.clone(),
+                        CurrentAccountSubject::Anonymous(_) => {
+                            panic!("Cannot launch API server with anonymous account")
+                        }
+                    };
+
                     Box::new(APIServerRunCommand::new(
                         base_catalog.clone(),
                         workspace_svc.is_multi_tenant_workspace(),
                         cli_catalog.get_one()?,
                         server_matches.get_one("address").map(|a| *a),
                         server_matches.get_one("http-port").map(|p| *p),
+                        server_matches.get_flag("get-token"),
+                        current_account_name,
                     ))
                 }
                 Some(("gql-query", query_matches)) => Box::new(APIServerGqlQueryCommand::new(
--- a/src/app/cli/src/cli_commands.rs
+++ b/src/app/cli/src/cli_commands.rs
@@ -438,8 +438,6 @@ pub fn get_command(
                 None => {
                     let workspace_svc = cli_catalog.get_one::<WorkspaceService>()?;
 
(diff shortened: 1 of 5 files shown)


src/app/cli/src/explore/api_server.rs
```diff
@@ -26,14 +30,16 @@ pub struct APIServer {
     task_executor: Arc<dyn TaskExecutor>,
     flow_service: Arc<dyn FlowService>,
     time_source: Arc<dyn SystemTimeSource>,
+    access_token: String,
 }
 
 impl APIServer {
-    pub fn new(
+    pub async fn new(
```
zaychenko-sergei, line 37: 1) Can we avoid async new? It's not like a serious rule, but usually constructors should be sync and should not return errors. 

2) Perhaps, the token generation logic might be moved to command level. `APIServer` does not use it anyhow, it just generates it and provides a getter, but no other usage. This is an indication it's likely not owning this responsibility.

src/app/cli/src/explore/api_server.rs
```diff
@@ -45,6 +51,23 @@ impl APIServer {
...
+
+        let auth_svc = base_catalog.get_one::<dyn AuthenticationService>().unwrap();
+        let access_token = auth_svc
+            .login(
+                &accounts::LOGIN_METHOD_PASSWORD.to_string(),
+                serde_json::to_string::<accounts::PasswordLoginCredentials>(&login_credentials)
+                    .unwrap(),
+            )
+            .await
+            .unwrap()
```
zaychenko-sergei, line 68: Convert these unwraps into `int_err()` once you move the logic into command.
The tool should not simply die with a stack trace, instead we have a mechanism of internal errors that map to CLI error handlers.

zaychenko-sergei (changes requested):

src/app/cli/src/explore/api_server.rs
rmn-boiko: Noticed for the future as well

src/app/cli/src/cli_commands.rs
```diff
@@ -437,12 +437,17 @@ pub fn get_command(
...
+                    let current_account_subject = cli_catalog.get_one::<CurrentAccountSubject>()?;
+
                     Box::new(APIServerRunCommand::new(
                         base_catalog.clone(),
                         workspace_svc.is_multi_tenant_workspace(),
                         cli_catalog.get_one()?,
                         server_matches.get_one("address").map(|a| *a),
                         server_matches.get_one("http-port").map(|p| *p),
+                        server_matches.get_flag("get-token"),
+                        current_account_subject,
```
zaychenko-sergei, line 450: It's enough to write `cli_catalog.get_one()?`, the automatic type deduction will do the magic.

src/app/cli/src/commands/system_api_server_run_command.rs
```diff
@@ -63,6 +75,28 @@ impl Command for APIServerRunCommand {
...
+        let auth_svc = self.catalog.get_one::<dyn AuthenticationService>().unwrap();
+        let access_token = auth_svc
+            .login(
+                &accounts::LOGIN_METHOD_PASSWORD.to_string(),
+                serde_json::to_string::<accounts::PasswordLoginCredentials>(&login_credentials)
+                    .int_err()?,
+            )
+            .await
+            .int_err()?
+            .access_token;
```
zaychenko-sergei, line 99: The method `run` is getting a bit long, it basically doubled. Perhaps sub-methods for the large parts would make it more readable. At least a private method for the the new code.

zaychenko-sergei (approved):
Don't forget CHANGELOG.md

PR #409, opened 2023-12-29
Add initial Github templates

s373r:
Closes https://github.com/kamu-data/planning-internal/issues/54

### Additional materials

Documentation entry point: [About issue and pull request templates - GitHub Docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,68 @@
...
+      options:
+        - label: I have searched the existing issues
+          required: true
+  - type: textarea
+    attributes:
+      label: Version
+      placeholder: |
+        $ kamu version
+        appVersion: 0.150.0
+        buildTimestamp: 2023-12-29T10:41:15.288818966Z
```
zaychenko-sergei, line 19: Maybe also GIT hash:
`gitSha: 0db95273bed61f322259ff6f09906b8755915fbb`

zaychenko-sergei (approved):

zaychenko-sergei:
I assume similar templates will be needed in other repositories too.
We should also have it at least in the UI repository.

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+
+Closes: <#issue>
+
+<!--- Describe your changes in detail -->
+
+## Checklist before requesting a review
+- [ ] [CHANGELOG.md](./CHANGELOG.md) updated
+- [ ] API changes are backwards-compatible
+- [ ] Workspace layout changes include a migration
+- [ ] Documentation updated
```
sergiimk, line 11: Perhaps: `Documentation update PR: <link or N/A>` to force the person to link the docs changes

.github/ISSUE_TEMPLATE/bug_report.yaml
sergiimk: I believe the suggestion is for reporter to paste the entire ouput of `kamu version` command.

I suggest to say that directly, e.g. here's an example from `podamn` repo:
image

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,68 @@
...
+title: "<title>"
+labels: ["Bug"]
+body:
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      description: Please search to see if an issue already exists for the bug you encountered.
+      options:
+        - label: I have searched the existing issues
+          required: true
```
sergiimk, line 12: Will this checkbox be visible in the resulting issue?

Would it be better just to have a preamble mentioning to search for similar issues, but without some input?

Example: https://github.com/containers/podman/issues/new/choose

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,68 @@
...
+      description: A concise description of what you expected to happen.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Relevant log output
+      description: Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
+      render: sh
+    validations:
+      required: true
```
sergiimk, line 59: Logs are probably best attached as files - can we do that?

We should also be more specific and ask for `.kamu/run/kamu.log` **right after the execution of faulty command** (as we clean up the logs on start each time)

sergiimk: Probably shouldn't be required also as some info in logs may be sensitve.

sergiimk (changes requested):
Thanks for championing this @s373r!

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: > Will this checkbox be visible in the resulting issue?

yes

> Would it be better just to have a preamble mentioning to search for similar issues, but without some input?

I proceeded from the premise that it should be an obvious first action that requires minimal effort (just a click)

> Example: https://github.com/containers/podman/issues/new/choose

**Q:** do you mean [this](https://github.com/containers/podman/issues/new?assignees=&labels=kind%2Fbug%2Ctriage-needed&projects=&template=bug_report.yaml) bug template?

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: Fixed via 5232927e7e596bafca6a35caae7754e7400490a2

.github/pull_request_template.md
s373r: Fixed via 5232927e7e596bafca6a35caae7754e7400490a2

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: > Logs are probably best attached as files - can we do that?

Yes, this input area is regular Github comment area, like the following 
image


---
Fixed via b32cb992db332fc98c9bda3cdc508e092d396f5e

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,70 @@
...
+      label: Expected Behavior
+      description: A concise description of what you expected to happen.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Logs
+      description: |
+        Please attach `.kamu/run/kamu.log` right after the execution of faulty command.
+        Another option: please copy and paste all relevant log data.
```
sergiimk, line 59: `Alternatively, please copy and paste all relevant log data`

.github/ISSUE_TEMPLATE/bug_report.yaml
sergiimk: Yeah, that's the template I was referring to.

I think it's good to push people to search for similar issues, but I wouldn't want any output related to this to end up in the issue itself. I think it would be distracting.

sergiimk (approved):
Few minor suggestions, but overall I'm happy to start using this as-is.

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: Will be addressed in the follow-up PR

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: Will be addressed in the follow-up PR

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: Fixed in https://github.com/kamu-data/kamu-cli/pull/422

.github/ISSUE_TEMPLATE/bug_report.yaml
s373r: Fixed in https://github.com/kamu-data/kamu-cli/pull/422

PR #412, opened 2023-12-30
Remove Spark ingest and introduce RawQuery engine operation

sergiimk:
This PR removes legacy Spark ingest in favor of Datafusion.

Since we don't want to restrict people to only use DataFusion SQL in their `preprocess` step of the ingest we:
- rename old engine operation `execute_query()` to `execute_transform()`
- introduce a new engine operation `execute_raw_query()`
  - unlike `execute_transform()` this operation accepts only input Parquet files, and query - it works in a batch mode and with any data

This way if `preprocess` specifies an engine other than `datafusion` - the ingest process will:
- dump the intermediate results into Parquet file
- spawn an engine
- call `execute_raw_query()` operation on it
- use its result in the `merge` step

When using `datafision` the ingest will be done in-process, without spawning an engine.

Related PRs:
- ODF engine schemas: https://github.com/open-data-fabric/open-data-fabric/pull/74
- Datafusion engine update: https://github.com/kamu-data/kamu-engine-datafusion/pull/6
- Spark engine update: https://github.com/kamu-data/kamu-engine-spark/pull/8
- Flink engine update: https://github.com/kamu-data/kamu-engine-flink/pull/9


s373r (approved):
Great job! :1st_place_medal:

src/infra/core/tests/tests/ingest/test_polling_ingest.rs
```diff
@@ -954,6 +869,188 @@ async fn test_ingest_polling_datafusion_bad_column_names_rename() {
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+#[test_group::group(containerized, engine, ingest, spark)]
+#[test_log::test(tokio::test)]
+async fn test_ingest_polling_preprocess_with_spark() {
```
zaychenko-sergei, line 874: Large number of duplicated actions between Spark and Flink test.

zaychenko-sergei (approved):

src/infra/core/tests/tests/ingest/test_polling_ingest.rs
sergiimk: yeah, I tried generalizing the test, but since different engines can have slightly different queries and produce slightly different schemas - most of parameters end up to be non-reusable and the amount of lines of code saved is not that big.

Issue #396, opened 2023-12-27
Flows: Fully support CRON expressions in the Flows System


zaychenko-sergei:
See the following file: `src/domain/flow-system/src/entities/shared/schedule.rs`.

`Schedule` enum defines an operation to get the next matching time from the scheduling rule, called `next_activation_time`.
It is implemented for time-delta configurations, but is left unimplemented for CRON expressions.

To do in this ticket:
  - implement and test this key scheduling function (should be doable via simple unit tests)
  - provide validation of CRON expressions, replace data type that stores is, if necessary (currently it accepts any string)
  - if CRON expression is invalid, it should not be accepted by the GraphQL API, and should return a nicely formatted error message
  
Note: consider using a ready implementation library to interpret CRON expressions, i.e., there is https://github.com/zslayton/cron, but you are not restricted to use it, if there is anything better.

PR #413, opened 2024-01-03
Add cron schedule implementation for flow-system

rmn-boiko:
Related issue https://github.com/kamu-data/kamu-cli/issues/396


src/domain/flow-system/src/entities/shared/schedule.rs
```diff
@@ -36,12 +40,52 @@ pub struct ScheduleCronExpression {
...
+                let schedule = match CronSchedule::from_str(&ce.cron_expression) {
+                    Err(_) => {
+                        return Err(ScheduleError::InvalidCronExptression(
+                            InvalidCronExptressionError {
+                                expression: ce.cron_expression.clone(),
+                            },
+                        ));
+                    }
+                    Ok(cron_schedule) => cron_schedule,
+                };
```
zaychenko-sergei, line 80: The validation might be too late for this moment.
We want to detect it when it's configured, not when it is used.
I would expect this validation at `Schedule` construction time, and converstion into GraphQL API error, when called from there.

In addition, converting to CRON schedule each time might be expensive for every `next_activation_time` call.

zaychenko-sergei (changes requested):

src/domain/flow-system/src/entities/shared/schedule.rs
```diff
@@ -36,12 +40,52 @@ pub struct ScheduleCronExpression {
...
+#[error("Cron expression {expression} is invalid")]
+pub struct InvalidCronExptressionError {
+    pub expression: String,
+}
+
+#[derive(Error, Debug)]
+#[error("Cron expression {expression} iteration has been exceed")]
+pub struct CronExpressionIterationError {
+    pub expression: String,
+}
```
zaychenko-sergei, line 62: Try to add some tests to activate these errors. Perhaps, a good place is the GraphQL test suite for the API that configures flow schedules.

rmn-boiko: Yeah, I added such a test, but only now I realize that it will not trigger during GraphQL setup. Moving it to a higher level right now.

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
```diff
@@ -316,6 +316,51 @@ async fn test_crud_cron_root_dataset() {
...
+        .execute(
+            async_graphql::Request::new(mutation_code.clone())
+                .data(harness.catalog_authorized.clone()),
+        )
+        .await;
+    assert!(res.is_err(), "{:?}", res);
+    assert_eq!(
+        res.errors[0].message,
+        format!(
+            "Cron expression {} iteration has been exceed",
```
zaychenko-sergei, line 360: exceeded

src/domain/flow-system/src/entities/shared/schedule.rs
```diff
@@ -29,20 +33,63 @@ pub struct ScheduleTimeDelta {
...
+        match schedule.upcoming(Utc).next() {
+            Some(_) => Ok(schedule),
+            None => Err(ScheduleError::CronExpressionIterationExceed(
+                CronExpressionIterationError {
+                    expression: cron_expression.clone(),
+                },
+            )),
+        }
+    }
+    pub fn next_activation_time(&self, now: DateTime<Utc>) -> Result<DateTime<Utc>, ScheduleError> {
```
zaychenko-sergei, line 82: 1) A blank line missing
2) I don't think we should mix error types between these 2 methods, since parsing error will not happen at this stage.

src/infra/flow-system-inmem/src/services/flow/flow_service_inmem.rs
```diff
@@ -214,7 +214,7 @@ impl FlowServiceInMemory {
             None => {
                 let mut flow = self.make_new_flow(flow_key.clone(), trigger).await?;
 
-                let next_activation_time = schedule.next_activation_time(start_time);
+                let next_activation_time = schedule.next_activation_time(start_time).int_err()?;
```
zaychenko-sergei, line 217: This needs more careful handling. If CRON expression may get out of next values, we should not panic, and should simply skip scheduling the next flow, as it will never happen again.

src/infra/flow-system-inmem/tests/tests/test_schedule.rs
```diff
@@ -0,0 +1,58 @@
...
+use chrono::prelude::*;
+use chrono::Utc;
+use kamu_flow_system::{
+    CronExpressionIterationError,
+    InvalidCronExpressionError,
+    Schedule,
+    ScheduleError,
+};
+
+#[test]
```
zaychenko-sergei, line 21: Maybe, convert this to traditional unit test module located in `Schedule` file ?

zaychenko-sergei (changes requested):

zaychenko-sergei (approved):

Issue #276, opened 2023-11-14
Introduce Admin users


zaychenko-sergei:
- Core:
  - Extend AccountInfo structure with a boolean property (isAdmin)
  - Extend CurrentAccountSubject with a similar boolean property
- OSO Authorizer:
  - Modeling admin property in user structure
  - Admin users should be able to read or write any dataset regardless of ownership, privacy details.
- HTTP:
  - Support admin flag propagation in AuthenticationMiddleware
- GraphQL:
  - Extend GraphQL API equivalent of AccountInfo structure
  - Introduce AdminGuard for GraphQL queries allowed only for admins
  - Create an API sketch for admin-specific functionality without implementation - just to check the guard works fine:
  - list all tasks in the system regardless of the dataset (new API)
- CLI:
  - UsersConfig structure should hold optional isAdmin marker, false by default. It should define the built-in account’s data.
  - Support flag propagation in AccountService.
- OAuth:
  - No GitHub-driven admins model at this stage. All users stay with current privileges.

PR #414, opened 2024-01-03
Introduce Admin users

s373r:
Related issue: https://github.com/kamu-data/kamu-cli/issues/276

### DIFF
--- a/src/adapter/graphql/src/queries/admin/admin.rs
+++ b/src/adapter/graphql/src/queries/admin/admin.rs
@@ -7,18 +7,15 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-mod accounts;
-mod admin;
-mod auth;
-mod data;
-mod datasets;
-mod search;
-mod tasks;
+use crate::prelude::*;
+use crate::AdminGuard;
 
-pub(crate) use accounts::*;
-pub(crate) use admin::*;
-pub(crate) use auth::*;
-pub(crate) use data::*;
-pub(crate) use datasets::*;
-pub(crate) use search::*;
-pub(crate) use tasks::*;
+pub struct Admin;
+
+#[Object]
+impl Admin {
+    #[graphql(guard = "AdminGuard::new()")]
+    async fn self_test(&self, _ctx: &Context<'_>) -> Result<String> {
+        Ok("OK".to_string())
+    }
+}
--- a/src/app/cli/src/services/accounts/models.rs
+++ b/src/app/cli/src/services/accounts/models.rs
@@ -62,7 +62,7 @@ impl CurrentAccountIndication {
     }
 
     pub fn to_current_account_subject(&self) -> CurrentAccountSubject {
-        let is_admin = false;
+        let is_admin = true;
 
         CurrentAccountSubject::logged(AccountName::from(self.account_name.clone()), is_admin)
     }
--- a/src/app/cli/src/services/accounts/models.rs
+++ b/src/app/cli/src/services/accounts/models.rs
@@ -62,7 +62,9 @@ impl CurrentAccountIndication {
     }
 
     pub fn to_current_account_subject(&self) -> CurrentAccountSubject {
-        CurrentAccountSubject::logged(AccountName::from(self.account_name.clone()))
+        let is_admin = false;
+
+        CurrentAccountSubject::logged(AccountName::from(self.account_name.clone()), is_admin)
     }
 }
 
--- a/src/app/cli/src/services/accounts/models.rs
+++ b/src/app/cli/src/services/accounts/models.rs
@@ -42,10 +42,16 @@ pub struct CurrentAccountIndication {
     pub account_name: AccountName,
     pub user_name: String,
     pub specified_explicitly: bool,
+    is_admin: bool,
 }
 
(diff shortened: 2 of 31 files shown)


src/app/cli/src/services/accounts/models.rs
```diff
@@ -62,7 +62,9 @@ impl CurrentAccountIndication {
     }
 
     pub fn to_current_account_subject(&self) -> CurrentAccountSubject {
-        CurrentAccountSubject::logged(AccountName::from(self.account_name.clone()))
+        let is_admin = false;
+
+        CurrentAccountSubject::logged(AccountName::from(self.account_name.clone()), is_admin)
```
sergiimk, line 67: If I understand correctly this path is used by CLI. Don't we want the user launching e.g. `kamu ui` to be considered an admin of their own local workspace?

s373r: fair point, thought about it too, but decided that to be an admin you need explicit `is_admin` :thinking: 

yes, it would be better to give admin rights in this case.

s373r: Fixed via 6b6b5932a59fc506238d95c3b369b715f664ba56

src/adapter/graphql/src/queries/admin/admin.rs
```diff
@@ -0,0 +1,21 @@
...
+use crate::AdminGuard;
+
+pub struct Admin;
+
+#[Object]
+impl Admin {
+    #[graphql(guard = "AdminGuard::new()")]
+    async fn self_test(&self, _ctx: &Context<'_>) -> Result<String> {
+        Ok("OK".to_string())
+    }
```
zaychenko-sergei, line 20: It might be a nice demo for how to use the new guard class, but I also expect a very simple boolean attribute on the `Account` schema.

src/app/cli/src/services/accounts/models.rs
zaychenko-sergei: It might be a good idea if a single-tenant repository user (`"kamu"`) is an admin.
But for a multi-tenant workspace I expected this marker to come from `UsersConfig`, so that we can configure both admin and non-admin users for our demos, and that would imply having a `false` value by default, unless specified in configuration explicitly.

zaychenko-sergei: This was actually very explicitly written in the ticket:

> UsersConfig structure should hold optional isAdmin marker, false by default. It should define the built-in account’s data

zaychenko-sergei (changes requested):

zaychenko-sergei:
The tests are failing, because regenerated GraphQL schema does not match what is committed.

src/app/cli/src/services/accounts/models.rs
s373r: > But for a multi-tenant workspace I expected this marker to come from UsersConfig, so that we can configure both admin and non-admin users for our demos, and that would imply having a false value by default, unless specified in configuration explicitly
> ...
> This was actually very explicitly written in the ticket:

Yes, and it's actually done

```rust
#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
#[serde(deny_unknown_fields, rename_all = "camelCase")]
pub struct AccountInfo {
    pub account_id: AccountID,
    pub account_name: AccountName,
    pub account_type: AccountType,
    pub display_name: AccountDisplayName,
    pub avatar_url: Option<String>,
    #[serde(default)]
    pub is_admin: bool, // <--- NEW
}
```
(c) https://github.com/kamu-data/kamu-cli/pull/414/files#diff-26be9abe208e5f3eed1d8c5b313725b924d46958959f284abc4738b71f98e060R30-R35

This way, we can set administrator rights for desired user(s):
```yaml
---
version: 1
kind: CLIConfig
content:
  engine:
    runtime: podman
  users:
    predefined:
      - accountId: 1337
        accountName: s373r
        displayName: Dima Pristupa
        accountType: User
        avatarUrl: https://avatars.githubusercontent.com/u/50896974?s=200&v=4
        isAdmin: true # <--- NEW: `false` if not present
```

src/adapter/graphql/src/queries/admin/admin.rs
s373r: Done via https://github.com/kamu-data/kamu-cli/pull/414/commits/0e7bec378591a878b423ae3cd08c9ba43e0234d3

<details>
  <summary>Example request/response pair</summary>

Request:
```graphql
query {
    accounts {
        byName(name: "s373r") {
            displayName
            isAdmin
            id
            accountName
            displayName
            avatarUrl
        }
    }
}
```

Response:
```json
{
  "data": {
    "accounts": {
      "byName": {
        "displayName": "Dima Pristupa",
        "isAdmin": true,
        "id": "1337",
        "accountName": "s373r",
        "avatarUrl": "https://avatars.githubusercontent.com/u/50896974?s=200&v=4"
      }
    }
  },
  "extensions": {}
}
```
</details>

zaychenko-sergei (approved):
CHANGELOG change added to already released version, start a new one

src/app/cli/src/services/accounts/models.rs
s373r: Done

PR #415, opened 2024-01-03
GraphQL API to manually trigger dataset flow

### LINKED ISSUE #283: Flows: provide GraphQL API to schedule manual dataset flow
Must have a Writer privilege on the dataset.
Result of an API call is in moving this dataset to `FlowScheduled` state:
  - polling update
  - transformation update
Task gets scheduled and visible in the Task System’s queue.
Task gets executed when executor is ready.

zaychenko-sergei:
Closes #283


resources/schema.gql
```diff
@@ -323,6 +327,7 @@ type DatasetFlows {
 
 type DatasetFlowsMut {
 	configs: DatasetFlowConfigsMut!
+	runs: DatasetFlowRunsMut!
```
sergiimk, line 330: Since we have system flows that are not associated with some specific dataset, would it be better to make triggering a separate mutation in the root scope, just like we have for `TasksMut`?

I.e.:
```
mutation {
  tasks {
    cancelTask()
    createUpdateDatasetTask()
    createProbeTask()
  } 

  flows {
    triggerFlow()
  }
}
```
Otherwise we might end up with both.

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
```diff
@@ -0,0 +1,100 @@
...
+    async fn trigger_flow(
+        &self,
+        ctx: &Context<'_>,
+        dataset_flow_type: DatasetFlowType,
+    ) -> Result<TriggerFlowResult> {
+        if let Some(e) =
+            ensure_expected_dataset_kind(ctx, &self.dataset_handle, dataset_flow_type).await?
+        {
+            return Ok(TriggerFlowResult::IncompatibleDatasetKind(e));
+        }
```
sergiimk, line 46: Strange to see such check in API layer. Should this be checked by `FlowService` itself and the error mapped?

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        ensure_scheduling_permission(ctx, &self.dataset_handle).await?;
```
sergiimk, line 48: Same here for permissions - shouldn't this be in the service?

sergiimk (changes requested):

resources/schema.gql
sergiimk: agreed on Slack that we will treat per-dataset and admin tasks in mutations separately

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
sergiimk: agreed on slack to postpone this and perhaps later extract this into a `DatasetFlowService`

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
sergiimk: agreed on slack to postpone this and perhaps later extract this into a `DatasetFlowService`

sergiimk (approved):

rmn-boiko (approved):

Issue #330, opened 2023-12-13
Add system info / diagnose commands


sergiimk:
We could catch a lot of issues in user environment early with some simple checks. We can also gather a lot of information to simplify bug reporting.

- [x]  Info
    - [x]  kamu extended version
    - [x]  workspace path
    - [x]  docker/podman versions
- [x]  diagnose
    - [x]  Check if docker/podman is installed
    - [x]  Do a basic permission test of podman (rootless & SELinux issue)
    - [x]  Workspace consistency check
- [x]  Error messages
    - [x]  Point to install instructions
    - [x]  Suggest running `diagnose` on errors / panics
    - [x]  Suggest including `kamu system info` and filing an issue

PR #426, opened 2024-01-09
Add info and diagnose commands

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/330


src/app/cli/src/app.rs
```diff
@@ -437,8 +437,11 @@ fn configure_logging(output_config: &OutputConfig, workspace_layout: &WorkspaceL
             "\n{}",
             console::style(
                 "Oh no, looks like kamu has crashed!\n\
+                Please follow the next install intructions https://docs.kamu.dev/cli/get-started/installation/. \n\
+                Also, please, run `kamu system diagnose` to check does your environment match the requirements. \n\
```
sergiimk, line 441: Pointing to installation instructions after crash seems strange. I think only `diagnose` should do that if it detects some issues with the environment.

I suggest using this message:
```
Oh no, looks like kamu has crashed!
Please try running `kamu system diagnose` to verify your environment.
If the issue persists, help us by reporting this problem at https://github.com/kamu-data/kamu-cli/issues.
```

And also expanding @s373r's recent GH issue template to mention running `diagnose` command.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RunCheck {
+    pub container_runtime_type: ContainerRuntimeType,
+    pub is_installed: bool,
+    pub is_rootless: bool,
+    pub workspace_dir: Option<String>,
+    pub is_workspace_consitence: bool,
```
sergiimk, line 74: typo `consitence`

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,109 @@
...
+            Err(_) => false,
+        };
+
+        Self {
+            container_runtime_type: container_runtime.config.runtime,
+            is_installed: !is_container_command_err,
+            is_rootless,
+            workspace_dir: workspace_layout.root_dir.to_str().map(String::from),
+            is_workspace_consitence: current_workspace.is_in_workspace()
+                && current_workspace.is_upgrade_needed().unwrap_or(false),
```
sergiimk, line 106: - I don't think checking workspace version is enough for "consistency" check - I think we should at least iterate all datasets and check hashes of all metadata blocks and linked objects.

- Reporting inconsistency as `bool` doesn't provide any indication of what the issue is - we need to tell user what's wrong and ideally how to fix it.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,109 @@
...
+        let current_workspace = WorkspaceService::new(Arc::new(workspace_layout.clone()));
+
+        let is_rootless = match container_runtime
+            .custom_cmd(r"info".to_string())
+            .output()
+            .await
+        {
+            Ok(container_info) => String::from_utf8(container_info.stdout)
+                .unwrap()
+                .contains("rootless: true"),
```
sergiimk, line 96: Should we also tests that we can spawn some container and that we can mount a file into it?

`rootless: true` is hardly an indication that podman can function correctly.

sergiimk (changes requested):
Let's discuss design of these commands on Slack

src/app/cli/src/commands/system_info_command.rs
```diff
@@ -161,6 +147,11 @@ impl BuildInfo {
             cargo_target_triple: option_env!("VERGEN_CARGO_TARGET_TRIPLE"),
             cargo_features: option_env!("VERGEN_CARGO_FEATURES"),
             cargo_opt_level: option_env!("VERGEN_CARGO_OPT_LEVEL"),
+            workspace_dir: WorkspaceService::find_workspace()
+                .root_dir
+                .to_str()
+                .map(String::from),
+            container_version: container_version_output,
```
sergiimk, line 154: This shouldn't be a part of `BuildInfo` - this struct is meant to carry information about how `kamu` binary was compiled.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+    ) -> Result<(), CLIError> {
+        let mut out = std::io::stdout();
+
+        self.run_container_check(&mut out).await?;
+        self.run_container_pull_check(&mut out).await?;
+        self.run_container_rootless_check(&mut out).await?;
+        self.run_container_volume_mount_check(&mut out).await?;
+        self.run_workspace_consistency_check(dataset_repo, verification_svc, &mut out)
+            .await?;
+        Ok(())
```
zaychenko-sergei, line 195: Here is what it prints - the last reported line interfers with the next shell prompt:

```
sergeiz@ubuntu:~/kamu/testing/test3$ kamu system diagnose
container installed... ok
container can pull images... ok
container rootless check... ok
container volume mounts work... ok
workspace consistent... oksergeiz@ubuntu:~/kamu/testing/test3$
```

zaychenko-sergei: Plz improve the messages, so it's clearer which container system is queried.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    async fn run_container_volume_mount_check(&self, f: &mut Stdout) -> Result<(), CLIError> {
+        write!(f, "\ncontainer volume mounts work... ")?;
+        let temp_dir = tempfile::tempdir()?;
+        let cwd = Path::new(".").canonicalize()?;
+
+        let file_path = temp_dir.path().join("tmp.txt");
+        let _ = File::create(file_path)?;
+
+        let is_container_rootless = match self
```
zaychenko-sergei, line 132: This variable was copy-pasted from previous test.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn new() -> Self {
+        Self {
+            container_runtime: ContainerRuntime::default(),
+        }
+    }
+
+    async fn run_container_check(&self, f: &mut Stdout) -> Result<(), CLIError> {
+        write!(f, "container installed... ")?;
+
+        let is_container_installed = match self
```
zaychenko-sergei, line 77: When reading a name that starts from `"is_"` I would naturally expect something boolean.

src/app/cli/src/app.rs
```diff
@@ -436,10 +436,13 @@ fn configure_logging(output_config: &OutputConfig, workspace_layout: &WorkspaceL
...
             console::style(
-                "Oh no, looks like kamu has crashed!\n\
-                Please help us by reporting this problem at \
-                https://github.com/kamu-data/kamu-cli/issues"
-            ).bold()
+                "Oh no, looks like kamu has crashed!\n
+                Please try running `kamu system \
+                 diagnose` to verify your environment. \n
+                If the issue persists, help \
+                 us by reporting this problem at https://github.com/kamu-data/kamu-cli/issues"
```
zaychenko-sergei, line 443: Should we mention `kamu system info` as well as a great source of information for issue reporting?

src/app/cli/src/cli_parser.rs
```diff
@@ -1184,6 +1184,15 @@ pub fn cli() -> Command {
...
                             ]),
+                        Command::new("diagnose")
+                            .about("Run basic system diagnose check")
+                            .args([
+                                Arg::new("output-format")
+                                    .long("output-format")
+                                    .short('o')
+                                    .value_name("FMT")
+                                    .value_parser(["shell", "json", "yaml"]),
+                            ]),
```
zaychenko-sergei, line 1195: This switch is not passed into the command, and is not influencing the output report anyhow.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+pub struct RunCheck {
+    container_runtime: ContainerRuntime,
+}
+
+impl RunCheck {
+    fn new() -> Self {
+        Self {
+            container_runtime: ContainerRuntime::default(),
```
zaychenko-sergei, line 70: `ContainerRuntime` is an injectable thing as well.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+                Ok(_) => style("ok").green(),
+                Err(_) => style("failed").red(),
+            };
+
+        write!(f, "{is_container_image_pulled}")?;
+        Ok(())
+    }
+
+    async fn run_container_rootless_check(&self, f: &mut Stdout) -> Result<(), CLIError> {
+        write!(f, "\ncontainer rootless check... ")?;
```
zaychenko-sergei, line 105: Will previous `pull images` test even pass without a rootless property properly configured? If no, then move it up.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+        };
+        for dataset in datasets {
+            if let Err(_) = verification_svc
+                .verify(
+                    &dataset.as_local_ref(),
+                    (None, None),
+                    verify_options.clone(),
+                    None,
+                )
+                .await
```
zaychenko-sergei, line 171: For a large enough workspace, this is not printing any progress for a long time.
As a minimum, we must indicate the dataset we started to check.

But also, the verification should show some progress indication. See how `kamu verify` command displays it is busy - implement something similar here too, otherwise I would kill the diagnosis faster than it finishes.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        write!(f, "{is_container_rootless}")?;
+        Ok(())
+    }
+
+    async fn run_workspace_consistency_check(
+        &self,
+        dataset_repo: Arc<dyn DatasetRepository>,
+        verification_svc: Arc<dyn VerificationService>,
+        f: &mut Stdout,
```
zaychenko-sergei, line 154: Maybe we should use a more generic type, i.e. something that supports `std::io::Write` trait ? Who knows, if we would like to use some other type of I/O stream in future.

src/app/cli/src/output/output_write.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use crate::OutputConfig;
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub fn write_output<T: serde::Serialize>(
+    value: T,
+    output_config: &OutputConfig,
+    output_format: Option<impl AsRef<str>>,
+) -> Result<(), InternalError> {
```
zaychenko-sergei, line 20: It's nice you've isolated this, but it isn't used anywhere except previous place.
Perhaps, your idea was to re-use it in `kamu system diagnose`.

zaychenko-sergei (changes requested):

src/app/cli/src/cli_parser.rs
rmn-boiko: Forgot to remove it after last changes

src/app/cli/src/output/output_write.rs
rmn-boiko: Yeah, In first version it was shared for both `info` and `diagnose`. Rollback these changes

src/utils/container-runtime/src/runtime.rs
```diff
@@ -119,6 +119,12 @@ impl ContainerRuntime {
         Ok(())
     }
 
+    pub fn custom_cmd(&self, command: String) -> Command {
+        let mut cmd = self.new_command();
+        cmd.arg(command);
+        cmd
+    }
```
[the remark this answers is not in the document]
rmn-boiko: Seems no error is possible

src/app/cli/src/commands/system_diagnose_command.rs
rmn-boiko: Changed a bit logic, to check is it possible to run a container. About rootles this chack will fail even on the first step

zaychenko-sergei (approved):

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,246 @@
...
+        write!(f, "container installed... ")?;
+
+        let container_installed_check_status = match self
+            .container_runtime
+            .custom_cmd("--version".to_string())
+            .output()
+            .await
+        {
+            Ok(_) => style("ok").green(),
+            Err(_) => style("failed").red(),
```
sergiimk, line 105: In case of failure should we also display `stderr` of the command?

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,246 @@
...
+        };
+
+        write!(f, "{container_installed_check_status}")?;
+        Ok(())
+    }
+
+    async fn container_pull_check(&self, f: &mut dyn std::io::Write) -> Result<(), CLIError> {
+        write!(f, "container can pull images... ")?;
```
sergiimk, line 113: same here re podman/docker

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,246 @@
...
+        Ok(())
+    }
+
+    async fn container_volume_mount_check(
+        &self,
+        f: &mut dyn std::io::Write,
+    ) -> Result<(), CLIError> {
+        write!(f, "container volume mounts work... ")?;
+        let temp_dir = tempfile::tempdir()?;
+        let cwd = Path::new(".").canonicalize()?;
```
sergiimk, line 154: I'm surprised this worked for you. As far as I remember, on MacOS by default only the user home directory is mounted into Docker VM and files mounted from `/tmp` will not be visible to container. This is the reason we create all files for engines in `.kamu/run` directory.

Actually this might be the case with your Mac - you only check that running container succeeds succeeds, but it will even if the path does not exist.

I think we should run container with `cat my-tempfile` command and verify that container could actually read the temporary file.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,246 @@
...
+        listener: Option<Arc<VerificationMultiProgress>>,
+        f: &mut dyn std::io::Write,
+    ) -> Result<(), CLIError> {
+        write!(f, "workspace consistent... ")?;
+        let datasets: Vec<_> = self.dataset_repo.get_all_datasets().try_collect().await?;
+
+        let verify_options = VerificationOptions {
+            check_integrity: true,
+            replay_transformations: false,
+        };
```
sergiimk, line 190: When talking to @zaychenko-sergei we realized that currently these options will include verification of both physical and logical hashes of data.

Logical hashing is pretty slow currently and can take minutes on large data. I think we should extend `VerificationOptions` with `check_logical_hashes` flag and skip this validation during `system diagnose` command.

src/utils/container-runtime/src/runtime.rs
```diff
@@ -119,6 +119,12 @@ impl ContainerRuntime {
         Ok(())
     }
 
+    pub fn custom_cmd(&self, command: String) -> Command {
```
sergiimk, line 122: I would prefer to have a `version` or `info` methods instead. This type should isolate the user from knowing specific flags and arguments.

sergiimk (changes requested):

src/app/cli/src/commands/system_info_command.rs
```diff
@@ -164,3 +184,58 @@ impl BuildInfo {
...
+                version: 0,
+                root_dir: "".to_string(),
+            };
+        }
+        let root_dir = match workspace_svc.layout() {
+            Some(wl) => wl.root_dir.to_str().map(String::from).unwrap(),
+            None => "".to_string(),
+        };
+        Self {
+            version: workspace_svc.workspace_version().unwrap().unwrap().into(),
```
sergiimk, line 208: You should consider that `system info` is a command that users will run when something is going seriously wrong and their workspace might be in a bad state.

The last thing we want is for this command to crash and provide us zero information.

So I would be extremely careful with `unwrap()`s here and provide fallbacks everywhere. One missing value is OK, but crashing would be unacceptable.

src/app/cli/src/commands/system_info_command.rs
```diff
@@ -164,3 +184,58 @@ impl BuildInfo {
...
+    pub info: Value,
+}
+
+impl ContainerRuntimeInfo {
+    pub async fn collect(container_runtime: Arc<ContainerRuntime>) -> Self {
+        let container_info_output = match container_runtime.info().output().await {
+            Ok(container_info) => {
+                if container_info.status.success() {
+                    let output = String::from_utf8(container_info.stdout).unwrap();
+                    Value::from_str(output.as_str()).unwrap()
```
sergiimk, line 226: Why did you chose to handle process errors, but `unwrap()` a potential deserialization error?

src/app/cli/src/commands/system_info_command.rs
```diff
@@ -164,3 +184,58 @@ impl BuildInfo {
...
+                } else {
+                    Value::from_str("{\"error\":\"Unable to parse container runtime info result\"}")
+                        .unwrap()
```
sergiimk, line 229: Instead of parsing a string you could express this as
```rust
serde_json::json!({"error": "blah"})
```
which would construct a `Value` at compile time.

src/app/cli/src/commands/system_info_command.rs
sergiimk, line 226: I was pretty stumbled by `Value::from_str()` - it wasn't clear for me whether we are parting a string to JSON or creating a JSON string (`Value::String`) with this content.

Perhaps `serde_json::from_str(s)?` would be less ambiguous.

src/infra/core/src/verification_service_impl.rs
```diff
@@ -119,6 +120,9 @@ impl VerificationServiceImpl {
                             },
                         ));
                     } else {
+                        if !check_logical_hashes {
+                            break;
```
sergiimk, line 124: This looks like a bug. You are terminating the entire verification loop upon encountering the first block with data.

sergiimk (changes requested):
Proposed a small refactoring and identified one serious bug.

sergiimk (changes requested):
Almost there! Left a few more suggestions to polish the design of system checks.

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -0,0 +1,325 @@
...
+    fn name(&self) -> String {
+        format!(
+            "{} rootless run check",
+            self.container_runtime.config.runtime
+        )
+    }
+    async fn run(&self) -> Result<(), DiagCheckError> {
+        let run_args = RunArgs {
+            image: BUSYBOX.to_string(),
+            container_name: Some("check-rootless-run".to_owned()),
```
sergiimk, line 208: Let's prefix the name with `kamu-` so that user knew what this container is about, and also add a random suffix, i.e. `kamu-check-rootless-af4s6` - this way if somehow we end up running two of these checks in parallel (e.g. during tests) they wouldn't conlfict.

sergiimk (commented):
This is looting really good :+1: :100: 

I don't have any further comments.

@zaychenko-sergei please take another look.

zaychenko-sergei (approved):
Made some cosmetic changes to improve readability

Issue #332, opened 2023-12-13
Introduce clippy linting


sergiimk:
## Requirements

Far too long we have been ignoring this. Now that there are more people are working on this project it's time to stricten-up our standards.

- Add CI step for `clippy`
- Assess how it affects the build performance
- Propose a gradual path to introduce it over time

### Game plan
- [x] `1st phase`
  - [x] Codegen area
    - [x] Update codegen scripts https://github.com/open-data-fabric/open-data-fabric/pull/76
    - [x] Configure ignored files (`*_generated.rs`) ~~OR fix it right after generation~~ https://github.com/open-data-fabric/open-data-fabric/pull/76
      - Context: https://github.com/kamu-data/kamu-cli/pull/446#discussion_r1453216369
  - [x] Fix the current code with auto-fix `clippy` functional
    - [x] https://github.com/kamu-data/kamu-cli/pull/437
    - [x] https://github.com/kamu-data/kamu-cli/pull/439
  - [x] Fix manually rest warnings https://github.com/kamu-data/kamu-cli/pull/446
    - [x] https://github.com/kamu-data/kamu-cli/pull/447
- [x] Waiting for the merge https://github.com/kamu-data/kamu-cli/pull/407
- [ ] `2nd phase`
  - [x] `[CI]` Activate failing for the brand-new CI step if warnings present
  - [x] `[CI]` Add CI step for `cargo clippy` w/o failing (to measure times, +2m 55s to "Lint / Formatting" step) https://github.com/kamu-data/kamu-cli/pull/453
  - [x] Activate `pedantic` ruleset
    - [x] Automatic fixes
      - [x] https://github.com/kamu-data/kamu-cli/pull/460
      - [x] https://github.com/open-data-fabric/open-data-fabric/pull/79
    - [x] Manual fixes
      - [x] https://github.com/kamu-data/kamu-cli/pull/464
      - [x] https://github.com/kamu-data/kamu-cli/pull/466
    - [x] `[CI]` Enable pedantic ruleset 
  - [x] Discuss problematic/questionable `clippy` rules
  - [x] Reactivate disabled `clippy` rules
    - [x] https://github.com/kamu-data/kamu-cli/pull/473
    - [x] https://github.com/kamu-data/kamu-cli/pull/489
    - [x] https://github.com/kamu-data/kamu-cli/pull/491
  - [x] Aside activities
    - [x] Update `DEVELOPER.md` and `make lint`
[...]

PR #437, opened 2024-01-11
cargo clippy integration: major part of auto fixes

s373r:
## Description

### DIFF
--- a/src/adapter/flight-sql/src/service.rs
+++ b/src/adapter/flight-sql/src/service.rs
@@ -488,7 +488,7 @@ impl KamuFlightSqlService {
 
     fn schema_to_arrow(&self, arrow_schema: &Schema) -> Result<Vec<u8>, Status> {
         let options = IpcWriteOptions::default();
-        let pair = SchemaAsIpc::new(&arrow_schema, &options);
+        let pair = SchemaAsIpc::new(arrow_schema, &options);
         let data_gen = IpcDataGenerator::default();
         let encoded_data = data_gen.schema_to_bytes(pair.0, pair.1);
         let mut schema_bytes = vec![];
--- a/src/adapter/flight-sql/tests/tests/test_service.rs
+++ b/src/adapter/flight-sql/tests/tests/test_service.rs
@@ -86,8 +86,7 @@ async fn get_client(addr: &SocketAddr) -> FlightSqlServiceClient<Channel> {
         .await
         .expect("error connecting");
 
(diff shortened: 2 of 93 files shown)


src/adapter/graphql/src/scalars/odf_generated.rs
```diff
@@ -85,8 +85,8 @@ pub struct AttachmentEmbedded {
 impl From<odf::AttachmentEmbedded> for AttachmentEmbedded {
     fn from(v: odf::AttachmentEmbedded) -> Self {
         Self {
-            path: v.path.into(),
-            content: v.content.into(),
+            path: v.path,
```
sergiimk, line 88: Note that you are changing a code-generated file

sergiimk (commented):
Changes look pretty straightforward, just need to figure out what to do for codegen files. :+1: 

Few questions:
1) Is this all fixes or more is coming?
2) Should we add a clippy run to `make lint` target?
3) Should we run it in CI
4) Is there a way to enable clippy warnings in VSCode and if yes should we mention it in `DEVELOPER.md`

Note: I'm a little worried about diff explosion with my `breaking-changes` branch, so if there will be more radical changes coming - let's postpone them after the merge.

s373r:
@sergiimk ,

> Changes look pretty straightforward, just need to figure out what to do for codegen files. :+1: 

I have already some ideas for that (uncommited), will be implemented in the follow-up PRs (just to reduce diffs and cognitive attack to reviewers)

> Few questions:
> 1) Is this all fixes or more is coming?
yes

> 2) Should we add a clippy run to `make lint` target?
for sure, will be added

> 3) Should we run it in CI
in the separate PR

> 4) Is there a way to enable clippy warnings in VSCode and if yes should we mention it in `DEVELOPER.md`
will investigate that

> Note: I'm a little worried about diff explosion with my `breaking-changes` branch, so if there will be more radical changes coming - let's postpone them after the merge.
I'm preparing a base for automatic changes, so, we can choose different strategies if there are a lot of merge conflicts -- for example, I can extract and provide changes (patches, diffs) to run linting and fixes that you apply to your branch (for example)

src/adapter/graphql/src/scalars/odf_generated.rs
s373r: I'll exclude this file for now -- a good catch! :medal_sports:

sergiimk (approved):

Issue #383, opened 2023-12-23
Improve design of `FlowService` tests and re-enable them on Windows platform


zaychenko-sergei:
1. Currently `FlowService` tests are based on manipulating with real time. Although a big effort was taken to stabilize them, they are not guaranteed to be determenistic. 
2. Currently `FlowService` tests are disabled on Windods platform, due to less predictable timing behavior.

The idea is to implement artificial control over time in the `SystemTimeSource` trait:
 - the trait needs to be extended with a method, saying "wake me up after this interval"
 - the real implementation will have to simply sleep using Tokio async sleep
 - the stub implementation will have to block until the artifficial time shifts to desired point by another task/thread
  (use https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html as an inspiration)

Tests implementation will need to be revised around managing artifficial time instead of currentl sleep-based strategy, before test stimuli are applied on the `FlowService`.

PR #451, opened 2024-01-18
Improve design of FlowService tests and re-enable them on Windows platform

s373r:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/383

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,18 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Changed
+- Use artificial control over time in `FlowService` tests to stabilize their behavior
+
 ## [0.154.1] - 2024-01-24
 ### Fixed
 - Eliminated issue when launching transform flows for derived datasets after upstream dataset is updated.
 
 ## [0.154.0] - 2024-01-24
 ### Added
-- `kamu logout` command accepts `--all` switch to drop all current server sessions.
+- `kamu logout` command accepts `--all` switch to drop all current server sessions
 ### Changed
 - `kamu login` and `kamu logout` commands accept server URL as a positional argument, not as `-s` switch
 - Improved output reporting of `kamu logout` command
 ### Fixed
-- `kamu login` and `kamu logout` commands properly handle plaform URLs without an explicit schema (`https://` attached by default)
+- `kamu login` and `kamu logout` commands properly handle platform URLs without an explicit schema (`https://` attached by default)
 - Flow configuration API no longer crashes on specific input combinations when interval expressed in smaller time units 
    exceed minimal boundary that forms a bigger time unit
 - Fixed runtime crashes related to background execution of automatically scheduled tasks
--- a/src/adapter/http/tests/harness/server_side_harness.rs
+++ b/src/adapter/http/tests/harness/server_side_harness.rs
@@ -16,8 +16,8 @@ use kamu::domain::{
     auth,
     CurrentAccountSubject,
     DatasetRepository,
-    FakeSystemTimeSource,
     InternalError,
+    SystemTimeSourceStub,
 };
 use kamu::testing::{MockAuthenticationService, MockDatasetActionAuthorizer};
 use kamu::DatasetLayout;
@@ -44,7 +44,7 @@ pub(crate) trait ServerSideHarness {
         self.dataset_url_with_scheme(dataset_alias, "odf+http")
     }
 
-    fn system_time_source(&self) -> &FakeSystemTimeSource;
+    fn system_time_source(&self) -> &SystemTimeSourceStub;
 
     async fn api_server_run(self) -> Result<(), InternalError>;
 }
--- a/src/adapter/http/tests/harness/server_side_harness.rs
+++ b/src/adapter/http/tests/harness/server_side_harness.rs
@@ -16,8 +16,8 @@ use kamu::domain::{
     auth,
     CurrentAccountSubject,
     DatasetRepository,
+    FakeSystemTimeSource,
     InternalError,
-    SystemTimeSourceStub,
 };
 use kamu::testing::{MockAuthenticationService, MockDatasetActionAuthorizer};
 use kamu::DatasetLayout;
@@ -44,7 +44,7 @@ pub(crate) trait ServerSideHarness {
         self.dataset_url_with_scheme(dataset_alias, "odf+http")
     }
 
-    fn system_time_source(&self) -> &SystemTimeSourceStub;
+    fn system_time_source(&self) -> &FakeSystemTimeSource;
 
     async fn api_server_run(self) -> Result<(), InternalError>;
 }
--- a/src/domain/core/src/utils/time_source.rs
+++ b/src/domain/core/src/utils/time_source.rs
@@ -7,6 +7,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::cmp::{Ordering, Reverse};
+use std::collections::BinaryHeap;
 use std::sync::{Arc, Mutex};
 
 use chrono::{DateTime, Duration, Utc};
@@ -19,7 +21,7 @@ use tokio::sync::oneshot;
 pub trait SystemTimeSource: Send + Sync {
     fn now(&self) -> DateTime<Utc>;
 
-    async fn wake_after(&self, duration: Duration);
+    async fn sleep(&self, duration: Duration);
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////
@@ -34,7 +36,7 @@ impl SystemTimeSource for SystemTimeSourceDefault {
         Utc::now()
     }
 
-    async fn wake_after(&self, duration: Duration) {
+    async fn sleep(&self, duration: Duration) {
         let std_duration = duration.to_std().unwrap();
 
         tokio::time::sleep(std_duration).await;
@@ -43,40 +45,59 @@ impl SystemTimeSource for SystemTimeSourceDefault {
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
-type AwaitingCallerId = usize;
-type WokeUpCallersIds = Vec<AwaitingCallerId>;
-
 #[derive(Debug)]
 struct AwaitingCaller {
-    id: AwaitingCallerId,
-    time_to_wake_up: Duration,
+    wake_up_time: DateTime<Utc>,
     waker_tx: Option<oneshot::Sender<()>>,
 }
 
-#[derive(Debug, Default)]
+impl Eq for AwaitingCaller {}
+
+impl PartialEq<Self> for AwaitingCaller {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other) == Ordering::Equal
+    }
+}
+
+impl PartialOrd<Self> for AwaitingCaller {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for AwaitingCaller {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.wake_up_time.cmp(&other.wake_up_time)
+    }
+}
+
+#[derive(Debug)]
 struct SystemTimeSourceStubState {
-    t: Option<DateTime<Utc>>,
-    next_caller_id: AwaitingCallerId,
-    awaiting_callers: Vec<AwaitingCaller>,
+    t: DateTime<Utc>,
+    awaiting_callers: BinaryHeap<Reverse<AwaitingCaller>>,
 }
 
 #[derive(Debug, Clone)]
-pub struct SystemTimeSourceStub {
+pub struct FakeSystemTimeSource {
     state: Arc<Mutex<SystemTimeSourceStubState>>,
 }
 
-impl SystemTimeSourceStub {
+impl FakeSystemTimeSource {
     pub fn new() -> Self {
+        let state = SystemTimeSourceStubState {
+            t: Utc::now(),
+            awaiting_callers: BinaryHeap::new(),
+        };
+
         Self {
-            state: Arc::new(Mutex::new(SystemTimeSourceStubState::default())),
+            state: Arc::new(Mutex::new(state)),
         }
     }
 
     pub fn new_set(t: DateTime<Utc>) -> Self {
         let state = SystemTimeSourceStubState {
-            t: Some(t),
-            next_caller_id: 1,
-            awaiting_callers: vec![],
+            t,
+            awaiting_callers: BinaryHeap::new(),
         };
 
         Self {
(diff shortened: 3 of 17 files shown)


src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
 #[dill::interface(dyn SystemTimeSource)]
 pub struct SystemTimeSourceDefault;
 
+#[async_trait::async_trait]
 impl SystemTimeSource for SystemTimeSourceDefault {
     fn now(&self) -> DateTime<Utc> {
         Utc::now()
     }
+
+    async fn wake_after(&self, duration: Duration) {
```
sergiimk, line 37: why not call it `sleep`?

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
+        let (tx, rx) = oneshot::channel();
+
+        {
+            let mut state = self.state.lock().unwrap();
+
+            let new_awaiting_caller = AwaitingCaller {
+                id: state.next_caller_id,
```
sergiimk, line 146: Why do we need `caller_id`? I don't see it being read anywhere.

sergiimk: Ah I see it's used in tests... I wonder if instead of checking the list of caller IDs that were woken up you could `poll()` the actual caller futures and see that only those you expect are ready.

s373r: If I understand your proposition correctly, it doesn't really work for us, because in `SystemTimeSourceStub::simulate_time_passage()` we mark futures we need as ready.

I mean, we can't check if the futures are ready, because we determine when they are ready

src/domain/core/src/utils/time_source.rs
s373r: In general, yes, this is the most expected name for this method -- I chose the current name under impression of the ticket requirements.

@zaychenko-sergei , please share your thoughts -- or do I need to just use `sleep()`?

src/domain/core/src/utils/time_source.rs
sergiimk: Currently your tests do something like this:
```rust
let f = ts.wake_after(10s);
let woken_ids = ts.simulate_time_passage(15s).await;
assert_eq!(woken_ids, [1]);
timeout(f).await.unwrap();
```
What I'm proposing is:
```rust
let f = ts.wake_after(10s);
ts.simulate_time_passage(15s).await;
assert_eq!(f.poll(), Poll::Ready);
```

Now, I'm not 100% sure this will work, but if it does:
- it removes the need for any kind of IDs
- it tells us exactly that this future was completed - you'll be able to remove all the `timeout` logic

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
     }
 
     pub fn unset(&self) {
-        *self.t.lock().unwrap() = None;
+        let mut state = self.state.lock().unwrap();
+
+        state.t = None;
+    }
+
+    pub fn simulate_time_passage(&self, time_quantum: Duration) -> WokeUpCallersIds {
```
sergiimk, line 99: Let's call this method `advance`. We already know we're working with a time source, so "time" is redundant, we also know that we are using a stub which simulates things, so the name just looks too verbose.

src/domain/core/src/utils/time_source.rs
s373r: I see -- a good idea, will check

src/domain/core/src/utils/time_source.rs
s373r: Agreed -- will rename

src/domain/core/src/utils/time_source.rs
s373r: `advance()` was chosen
based on https://github.com/kamu-data/kamu-cli/pull/451#discussion_r1459404953

src/domain/core/src/utils/time_source.rs
s373r: I've experimented with different options, in general, if I allocate more time, it can be done but with ugly workarounds and honestly, not a guaranteed result.

This way we will add a lot of complexity, both in writing/reading for tests (requires `'static` `Future`'s, creating execution contexts for `Future::poll()`, absorbing futures as values in future runners), and in support.

I'd like to suggest keeping the current approach -- agreed that returning IDs is not the most elegant solution, but I think we can afford some liberties for test code

sergiimk: Reached out on Slack with some ideas.

For posterity, I managed to express one of the tests like this:
```rust
#[tokio::test]
async fn test_stub_wake_after_with_several_simulate_time_passage_x() {
    let t0 = point_in_time_in_a_parallel_universe();
    let system_time_source = SystemTimeSourceStub::new_set(t0);

    // 1. Start waiting 4 parallel futures:
    //    - a)  60 seconds
    //    - b)  90 seconds
    //    - c) 120 seconds
    //    - d) 120 seconds
    let mut futures = vec![
        Some(system_time_source.wake_after(Duration::seconds(60))),
        Some(system_time_source.wake_after(Duration::seconds(90))),
        Some(system_time_source.wake_after(Duration::seconds(120))),
        Some(system_time_source.wake_after(Duration::seconds(120))),
    ];

    // All futures are not ready
    assert_eq!(
        check_futures_for_completion(&mut futures),
        [false, false, false, false]
    );

    // 2. Simulate +30 seconds:
    //    - a) +00:30/01:00 - waiting
    //    - b) +00:30/01:30 - waiting
    //    - c) +00:30/02:00 - waiting
    //    - d) +00:30/02:00 - waiting
    system_time_source.advance(Duration::seconds(30));
    assert_eq!(
        check_futures_for_completion(&mut futures),
        [false, false, false, false]
    );

    // 3. Simulate +30 seconds:
    //    - a) +01:00/01:00 - done
    //    - b) +01:00/01:30 - waiting
    //    - c) +01:00/02:00 - waiting
    //    - d) +01:00/02:00 - waiting
    system_time_source.advance(Duration::seconds(30));
    assert_eq!(
        check_futures_for_completion(&mut futures),
        [true, false, false, false]
    );

    // 4. Simulate +30 seconds:
    //    - a) +01:30/01:00 - done (before)
    //    - b) +01:30/01:30 - done
    //    - c) +01:30/02:00 - waiting
    //    - d) +01:30/02:00 - waiting
    system_time_source.advance(Duration::seconds(30));
    assert_eq!(
        check_futures_for_completion(&mut futures),
        [true, true, false, false]
    );

    // 4. Simulate +30 seconds:
    //    - a) +01:30/01:00 - done (before)
    //    - b) +01:30/01:30 - done (before)
    //    - c) +01:30/02:00 - done
    //    - d) +01:30/02:00 - done
    system_time_source.advance(Duration::seconds(30));
    assert_eq!(
        check_futures_for_completion(&mut futures),
        [true, true, true, true]
    );
}
```
with this function doing the heavy lifting:
```rust
fn check_futures_for_completion(
    futures: &mut Vec<Option<Pin<Box<dyn Future<Output = ()> + Send + '_>>>>,
) -> Vec<bool> {
    let mut res = Vec::new();
    for i in 0..futures.len() {
        if let Some(f) = &mut futures[i] {
            if f.now_or_never().is_some() {
                res.push(true);
                futures[i] = None;
            } else {
                res.push(false);
            }
        } else {
            res.push(true);
        }
    }
    res
}
```

src/infra/flow-system-inmem/tests/tests/test_flow_service_inmem.rs
```diff
@@ -1287,6 +1297,39 @@ impl FlowHarness {
...
+        fn div_up(a: i64, b: i64) -> i64 {
+            (a + (b - 1)) / b
+        }
+
+        const TIME_INCREMENT: Duration = Duration::milliseconds(SCHEDULING_ALIGNMENT_MS);
+        const CONTEXT_SWITCHING_SLEEP_DURATION: StdDuration = if cfg!(windows) {
+            // Windows sleep() is accurate up to about ~20-50 milliseconds
+            StdDuration::from_millis(50)
+        } else {
+            StdDuration::from_millis(5)
```
zaychenko-sergei, line 1319: Does it matter for faked time at all?

s373r: yes, we need pauses so that asychronous code has time to work (context switching)

also, when we have multiple calls for a very short period of time, we cannot bypass them with a single `sleep` call

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
-            t: Arc::new(Mutex::new(Some(t))),
+            state: Arc::new(Mutex::new(state)),
         }
     }
 
     pub fn set(&self, t: DateTime<Utc>) {
-        *self.t.lock().unwrap() = Some(t);
+        let mut state = self.state.lock().unwrap();
+
+        state.t = Some(t);
```
zaychenko-sergei, line 90: How does `set` method integrate with the awaiting callers?
Also, what if we set a time that is earlier than previous value of `t` ?

src/domain/core/src/utils/time_source.rs
zaychenko-sergei: Name is a least of interesting problems here. I'm okay with any of these suggestions.

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+type AwaitingCallerId = usize;
+type WokeUpCallersIds = Vec<AwaitingCallerId>;
+
+#[derive(Debug)]
+struct AwaitingCaller {
+    id: AwaitingCallerId,
+    time_to_wake_up: Duration,
```
zaychenko-sergei, line 52: It's not a "time", it's an offset from `t`. If the idea was to store time, the type should not be `Duration`.

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
 pub struct SystemTimeSourceStub {
-    t: Arc<Mutex<Option<DateTime<Utc>>>>,
+    state: Arc<Mutex<SystemTimeSourceStubState>>,
 }
 
 impl SystemTimeSourceStub {
     pub fn new() -> Self {
         Self {
-            t: Arc::new(Mutex::new(None)),
+            state: Arc::new(Mutex::new(SystemTimeSourceStubState::default())),
```
zaychenko-sergei, line 71: I have big doubts that with the logic we are introducing it's correct to allow `None` for `t`.

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
+    t: Option<DateTime<Utc>>,
+    next_caller_id: AwaitingCallerId,
+    awaiting_callers: Vec<AwaitingCaller>,
+}
+
 #[derive(Debug, Clone)]
 pub struct SystemTimeSourceStub {
-    t: Arc<Mutex<Option<DateTime<Utc>>>>,
+    state: Arc<Mutex<SystemTimeSourceStubState>>,
 }
```
zaychenko-sergei, line 66: Looking at the introduced algorithm, it's no longer correct to call this time source a "stub". It's becoming a "fake". Perhaps some of the very simple tests that only need to freeze time for predictable reports may be fine with the old stub, while the flow system requires a totally different test double with what is suggested here.

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
+
+                tx.send(()).unwrap();
+            }
+
+            !is_time_up
+        });
+
+        if let Some(ref mut t) = state.t {
+            *t += time_quantum;
+        }
```
zaychenko-sergei, line 122: Maybe an absolute awakening time would be easier to store rather then storing offsets and moving the after every `advance` ?

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +28,132 @@ pub trait SystemTimeSource: Send + Sync {
...
+        let mut state = self.state.lock().unwrap();
+
+        state.t = None;
+    }
+
+    pub fn advance(&self, time_quantum: Duration) -> WokeUpCallersIds {
+        let mut state = self.state.lock().unwrap();
+        let mut woke_up_callers_ids = vec![];
+
+        state.awaiting_callers.retain_mut(|awaiting_caller| {
```
zaychenko-sergei, line 103: If I understood correctly, this is scanning the entire list of awakenings to apply the predicate + it touches the element. What if we consider implementing it in the form of a prioririty queue based on binary heap? We would get O(1) sorted access.

src/domain/core/src/utils/time_source.rs
zaychenko-sergei: I have my doubts.

Doubt #1: coming back to original example, let''s slightly modify it:

```
let f1 = ts.wake_after(10s); 
let f2 = ts.wake_after(5s);
let woken_ids = ts.simulate_time_passage(15s).await; 
assert_eq!(woken_ids, [2, 1]);
timeout(f1).await.unwrap();
timeout(f2).await.unwrap();
```

The "passed" time jumps too quick, and swallows 2 time slots in the past.
If we allow this as currently, I would expect the awakenings to happen in a certain order that is more or less preserving time relations. Okay, we don't have to restrict the order between 2 futures scheduled after 10s both, but when the time relation is clear, I think we should respect it.

Doubt #2: whether swallowing a timeslot should be allowed at all.
Or, in other words, moving time to a point after an existing scheduled awakening, could be a `panic!`.
If something is planned in 5s, maybe it should be activated exactly at 5s moment only.
Since we are talking about the tests, not the real time, we control the time granularity very well.
Notice the code in flow system that does time rounding and respects minimal scheduler step.
I think it a reasonable delay vs step time is choosen, this could be very accurate and effecient.

src/infra/flow-system-inmem/tests/tests/test_flow_service_inmem.rs
zaychenko-sergei: I don't get it. These are tests with faked time, not real time. Why do we care about real time issues on Windows ?

src/domain/core/src/utils/time_source.rs
s373r: Done

src/domain/core/src/utils/time_source.rs
s373r: renamed to `sleep()`

src/domain/core/src/utils/time_source.rs
s373r: good remark -- renamed `FakeSystemTimeSource`

src/infra/flow-system-inmem/tests/tests/test_flow_service_inmem.rs
s373r: Oh, I see -- you are right, that was extra things -- replaces with `yield_now()` (we definitely don't need to sleep here at all)

src/domain/core/src/utils/time_source.rs
s373r: > use check_futures_for_completion()

thanks for a great tip -- done

s373r: > If we allow this as currently, I would expect the awakenings to happen in a certain order that is more or less preserving time relations. Okay, we don't have to restrict the order between 2 futures scheduled after 10s both, but when the time relation is clear, I think we should respect it.

this functionality (respect for order) is implemented, but without returning IDs

> Doubt 2: whether swallowing a timeslot should be allowed at all.

I understand the motivation, but in this test case where we don't know exactly who will call `sleep()` and where and when.
Thus, making timestamp shifts exactly to the next time we need to release an awaiting caller will be very problematic.

I'd like to suggest not making this test implementation so demanding

src/domain/core/src/utils/time_source.rs
s373r: > How does set method integrate with the awaiting callers?

currently fully integrates

> Also, what if we set a time that is earlier than previous value of t ?

will be a panic

src/domain/core/src/utils/time_source.rs
s373r: fixed

src/domain/core/src/utils/time_source.rs
s373r: and it's very right that you have -- it was a mistake to leave it as `Optional` 

fixed

src/domain/core/src/utils/time_source.rs
s373r: done

src/domain/core/src/utils/time_source.rs
s373r: fixed, thanks for such good comment

CHANGELOG.md
```diff
@@ -4,14 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Changed
+- Use artificial control over time in `FlowService` tests to stabilize their behavior on Windows
```
zaychenko-sergei, line 9: Not only on Windows, but on Windows it was really tough to pass.

src/adapter/http/tests/harness/server_side_harness.rs
```diff
@@ -44,7 +44,7 @@ pub(crate) trait ServerSideHarness {
         self.dataset_url_with_scheme(dataset_alias, "odf+http")
     }
 
-    fn system_time_source(&self) -> &SystemTimeSourceStub;
+    fn system_time_source(&self) -> &FakeSystemTimeSource;
```
zaychenko-sergei, line 47: I was talking about keeping `SystemTimeSourceStub` for simple cases we had before. It would have only:
  - set logic
  - unset logic
  - default constructor with current time
  
While the advance fake version would be another implementation struct that we will only use in flow service tests for now. And that should not contain free default time, always starting from fixed test time and advancing via `sleep` method only.

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +30,155 @@ pub trait SystemTimeSource: Send + Sync {
...
+        tokio::time::sleep(std_duration).await;
+    }
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Debug)]
+struct AwaitingCaller {
+    wake_up_time: DateTime<Utc>,
+    waker_tx: Option<oneshot::Sender<()>>,
```
zaychenko-sergei, line 51: I think you earlier had a sequence ID. It might be useful to define some stricter order for priority queue. Time first, then the ID. As `waker_tx` is unlikely usable for sorting.

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +30,155 @@ pub trait SystemTimeSource: Send + Sync {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for AwaitingCaller {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.wake_up_time.cmp(&other.wake_up_time)
+    }
+}
```
zaychenko-sergei, line 72: Isn't auto-derive comparison doing the same?

src/domain/core/src/utils/time_source.rs
```diff
@@ -24,46 +30,155 @@ pub trait SystemTimeSource: Send + Sync {
...
         }
     }
 
     pub fn set(&self, t: DateTime<Utc>) {
-        *self.t.lock().unwrap() = Some(t);
+        let mut state = self.state.lock().unwrap();
+
+        if state.t > t {
+            panic!(
+                "The previous time [{}] is more than new time [{t}]",
```
zaychenko-sergei, line 113: "New time is earlier than previously seen".

Although, if you split the stub and fake, as I asked, I don't think we need a `set` public method at all, and could do everything in `advance` implementation.

src/adapter/http/tests/harness/server_side_harness.rs
s373r: Yes, thanks for the comment, I misunderstood last time -- done for now

src/domain/core/src/utils/time_source.rs
s373r: Done

src/domain/core/src/utils/time_source.rs
s373r: Unfortunately without additional crates we can't do this

src/domain/core/src/utils/time_source.rs
s373r: Fixed

zaychenko-sergei (approved):

PR #452, opened 2024-01-18
Add verify pattern logic

rmn-boiko:
## Description

Closes: <#issue>


src/app/cli/src/cli_commands.rs
```diff
@@ -546,3 +546,19 @@ where
...
+where
+    I: IntoIterator<Item = String>,
+{
+    let mut result_refs = Vec::new();
+    for dataset_ref in dataset_refs.into_iter() {
+        result_refs.push(dataset_ref);
+    }
+
+    return Ok(result_refs);
+}
```
zaychenko-sergei, line 564: What exactly this would validate?
In non-wildcard version we are checking for dataset existence in the workspace.
In wildcard this is not applicable.

I assume there might be nothing to validate. We just need to ensure parsing of arguments in value parser is done correctly.

src/app/cli/src/cli_parser.rs
```diff
@@ -1334,14 +1334,32 @@ fn value_parse_dataset_name(s: &str) -> Result<DatasetName, String> {
 }
 
 fn value_parse_dataset_ref_local(s: &str) -> Result<DatasetRef, String> {
-    match DatasetRef::try_from(s) {
+    let dataset_ref_pattern = DatasetRefPattern {
+        pattern: s.to_string(),
+        ..DatasetRefPattern::default()
+    };
+    let dataset_ref = s.to_string().replace(dataset_ref_pattern.wildcard, "ref");
+    match DatasetRef::try_from(dataset_ref.as_str()) {
```
zaychenko-sergei, line 1342: Why did we have to touch the original non-wildcard parser at all?

src/app/cli/src/cli_parser.rs
```diff
@@ -1334,14 +1334,32 @@ fn value_parse_dataset_name(s: &str) -> Result<DatasetName, String> {
...
             Err("Local reference should be in form: `did:odf:...` or `my.dataset.id`".to_string())
         }
     }
 }
 
+fn value_parse_dataset_ref_pattern_local(s: &str) -> Result<String, String> {
+    let arg = s.to_string();
+    if arg.contains(DatasetRefPattern::default().wildcard) {
+        return Ok(arg);
+    }
```
zaychenko-sergei, line 1354: I don't think it's enough. You can have a complete b*t with wildcard symbol, and it will pass.
I think we should implement another grammar in `src/domain/opendatafabric/src/identity/grammar.rs`

src/app/cli/src/commands/verify_command.rs
```diff
@@ -9,23 +9,29 @@
...
+>;
 
 ///////////////////////////////////////////////////////////////////////////////
 
 pub struct VerifyCommand {
     dataset_repo: Arc<dyn DatasetRepository>,
     verification_svc: Arc<dyn VerificationService>,
     output_config: Arc<OutputConfig>,
-    refs: Vec<DatasetRef>,
+    refs: Vec<String>,
```
zaychenko-sergei, line 34: We should rather have a dedicated type based on grammar, like `DatasetName`.
It will be still mostly a string inside technically, but an explicit type has a logical meaning and validation guarantee

src/app/cli/src/commands/system_diagnose_command.rs
```diff
@@ -281,11 +281,11 @@ impl DiagnosticCheck for CheckWorkspaceConsistent {
         });
         let datasets: Vec<_> = self.dataset_repo.get_all_datasets().try_collect().await?;
 
-        let verify_options = VerificationOptions {
+        let verify_options = Arc::new(VerificationOptions {
```
zaychenko-sergei, line 284: This insertion of `Arc` all over the interface is a wrong idea.
The problem you've been trying to solve with the task spawning can be expressed much easier - clone options locally before spawn callback:

```
 let handles: Vec<_> = requests
            .into_iter()
            .map(|r| {
                ...
                let clone_options = options.clone();
```

src/app/cli/src/commands/verify_command.rs
```diff
@@ -9,23 +9,29 @@
...
 use super::{BatchError, CLIError, Command};
 use crate::output::OutputConfig;
 use crate::VerificationMultiProgress;
 
-type GenericVerificationResult =
-    Result<Vec<(DatasetRef, Result<VerificationResult, VerificationError>)>, CLIError>;
+type GenericVerificationResult = Result<
+    Vec<(
+        DatasetRef,
+        Arc<Result<VerificationResult, VerificationError>>,
```
zaychenko-sergei, line 23: Why the result requires `Arc`? I think it can be avoided. We never access it from multiple threads, just transfer or copy the object

src/app/cli/src/commands/verify_command.rs
```diff
@@ -73,27 +79,73 @@ impl VerifyCommand {
...
     ) -> GenericVerificationResult {
-        let dataset_handle = self
+        let dataset_ref_arg = self.refs.first().unwrap();
+        let dataset_ref_pattern = DatasetRefPattern {
+            pattern: dataset_ref_arg.to_string(),
+            ..DatasetRefPattern::default()
+        };
+        if !dataset_ref_arg
+            .to_string()
+            .contains(dataset_ref_pattern.wildcard)
```
zaychenko-sergei, line 92: Why would this need a separate algorithm? A wildcard without instance of `%` will return the same results as listing all datasets.

zaychenko-sergei: One more thing:
 - in wildcard mode this is okay to say if pattern did not match anything - this needs a warning though
 - in non-wildcard mode, we should still have an error saying dataset is not found.

src/domain/core/src/repos/dataset_repository.rs
```diff
@@ -52,6 +52,11 @@ pub trait DatasetRepository: DatasetRegistry + Sync + Send {
 
     fn get_all_datasets(&self) -> DatasetHandleStream<'_>;
 
+    fn get_all_datasets_filtered(
+        &self,
+        dataset_ref_pattern: DatasetRefPattern,
+    ) -> DatasetHandleStream<'_>;
```
zaychenko-sergei, line 58: I think the filtering algorithm does not depend on repository kind - so it should not be expanded in implementatino structs anyhow differently.

It could be a static method with default implementation, or even a global utility function, which takes a stream of dataset handles + pattern, and filters them depending if they satisfy the pattern.

src/domain/core/src/services/verification_service.rs
```diff
@@ -55,6 +55,12 @@ pub enum VerificationResult {
     Valid,
 }
 
+#[derive(Debug)]
+pub struct VerificationMultiResult {
+    pub verification_result: Arc<Result<VerificationResult, VerificationError>>,
```
zaychenko-sergei, line 60: I strongly don't think an `Arc` is required here. It's not something we would share between threads.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -399,7 +417,7 @@ trait DatasetStorageStrategy: Sync + Send {
 
     fn get_dataset_path(&self, dataset_handle: &DatasetHandle) -> PathBuf;
 
-    fn get_all_datasets(&self) -> DatasetHandleStream<'_>;
+    fn get_all_datasets<'s>(&'s self) -> DatasetHandleStream<'s>;
```
zaychenko-sergei, line 420: Why?

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -195,7 +195,14 @@ impl DatasetRepository for DatasetRepositoryS3 {
...
+    fn get_all_datasets<'s>(&'s self) -> DatasetHandleStream<'s> {
+        self.stream_datasets_if(|_| true)
+    }
+
+    fn get_all_datasets_filtered<'s>(
+        &'s self,
+        _dataset_ref_pattern: DatasetRefPattern,
+    ) -> DatasetHandleStream<'s> {
         self.stream_datasets_if(|_| true)
     }
```
zaychenko-sergei, line 207: This is incorrect, but see my comment in trait first

src/infra/core/src/verification_service_impl.rs
```diff
@@ -324,11 +325,50 @@ impl VerificationService for VerificationServiceImpl {
...
+                                )
+                                .await,
+                        );
+                        VerificationMultiResult {
+                            verification_result: res,
+                            dataset_handle: dataset_handle.clone(),
+                        }
+                    })
+                    .await // Now you can await the result of spawn_blocking
+                    .unwrap() // Handle the Result properly in your code
```
zaychenko-sergei, line 367: Is `.unwrap()` really a good handling here? I guess we need some error mapping and propagation of `Result`.

zaychenko-sergei (changes requested):

src/infra/core/src/verification_service_impl.rs
rmn-boiko: Yeah, such as we are handling each error In separate verify we may cover it with internal error

src/app/cli/src/commands/verify_command.rs
rmn-boiko: There was small issue with error handling in verify_command, and it was some kind of temporary solution. Removed

src/app/cli/src/cli_validation.rs
```diff
@@ -0,0 +1,167 @@
...
+    RepoName,
+};
+use thiserror::Error;
+use url::Url;
+
+#[derive(Clone)]
+pub struct DatasetPatternValidationRes {
+    pub pattern: DatasetPattern,
+    pub is_dataset_ref: bool,
+}
```
sergiimk, line 28: Not sure why we need `DatasetPatternValidationRes` type. Is this purely for optimization when instead of a pattern we get a reference without any wildcards?

If so, it might be better for `DatasetRefPattern` type to have `has_wildcards() -> bool` method and `as_dataset_ref() -> Option<DatasetRef>` methods.

This would enable code like:

```rust
fn get_datasets_matching(self, pattern: DatasetRefPattern) -> Stream<DatasetHandle> {
  if !pattern.has_wildcards() {
    // Optimization for when a pattern refers to one dataset
    stream::once(self.get_dataset(pattern.as_dataset_ref().unwrap()))
  } else {
    // <do the expensive matching>
  } 
}
```

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -223,6 +223,16 @@ impl DatasetName {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+///////////////////////////////////////////////////////////////////////////////
+
+newtype_str!(
+    DatasetPattern,
+    Grammar::match_dataset_pattern,
+    DatasetPatternSerdeVisitor
+);
```
sergiimk, line 232: `DatasetPattern` is a bad name as `Dataset` type is not some identifier - it's a complex aggregate object.

If this is a pattern that only matches dataset names we should name it `DatasetNamePattern`.

src/adapter/graphql/src/scalars/dataset_id_name.rs
```diff
@@ -116,6 +116,53 @@ impl ScalarType for DatasetName {
     }
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////
+// DatasetPatter
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct DatasetPattern(odf::DatasetPattern);
```
zaychenko-sergei, line 124: Where do we use it in GraphQL? Or is this auto-generated?

src/app/cli/src/cli_commands.rs
```diff
@@ -541,3 +541,19 @@ where
...
+fn validate_many_dataset_wildcards<I>(
+    _catalog: &dill::Catalog,
+    dataset_refs: I,
+) -> Result<Vec<DatasetRefPattern>, CLIError>
+where
+    I: IntoIterator<Item = DatasetRefPattern>,
+{
+    let mut result_refs = Vec::new();
+    for dataset_ref in dataset_refs.into_iter() {
+        result_refs.push(dataset_ref);
```
sergiimk, line 555: `validate_many_dataset_refs` function exists only because we sometimes allow CLI to work in the "multi-tenant" mode. This is a mode we only use for testing, where inside of your local workspace you can created datasets that belong to different "accounts". These accounts are fake, but it gives us ability to test some multi-tenant features in the comfort of local CLI. 

`validate_many_dataset_refs` simply checks that `account/dataset` reference format is used only in multi-tenant workspaces.

In your case with `validate_many_dataset_wildcards` since `DatasetRefPattern` is a superset of `DatasetRef` type and can be `my.datasets.%` or `alice/my.datasets.%` it also can be single- or multi-tenant, thus we should replicate the same check for it.

src/app/cli/src/cli_validation.rs
```diff
@@ -0,0 +1,163 @@
...
+        match DatasetRef::try_from(s) {
+            Ok(_dsr) => _dsr,
+            Err(_) => {
+                return Err(ValidationDatasetRefError::Failed(ValidationError {
+                    message: "Local reference should be in form: `did:odf:...` or `my.dataset.id`"
+                        .to_string(),
+                }))
+            }
+        };
+    }
```
sergiimk, line 78: Once we get the grammar right I expect this function to be as simple as:
```rust
match DatasetRefPattern::try_from(s) {
        Ok(v) => Ok(v),
        Err(_) => {
            Err("Local reference should be in form: `did:odf:...`, `my.dataset.id`, or a wildcard pattern `my.dataset.%`".to_string())
        }
    }
```

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,38 @@ impl std::cmp::PartialOrd for DatasetRefAny {
         Some(self.cmp(other))
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct DatasetRefPattern {
+    pub wildcard: char,
+    pub pattern: DatasetNamePattern,
+}
```
sergiimk, line 749: I think our types are not decomposed correctly.

Our CLI commands currently accept `DatasetRef`, which can be:
- `DatasetRef::ID(DatasetID("did:odf:xxx"))`
- `DatasetRef::Alias(DatasetAlias(None, DatasetName("my.dataset")))`
- `DatasetRef::Alias(DatasetAlias(Some(AccountName("alice")), DatasetName("my.dataset")))`
- `DatasetRef::Handle(DatasetHandle(id, alias))` - for internal use when a reference is already resolved to a specific dataset (to avoid expensive resolving multiple times)

Our goal is to accept **everything that was supported before** PLUS patterns.

Perhaps the simplest way to achieve all this is:
```rust
pub enum DatasetRefPattern {
  Ref(DatasetRef),
  Pattern(Option<AccountName>, DatasetNamePattern),
}
```

In our parsing function if we manage to parse a valid `DatasetRef` - we know it's just a regular reference to a single dataset. Otherwise we try to parse it as a pattern with optional account name.

sergiimk: I recommend you to right away add the tests that parse all these different cases into `DatasetRefPattern`. Let's use the same examples you specified in your RFC as test inputs.

src/app/cli/src/commands/verify_command.rs
```diff
@@ -76,24 +78,56 @@ impl VerifyCommand {
...
+                        (None, None),
+                        options,
+                        listener,
+                    )
+                    .await;
+
+                Ok(vec![Ok(VerificationMultiResult {
+                    dataset_handle,
+                    verification_result: res,
+                })])
```
zaychenko-sergei, line 101: Let's move this into a method too. It's usually much more readable when all parts of a function work at the same level of abstraction.

src/app/cli/src/cli_commands.rs
```diff
@@ -542,3 +564,19 @@ where
...
+
+// ToDo add validation for DatasetRefPattern and DatasetRef expresion
+fn validate_many_dataset_wildcards<I>(
+    catalog: &dill::Catalog,
+    dataset_ref_patterns: I,
+) -> Result<Vec<DatasetRefPattern>, CLIError>
+where
+    I: IntoIterator<Item = DatasetRefPattern>,
+{
+    let mut result_refs = Vec::new();
```
zaychenko-sergei, line 576: The result is no longer a ref, name of the collection is a bit confusing. Maybe `result_patterns`.

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,130 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DatasetRefPattern {
+    Ref(DatasetRef),
+    Pattern(Option<AccountName>, DatasetNamePattern),
+}
+
+impl DatasetRefPattern {
+    pub fn has_wildcard(s: &str, wildcard_symbol: char) -> bool {
+        s.contains(wildcard_symbol)
+    }
```
zaychenko-sergei, line 754: It's used only in the `FromStr` implemetation. Perhaps, we can just call `contains` directly on the string.

src/app/cli/src/output/verify_progress.rs
```diff
@@ -54,6 +54,9 @@ impl VerificationMultiListener for VerificationMultiProgress {
             self.multi_progress.clone(),
         )))
     }
+    fn begin_multi_verify(&self) -> Option<Arc<dyn VerificationMultiListener>> {
```
zaychenko-sergei, line 57: Blank line before

src/domain/core/src/services/verification_service.rs
```diff
@@ -55,6 +55,12 @@ pub enum VerificationResult {
     Valid,
 }
 
+#[derive(Debug)]
+pub struct VerificationMultiResult {
```
zaychenko-sergei, line 59: Why do you call it "multi"? It actually represents record of 1 dataset verification.

src/domain/core/src/services/verification_service.rs
```diff
@@ -140,6 +146,9 @@ pub trait VerificationMultiListener: Send + Sync {
     fn begin_verify(&self, _dataset: &DatasetHandle) -> Option<Arc<dyn VerificationListener>> {
         None
     }
+    fn begin_multi_verify(&self) -> Option<Arc<dyn VerificationMultiListener>> {
```
zaychenko-sergei, line 149: Blank

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,126 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+        }
+    }
+}
+
+super::dataset_identity::impl_parse_error!(DatasetRefPattern);
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
```
zaychenko-sergei, line 788: Much more advanced coverage is necessary to cover patterns grammar.

src/app/cli/src/cli_commands.rs
```diff
@@ -542,3 +564,18 @@ where
...
+) -> Result<Vec<DatasetRefPattern>, CLIError>
+where
+    I: IntoIterator<Item = DatasetRefPattern>,
+{
+    let mut result_ref_patterns = Vec::new();
+    for dataset_ref_pattern in dataset_ref_patterns.into_iter() {
+        result_ref_patterns.push(validate_dataset_ref_wildcard(catalog, dataset_ref_pattern)?);
+    }
+
+    Ok(result_ref_patterns)
```
sergiimk, line 580: I think this can be simplified to something like:
```rust
dataset_ref_patterns.into_iter().for_each(|p| validate(catalog, p)).collect()?
```

src/app/cli/src/cli_value_parser.rs
```diff
@@ -0,0 +1,148 @@
...
+    RepoName,
+};
+use thiserror::Error;
+use url::Url;
+
+#[derive(Clone)]
+pub struct DatasetPatternValidationRes {
+    pub pattern: DatasetNamePattern,
+    pub is_dataset_ref: bool,
+}
```
sergiimk, line 30: I don't see where this type is used

src/app/cli/src/cli_value_parser.rs
```diff
@@ -0,0 +1,148 @@
...
+pub fn value_parse_dataset_ref_pattern_local(
+    s: &str,
+) -> Result<DatasetRefPattern, ValidationDatasetRefError> {
+    match DatasetRefPattern::from_str(s) {
+        Ok(drp) => Ok(drp),
+        Err(_) => Err(ValidationDatasetRefError::Failed(ValidationError {
+            message: "Local reference should be in form: `did:odf:...`, `my.dataset.id`, or a \
+                      wildcard pattern `my.dataset.%`"
+                .to_string(),
+        })),
```
sergiimk, line 63: Why not return an error as `String` here just like in other validators. It doesn't look like `ValidationDatasetRefError` type gives us anything extra.

src/app/cli/src/cli_value_parser.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+pub fn value_parse_dataset_name(s: &str) -> Result<DatasetName, String> {
```
sergiimk, line 67: Since these validators are only useful in CLI - we should use `pub(crate)` visibility.

`pub` should only be used for things we want to export from the crate.

If your `DatasetPatternValidationRes` type above was `pub(crate)` - compiler would let you know that its not used anywhere, otherwise it has to assume that someone who links `kamu-cli` as a library may use this type.

src/app/cli/src/error.rs
```diff
@@ -249,7 +249,7 @@ pub struct NotInMultiTenantWorkspace;
 #[derive(Error, Clone, PartialEq, Eq, Debug)]
 #[error("Multi-tenant reference is unexpected in single-tenant workspace: {dataset_ref}")]
 pub struct MultiTenantRefUnexpectedError {
-    pub dataset_ref: opendatafabric::DatasetRef,
+    pub dataset_ref: String,
```
sergiimk, line 252: Why not use `DatasetRefPattern` here? Since `DatasetRef` type is a strict subset of `DatasetRefPattern` - it can be used to represent both.

src/app/cli/src/cli_parser.rs
```diff
@@ -1281,7 +1281,7 @@ pub fn cli() -> Command {
                             .index(1)
                             .num_args(1..)
                             .required(true)
-                            .value_parser(value_parse_dataset_ref_local)
+                            .value_parser(value_parse_dataset_ref_pattern_local)
```
sergiimk, line 1284: Let's also update the `after_help` section to show an example of using patterns

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,126 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
 }
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DatasetRefPattern {
+    Ref(DatasetRef),
+    Pattern(Option<AccountName>, DatasetNamePattern),
+}
+
+impl DatasetRefPattern {
+    pub fn match_pattern(dataset_ref: &str, pattern: &str) -> Result<bool, InvalidPatternError> {
```
sergiimk, line 752: The type is already called `DatasetRefPattern`, so calling the method `match_pattern` is redundant. I propose just `matches()` or `is_match()` for similarity with [regex crate](https://docs.rs/regex/latest/regex/struct.Regex.html#method.is_match).

sergiimk: Also why are we accepting `dataset_ref` as `str` and not `DatasetRef`? I think this can lead to multiple problems.

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,126 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+            Ok(dr) => Ok(Self::Ref(dr)),
+            Err(_) => Err(Self::Err::new(s)),
+        }
+    }
+}
+
+super::dataset_identity::impl_parse_error!(DatasetRefPattern);
+
+#[cfg(test)]
+mod tests {
```
sergiimk, line 786: We currently don't use `mod tests` anywhere in the project, instead we put all tests in the separate `tests/` directory.

This is just a stylistic choice we made very early on, and I'd be OK to revisit this decision, but let's move these tests into `test/` for now and have this discussion separately.

src/domain/opendatafabric/src/identity/grammar.rs
```diff
@@ -114,6 +153,60 @@ impl Grammar {
...
+
+    // EndElement = ("." | "-")? HostName
+    pub fn match_hostname_pattern_end_element(s: &str) -> Option<(&str, &str)> {
+        let (h1, t1) = Self::match_opt(s, |s| Self::match_one_of_chars(s, &['.', '-']))?;
+        let (h2, t2) = Self::match_hostname(t1)?;
+        Some((&s[..h1.len() + h2.len()], t2))
+    }
+
+    // Pattern = StartElement? ("%"+ MiddleElement)* "%"+ EndElement?
+    pub fn match_hostname_pattern(s: &str) -> Option<(&str, &str)> {
```
sergiimk, line 190: I'd like to revisit this again when we have a much better test coverage.

But my gut reaction is that this might be an overkill:
- Invalid pattern will never result in us creating some incorrectly named object in the system - thanks to existing grammar for `DatasetName` we know for sure that all dataset names are already valid `Hostname`s
- The worst think that can happen in case of an invalid patterns like `%a..b` is that **it will simply not match anything**
- This makes me question the point of strict pattern validation. If user has invalid elements in the pattern like `a..b` is it any better to complain about invalid pattern than to return empty result?

CC @zaychenko-sergei

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,126 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DatasetRefPattern {
+    Ref(DatasetRef),
+    Pattern(Option<AccountName>, DatasetNamePattern),
+}
+
+impl DatasetRefPattern {
+    pub fn match_pattern(dataset_ref: &str, pattern: &str) -> Result<bool, InvalidPatternError> {
+        Like::<false>::like(dataset_ref, pattern)
```
sergiimk, line 753: @zaychenko-sergei 
This actually brings up a very important point - should comparing all aliases be case-insensitive?

Since we already use dataset names as subdomains sometimes - I think the answer is YES, and we should amend both the spec and equality comparisons of names/aliases.

src/domain/opendatafabric/src/identity/dataset_refs.rs
sergiimk: I also don't see any tests for matching itself.

src/domain/opendatafabric/src/identity/dataset_refs.rs
zaychenko-sergei: Not true, we do have this type of tests. They are very useful for well-isolated classes or functions that do not require catalog dependencies.

src/domain/opendatafabric/src/identity/grammar.rs
sergiimk: Discussed this with @zaychenko-sergei further:

Having this grammar makes behavior more consistent with `DatasetRef` where invalid references fail validation rather than not match anything.

My concern was primarily the complexity of expressing this grammar, but given that:
- it's already been written
- implementation itself is not that complex
- and can be easily extended to other aspects of `LIKE` syntax (e.g. matching any character by underscore `_`)

Then there is no reason not to keep it.

src/app/cli/src/cli_value_parser.rs
rmn-boiko: Sorry, I cleaned up this, but didnt push yet

src/app/cli/src/commands/verify_command.rs
```diff
@@ -138,9 +172,22 @@ impl Command for VerifyCommand {
         let mut valid = 0;
         let mut errors = 0;
 
-        for (_, res) in &verification_results {
+        if verification_results.is_empty() {
+            eprintln!(
+                "{}",
+                console::style("There are no datasets matching patter")
```
sergiimk, line 178: Typo: `patter`

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +741,140 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DatasetRefPattern {
+    Ref(DatasetRef),
+    Pattern(Option<AccountName>, DatasetNamePattern),
+}
+
+impl DatasetRefPattern {
+    pub fn is_match(dataset_ref: DatasetRef, pattern: &str) -> Result<bool, InvalidPatternError> {
+        Like::<false>::like(dataset_ref.to_string().as_str(), pattern)
+    }
```
sergiimk, line 754: 1) I still don't see any tests for this function

2) `dataset_ref: DatasetRef` can be passed by reference instead of by value

3) Why is this a static function and not a method? Wouldn't it be better to have it as:
  `fn is_match(&self, dataset_ref: &DatasetRef)`

    Otherwise the user of this API will be forced to extract `pattern: &str` somehow from `DatasetRefPattern`, which will involve complex matching / destructuring etc.

4) Consider what will happen if `dataset_ref` is in form of `did:odf:xxx`. In RFC we said we don't ever want to pattern-match IDs, but there's nothing protecting us against this here. (need tests!)

src/domain/opendatafabric/src/identity/dataset_refs.rs
sergiimk: > we do have this type of tests

That one suite of `mod tests` that was introduced in flow system ignored the unwritten convention. 99% of our tests still reside in `tests`.

Like I said I'm not against revisiting this and starting to use `mod tests` for new code, but since all other types in this module have their tests in `test_dataset_identity.rs` - I think using `mod tests` just for one of these types will be confusing.

sergiimk (changes requested):

src/app/cli/src/cli_parser.rs
```diff
@@ -324,6 +324,10 @@ pub fn cli() -> Command {
                         Delete a local dataset:
 
                             kamu delete my.dataset
+
+                        Delete a local datasets matching pattern:
```
sergiimk, line 328: `a` is an article used only with singular countables nouns, so `delete a dataset`, but `delete datasets`

src/app/cli/src/cli_parser.rs
```diff
@@ -1305,6 +1309,10 @@ pub fn cli() -> Command {
 
                             kamu verify com.example.deriv
 
+                        Verify the data in a datasets matching pattern:
```
sergiimk, line 1312: remove `a` article

src/domain/opendatafabric/src/identity/dataset_refs.rs
sergiimk: 5. Since we already validated the pattern via our strict grammar in theory it should never return `InvalidPatternError`

src/domain/opendatafabric/src/identity/dataset_refs.rs
sergiimk: Created a proposal https://github.com/kamu-data/kamu-cli/issues/469

zaychenko-sergei (approved):
No extra comments from my side. We discussed that `push` and `pull` commands will be done in a separate ticket

src/app/cli/src/commands/delete_command.rs
```diff
@@ -46,28 +48,24 @@ impl DeleteCommand {
...
+            .try_collect()
+            .await?
         };
 
-        // Check references exist
-        // TODO: PERF: Create a batch version of `resolve_dataset_ref`
-        for dataset_ref in &self.dataset_refs {
-            match self.dataset_repo.resolve_dataset_ref(dataset_ref).await {
-                Ok(_) => Ok(()),
-                Err(GetDatasetError::NotFound(e)) => Err(CLIError::usage_error_from(e)),
```
sergiimk, line 66: Can you test `kamu delete dataset-name-that-doesnt-exist`? Previously `NotFound` error was translated into `CLIError::Usage` as you can see here, but now it looks like it will surface as internal error.

src/domain/core/src/services/verification_service.rs
```diff
@@ -140,6 +146,10 @@ pub trait VerificationMultiListener: Send + Sync {
     fn begin_verify(&self, _dataset: &DatasetHandle) -> Option<Arc<dyn VerificationListener>> {
         None
     }
+
+    fn begin_multi_verify(&self) -> Option<Arc<dyn VerificationMultiListener>> {
+        None
```
sergiimk, line 151: What is the point of this method?

`begin_verify` is a way to convert from "multi-listener" into a "single-listener".

This new method returns "multi-listener" from what is already a "multi-listener"...

We don't have such a method in any of the other listener interfaces.

src/app/cli/src/output/verify_progress.rs
```diff
@@ -54,6 +54,10 @@ impl VerificationMultiListener for VerificationMultiProgress {
             self.multi_progress.clone(),
         )))
     }
+
+    fn begin_multi_verify(&self) -> Option<Arc<dyn VerificationMultiListener>> {
+        Some(Arc::new(VerificationMultiProgress::new()))
```
sergiimk, line 59: Not sure why you need an existing `VerificationMultiProgress` to construct a new `VerificationMultiProgress` instance here.

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -740,3 +740,77 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+            },
+            Self::Pattern(account_name_maybe, dataset_name_pattern) => {
+                (account_name_maybe.is_none()
+                    || *account_name_maybe == dataset_hande.alias.account_name)
+                    && dataset_name_pattern.is_match(&dataset_hande.alias.dataset_name)
+            }
+        }
+    }
+
+    pub fn dataset_ref(&self) -> Option<&DatasetRef> {
```
sergiimk, line 766: `as_dataset_ref` might be a better name to underline that we're not accessing some field, but re-interpreting this instance as an enum variant.

sergiimk (changes requested):

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -0,0 +1,48 @@
...
+        .collect();
+
+    if glob_dataset_ref_patterns.is_empty() {
+        Box::pin(async_stream::try_stream! {
+            for dataset_ref_pattern in dataset_ref_patterns.iter() {
+                // Check references exist
+                // TODO: PERF: Create a batch version of `resolve_dataset_ref`
+                match dataset_repo.resolve_dataset_ref(dataset_ref_pattern.dataset_ref().unwrap()).await {
+                    Ok(dataset_handle) => yield dataset_handle,
+                    Err(_) => continue,
```
sergiimk, line 32: We shouldn't ignore all errors here

zaychenko-sergei (approved):

PR #453, opened 2024-01-19
CI: clippy integration

s373r:
## Description


.github/workflows/build.yaml
```diff
@@ -30,6 +30,11 @@ jobs:
       - name: license headers
         run: |
           cargo test -p kamu-repo-tools -- license_header
+      # TODO: Remove error suppression
+      #       https://github.com/kamu-data/kamu-cli/issues/332
+      - name: clippy
+        run: |
+          cargo clippy --all --all-targets -- -D warnings || true
```
sergiimk, line 37: I took your branch and ran this command, and it produced no warnings or errors. If that's what you see too, what prevents us from enabling this lint?

s373r: > I took your branch and ran this command, and it produced no warnings or errors. If that's what you see too, what prevents us from enabling this lint?

hm, I just made a plan to include enabling a little later, as I expected a lot of warnings/errors.
There were indeed a lot of bugs, but they've all been fixed -- so I don't see any barriers to inclusion.

Will do

s373r: Done

.github/workflows/build.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions-rs/toolchain@v1
         with:
-          components: rustfmt
+          components: rustfmt, clippy
```
sergiimk, line 24: This lint is called `Lint / Formatting` but adding `clippy` to it changes its purpose.
I think we should ether rename the lint to `Lint / Code` or separate clippy into its own job.

s373r: A good catch -- I thought (by mistake) that it was both lint and formatting, so I thought I didn't need to change the name. Now, after your remark, I realize that it is a group and a subgroup

Fixed

zaychenko-sergei (commented):
Looking at the logs, we are running linting for the hundreds of external dependencies.
I think we should only handle our own crates, if that's technically possible.

sergiimk:
> Looking at the logs, we are running linting for the hundreds of external dependencies. I think we should only handle our own crates, if that's technically possible.

I had the same reaction, but then though "Nah, it can't be running lints for all dependencies, it probably only builds them for AST".

Turns out there is an option:
```
--no-deps                Run Clippy only on the given crate, without linting the dependencies
```

Makefile
```diff
@@ -13,7 +13,7 @@ lint:
 	cargo fmt --check
 	cargo test -p kamu-repo-tools
 	cargo deny check
-	cargo clippy all
+	cargo clippy --all --all-targets
```
zaychenko-sergei, line 16: The docs on `cargo check` say switch `--all` should be replaced with `--workspace`:

```
 --all               Alias for --workspace (deprecated)
 ```

s373r: a good catch -- replaced

zaychenko-sergei:
> > Looking at the logs, we are running linting for the hundreds of external dependencies. I think we should only handle our own crates, if that's technically possible.
> 
> I had the same reaction, but then though "Nah, it can't be running lints for all dependencies, it probably only builds them for AST".
> 
> Turns out there is an option:
> 
> ```
> --no-deps                Run Clippy only on the given crate, without linting the dependencies
> ```

This switch is useless for us, it's role is to lint 1 isolated workspace crate, excluding dependencies from the same workspace. Regarding external dependencies - they are checked once (later skipped incrementally), as Clippy needs to link the model properly.

zaychenko-sergei (approved):

Issue #448, opened 2024-01-18
login/logout commands crash


sergiimk:
### Describe the bug

`kamu login/logout` commands crash when `--server` doesn't specify protocol scheme

### Steps To Reproduce

```
kamu login platform.demo.kamu.dev
```

### Current Behavior

```
thread 'main' panicked at src/app/cli/src/cli_commands.rs:224:40:
called `Result::unwrap()` on an `Err` value: RelativeUrlWithoutBase
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/core/src/result.rs:1073:23
   4: kamu_cli::cli_commands::get_command::{{closure}}
             at /home/smikhtoniuk/Work/projects/kamu/kamu-cli/src/app/cli/src/cli_commands.rs:224:26
   5: core::option::Option<T>::map
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/core/src/option.rs:1072:29
   6: kamu_cli::cli_commands::get_command
             at /home/smikhtoniuk/Work/projects/kamu/kamu-cli/src/app/cli/src/cli_commands.rs:222:13
   7: kamu_cli::app::run::{{closure}}
             at /home/smikhtoniuk/Work/projects/kamu/kamu-cli/src/app/cli/src/app.rs:110:24
   8: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /home/smikhtoniuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.35.1/src/runtime/park.rs:282:63
   9: tokio::runtime::coop::with_budget
             at /home/smikhtoniuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.35.1/src/runtime/coop.rs:107:5
  10: tokio::runtime::coop::budget
[...]

PR #454, opened 2024-01-19
448 loginlogout commands crash

zaychenko-sergei:
## Description

Closes: #448 

### DIFF
--- a/src/app/cli/src/cli_commands.rs
+++ b/src/app/cli/src/cli_commands.rs
@@ -206,10 +206,7 @@ pub fn get_command(
             } else {
                 odf_server::AccessTokenStoreScope::Workspace
             },
-            // TODO: improve URL parser
-            submatches
-                .get_one::<String>("server")
-                .map(|s| Url::parse(s).unwrap()),
+            submatches.get_one::<String>("server").cloned(),
         )),
         Some(("logout", submatches)) => Box::new(LogoutCommand::new(
             cli_catalog.get_one()?,
--- a/src/app/cli/src/cli_commands.rs
+++ b/src/app/cli/src/cli_commands.rs
@@ -216,6 +216,7 @@ pub fn get_command(
                 odf_server::AccessTokenStoreScope::Workspace
             },
             submatches.get_one::<Url>("server").cloned(),
+            submatches.get_flag("all"),
         )),
         Some(("new", submatches)) => Box::new(NewDatasetCommand::new(
             submatches.get_one::<DatasetName>("name").unwrap().clone(),
--- a/src/app/cli/src/cli_commands.rs
+++ b/src/app/cli/src/cli_commands.rs
@@ -206,7 +206,7 @@ pub fn get_command(
             } else {
                 odf_server::AccessTokenStoreScope::Workspace
             },
-            parse_server_url(submatches.get_one::<String>("server"))?,
+            submatches.get_one::<Url>("server").cloned(),
         )),
         Some(("logout", submatches)) => Box::new(LogoutCommand::new(
             cli_catalog.get_one()?,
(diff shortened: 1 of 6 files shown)


src/app/cli/src/cli_commands.rs
```diff
@@ -218,10 +215,7 @@ pub fn get_command(
             } else {
                 odf_server::AccessTokenStoreScope::Workspace
             },
-            // TODO: improve URL parser
-            submatches
-                .get_one::<String>("server")
-                .map(|s| Url::parse(s).unwrap()),
+            parse_server_url(submatches.get_one::<String>("server"))?,
```
sergiimk, line 218: Do you need to specify `::<String>` here? I'd expect that type inference should work fine in this simple case.

src/app/cli/src/cli_commands.rs
```diff
@@ -546,3 +540,25 @@ where
 
     Ok(result_refs)
 }
+
+fn parse_server_url(maybe_url_str: Option<&String>) -> Result<Option<Url>, CLIError> {
```
sergiimk, line 544: Can we move this function on a `cli_parser.rs` level (similarly to `value_parse_dataset_ref_local`), to fail argument parsing if URLs are invalid, and to extract already parsed `Url` types instead of raw strings.

src/app/cli/src/cli_commands.rs
```diff
@@ -206,10 +206,7 @@ pub fn get_command(
             } else {
                 odf_server::AccessTokenStoreScope::Workspace
             },
-            // TODO: improve URL parser
-            submatches
-                .get_one::<String>("server")
-                .map(|s| Url::parse(s).unwrap()),
+            parse_server_url(submatches.get_one::<String>("server"))?,
```
sergiimk, line 209: Can I also ask for a small command behavior change?

Can we please make the `server` argument positional instead of a named, so we could write:
```
kamu login platform.kamu.dev
```
instead of:
```
kamu login --server platform.kamu.dev
```
I really want to keep this command similar to `docker login`.

src/app/cli/src/commands/logout_command.rs
```diff
@@ -33,15 +33,19 @@ impl LogoutCommand {
...
 }
 
 #[async_trait::async_trait(?Send)]
 impl Command for LogoutCommand {
     async fn run(&mut self) -> Result<(), CLIError> {
-        let odf_server_frontend_url = self
-            .server
-            .clone()
-            .unwrap_or_else(|| Url::parse(odf_server::DEFAULT_ODF_FRONTEND_URL).unwrap());
+        let odf_server_frontend_url = self.get_server_url();
```
sergiimk, line 48: I was a bit confused when I ran:
```
kamu logout
```
and command produced no output while all tokens remained present. Took me a second to realize that it probably tries to log me out of some default server, which I'm not logged into.

Perhaps we can add a confirmation message, e.g.:
```
Logged out of <server>
```

Also a `logout --all` flag would be really nice to have.

sergiimk (changes requested):

src/app/cli/src/cli_commands.rs
zaychenko-sergei: Done

src/app/cli/src/cli_commands.rs
zaychenko-sergei: Moved

src/app/cli/src/cli_commands.rs
zaychenko-sergei: `String` was not required, but `Url` is now required

src/app/cli/src/commands/logout_command.rs
zaychenko-sergei: Done both

sergiimk (approved):
:+1:

Issue #458, opened 2024-01-22
Flow configuration API is crashing when accepting large time delta values


zaychenko-sergei:
This is what is printed for entry of "169 days":

```
thread 'tokio-runtime-worker' panicked at src/adapter/graphql/src/scalars/flow_configuration.rs:117:13:
assertion failed: (value - chrono::Duration::hours(num_hours)).is_zero()
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/90e321d82a0a9c3d0e3f180d4d17541b729072e0/library/core/src/panicking.rs:142:5
   3: <kamu_adapter_graphql::scalars::flow_configuration::TimeDelta as core::convert::From<chrono::duration::Duration>>::from
   4: <kamu_adapter_graphql::scalars::flow_configuration::FlowConfiguration as core::convert::From<kamu_flow_system::entities::flow_configuration::flow_configuration_state::FlowConfigurationState>>::from
   5: <kamu_adapter_graphql::mutations::flows_mut::dataset_flow_configs_mut::DatasetFlowConfigsMut as async_graphql::resolver_utils::container::ContainerType>::resolve_field::{{closure}}

```

Stabilize API on various expression kinds.

PR #461, opened 2024-01-22
Flow configuration: improved TimeDelta validation & internal conversions

zaychenko-sergei:
## Description

Closes: #458

Improved TimeDelta validation & internal conversions.
Added solid test coverage.

- [x] Workspace layout changes include a migration: : not needed
- [x] Documentation update PR: N/A
- [x] Dataset pipelines update scheduled if needed: not needed

### DIFF
--- a/src/adapter/graphql/src/scalars/flow_configuration.rs
+++ b/src/adapter/graphql/src/scalars/flow_configuration.rs
@@ -127,7 +127,7 @@ impl From<chrono::Duration> for TimeDelta {
         }
 
         unreachable!(
-            "Expecting intervals not tinier than 1 minute that are clearly dividable by unit"
+            "Expecting intervals not smaller than 1 minute that are clearly dividable by unit"
         );
     }
 }
--- a/src/adapter/graphql/src/scalars/flow_configuration.rs
+++ b/src/adapter/graphql/src/scalars/flow_configuration.rs
@@ -95,8 +95,7 @@ pub enum TimeUnit {
 impl From<chrono::Duration> for TimeDelta {
     fn from(value: chrono::Duration) -> Self {
         let num_weeks = value.num_weeks();
-        if num_weeks > 0 {
-            assert!((value - chrono::Duration::weeks(num_weeks)).is_zero());
+        if (value - chrono::Duration::weeks(num_weeks)).is_zero() {
             return Self {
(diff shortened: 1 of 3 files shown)


src/adapter/graphql/src/scalars/flow_configuration.rs
```diff
@@ -95,42 +95,40 @@ pub enum TimeUnit {
...
+        if (value - chrono::Duration::minutes(num_minutes)).is_zero() {
             return Self {
                 every: num_minutes as u32,
                 unit: TimeUnit::Minutes,
             };
         }
 
-        unreachable!("Expecting intervals not tinier than 1 minute");
+        unreachable!(
+            "Expecting intervals not tinier than 1 minute that are clearly dividable by unit"
```
sergiimk, line 130: `tinier -> smaller`

src/adapter/graphql/src/scalars/flow_configuration.rs
```diff
@@ -95,42 +95,40 @@ pub enum TimeUnit {
...
-            assert!((value - chrono::Duration::minutes(num_minutes)).is_zero());
+        if (value - chrono::Duration::minutes(num_minutes)).is_zero() {
             return Self {
                 every: num_minutes as u32,
                 unit: TimeUnit::Minutes,
             };
         }
 
-        unreachable!("Expecting intervals not tinier than 1 minute");
+        unreachable!(
```
sergiimk, line 129: `unreachable` is usually for things you are certain won't happen. If our `From` can fail on a whole bunch of `Duration` values perhaps it should be a `TryFrom` that returns an error rather than panics the whole program?

sergiimk (approved):
Change is OK, but suggested a few ways to improve.

src/adapter/graphql/src/scalars/flow_configuration.rs
zaychenko-sergei: improved wording

src/adapter/graphql/src/scalars/flow_configuration.rs
zaychenko-sergei: I'm certain the things won't happen, as, assuming correct implementation, API only returns values that were previously set via API, and those cannot include mixed unit itervals like 1h 25 min

Issue #459, opened 2024-01-22
CRON expressions for flow scheduling should not have "seconds" parts


zaychenko-sergei:
Currently we require a full syntax of CRON that includes "seconds" component.
This is pointless for the application, minimal scheduling unit would be 1 minute.

Adjust the validation and interpretation of CRON expressions to avoid seconds.

PR #462, opened 2024-01-23
API for CRON expressions is using struct with explciit CRON part fields

zaychenko-sergei:
## Description

Closes: #459 

Replacing string-based CRON expressions with a data structure with explicit CRON fields.

- [ ] API changes are backwards-compatible - NOT COMPATIBLE, but no consumer yet
- [x] Workspace layout changes include a migration - not needed
- [x] Documentation update PR: not needed
- [x] Dataset pipelines update scheduled if needed - not needed

### DIFF
--- a/resources/schema.gql
+++ b/resources/schema.gql
@@ -164,7 +164,19 @@ type CreateDatasetResultSuccess implements CreateDatasetResult & CreateDatasetFr
 }
 
 type CronExpression {
-	cronExpression: String!
+	min: String!
+	hour: String!
+	dayOfMonth: String!
+	month: String!
+	dayOfWeek: String!
+}
+
+input CronExpressionInput {
+	min: String!
+	hour: String!
+	dayOfMonth: String!
+	month: String!
+	dayOfWeek: String!
 }
 
 type DataBatch {
@@ -1139,7 +1151,7 @@ type RequestHeader {
 
 input ScheduleInput @oneOf {
 	timeDelta: TimeDeltaInput
-	cronExpression: String
+	cronExpression: CronExpressionInput
 }
 
 type Search {
--- a/src/domain/flow-system/src/entities/shared/schedule.rs
+++ b/src/domain/flow-system/src/entities/shared/schedule.rs
@@ -21,7 +21,7 @@ pub enum Schedule {
     /// Time-delta based schedule
     TimeDelta(ScheduleTimeDelta),
     /// Cron-based schedule
-    CronExpression(CronSchedule),
+    CronExpression(ScheduleCronExpression),
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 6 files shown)


src/domain/flow-system/src/entities/shared/schedule.rs
```diff
@@ -113,52 +152,36 @@ mod tests {
...
             .with_ymd_and_hms(current_year + 1, 1, 1, 0, 0, 0)
             .unwrap();
 
-        let cron_time = cron_expression.next_activation_time(Utc::now()).unwrap();
-
+        let cron_time = cron_schedule.next_activation_time(Utc::now()).unwrap();
         assert_eq!(cron_time, expected_time);
     }
-
-    // Should return None if cron expression has no more iteration
```
rmn-boiko, line 156: I just realized. Do we have any logic to remove flow if cron expression has expired?

src/domain/flow-system/src/entities/shared/schedule.rs
```diff
@@ -57,24 +95,31 @@ pub struct CronExpressionIterationError {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 impl Schedule {
-    pub fn validate_cron_expression(
-        cron_expression: String,
-    ) -> Result<CronSchedule, ScheduleError> {
-        let schedule = match CronSchedule::from_str(&cron_expression) {
+    pub fn new_cron_schedule(source: CronExpressionSource) -> Result<Schedule, ScheduleError> {
+        let cron_as_string_with_dummy_seconds = format!(
```
rmn-boiko, line 99: Just a proposal, for me looks useful to extend `CronExpressinSource` with `as_cron_string()` method

rmn-boiko (approved):

src/domain/flow-system/src/entities/shared/schedule.rs
zaychenko-sergei: We don't  allow years now. How can it expire?

src/domain/flow-system/src/entities/shared/schedule.rs
zaychenko-sergei: Made a method, might be useful

resources/schema.gql
```diff
@@ -164,7 +164,19 @@ type CreateDatasetResultSuccess implements CreateDatasetResult & CreateDatasetFr
 }
 
 type CronExpression {
-	cronExpression: String!
+	min: String!
```
sergiimk, line 167: My preference would be to leave cron parameter a string and instead specify in the documentation:
- the formats that we support
  - 5-term with minutes
  - 6-term with seconds
- add a note that we don't guarantee any timings and that all scheduling is best-effort

 If you decide to proceed with this structure my only suggestion is to use `minute` instead of `min`, as all other terms have full names already.

zaychenko-sergei:
Decided to try another approach

Issue #228, opened 2023-08-29
DataFusion-based SQL console


sergiimk:
Now that we have proven that DataFusion can work well for ingest, the user experience would also benefit from it in `kamu sql`.

DataFusion already has a great CLI which we could integrate with (hopefully) little effort.

Task scope:
- Link `datafusion-cli` as a dependency of `kamu-cli`
- Allow launching `datafusion-cli` via `kamu sql --engine datafusion`
- Integrate `datafusion-cli` with `SessionContext` provided by the `QueryService`, giving CLI access to kamu datasets in the workspace
  - *Note: this may look similar to FlightSQL integration*
- Write a tests for executing a simple query on a sample dataset
  - *Note: this may be hard if CLI does not allow redirecting the STDOUT to a different stream*
- Make `datafusion` the default engine for `kamu sql` command

PR #463, opened 2024-01-24
DataFusion-based SQL console #228:

Wizzy-wooz:
## Description

Closes: [<#issue>](https://github.com/kamu-data/kamu-cli/issues/228) DataFusion-based SQL console #228

Job done:
1. Linked datafusion-cli as a dependency of kamu-cli.
2. Allowed launching datafusion-cli via kamu sql, made datafusion the default engine for kamu sql command.
3. added BSL-1.0 to licenses after running cargo deny check.

## Testing:

1. kamu sql => datafusion-cli:
image

2. kamu --sql --command "select 1": datafusion engine:

image

**NB! Testing will be added in next PR.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,34 +164,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "080e9890a082662b09c1ad45f567faeeb47f22b5fb23895fbe1e651e718e25ca"
 
 [[package]]
-name = "apache-avro"
-version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ceb7c683b2f8f40970b70e39ff8be514c95b96fcb9c4af87e1ed2cb2e10801a0"
-dependencies = [
- "bzip2",
- "crc32fast",
- "digest",
- "lazy_static",
- "libflate 2.0.0",
- "log",
- "num-bigint",
- "quad-rand",
- "rand",
- "regex-lite",
- "serde",
- "serde_json",
- "snap",
- "strum",
- "strum_macros 0.25.3",
- "thiserror",
- "typed-builder",
- "uuid",
- "xz2",
- "zstd 0.12.4",
-]
-
-[[package]]
 name = "approx"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -636,17 +608,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


deny.toml
```diff
@@ -43,6 +43,7 @@ allow = [
     "0BSD",
     "BSD-2-Clause",
     "BSD-3-Clause",
+    "BSL-1.0",
```
sergiimk, line 46: For next reader - this is [Boost Source License](https://choosealicense.com/licenses/bsl-1.0/) and is pure OSS.

src/app/cli/src/commands/sql_shell_command.rs
```diff
@@ -124,17 +129,30 @@ impl SqlShellCommand {
 
         Ok(())
     }
+
+    async fn run_datafusion_cli_command(&self) -> Result<(), CLIError> {
+        let mut print_options = PrintOptions {
+            format: PrintFormat::Table,
+            quiet: true,
```
sergiimk, line 136: Do you know what this option does? 

I'd suspect that quiet mode is used for executing scripts not to produse some verbose messages like table creation confirmations etc.

Perhaps in interactive shell mode we should be verbose.

src/app/cli/src/commands/sql_shell_command.rs
```diff
@@ -124,17 +129,30 @@ impl SqlShellCommand {
...
 impl Command for SqlShellCommand {
     async fn run(&mut self) -> Result<(), CLIError> {
         match (self.engine.as_deref(), &self.command, &self.url) {
-            (Some("datafusion"), Some(_), None) => self.run_datafusion_command().await,
-            (Some("datafusion"), _, _) => Err(CLIError::usage_error(
-                "DataFusion engine currently doesn't have a shell and only supports single \
-                 command execution",
-            )),
+            (None, None, None) => self.run_datafusion_cli_command().await,
+            (None, Some(_), None) => self.run_datafusion_command().await,
```
sergiimk, line 155: First match should be `None | Some("datafusion")` since we still want to support cases when the engine is specified explicitly, e.g.:
`kamu sql --engine datafusion`

src/app/cli/src/error.rs
```diff
@@ -213,6 +214,12 @@ impl From<InternalError> for CLIError {
     }
 }
 
+impl From<ReadlineError> for CLIError {
```
sergiimk, line 217: Where do we use this conversion?

I see that the error from `exec_from_repl` is mapped explicitly.

If this is not necessary any more - we could also remove the explicit dependency on `rustyline` from `Cargo.toml`.

sergiimk (changes requested):

src/app/cli/src/commands/sql_shell_command.rs
Wizzy-wooz: taken from   https://arrow.apache.org/datafusion/user-guide/cli.html:  -q, --quiet
            Reduce printing other than the results and work quietly

Wizzy-wooz: I don't mind changing it to false then we have more info.

src/app/cli/src/commands/sql_shell_command.rs
Wizzy-wooz: how about (Some("datafusion"), None, None) | (None, None, None) => self.run_datafusion_cli_command().await - ?

src/app/cli/src/commands/sql_shell_command.rs
sergiimk: I tested locally and all seems to work as expected:
- I can query things
- `join` things
- `explain` queries
- and I get helpful error messages on invalid schema and query mistakes 

I tried with `quiet: true/false` and didn't notice any difference.

src/app/cli/src/commands/sql_shell_command.rs
sergiimk: an even better option might be to move what the default engine is outside of the match, e.g.:
```rust
let engine = self.engine.as_deref().unwrap_or("datafusion");
match (....) {
  ("datafusion", _, _) => ...,
  ("spark", _, _) => ...,
}
```

sergiimk:
Also, please add an entry to `CHANGELOG.md` and rebase your branch to `master`

Cargo.lock
```diff
@@ -1866,13 +2262,37 @@ dependencies = [
...
 
+[[package]]
+name = "datafusion-cli"
+version = "33.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f51d09ed6f01e41a983cf720228b4f0811d755b3f52776dd5170c744bd5d626e"
+dependencies = [
+ "arrow",
+ "async-trait",
+ "aws-config 0.55.3",
```
sergiimk, line 2273: Also just noticed that we are compiling two versions of many of the `aws-*` dependencies.

Looks like `kamu` depends on `aws-* = 0.57` while `datafusion-cli` pulls in `aws-* = 0.55`.

Unfortunately it doesn't look like we can disable AWS as a feature in `datafusion-cli`:
https://github.com/apache/arrow-datafusion/blob/main/datafusion-cli/Cargo.toml

@Wizzy-wooz  - leave this problem to us. 

I think our options are:
- Accept duplication of dependencies for now
- Submit a patch to `datafusion-cli` to bump the version of `aws-*` libraries
- Submit a patch to make AWS feature optional

src/app/cli/src/commands/sql_shell_command.rs
sergiimk: You fixed it in one place but the same problem exists for `run_datafusion_command`

Wizzy-wooz: On it

Wizzy-wooz: done

zaychenko-sergei (approved):

Cargo.lock
```diff
@@ -1,7185 +0,0 @@
-# This file is automatically @generated by Cargo.
```
sergiimk, line 1: We shouldn't delete the lock file

sergiimk (commented):
Please restore the `Cargo.lock` file.
Other than that all looks good - we just need to figure out what to do with dependency duplication issue.

Cargo.lock
sergiimk: Looks like rebase is still needed too.

Wizzy-wooz: Working on it - it has happened after my merge...

Wizzy-wooz: Done

Cargo.lock
sergiimk: Decided to copy the code over for reasons described in new crate's README.

PR #473, opened 2024-01-30
cargo: fix "pedantic::result_large_err" warnings

s373r:
## Description

### DIFF
--- a/src/domain/core/src/entities/metadata_chain.rs
+++ b/src/domain/core/src/entities/metadata_chain.rs
@@ -376,7 +376,7 @@ pub enum AppendError {
     #[error(transparent)]
     RefCASFailed(#[from] RefCASError),
     #[error(transparent)]
-    InvalidBlock(#[from] AppendValidationError),
+    InvalidBlock(#[from] Box<AppendValidationError>),
     #[error(transparent)]
     Access(
         #[from]
@@ -389,6 +389,12 @@ pub enum AppendError {
         #[backtrace]
         InternalError,
     ),
+}
+
+impl From<AppendValidationError> for AppendError {
+    fn from(e: AppendValidationError) -> Self {
+        Self::InvalidBlock(Box::new(e))
+    }
 }
 
 impl From<SetRefErrorRepo> for AppendError {
@@ -487,16 +493,14 @@ pub enum AppendValidationError {
 ///////////////////////////////////////////////////////////////////////////////
 
 #[derive(Error, PartialEq, Eq, Debug)]
-#[error("Invalid event: {message}: {event:?}")]
+#[error("Invalid event: {message}")]
 pub struct InvalidEventError {
-    event: MetadataEvent,
     message: String,
 }
 
 impl InvalidEventError {
-    pub fn new(event: impl Into<MetadataEvent>, message: impl Into<String>) -> Self {
+    pub fn new(message: impl Into<String>) -> Self {
         Self {
-            event: event.into(),
             message: message.into(),
         }
     }
--- a/src/domain/core/src/entities/metadata_chain.rs
+++ b/src/domain/core/src/entities/metadata_chain.rs
@@ -481,12 +481,12 @@ pub enum AppendValidationError {
     #[error(transparent)]
     InvalidEvent(#[from] InvalidEventError),
     #[error(transparent)]
-    NoOpEvent(#[from] Box<NoOpEventError>),
+    NoOpEvent(#[from] NoOpEventError),
 }
 
 impl AppendValidationError {
     pub fn no_op_event(event: impl Into<MetadataEvent>, message: impl Into<String>) -> Self {
-        Self::NoOpEvent(Box::new(NoOpEventError::new(event, message)))
+        Self::NoOpEvent(NoOpEventError::new(event, message))
     }
 }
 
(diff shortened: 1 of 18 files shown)


src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -376,7 +376,7 @@ pub enum AppendError {
     #[error(transparent)]
     RefCASFailed(#[from] RefCASError),
     #[error(transparent)]
-    InvalidBlock(#[from] AppendValidationError),
+    InvalidBlock(#[from] Box<AppendValidationError>),
```
sergiimk, line 379: This looks a quite strange, as at this level reader cannot tell why all other errors are included by-value but this one is boxed. I think a better fix would be to box the whatever field inside the `AppendValidationError` that makes it so big (I assume its `MetadataEvent`).

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -487,16 +493,14 @@ pub enum AppendValidationError {
 ///////////////////////////////////////////////////////////////////////////////
 
 #[derive(Error, PartialEq, Eq, Debug)]
-#[error("Invalid event: {message}: {event:?}")]
+#[error("Invalid event: {message}")]
 pub struct InvalidEventError {
-    event: MetadataEvent,
```
sergiimk, line 492: Why are we removing this? This field will gets logged by our error handlers and is a source of valuable information. If anything it probably should be made `pub` to underline that its meat to be used by those handling this error.

sergiimk (commented):
I feel like this change can be a lot smaller if we box fields and not errors as a whole

src/domain/core/src/entities/metadata_chain.rs
s373r: After examining the code, came to the conclusion that this field is used solely for logging -- so I included a representation of the event in the message:

https://github.com/kamu-data/kamu-cli/pull/473/commits/ef9517e5da0bd248d83f61017824e1ae357b8ae2#diff-82e550963850c5eab67f442e4a683a96f7f6a7eb462d8e04d5ec0788679a24ffR19-R24

src/domain/core/src/entities/metadata_chain.rs
s373r: Indeed, I could put the error one level lower and it would be more practical -- thanks for the great tip!

will be implemented

resources/cli-reference.md
```diff
@@ -245,6 +245,10 @@ Delete a local dataset:
 
     kamu delete my.dataset
 
+Delete local datasets matching pattern:
+
+    kamu delete my.dataset.%
+
```
zaychenko-sergei, line 251: Careful here, would conflict with master where I've fixed that already

src/domain/core/src/entities/metadata_chain.rs
zaychenko-sergei: Yes, please Box the event, there are useful fields to see in platform server logs

zaychenko-sergei: I see, the message now includes a full debug trace of the event already, so this looks fine then.

src/domain/core/src/entities/metadata_chain.rs
s373r: Done

src/domain/core/src/entities/metadata_chain.rs
s373r: Added code comments to make this decision clearer for future readers

resources/cli-reference.md
s373r: This commit has been dropped

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -481,22 +481,29 @@ pub enum AppendValidationError {
     #[error(transparent)]
     InvalidEvent(#[from] InvalidEventError),
     #[error(transparent)]
-    NoOpEvent(#[from] NoOpEventError),
+    NoOpEvent(#[from] Box<NoOpEventError>),
```
zaychenko-sergei, line 484: Maybe Box for the event structure inside the error?

zaychenko-sergei (changes requested):

src/domain/core/src/entities/metadata_chain.rs
s373r: Done https://github.com/kamu-data/kamu-cli/pull/473/commits/122e88a7b4092a9507fea49cda1f10a747119054

src/domain/core/src/entities/metadata_chain.rs
sergiimk: Why do we have two different approaches between `InvalidEventError` and `NoOpEventError`?

Personally I like having a boxed `event` in the error rather than pre-formatting it into the message, as it gives more flexibility on how to present the error.

sergiimk (changes requested):

src/domain/core/src/entities/metadata_chain.rs
s373r: > Personally I like having a boxed event in the error rather than pre-formatting it into the message, as it gives more flexibility on how to present the error.

agree, consistency is paramount -- will use `Box<T>`

sergiimk (approved):
:+1:

zaychenko-sergei (approved):

Issue #355, opened 2023-12-13
Avoid using fixed container and network names


sergiimk:
Verify that in all places we give unique names to containers and network so that it was possible to run any command / operation concurrently.

PR #480, opened 2024-02-01
Remove static container names and networks

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/355


src/infra/core/src/ingest/fetch_service.rs
```diff
@@ -313,6 +313,7 @@ impl FetchService {
                 if #[cfg(unix)] {
                     self.container_runtime
                         .run_attached(&fetch.image)
+                        .container_name_prefix("kamu-permissions-")
```
s373r, line 316: I suggest using `get_random_name()` inside `container_name_prefix()`

s373r, line 316: Optional: rename `container_name_prefix()` to `random_container_name_with_prefix()` (or similar)

sergiimk: +1 to better name and hiding name generation fully in `container-runtime` crate

we should also provide a similar method for creating randomized networks

sergiimk (commented):
Make sure to thoroughly test the affected commands. I would suggest creating two workspaces and running those commands in parallel to see that the work and don't conflict.

rmn-boiko:
> Make sure to thoroughly test the affected commands. I would suggest creating two workspaces and running those commands in parallel to see that the work and don't conflict.



> Make sure to thoroughly test the affected commands. I would suggest creating two workspaces and running those commands in parallel to see that the work and don't conflict.

I was looking for each `run_attached`, `container_name`, `RunArgs` and change in each place

s373r (approved):

src/app/cli/src/explore/sql_shell_impl.rs
```diff
@@ -78,11 +78,13 @@ impl SqlShellImpl {
                 .stderr(std::fs::File::create(&spark_stderr_path).int_err()?);
 
             container_builder = if let Some(p) = port {
-                container_builder.network("host").map_port_with_address(
-                    address.map_or(String::from("127.0.0.1"), ToString::to_string),
-                    p,
-                    10000,
-                )
+                container_builder
+                    .random_container_network_with_prefix("host-")
```
sergiimk, line 82: Network `host` has a special meaning here - it tells docker to expose the container directly on the **same network as your host machine** without any additional layers. We do this when user specifies `port`.

Also `host` network always exists. And because you're giving containers random names - it should be already possible to run multiple containers on `host` network as long they specify different ports. 

We should revert this block.

src/utils/container-runtime/src/container.rs
```diff
@@ -51,20 +51,17 @@ impl ContainerRunCommand {
...
-
-        let mut name = String::with_capacity(10 + prefix.as_ref().len());
-        name.push_str(prefix.as_ref());
-        name.extend(
-            rand::thread_rng()
-                .sample_iter(&Alphanumeric)
-                .take(10)
-                .map(char::from),
-        );
+    pub fn random_container_network_with_prefix(mut self, prefix: impl AsRef<str>) -> Self {
```
sergiimk, line 54: I don't think running a container auto-creates a network:
```sh
docker run -it --rm --network xxx ghcr.io/kamu-data/kamu-base:latest-with-data
docker: Error response from daemon: network xxx not found.
```
Thus I don't see how this method can work.
Something needs to create a network with random name first and then pass its name into containers.

src/app/cli/src/explore/notebook_server_impl.rs
```diff
@@ -73,20 +73,20 @@ impl NotebookServerImpl {
             "Exposing Notebook server on a host network interface is not yet supported"
         );
 
-        let network_name = "kamu";
+        let network_name = get_random_name_with_prefix("kamu-");
 
         // Delete network if exists from previous run
         let _ = self
```
sergiimk, line 79: Looks like this is not needed any more - the chances of us finding a network with the same name are extremely low.

src/app/cli/src/explore/notebook_server_impl.rs
```diff
@@ -73,20 +73,20 @@ impl NotebookServerImpl {
...
+            .remove_network_cmd(&network_name)
             .stdout(std::process::Stdio::null())
             .stderr(std::process::Stdio::null())
             .output()
             .await;
 
         let network = self
             .container_runtime
-            .create_network(network_name)
+            .create_network(&network_name)
```
sergiimk, line 89: Should we have `create_network_with_random_name()` method?

rmn-boiko: We pass this in a few containers inside this method. So it probably should assigned once at the top and passed like that

sergiimk: If the method returns `NetworkHandle` then we could extract the generated name from there. (Note `NetworkHandle` stores the name, but does not expose it currently, but I think it could).

src/utils/container-runtime/src/container.rs
rmn-boiko: this method is equal to `network` method which sets network parameters (but in this case with randomization). I'll create on for creation as well

src/utils/container-runtime/src/handles.rs
```diff
@@ -30,6 +30,10 @@ impl NetworkHandle {
         }
     }
 
+    pub fn name(&self) -> Option<String> {
+        self.network_name.clone()
```
sergiimk, line 34: Why allocate memory instead of returning `&str`?

Also notice that `network_name` only becomes `None` after the handle is dropped (or `free` is called), so we can use `unwrap` instead of returning an `Option`, as using network name after freeing the handle should not ever happen.

zaychenko-sergei (approved):

Issue #389, opened 2023-12-27
In-memory MetadataChain caching


sergiimk:
Currently many operations iterate the metadata chain multiple times, causing lots of read traffic to the repo.

Biggest offender are likely ingest and derivative transform, as they currently do a full-scan of metadata when preparing an operation, and then potentially multiple full and partial scans when committing the new block to properly validate it.

There are multiple ways to improve this situation, including e.g more intelligent metadata traversal with skip lists.

This ticket is to introduce an in-memory block cache to `MetadataChain` that are kept for the duration of an operation / request (e.g. while `MetadataChain` instance is alive).

PR #484, opened 2024-02-02
Implement In-memory MetadataChain caching

s373r:
## Description

Related issue: https://github.com/kamu-data/kamu-cli/issues/389

- [x] Unit-tests added


src/infra/core/src/repos/dataset_factory_impl.rs
```diff
@@ -56,11 +58,23 @@ impl DatasetFactoryImpl {
...
+                    obj_repo.clone(),
+                    ReferenceRepositoryImpl::new(NamedObjectRepositoryLocalFS::new(
+                        layout.refs_dir,
+                    )),
+                    GetMetadataBlockStrategyCacheInMemoryFacade::new_boxed(
+                        GetMetadataBlockStrategyImpl::new_boxed(obj_repo),
+                        hash_metadata_block_cache_map.clone(),
+                    ),
+                ),
+                hash_metadata_block_cache_map,
```
sergiimk, line 77: This is extremely complex - let's discuss other approaches on Slack

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -22,7 +24,7 @@ use crate::repos::{SetRefError as SetRefErrorRepo, *};
 #[async_trait]
 pub trait MetadataChain: Send + Sync {
     /// Resolves reference to the block hash it's pointing to
-    async fn get_ref(&self, r: &BlockRef) -> Result<Multihash, GetRefError>;
+    async fn get_hash_by_ref(&self, r: &BlockRef) -> Result<Multihash, GetRefError>;
```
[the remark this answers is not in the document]
sergiimk: PS: I would suggest `resolve_ref` - short and consistent with `resolve_dataset_ref`

src/infra/core/src/repos/metadata_chain_cache_inmen_facade.rs
```diff
@@ -0,0 +1,143 @@
...
+    SetRefError,
+    SetRefOpts,
+};
+use opendatafabric::{MetadataBlock, Multihash};
+
+use crate::HashMetadataBlockCacheMap;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct MetadataChainCacheFacadeInMemory<WrappedChain> {
```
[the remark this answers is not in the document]
sergiimk: Also typo in the file name: `inmen`

src/infra/core/src/repos/metadata_chain_cache_inmen_facade.rs
```diff
@@ -0,0 +1,143 @@
...
+        hash: &Multihash,
+        opts: SetRefOpts<'a>,
+    ) -> Result<(), SetRefError> {
+        let set_ref_result = self.chain.set_ref(r, hash, opts).await;
+
+        if set_ref_result.is_ok() {
+            self.ref_cache.insert(r.clone(), hash.clone());
+        }
+
+        set_ref_result
```
sergiimk, line 114: `MetadataChain` is `Sync` - what happens if we have multiple threads here?

src/domain/core/src/entities/metadata_chain.rs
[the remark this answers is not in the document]
s373r: > Note for future: it's better separate mass-renamings like this one into a separate PR or agree on them verbally and make them just before merging the branch. 
> This would save a lot of time for reviewers, as I'm struggling to understand how caching is actually implemented in the sea of ~50 files that were changed.

Yes, you're right, of course -- next time such changes will be put in separate PRs. PRs must be atomic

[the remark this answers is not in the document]
s373r: > PS: I would suggest resolve_ref - short and consistent with resolve_dataset_ref

will use

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -391,6 +393,18 @@ pub enum AppendError {
     ),
 }
 
+impl From<GetBlockError> for AppendError {
+    fn from(v: GetBlockError) -> Self {
+        match v {
+            GetBlockError::NotFound(e) => AppendValidationError::PrevBlockNotFound(e).into(),
```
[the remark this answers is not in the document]
s373r: > so while in one case NotFound will correspond to PrevBlockNotFound

I've extracted the repeating code from two places:
- `MetadataChainImpl::validate_append_sequence_numbers_integrity()`
- `MetadataChainImpl::validate_append_system_time_is_monotonic()`

> In this case I'd say NotFound is ambiguous

But yes, I agree with you that the ambiguity is present.

I see two options here: separate it into a separate function or return code duplication. I would prefer the first option.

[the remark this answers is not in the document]
s373r: Fixed via https://github.com/kamu-data/kamu-cli/pull/484/commits/78c318f06cee42ad17187655b4172744d0b52825

src/domain/core/src/entities/metadata_stream.rs
```diff
@@ -22,7 +22,9 @@ pub type DynMetadataStream<'a> = Pin<Box<dyn MetadataStream<'a> + Send + 'a>>;
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
-type MetadataStreamItem = Result<(Multihash, MetadataBlock), IterBlocksError>;
+pub type MetadataBlockPair = (Multihash, MetadataBlock);
```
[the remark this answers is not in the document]
s373r: yes, those names will be clearer to readers -- chose `HashedMetadataBlock`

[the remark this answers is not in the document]
s373r: Done via ade0817aab99e82342b10990df1705fd965633f7

src/infra/core/src/repos/metadata_chain_cache_inmen_facade.rs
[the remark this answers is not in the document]
s373r: `MetadataChainCacheFacadeInMemory` has been deleted.

Used the pattern for `GetMetadataBlockStrategyCachingInMem` & `GetMetadataBlockStrategyCachingInMem`

--

Done

src/infra/core/src/repos/metadata_chain_cache_inmen_facade.rs
s373r: At this point, `MetadataChainCacheFacadeInMemory` has been removed and two reentrant classes are left (due to the use of `DashMap`):
- `GetMetadataBlockStrategyCachingInMem`
- `ReferenceRepositoryCachingInMem`

src/adapter/graphql/tests/tests/test_gql_metadata_chain.rs
```diff
@@ -403,8 +405,10 @@ async fn metadata_update_readme_new() {
 
     assert_result(res, "CommitResultSuccess");
 
+    let dataset = dataset_repo.get_dataset(&dataset_ref).await.unwrap();
```
zaychenko-sergei, line 408: Make a comment that we need to start a new transaction.
Maybe also scope the parts of the tests with separated blocks, so that the lifetime of the dataset/cache is clear.

src/domain/core/src/entities/metadata_block_repository.rs
```diff
@@ -0,0 +1,37 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait]
+pub trait MetadataBlockRepository: Send + Sync {
+    async fn contains(&self, hash: &Multihash) -> Result<bool, ContainsError>;
+
+    async fn get(&self, hash: &Multihash) -> Result<MetadataBlock, GetBlockError>;
+
+    async fn insert<'a>(
```
zaychenko-sergei, line 30: I think each of these methods should include word "block" in the name.

src/domain/core/src/entities/metadata_block_repository.rs
```diff
@@ -0,0 +1,37 @@
...
+        &'a self,
+        block: &MetadataBlock,
+        options: InsertOpts<'a>,
+    ) -> Result<InsertResult, InsertError>;
+
+    fn as_object_repo(&self) -> &dyn ObjectRepository;
```
zaychenko-sergei, line 36: Oops. Unexpected discovery.
This is used in the implementation of transfer protocols to:
  - extract block size
  - extract bytes of the block
  - retreive some download URLs
  
It's now becoming obvious the transfer protocol is effectively ignoring the caching layer we've just built.

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
```diff
@@ -0,0 +1,94 @@
...
+impl<ObjRepo> MetadataBlockRepository for MetadataBlockRepositoryCachingInMem<ObjRepo>
+where
+    ObjRepo: ObjectRepository + Sync + Send,
+{
+    async fn contains(&self, hash: &Multihash) -> Result<bool, ContainsError> {
+        if self.cache.contains_key(hash) {
+            return Ok(true);
+        }
+
+        self.wrapped.contains(hash).await
```
zaychenko-sergei, line 60: Can we implement `contains` via `get`, so that `contains` query caches the block?
In case of a cache miss that still goes to the disk/s3 storage and makes a query of comparable cost as full object retreival (for blocks specificly, as reading 400byes or reading file size is assumed to roughly be the same cost).

zaychenko-sergei (changes requested):

src/adapter/http/tests/tests/test_data.rs
```diff
@@ -74,13 +74,11 @@ async fn test_data_push_ingest_handler() {
         })
         .await
         .unwrap();
-
-    let dataset_url =
-        server_harness.dataset_url_with_scheme(&create_result.dataset_handle.alias, "http");
+    let dataset_ref = create_result.dataset_handle.as_local_ref();
+    let dataset_url = server_harness.dataset_url_with_scheme(dataset_ref.alias().unwrap(), "http");
```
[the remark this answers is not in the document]
sergiimk: Pelase see discussion about this on Slack

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
s373r: Absolutely makes sense -- a good catch

src/domain/core/src/entities/metadata_block_repository.rs
s373r: Done

s373r: Now we have:
- `MetadataBlockRepository::get_block()`
- `MetadataBlockRepository::get_block_data()`
- `MetadataBlockRepository::insert_block_data()`

src/domain/core/src/entities/metadata_block_repository.rs
s373r: > This is used in the implementation of transfer protocols to:
> - extract block size
> - extract bytes of the block

`MetadataBlockRepository` is now used

> - retreive some download URLs

`ObjectRepository` continues to be used

src/adapter/http/tests/tests/test_data.rs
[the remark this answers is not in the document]
s373r: Ref caching has been deleted

src/domain/core/src/entities/mod.rs
```diff
@@ -12,11 +12,13 @@ pub mod dataset;
 pub mod dataset_summary;
 pub mod engine;
 pub mod events;
+pub mod metadata_block_repository;
```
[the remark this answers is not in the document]
s373r: At the time of your comment, we really shouldn't have stored `MetadataBlockRepository` trait in domain area.

But now, I had to add `MetadataChain::as_metadata_block_repository()` method that requires the trait to be in the area.

src/adapter/http/src/smart_protocol/protocol_dataset_helper.rs
```diff
@@ -178,19 +178,17 @@ pub async fn prepare_dataset_metadata_batch(
...
         .map(|(hash, bytes)| {
             // TODO: Avoid depending on specific implementation of MetadataChain.
             // This is currently necessary because we need to be able to deserialize blocks
             // BEFORE an instance of MetadataChain exists. Consider injecting a
             // configurable block deserializer.
-            match kamu::MetadataChainImpl::<(), ()>::deserialize_block(&hash, &bytes) {
-                Ok(block) => Ok((hash, block)),
-                Err(err) => Err(err),
-            }
+            deserialize_metadata_block(&hash, &bytes).map(|block| (hash, block))
```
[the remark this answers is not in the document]
s373r: I understand your point, really need to find this function a new home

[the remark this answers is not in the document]
s373r: Done

src/domain/core/src/entities/mod.rs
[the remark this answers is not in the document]
sergiimk: I think `domain` is the right place for this trait, but can't it be in `domain/core/src/repos/` instead? It can still be imported by `MetadataChain` as it's within one crate.

src/domain/core/src/entities/metadata_block_repository.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait]
+pub trait MetadataBlockRepository: Send + Sync {
+    async fn contains(&self, hash: &Multihash) -> Result<bool, ContainsError>;
+
+    async fn get_block(&self, hash: &Multihash) -> Result<MetadataBlock, GetBlockError>;
+
+    async fn get_block_data(&self, hash: &Multihash) -> Result<Bytes, GetError>;
```
sergiimk, line 32: Seems unfortunate for this type to now work with two different levels of abstraction: blocks and bytes...

If we store bytes in the cache we might as well remove `MetadataBlockRepository` as such and go back to caching on `ObjectRepository` level, since the goal of adding `MetadataBlockRepository` was to avoid the deserialization overhead on cache hits.

@zaychenko-sergei perhaps transfer protocol should handle metadata blocks on a lower level:
- when iterating it could accumulate data as bytes
- when calculating sizes - it could create flatbuffer proxies over those bytes to read data directly without deserializing into DTOs (avoids lots of allocations)
- thus perhaps it could only scan metadata once, while maintaining its own "cache"

zaychenko-sergei: 1) Put it differently: cache could store both raw bytes and deserialized version simultaneously.  In addition, the deserialized version could also be lazily computed (thus never or at most once) to avoid ovearhead. It still keeps the new trait at the block abstraction level, and preserves effeciency for low-level access (bytes, size).

2) Protocol benefits from the implemented idea of dataset per-transaction cache. It's currently already doing 2 traversals:
   - at step 1 to create a size estimate
   - at step 2 to pack a tarball out of block bytes
As the `Dataset` object is kept between steps of the protocol, we can do just 1 set of reads with the cache involved. 

With 1 implemented as I describe, there is no deserialization overhead (lazy thing not computed).

3) The protocol should rather be kept isolated from storage/caching details, as it currently is. I would only go into custom option if we could sum sizes / zip stream for an interval range more effeciently than doing it individually in a loop. 

If we are fantasizing here, block could have a field meaning a sum of bytes in all the previouls blocks, thus size estimate could be computed  as a difference between these numbers in 2 selected blocks making a range.

src/domain/core/src/entities/mod.rs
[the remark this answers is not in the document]
s373r: Done

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
```diff
@@ -0,0 +1,116 @@
...
+};
+use opendatafabric::{MetadataBlock, Multihash};
+
+use crate::MetadataBlockRepositoryImpl;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct MetadataBlockRepositoryCachingInMem<ObjRepo> {
+    wrapped: MetadataBlockRepositoryImpl<ObjRepo>,
+    cache: DashMap<Multihash, Bytes>,
```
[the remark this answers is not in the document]
s373r: Done

src/domain/core/src/entities/metadata_block_repository.rs
s373r: > 1. Put it differently: cache could store both raw bytes and deserialized version simultaneously. In addition, the deserialized version could also be lazily computed (thus never or at most once) to avoid ovearhead. It still keeps the new trait at the block abstraction level, and preserves effeciency for low-level access (bytes, size).

Implemented

src/domain/core/src/entities/metadata_block_repository.rs
s373r: Done + <Result, Errors> for other methods

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
```diff
@@ -0,0 +1,181 @@
...
+    MetadataBlockRepository,
+    ObjectRepository,
+};
+use opendatafabric::{MetadataBlock, Multihash};
+
+use crate::MetadataBlockRepositoryImpl;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+struct MetadataBlockRepositoryCacheValue {
```
sergiimk, line 32: If a type is local to module I would recommend sticking to short names like `CachedValue` - there is no danger of ambiguity or collision on module level.

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -815,14 +755,13 @@ where
...
+        let block_data = FlatbuffersMetadataBlockSerializer
             .write_manifest(&block)
             .int_err()?;
-
         let res = self
-            .obj_repo
-            .insert_bytes(
-                &data,
+            .meta_block_repo
+            .insert_block_data(
```
sergiimk, line 763: We provide DTO-based `get` method, so why don't we hide the serialization of a block inside `MetadataBlockRepo::insert_block()` too to make interface symmetrical?

src/domain/core/src/repos/metadata_block_repository.rs
```diff
@@ -0,0 +1,245 @@
...
+        &self,
+        hash: &Multihash,
+        opts: ExternalTransferOpts,
+    ) -> Result<GetBlockExternalUrlResult, GetBlockExternalUrlError>;
+
+    async fn get_block_external_upload_url(
+        &self,
+        hash: &Multihash,
+        opts: ExternalTransferOpts,
+    ) -> Result<GetBlockExternalUrlResult, GetBlockExternalUrlError>;
```
sergiimk, line 62: Not sure about these methods - will bring up on Slack

sergiimk (commented):
I still don't see a test for caching behavior

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
s373r: After [the hotfix](https://github.com/kamu-data/kamu-cli/pull/496), we don't need this structure -- we can just cache blocks, not bytes

So it's no longer actual

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: After [the hotfix](https://github.com/kamu-data/kamu-cli/pull/496), we don't need this structure -- we can just cache blocks, not bytes

Thus, methods working with bytes (insert, delete, get size) are no longer needed -- only block-level methods are left in `MetadataBlockRepository`

src/domain/core/src/repos/metadata_block_repository.rs
s373r: After [the hotfix](https://github.com/kamu-data/kamu-cli/pull/496), we don't need this structure -- we can just cache blocks, not bytes

Thus, methods working with bytes (insert, delete, get size) are no longer needed -- only block-level methods are left in `MetadataBlockRepository`

src/infra/core/src/repos/metadata_block_repository_impl.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn as_object_repo(&self) -> &dyn ObjectRepository {
+        &self.obj_repo
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+// Helpers
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub trait MetadataBlockRepositoryExt<ObjRepo>
```
s373r, line 109: Используется для следующего фрагмента:
```rust
impl<WrappedRepo, ObjRepo> MetadataBlockRepositoryCachingInMem<WrappedRepo, ObjRepo>
where
    WrappedRepo: MetadataBlockRepositoryExt<ObjRepo> + Sync + Send,
    ObjRepo: ObjectRepository + Sync + Send,
{
    pub fn new(obj_repo: ObjRepo) -> Self {
        let wrapped = WrappedRepo::new(obj_repo); // <---

        MetadataBlockRepositoryCachingInMem::new_wrapped(wrapped)
    }
...
    ```

Comment (sergiimk) [reply src/infra/core/src/repos/metadata_block_repository_impl.rs:109]: I think this brings more complexity than it solves.

If in `dataset_factory.rs` instead of:
```rust
MetadataBlockRepositoryImplWithCache::new(ObjectRepositoryLocalFSSha3::new(
  layout.blocks_dir,
)),
```

we created instance as:
```rust
MetadataBlockRepositoryCachingInMem::new(
  MetadataBlockRepositoryImpl::new(
    ObjectRepositoryLocalFSSha3::new(layout.blocks_dir)
  )
)
```
It would:
- make `MetadataBlockRepositoryExt` trait not necessary
- make `pub type MetadataBlockRepositoryImplWithCache` alias not necessary
- also I think it would make compositional nature of cache more explicit - I find it really like to see how objects compose upon creation
- and also it might make it easier to introduce a per-process / redis cache in future - we will just slot in another layer of caching under InMem - in current code we won't be able to add another wrapper in the middle because `MetadataBlockRepositoryImpl` is hidden from the construction process

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -76,7 +76,10 @@ pub trait MetadataChain: Send + Sync {
     ) -> Result<Multihash, AppendError>;
 
     fn as_object_repo(&self) -> &dyn ObjectRepository;
```
zaychenko-sergei, line 78: I expect `as_object_repo` should be removed completely and `MetadataBlockRepository` would have those operations about `get_size` and `get_bytes`.

zaychenko-sergei (changes requested):

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
```diff
@@ -0,0 +1,119 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub type MetadataBlockRepositoryImplWithCache<ObjRepo> =
+    MetadataBlockRepositoryCachingInMem<MetadataBlockRepositoryImpl<ObjRepo>, ObjRepo>;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct MetadataBlockRepositoryCachingInMem<WrappedRepo, ObjRepo> {
+    wrapped: WrappedRepo,
+    cache: DashMap<Multihash, MetadataBlock>,
```
zaychenko-sergei, line 39: We discussed lazy computation of deserialized version + storing raw bytes initially.

src/domain/core/src/entities/metadata_chain.rs
s373r: Done, returned back

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
s373r: Done, returned back

zaychenko-sergei (approved):

src/infra/core/src/repos/metadata_block_repository_impl.rs
s373r: Done

zaychenko-sergei (approved):

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -286,6 +287,18 @@ pub enum GetBlockError {
     ),
 }
 
+impl From<GetError> for GetBlockError {
```
sergiimk, line 290: Minor: This error seems to logically belong next to `MetadataBlockRepo` trait, just like `GetBlockDataError`.

sergiimk (commented):
One minor comment, looks good otherwise

src/domain/core/src/entities/metadata_chain.rs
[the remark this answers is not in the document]
s373r: Done via https://github.com/kamu-data/kamu-cli/pull/505

PR #487, opened 2024-02-04
Simlified structure patterns for irrelevant fields

zaychenko-sergei:
## Description

Inspired by reading "Programming Rust" book.

- [ ] [CHANGELOG.md](./CHANGELOG.md) updated

### DIFF
--- a/src/app/cli/src/commands/log_command.rs
+++ b/src/app/cli/src/commands/log_command.rs
@@ -196,7 +196,7 @@ impl AsciiRenderer {
         )?;
 
         match &block.event {
-            MetadataEvent::SetDataSchema(e @ SetDataSchema { schema: _ }) => {
+            MetadataEvent::SetDataSchema(e @ SetDataSchema { .. }) => {
                 let schema = e.schema_as_arrow().unwrap();
                 let schema_str = kamu_data_utils::schema::format::format_schema_arrow(&schema);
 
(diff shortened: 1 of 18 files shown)


src/app/cli/src/commands/log_command.rs
```diff
@@ -343,25 +343,14 @@ impl AsciiRenderer {
                 }
                 self.render_property(output, 0, "WebsiteURL", &e.website_url)?;
             }
-            MetadataEvent::SetPollingSource(SetPollingSource {
-                fetch: _,
-                prepare: _,
-                read: _,
-                preprocess: _,
-                merge: _,
```
sergiimk, line 351: In this file I was specifying all fields on purpose to force whoever is changing metadata to consider this command. The `..` pattern is more like "don't care and won't ever care", so exhaustive match has its uses.

src/domain/task-system/src/entities/task_state.rs
```diff
@@ -68,19 +68,18 @@ impl Projection for TaskState {
...
+                    E::TaskRunning(TaskEventRunning { event_time, .. })
+                        if s.status == TaskStatus::Queued =>
+                    {
+                        Ok(Self {
+                            status: TaskStatus::Running,
+                            ran_at: Some(event_time),
+                            ..s
+                        })
+                    }
+                    E::TaskCancelled(TaskEventCancelled { event_time, .. })
```
sergiimk, line 80: Particularly in the case of event projections I think the old style is much better to force you to consider projection when adding a field to an event.

zaychenko-sergei: I actually started this change to avoid dummy fields in FlowsState. It's unlikely one will be changing events and not changing the state, this is a very local design decision.

src/app/cli/src/commands/log_command.rs
zaychenko-sergei: Looking at the statement body, it does not look like it was supposed to be ever extended.

s373r (approved):
Great, it's easier to read now

sergiimk (approved):

Issue #429, opened 2024-01-09
Implement `--recursive` flag support in CLI command


sergiimk:
Several of our commands currently don't handle `--recursive` flag, failing with `not implemented` error.

This ticket is to properly implement these cases:
- `kamu verify --recursive my.dataset` - should verify this and all upstream datasets in breadth-first order
- `kamu delete --recursive my.dataset` - should delete this and all downstream datasets in depth-first order
- `kamu delete a b c` - should figure out the dependencies between the datasets and delete them in a way that doesn't cause dangling references.

The implementation should reuse the newly-introduced dataset dependency graph.

Special consideration should be paid to cases when some of the dataset dependencies are not in the workspace (e.g. we pulled a remote derivative dataset, but didn't pull its inputs).

PR #493, opened 2024-02-07
Add recursive flag for verify and delete commands

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/429

### DIFF
--- a/src/app/cli/src/commands/delete_command.rs
+++ b/src/app/cli/src/commands/delete_command.rs
@@ -9,7 +9,7 @@
 
 use std::sync::Arc;
 
-use futures::{StreamExt, TryStreamExt};
+use futures::{future, StreamExt, TryStreamExt};
 use kamu::domain::*;
 use kamu::utils::datasets_filtering::filter_datasets_by_pattern;
 use opendatafabric::*;
@@ -93,14 +93,19 @@ impl Command for DeleteCommand {
         let confirmed = if self.no_confirmation {
             true
         } else {
+            let dataset_aliases = future::join_all(dataset_refs.iter().map(|dataset_id| async {
+                let dataset_hdl = self
+                    .dataset_repo
+                    .resolve_dataset_ref(dataset_id)
+                    .await
+                    .unwrap();
+                dataset_hdl.alias.to_string()
+            }))
+            .await;
             common::prompt_yes_no(&format!(
                 "{}: {}\n{}\nDo you wish to continue? [y/N]: ",
                 console::style("You are about to delete following dataset(s)").yellow(),
-                dataset_refs
-                    .iter()
-                    .map(ToString::to_string)
-                    .collect::<Vec<_>>()
-                    .join(", "),
+                dataset_aliases.join(", "),
                 console::style("This operation is irreversible!").yellow(),
             ))
         };
--- a/src/app/cli/src/commands/delete_command.rs
+++ b/src/app/cli/src/commands/delete_command.rs
@@ -62,7 +62,7 @@ impl Command for DeleteCommand {
                 self.dataset_repo.as_ref(),
                 self.dataset_ref_patterns.clone(),
             )
-            .map_ok(|dataset_handle| dataset_handle.as_local_ref().id().unwrap().clone())
+            .map_ok(|dataset_handle| dataset_handle.id)
             .try_collect()
             .await?;
 
--- a/src/app/cli/src/commands/delete_command.rs
+++ b/src/app/cli/src/commands/delete_command.rs
@@ -56,7 +56,7 @@ impl Command for DeleteCommand {
         }
 
         let dataset_refs: Vec<_> = if self.all {
-            unimplemented!("Recursive deletion is not yet supported")
+            unimplemented!("All deletion is not yet supported")
         } else {
             let dataset_ids: Vec<_> = filter_datasets_by_pattern(
                 self.dataset_repo.as_ref(),
@@ -68,7 +68,7 @@ impl Command for DeleteCommand {
 
             if self.recursive {
                 self.dependency_graph_service
-                    .get_all_downstream_dependencies(dataset_ids)
+                    .get_recursive_downstream_dependencies(dataset_ids)
                     .await
                     .int_err()?
                     .map(DatasetRef::ID)
@@ -105,7 +105,7 @@ impl Command for DeleteCommand {
             common::prompt_yes_no(&format!(
                 "{}: {}\n{}\nDo you wish to continue? [y/N]: ",
                 console::style("You are about to delete following dataset(s)").yellow(),
-                dataset_aliases.join(", "),
+                dataset_aliases.join("\n"),
                 console::style("This operation is irreversible!").yellow(),
             ))
         };
--- a/src/app/cli/src/commands/delete_command.rs
+++ b/src/app/cli/src/commands/delete_command.rs
@@ -9,7 +9,7 @@
 
 use std::sync::Arc;
 
-use futures::TryStreamExt;
+use futures::{StreamExt, TryStreamExt};
 use kamu::domain::*;
 use kamu::utils::datasets_filtering::filter_datasets_by_pattern;
 use opendatafabric::*;
@@ -19,6 +19,7 @@ use super::{common, CLIError, Command};
 pub struct DeleteCommand {
     dataset_repo: Arc<dyn DatasetRepository>,
     dataset_ref_patterns: Vec<DatasetRefPattern>,
+    dependency_graph_service: Arc<dyn DependencyGraphService>,
     all: bool,
     recursive: bool,
     no_confirmation: bool,
@@ -28,6 +29,7 @@ impl DeleteCommand {
     pub fn new<I>(
         dataset_repo: Arc<dyn DatasetRepository>,
         dataset_ref_patterns: I,
+        dependency_graph_service: Arc<dyn DependencyGraphService>,
         all: bool,
         recursive: bool,
         no_confirmation: bool,
@@ -38,6 +40,7 @@ impl DeleteCommand {
         Self {
             dataset_repo,
             dataset_ref_patterns: dataset_ref_patterns.into_iter().collect(),
+            dependency_graph_service,
             all,
             recursive,
             no_confirmation,
@@ -54,16 +57,29 @@ impl Command for DeleteCommand {
 
         let dataset_refs: Vec<_> = if self.all {
             unimplemented!("Recursive deletion is not yet supported")
-        } else if self.recursive {
-            unimplemented!("Recursive deletion is not yet supported")
         } else {
-            filter_datasets_by_pattern(
+            let dataset_ids: Vec<_> = filter_datasets_by_pattern(
                 self.dataset_repo.as_ref(),
                 self.dataset_ref_patterns.clone(),
             )
-            .map_ok(|dataset_handle| dataset_handle.as_local_ref())
+            .map_ok(|dataset_handle| dataset_handle.as_local_ref().id().unwrap().clone())
             .try_collect()
-            .await?
+            .await?;
+
+            if self.recursive {
+                self.dependency_graph_service
+                    .get_all_downstream_dependencies(dataset_ids)
+                    .await
+                    .int_err()?
+                    .map(DatasetRef::ID)
+                    .collect()
+                    .await
+            } else {
+                dataset_ids
+                    .iter()
+                    .map(|dataset_id| DatasetRef::ID(dataset_id.clone()))
+                    .collect()
+            }
         };
 
         if dataset_refs.is_empty() {
--- a/src/app/cli/src/commands/verify_command.rs
+++ b/src/app/cli/src/commands/verify_command.rs
@@ -158,7 +158,8 @@ impl VerifyCommand {
             .await)
     }
 
-    // Return
+    // Return tuple with filtered VerificationRequests(check existing)
+    //   with list of missed remote dependencies
     async fn check_remote_datasets(
         &self,
         verifiation_requests: Vec<VerificationRequest>,
--- a/src/app/cli/src/commands/verify_command.rs
+++ b/src/app/cli/src/commands/verify_command.rs
@@ -92,7 +92,7 @@ impl VerifyCommand {
 
         let requests: Vec<_> = if self.recursive {
             self.dependency_graph_service
-                .get_all_upstream_dependencies(dataset_ids)
+                .get_recursive_upstream_dependencies(dataset_ids)
                 .await
                 .int_err()?
                 .map(|dataset_id| VerificationRequest {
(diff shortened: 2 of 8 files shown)


src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,102 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+            .map(|dataset_id| {
+                state
+                    .get_dataset_node(dataset_id)
+                    .map_err(GetDownstreamDependenciesError::DatasetNotFound)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let mut node_indexes_result = HashSet::new();
+
+        for node_index in &nodes_to_search {
+            let mut bfs = Dfs::new(&state.datasets_graph, *node_index);
```
zaychenko-sergei, line 268: Two methods are only different by the choice of the traversal algorithm. Try to make a more general algorithm with the traversal method as a parameter (callback i.e.).

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,102 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+            .map(|dataset_id| {
+                state
+                    .get_dataset_node(dataset_id)
+                    .map_err(GetUpstreamDependenciesError::DatasetNotFound)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let mut node_indexes_result = HashSet::new();
+
+        for node_index in &nodes_to_search {
+            let mut bfs = Bfs::new(&state.datasets_graph, *node_index);
```
zaychenko-sergei, line 220: I think that `Bfs` vs `Dfs` only defines the order of visiting the graph nodes (breadth-first vs depth-first). It does not define directions of traversal like you assume here.

I don't think we care about the exact traversal algorithm here, all we want is a set, so pick any.
However, in order to invert the direction of traversal, I think you'll need something like [Reversed ](https://docs.rs/petgraph/latest/petgraph/visit/struct.Reversed.html), but you have to experiment.

src/app/cli/src/commands/verify_command.rs
```diff
@@ -79,17 +82,35 @@ impl VerifyCommand {
...
+            self.dependency_graph_service
+                .get_all_upstream_dependencies(dataset_ids)
+                .await
+                .int_err()?
+                .map(|dataset_id| VerificationRequest {
+                    dataset_ref: DatasetRef::ID(dataset_id),
+                    block_range: (None, None),
+                })
+                .collect()
+                .await
```
zaychenko-sergei, line 103: I wonder if we care about the order of verification here.
Right now the order is undetermined, since goes through a hash set.
But it seems to me that the order is important, and we need to verify the most upstream datasets first and fail early, if there is any problem with those - the datasets from next levels are unlikely good, if 1st level is wrong.

For "delete" command, the order definitely matters. Each deletion will check if any downstream datasets are still dangling, and will fail.

zaychenko-sergei (changes requested):
Think of maybe providing an API from dependency graph that allows graph traversals with a callback instead of trying to pre-collect nodes.

I think we need to provide a traversal that starts from the roots, so the deepest parent datasets (deepest upstream) are visited first, and the searched node is visited last.

src/app/cli/src/commands/delete_command.rs
```diff
@@ -54,16 +57,29 @@ impl Command for DeleteCommand {
...
-        } else if self.recursive {
-            unimplemented!("Recursive deletion is not yet supported")
         } else {
-            filter_datasets_by_pattern(
+            let dataset_ids: Vec<_> = filter_datasets_by_pattern(
                 self.dataset_repo.as_ref(),
                 self.dataset_ref_patterns.clone(),
             )
-            .map_ok(|dataset_handle| dataset_handle.as_local_ref())
+            .map_ok(|dataset_handle| dataset_handle.as_local_ref().id().unwrap().clone())
```
sergiimk, line 65: I think this whole block should return a `Vec<DatasetHandle>`:
- You already have handles after filtering by pattern, so by mapping them to IDs you lose information
- You later need handles (dataset aliases) to ask for user confirmation
- I assume you mapped everything to IDs only because `dependency_graph_service` returns IDs, but we could resolve the result of `get_all_downstream_dependencies` to handles instead.

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,100 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+                    .map_err(GetUpstreamDependenciesError::DatasetNotFound)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut result = Vec::new();
+        let mut visited_indexes = HashSet::new();
+
+        let mut reversed_graph = state.datasets_graph.clone();
+
+        reversed_graph.reverse();
```
sergiimk, line 223: Let's put a `// TODO: PERF: ...` comment here as this looks like a quite expensive operation.

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,100 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+        &self,
+        dataset_ids: Vec<DatasetID>,
+    ) -> Result<DatasetIDStream, GetUpstreamDependenciesError> {
+        self.ensure_datasets_initially_scanned()
+            .await
+            .int_err()
+            .map_err(GetUpstreamDependenciesError::Internal)
+            .unwrap();
+
+        let state = self.state.lock().await;
```
sergiimk, line 207: Since we are working on a cloned + reversed graph below, do we need to be holding the lock for the entire duration of the operation, or we could clone under the lock and release it?

src/domain/core/src/services/dependency_graph_service.rs
```diff
@@ -36,6 +36,20 @@ pub trait DependencyGraphService: Sync + Send {
...
+    /// Iterates over all levels of dataset's upstream dependencies
+    /// and return result including passed parameters
+    async fn get_all_upstream_dependencies(
+        &self,
+        dataset_ids: Vec<DatasetID>,
+    ) -> Result<DatasetIDStream, GetUpstreamDependenciesError>;
+
+    /// Iterates over all levels of dataset's downstream dependencies
+    /// and return result including passed parameters
+    async fn get_all_downstream_dependencies(
```
sergiimk, line 49: We probably should indicate in the method name whether this is depth- or breadth-first traversal

src/app/cli/src/commands/delete_command.rs
sergiimk: Looks like it's also tied to `get_all_downstream_dependencies` expecting `Vec<DatasetID>` as input, and we probably don't want to complicate the dependency graph API with `DatasetRef`s ....

I guess let's keep it as is unless we think of a better solution.

src/infra/core/src/dependency_graph_service_inmem.rs
rmn-boiko: I already have custom solution for recursive upstrean bradth-first search. I will provide it here as alternative

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,100 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+                    .get_dataset_node(dataset_id)
+                    .map_err(GetDownstreamDependenciesError::DatasetNotFound)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let mut node_indexes_result = HashSet::new();
+        let mut result = vec![];
+
+        for node_index in &nodes_to_search {
+            let mut current_node_result = vec![];
+            depth_first_search(&state.datasets_graph, Some(*node_index), |event| {
```
zaychenko-sergei, line 269: DFS implementation seems to accept multiple start nodes, did you try this option?
It is also has a built-in functionality to ignore duplicate visits.

src/infra/core/src/dependency_graph_service_inmem.rs
zaychenko-sergei: This rather looks as a wrong idea. The library allows traversing in any order.
Both DFS and BFS algorithms call a function called "neighbors".
By default this function is returning outgoing edges, but it's possible to switch to incoming edges.

When I mentioned graph reversing in a Slack discussion, I did not mean "go clone a graph each time and reverse it physically". I meant a `Reversed` adapter from the library, that does not copy anything:

```
        let reversed_graph = Reversed(&state.datasets_graph);
```

Note - it's taking a reference, thus not copying neither taking ownership.
The only note, to access `node_weight`, use `.0` suffix to extract the original graph from the adapter's tuple struct.

src/infra/core/src/dependency_graph_service_inmem.rs
zaychenko-sergei: The lock has to stay, see the comment about `Reversed` adapter that does not clone the graph.

zaychenko-sergei (changes requested):

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,100 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+                                .node_weight(node_index_event)
+                                .unwrap()
+                                .clone(),
+                        );
+                    }
+                }
+            });
+
+            current_node_result.reverse();
+            result.extend(current_node_result);
```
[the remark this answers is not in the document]
rmn-boiko: It will return `C` only once. I have test for such case https://github.com/kamu-data/kamu-cli/pull/493/files#diff-f3141fd7d49c5e827432aa059d504ce5745eb3160a00c1b26d5cb5decbf825a0R484

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,6 +194,100 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+                        node_indexes_result.insert(node_index_event);
+
+                        current_node_result.push(
+                            state
+                                .datasets_graph
+                                .node_weight(node_index_event)
+                                .unwrap()
+                                .clone(),
+                        );
+                    }
```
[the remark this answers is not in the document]
rmn-boiko: I rewrote full dfs algo.

src/infra/core/src/dependency_graph_service_inmem.rs
rmn-boiko: Yep. I've tried it before but faced issue, that node we are searching for always at the begging, found good solution for that and changed it!

src/domain/core/src/services/dependency_graph_service.rs
```diff
@@ -29,13 +29,37 @@ pub trait DependencyGraphService: Sync + Send {
...
+
+    async fn run_recursive_depth_first_search(
+        &self,
+        dataset_ids: Vec<DatasetID>,
+    ) -> Result<Vec<DatasetID>, GetDependenciesError>;
+
+    async fn run_recursive_breadth_first_search(
+        &self,
+        dataset_ids: Vec<DatasetID>,
+    ) -> Result<Vec<DatasetID>, GetDependenciesError>;
```
zaychenko-sergei, line 62: Do we need to expose these operations in the trait? They feel like internal.

src/infra/core/src/dependency_graph_service_inmem.rs
```diff
@@ -193,23 +194,125 @@ impl DependencyGraphService for DependencyGraphServiceInMemory {
...
+
+        let nodes_to_search = dataset_ids
+            .iter()
+            .map(|dataset_id| {
+                state
+                    .get_dataset_node(dataset_id)
+                    .map_err(GetDependenciesError::DatasetNotFound)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
```
zaychenko-sergei, line 243: Repeating part

src/app/cli/src/commands/delete_command.rs
```diff
@@ -53,17 +56,30 @@ impl Command for DeleteCommand {
...
-            unimplemented!("Recursive deletion is not yet supported")
+            unimplemented!("All deletion is not yet supported")
         } else {
-            filter_datasets_by_pattern(
+            let dataset_ids: Vec<_> = filter_datasets_by_pattern(
                 self.dataset_repo.as_ref(),
                 self.dataset_ref_patterns.clone(),
             )
-            .map_ok(|dataset_handle| dataset_handle.as_local_ref())
+            .map_ok(|dataset_handle| dataset_handle.as_local_ref().id().unwrap().clone())
```
zaychenko-sergei, line 65:
```
/// A resolved handle to the local dataset
#[derive(Clone, PartialEq, Eq, Hash)]
pub struct DatasetHandle {
    pub id: DatasetID,
    pub alias: DatasetAlias,
}
```

Just take ID directly, what's the sense of converting to local ref ?

zaychenko-sergei (approved):

zaychenko-sergei (changes requested):
As discussed in Slack, the choice of DFS vs BFS is not critical for use.
In both verify & delete commands we go in revesersed traversal direction when applying the action on the node. The only difference is a direction of edges - towards roots in verify and towards leaf derived in delete.

zaychenko-sergei:
I think verification process should stop after 1st error, @sergiimk what do you think?

```
sergeiz@ubuntu:~/kamu/testing/container/workspace$ kamu inspect lineage canada.daily-cases
canada.daily-cases: Derivative
└── canada.case-details: Derivative
    ├── alberta.case-details.hm: Derivative
    │   └── alberta.case-details: Root
    ├── british-columbia.case-details.hm: Derivative
    │   └── british-columbia.case-details: Root
    ├── ontario.case-details.hm: Derivative
    │   └── ontario.case-details: Root
    └── quebec.case-details.hm: Derivative
        └── quebec.case-details: Root

sergeiz@ubuntu:~/kamu/testing/container/workspace$ kamu verify --recursive canada.daily-cases
  [1/1] Dataset is valid (alberta.case-details)                                                                                                                                                              [1/1] Dataset is valid (british-columbia.case-details)                                                                                                                                                     [1/1] Dataset is valid (ontario.case-details)                                                                                                                                                              [1/1] Dataset is valid (quebec.case-details)                                                                                                                                                               [1/1] Dataset is valid (alberta.case-details.hm)                                                                                                                                                           [1/1] Dataset is valid (british-columbia.case-details.hm)                                                                                                                                                  [1/1] Dataset is valid (ontario.case-details.hm)                                                                                                                                                           [1/1] Dataset is valid (quebec.case-details.hm)                                                                                                                                                            [1/1] Error during transformation (canada.case-details)                                                                                                                                                    [1/1] Error during transformation (canada.daily-cases)                                                                                                                                                   Error: Failed to verify 2 dataset(s)
```

rmn-boiko:
> I think the verification process should stop after 1st error, @sergiimk what do you think?
> 
> image

The issue is if we get two unconnected trees and when we fail verification in the first tree, it should not skip full validation, but only the rest of the first tree. Such as we don't have any connections between datasets during the verification process implementation will bring a lot of new logic in `VerificationService`

zaychenko-sergei (approved):

Issue #495, opened 2024-02-08
Flows: implement returning actual counters describing completed work in flow description API


zaychenko-sergei:
Currently Flow API has reserved some fields, which describe the amount of flow run results, depending on the flow type:

```
union FlowDescription = FlowDescriptionDatasetPollingIngest | FlowDescriptionDatasetPushIngest | FlowDescriptionDatasetExecuteTransform | FlowDescriptionDatasetCompaction | FlowDescriptionSystemGC

type FlowDescriptionDatasetCompaction {
	datasetId: DatasetID!
	originalBlocksCount: Int!
	resultingBlocksCount: Int
}

type FlowDescriptionDatasetExecuteTransform {
	datasetId: DatasetID!
	transformedRecordsCount: Int
}

type FlowDescriptionDatasetPollingIngest {
	datasetId: DatasetID!
	ingestedRecordsCount: Int
}

type FlowDescriptionDatasetPushIngest {
	datasetId: DatasetID!
	sourceName: String
	inputRecordsCount: Int!
	ingestedRecordsCount: Int
}

type FlowDescriptionSystemGC {
	dummy: Boolean!
}
```

None of the counters actually works, this was earlier postponed due to ODF breaking changes branch.
Now it's time to implement those.

PR #500, opened 2024-02-12
495 flows implement returning actual counters describing completed work in flow description api

zaychenko-sergei:
## Description

Closes: #495

### DIFF
--- a/src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
+++ b/src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
@@ -70,7 +70,7 @@ impl DatasetFlowRunsMut {
             })?;
 
         Ok(TriggerFlowResult::Success(TriggerFlowSuccess {
-            flow: Flow::new(flow_state),
+            flow: Box::new(Flow::new(flow_state)),
         }))
     }
 
@@ -97,7 +97,7 @@ impl DatasetFlowRunsMut {
         match res {
             Ok(flow_state) => Ok(CancelScheduledTasksResult::Success(
                 CancelScheduledTasksSuccess {
-                    flow: Flow::new(flow_state),
+                    flow: Box::new(Flow::new(flow_state)),
                 },
             )),
             Err(e) => match e {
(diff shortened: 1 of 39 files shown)


src/domain/task-system/src/entities/task_status.rs
```diff
@@ -33,4 +35,18 @@ pub enum TaskOutcome {
...
+        matches!(self, Self::Success(_))
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TaskResult {
+    Empty,
+    PullResult(Box<PullResult>),
```
sergiimk, line 49: Is this boxing forced by clippy or is your choice?

I think in Rust the idiomatic way is to hide memory management inside the objects.

In other words, if `PullResult` becomes too big to keep on the stack (which I don't think it is yet) - then we would box the internal state of the `PullResult`, while allowing all upper layers to use it as a simple movable type.

This way `PullResult` encapsulates how exactly it manages memory, while users of this type only have to respect whether it's `Copy` / `Clone` / `Send`. And when it becomes to big to keep on the stack - it can be constructed directly on the heap, without touching any code that uses it.

I wouldn't want to see our enums sprawling with `Box<>`  and `Arc<>` for different variants because it's really hard to reason about them on this level.

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
```diff
@@ -125,7 +125,7 @@ enum TriggerFlowResult {
 #[derive(SimpleObject)]
 #[graphql(complex)]
 struct TriggerFlowSuccess {
-    pub flow: Flow,
+    pub flow: Box<Flow>,
```
sergiimk, line 128: Related to my other comment - a box on this level seems a bit weird. If the `Flow` is getting too big for the stack or is getting too expensive to move, then `the Flow` itself should probably box its state internally. It's unclear to me why we decide to box on the GQL level, but not elsewhere.

sergiimk (approved):
Feature looks good to merge :+1: 

My only comments are for us to come up with a good strategy for when to box things and on what level.

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
zaychenko-sergei: Note, it's not the `Flow` from the FS, it's the API `Flow` equivalent, which holds FS's `FlowState`. I agree, putting a `Box` 1 level below made the code cleaner.

Issue #472, opened 2024-01-30
Add support for wildcard references for pull and push commands


rmn-boiko:
An issue [Support wildcard references in dataset operations ](https://github.com/kamu-data/kamu-cli/issues/336) introduced a new type called `DatasetRefPattern`. This type enables CLI commands to accept expressions as arguments, allowing them to perform actions on multiple datasets that match the specified pattern. Initially, this feature was implemented for the `kamu verify` and `kamu delete` commands.

The objective is to extend this capability to include the `kamu pull` and `kamu push` commands.

PR #506, opened 2024-02-15
Add wildcard for pull and push

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/472


src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -821,4 +842,22 @@ impl std::str::FromStr for DatasetRefPattern {
...
+                Ok(dataset_pattern) => Ok(Self::Pattern(dataset_pattern)),
+                Err(_) => Err(Self::Err::new(s)),
+            };
+        }
+        match DatasetRefAny::from_str(s) {
+            Ok(dataset_ref) => Ok(Self::RefAny(dataset_ref)),
+            Err(_) => Err(Self::Err::new(s)),
+        }
+    }
+}
```
zaychenko-sergei, line 860: This style of parsing and `DatasetPattern` model is missing a case of multi-tenant remote alias:
   `RepoName/AccountName/DatasetNamePattern`

I guess we might need `DatasetRemotePattern` to cut off both names before dataset name pattern.

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -48,3 +51,48 @@ pub fn filter_datasets_by_pattern(
...
+                    )
+                })
+                .map_ok(|dataset_handle| dataset_handle.as_any_ref())
+                .map_err(Into::into)
+                .boxed(),
+        );
+    }
+
+    result
+}
```
zaychenko-sergei, line 98: I would suggest to simplify return type to be a simple stream.

Internally you coud simply concat 2 streams, I don't understand any need in this static string marker.

Stream 1:
```
Box::pin(async_stream::try_stream! {
   //Yield all non-pattern any-refs.
}
```

Stream2: If any patterns present, at stage 2, iterate datasets and call `is_match`.

Concatenate stream 1 and stream 2 with [`chain`](https://dtantsur.github.io/rust-openstack/tokio/stream/trait.StreamExt.html#method.chain)

zaychenko-sergei (changes requested):
Evaluate if it's really difficult to support patterns in :
  - `AccountName` (both local and any case)
  - `RepoName` (any case)
  
It seems that the base element grammar is identical, so adding patterns support should be trivial, assuming we don't allow pattern to hide separating `'/'` symbol.

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -739,79 +739,117 @@ impl std::cmp::PartialOrd for DatasetRefAny {
     }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// DatasetRefPattern
+////////////////////////////////////////////////////////////////////////////////
+
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub enum DatasetRefPattern {
+pub enum DatasetRefPatternLocal {
```
sergiimk, line 747: The old name seemed more consistent with the rest of code:
- `DatasetRef` - local reference
- `DatasetRefRemote` - remote reference

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -739,79 +739,117 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
-    Pattern(Option<AccountName>, DatasetNamePattern),
+    Pattern(DatasetPattern),
 }
 
-impl DatasetRefPattern {
-    pub fn is_match(&self, dataset_hande: &DatasetHandle) -> bool {
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DatasetRefPatternAny {
+    Pattern(DatasetPatternAny),
+    RefAny(DatasetRefAny),
```
sergiimk, line 755: Perhaps this type should look like this:
```rust
enum DatasetRefAnyPattern {
    Ref(DatasetRefAny),
    LocalPattern(Option<AccountName>, DatasetNamePattern),
    RemotePattern(RepoName, Option<AccountName>, DatasetNamePattern),
    AmbiguousPattern(Arc<str>, DatasetNamePattern),
}
```
This might eliminate the need for a lot of intermediate types.

src/domain/opendatafabric/src/identity/dataset_refs.rs
```diff
@@ -739,79 +739,117 @@ impl std::cmp::PartialOrd for DatasetRefAny {
...
+pub enum DatasetRefPatternLocal {
     Ref(DatasetRef),
-    Pattern(Option<AccountName>, DatasetNamePattern),
+    Pattern(DatasetPattern),
 }
 
-impl DatasetRefPattern {
-    pub fn is_match(&self, dataset_hande: &DatasetHandle) -> bool {
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DatasetRefPatternAny {
```
sergiimk, line 753: I think this should be called `DatasetRefAnyPattern` as it's a superset of the `DatasetRefAny` type.

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -238,8 +238,223 @@ impl DatasetNamePattern {
...
+                Err(_) => Err(Self::Err::new(s)),
+            },
+        }
+    }
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DatasetPatternAny {
```
sergiimk, line 328: I don't think this type is needed (see my comment on `DatasetRefPatternAny`)

src/app/cli/src/cli_parser.rs
```diff
@@ -702,7 +702,7 @@ pub fn cli() -> Command {
                         Arg::new("dataset")
                             .action(ArgAction::Append)
                             .index(1)
-                            .value_parser(value_parse_dataset_ref_any)
+                            .value_parser(value_parse_dataset_ref_pattern_any)
```
sergiimk, line 705: Please add an example of using patterns to command help

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -48,3 +60,73 @@ pub fn filter_datasets_by_pattern(
...
+    let clone_dataset_ref_patterns = dataset_ref_patterns.clone();
+
+    let static_stream = Box::pin(async_stream::try_stream! {
+        for dataset_ref_pattern in &dataset_ref_patterns.clone() {
+            if !dataset_ref_pattern.is_pattern() {
+                yield dataset_ref_pattern
+                    .as_dataset_ref_any()
+                    .unwrap()
+                    .clone();
+            } else if dataset_ref_pattern.is_remote() {
```
sergiimk, line 78: what happens to local patters like `dataset%` or `acc/dataset%`?

if I'm reading this correctly they will be skipped?

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -41,10 +63,145 @@ pub fn filter_datasets_by_pattern(
...
+        .chain(remote_patterns_stream)
+        .chain(local_patterns_stream)
+        .boxed()
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+fn get_static_datasets_stream(
+    static_refs: Vec<DatasetRefAnyPattern>,
+) -> FilteredDatasetRefAnyStream<'static> {
```
sergiimk, line 106: Since this function is private to the module and the stream you create here is later boxed in `filter_datasets_by_any_pattern` after chaining anyway, I think boxing here is not necessary. 

You might be able to return `impl TryStream<..>` and avoid allocation.

Same for `get_remote_datasets_stream`.

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -41,10 +63,145 @@ pub fn filter_datasets_by_pattern(
...
+    search_svc: Arc<dyn SearchService>,
+    remote_ref_patterns: Vec<DatasetRefAnyPattern>,
+    in_multitenant_mode: bool,
+) -> FilteredDatasetRefAnyStream<'static> {
+    Box::pin(async_stream::try_stream! {
+        for remote_ref_pattern in &remote_ref_patterns.clone() {
+            // TODO: potentially low performance solution,as it will always fully scan a remote repo.
+            // Should be improved after search will support wildcarding.
+            let maybe_repo_name = remote_ref_pattern.pattern_repo_name(in_multitenant_mode);
+            assert!(maybe_repo_name.is_some());
```
sergiimk, line 129: no point asserting if you later `unwrap()` - both result in a panic

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -41,10 +63,145 @@ pub fn filter_datasets_by_pattern(
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+fn get_remote_datasets_stream(
+    search_svc: Arc<dyn SearchService>,
+    remote_ref_patterns: Vec<DatasetRefAnyPattern>,
+    in_multitenant_mode: bool,
+) -> FilteredDatasetRefAnyStream<'static> {
+    Box::pin(async_stream::try_stream! {
+        for remote_ref_pattern in &remote_ref_patterns.clone() {
```
sergiimk, line 125: Why do we need to clone a `Vec` (that is already passed by value) just to iterate it by reference?

sergiimk (commented):
Few minor comments, but otherwise looks great :+1:

src/infra/core/src/utils/datasets_filtering.rs
zaychenko-sergei: I disagree. Mindful assertion is clearly showing an intent. `unwrap` often indicates draft code where we simply forgot to do error checking.

sergiimk: `expect` would probably be the best way to combine the two into one statement

src/infra/core/src/utils/datasets_filtering.rs
```diff
@@ -41,10 +63,146 @@ pub fn filter_datasets_by_pattern(
...
+        .boxed()
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+fn get_static_datasets_stream(
+    static_refs: Vec<DatasetRefAnyPattern>,
+) -> FilteredDatasetRefAnyStream<'static> {
+    Box::pin(async_stream::try_stream! {
+        for static_ref in &static_refs.clone() {
```
sergiimk, line 108: Unnecessary clone - we already own the `Vec` here.

Also if we use `static_refs.into_iter()` the below
```rust
yield static_ref
                .as_dataset_ref_any()
                .unwrap()
                .clone();
```
could be made into
```rust
yield static_ref
                .into_dataset_ref_any()
                .unwrap();
```
to avoid all the cloning.

src/infra/core/src/utils/datasets_filtering.rs
zaychenko-sergei: Corrected. 3 impls that get chained and boxed at the very end

Issue #401, opened 2023-12-27
Flows: properly support batching/throttling conditions


zaychenko-sergei:
Batching condition to execute Derived dataset flows, such as EXECUTE_QUERY, should have 3 properties:
 - min records to accumulate
 - max batching interval

So, the batching waiting stays waiting until the minimal number of records on the input accumulate, or until maximum batching interval time passes.

For the stake of simplicity, it should be enough to count total sum of records arrived at any input of the dataset - there is no need to have separate configuration values per specific input dataset.

Also, a watermark modification should be considered an important enough event to satisfy the batching condition.
 
Throttling period (flows are not run more frequent than given period) is a separate condition, and most likely, it will be not be editable in future for free datasets, rather it will become a system setting.

PR #507, opened 2024-02-15
401 flows properly support batching conditions

zaychenko-sergei:
## Description

Closes: #401.

Implemented batching conditions in the flow system for derived datasets.

- [ ] API changes are backwards-compatible - NO, UI needs update


resources/schema.gql
```diff
@@ -686,8 +691,8 @@ type FlowConfiguration {
 }
 
 type FlowConfigurationBatching {
```
sergiimk, line 693: Does this control ingest or transform batching, or both? Perhaps we should rename it to make this more explicit?

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_configs_mut.rs
```diff
@@ -199,6 +198,14 @@ impl From<TimeDeltaInput> for chrono::Duration {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+#[derive(InputObject)]
+struct BatchingConditionInput {
+    pub min_records_awaited: u64,
+    pub max_batching_interval: Option<TimeDeltaInput>,
```
sergiimk, line 204: Should the interval be required? Perhaps even capped at 24 hours.

I think we should reserve batching for dealing with high-frequency data. If we allow infinite interval batching that would mean that certain transform flow can potentially live forever ... which may mess with event store compactions, or monitoring/alerting or hanged flows.

I was thinking that at least `max_interval` will be driving the watermark propagation.

src/adapter/graphql/src/queries/flows/flow_start_condition.rs
```diff
@@ -43,13 +43,19 @@ impl From<fs::FlowStartConditionThrottling> for FlowStartConditionThrottling {
 
 #[derive(SimpleObject)]
 pub(crate) struct FlowStartConditionBatching {
-    pub threshold_new_records: usize,
+    pub active_batching_rule: FlowConfigurationBatching,
+    pub awaited_by_now: TimeDelta,
```
sergiimk, line 47: Why not track `batch_started_at` time instead?

Or if we require `max_interval` then this field could become something like `batch_deadline`.

Either way I think having absolute time in API is much nicer than relative deltas that are different on each request.

src/adapter/http/src/smart_protocol/messages.rs
```diff
@@ -185,12 +185,13 @@ pub struct MetadataBlocksBatch {
...
 
-/// Transfer size estimate
+/// Transfer plan
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
-pub struct TransferSizeEstimate {
+pub struct TransferPlan {
     pub num_blocks: u32,
     pub num_objects: u32,
     pub num_records: u64,
+    pub new_watermark: Option<DateTime<Utc>>,
```
sergiimk, line 194: I'm very much against exposing watermark on transfer protocol level - if looks out of place here.

Whoever is performing the transfer is free to analyze blocks to see if watermark changed or not, but let's not embed it into the protocol.

src/domain/core/src/services/pull_service.rs
```diff
@@ -169,6 +169,7 @@ pub enum PullResult {
         new_head: Multihash,
         num_blocks: u64,
         num_records: u64,
+        new_watermark: Option<DateTime<Utc>>,
```
sergiimk, line 172: While this is going in a direction of making pull results more informative, I'm starting to think that maybe we should do the opposite. 

Perhaps we should have all results only specify `old_head` and `new_head` and let whoever wants - can scan this block range and grab any information they want (number of records, watermark, schema changes etc.)

src/domain/flow-system/src/entities/flow/flow_state.rs
```diff
@@ -21,8 +21,8 @@ pub struct FlowState {
     pub flow_id: FlowID,
     /// Flow key
     pub flow_key: FlowKey,
-    /// Primary flow trigger
-    pub primary_trigger: FlowTrigger,
+    /// Triggers
+    pub triggers: Vec<FlowTrigger>,
```
sergiimk, line 25: yay, I was hoping this will happen eventually :)

src/domain/flow-system/src/entities/shared/batching_rule.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn min_records_awaited(&self) -> u64 {
+        self.min_records_awaited
+    }
+
+    #[inline]
+    pub fn max_batching_interval(&self) -> &Option<Duration> {
+        &self.max_batching_interval
+    }
+
+    pub fn evaluate<'a, I: Iterator<Item = &'a FlowTrigger>>(
```
sergiimk, line 42: I doubt we expect triggers to be in any more complex data structures than `Vec<>`, 
so perhaps we can accept `&[FlowTrigger]` instead of an iterator?

src/infra/core/src/sync_service_impl.rs
```diff
@@ -493,6 +506,23 @@ impl SyncServiceImpl {
...
+    fn new_watermark(block: &MetadataBlock) -> Option<DateTime<Utc>> {
+        match &block.event {
+            MetadataEvent::AddData(add_data) => {
+                if add_data.new_watermark.is_some() {
+                    return add_data.new_watermark;
+                }
+            }
+            MetadataEvent::ExecuteTransform(transform) => {
+                if transform.new_watermark.is_some() {
+                    return transform.new_watermark;
```
sergiimk, line 518: See description of the field: https://github.com/open-data-fabric/open-data-fabric/blob/master/schemas/metadata-events/AddData.json#L30

If `new_watermark` is present it doesn't necessarily mean that it was updated by the last operation. Once watermark appears it is "carried over" from one block to another. This is done so that when constructing next operation we didn't have to scan too deep to understand what the current watermark is.

src/domain/flow-system/src/entities/flow/flow_outcome.rs
```diff
@@ -46,14 +56,17 @@ impl From<&ts::TaskResult> for FlowResult {
...
                 }),
                 PullResult::Updated {
                     num_blocks,
                     num_records,
+                    new_watermark,
                     ..
                 } => Self::DatasetUpdate(FlowResultDatasetUpdate {
                     num_blocks,
                     num_records,
+                    watermark_modified: new_watermark.is_some(),
```
sergiimk, line 69: Don't think this will give results you expect (see my other comment about semantics of `new_watermark`).

src/domain/core/src/services/pull_service.rs
sergiimk: Nice consequence is that this would move "stats gathering" from push-pull ingest / transform / sync to just two places: CLI command output, and flow system.

Downside is the additional metadata reads.

sergiimk (changes requested):

resources/schema.gql
```diff
@@ -96,6 +96,11 @@ type AuthMut {
 	accountDetails(accessToken: String!): Account!
 }
 
+input BatchingConditionInput {
+	minRecordsAwaited: Int!
```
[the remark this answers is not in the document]
zaychenko-sergei: perhaps

resources/schema.gql
zaychenko-sergei: We don't know how push ingest will look like. Anyway, the configuration tries to be abstract from flow details. It could be  a totally different meaning, i.e., number of re-verified records for a potential verification flow.

zaychenko-sergei: Or it could be the compactiion control - one thing is to compact by schedule, another thing is to compact when accumulated a million of records, but not less than once in a month.

src/adapter/http/src/smart_protocol/messages.rs
zaychenko-sergei: I was trying to avoid the excessive re-scan.

src/infra/core/src/sync_service_impl.rs
zaychenko-sergei: Then why it is called so confusing? Name it `last_watermark` instead.
So, I guess I basically have to go through the range and detect changes in this field on the updated interval.

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_configs_mut.rs
zaychenko-sergei: Agree. Made the field required, and added upper bound validation result.

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_configs_mut.rs
```diff
@@ -199,6 +198,14 @@ impl From<TimeDeltaInput> for chrono::Duration {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+#[derive(InputObject)]
+struct BatchingConditionInput {
```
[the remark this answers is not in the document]
sergiimk: It wouldn't hurt to start writing them for new types.

src/infra/core/src/sync_service_impl.rs
sergiimk: If follows the `prev*` / `new*` convention for input/output fields of some operation introduced as part of the [RFC-014](https://github.com/open-data-fabric/open-data-fabric/blob/master/rfcs/014-minimize-offset-scanning.md). I can see how it can be confusing in this case, but also how `last_watermark` can be confusing for a transfrom - e.g. is it a watermark before or after the transformation?

src/domain/flow-system/src/entities/shared/batching_rule.rs
zaychenko-sergei: I originally started from the idea of walking over primary trigger and secondary triggers in events, but later simplified.

src/adapter/graphql/src/queries/flows/flow_start_condition.rs
zaychenko-sergei: Ended with `batching_deadline`.

resources/schema.gql
zaychenko-sergei: Decided to keep it as is for now. Will be more clear when adding new flow types.

src/adapter/http/src/smart_protocol/messages.rs
zaychenko-sergei: Reverted

src/infra/core/src/sync_service_impl.rs
zaychenko-sergei: Reverted

src/domain/core/src/services/pull_service.rs
zaychenko-sergei: Refactored to additional scanning outside `PullService`

src/infra/task-system-inmem/src/task_executor_inmem.rs
```diff
@@ -64,6 +76,131 @@ impl TaskExecutorInMemory {
             })
             .await
     }
+
+    async fn make_dataset_update_result(
+        &self,
+        dataset_id: &DatasetID,
+        pull_result: PullResult,
+    ) -> Result<TaskUpdateDatasetResult, InternalError> {
+        Ok(match pull_result {
```
sergiimk, line 85: Too much nesting here is making code hard to read. I'd suggest an early return on `PullResult::UpToDate`.

src/domain/flow-system/src/entities/shared/batching_rule.rs
```diff
@@ -0,0 +1,124 @@
...
+        let mut accumulated_records_count = 0;
+        let mut watermark_modified = false;
+
+        // Scan each accumulated trigger to decide
+        for trigger in triggers {
+            if let FlowTrigger::InputDatasetFlow(trigger) = trigger {
+                match &trigger.flow_result {
+                    FlowResult::Empty => {}
+                    FlowResult::DatasetUpdate(update) => {
+                        accumulated_records_count += update.num_records;
```
sergiimk, line 72: If we have datasets:
- A (root)
- B, C, D (derivative) all using A as an input

If A has 1,000 updates, will we have 3,000 triggers (per each downstream dataset) saved in our event store?

I thought that an update of an input will write a trigger once, and all subsequent updates to the same input will not produce a new trigger. I.e. maximum amount of `FlowTrigger::InputDatasetFlow` in our database is capped at the number of inputs used by all derivative datasets.

But here, from how you are summing up `num_records`, it looks like for every update a new trigger will be written. I don't suppose you are updating `num_records` of previous events.

If that's indeed the case this seems highly problematic for high-frequency root datasets that become very popular and are used as inputs of thousands of derivative datasets.

sergiimk (changes requested):

src/infra/task-system-inmem/src/task_executor_inmem.rs
zaychenko-sergei: Split on 2 methods. I've also reduce 1 level by utilizing `std::mem::drop` for the stream to unborrow refs.

src/domain/flow-system/src/entities/shared/batching_rule.rs
zaychenko-sergei: I've implemented reduction function for flow triggers that accumulates results of identical triggering inputs, however we will still have large amounts of ES-events in each derived flow, and that is something harder to solve.

Perhaps, any high frequency dataset:
- need a buffering at `PushSource` level anyway
- will anyway be a subject to trottling (currently 1 minute)

src/adapter/graphql/src/queries/flows/flow_start_condition.rs
```diff
@@ -19,12 +21,47 @@ pub(crate) enum FlowStartCondition {
...
+                            total_increment += dataset_changes_service
+                                .get_increment_since(
+                                    &dataset_trigger.dataset_id,
+                                    dataset_update.old_head.as_ref(),
+                                )
+                                .await
+                                .int_err()?;
+                        }
+                    }
+                }
```
sergiimk, line 54: Let's suppose the metadata chain of an input looks like this:
```
... <- b1 <- b2 <- b3 <- b4 <- b5
```

Our derivative dataset's flow contains this set of triggers:
```
InputDatasetFlow(DatasetUpdate(old_head=b2, new_head=b3))
InputDatasetFlow(DatasetUpdate(old_head=b3, new_head=b4))
```

While in our last `ExecuteTransform` the last processed input block is `b1`.

i.e. we **for whatever reason missed two transitions**:
```
b1 to b2
b4 to b5
```

If I read this right - this code seems to calculate something completely whacky like:
```
num_records (b2, b5] + num_records (b3, b5]
```

While it should either do:
- `num_records (b2, b3] + num_records (b3, b4]` - sum of intervals our triggers know about
- `num_records (b2, b5]` - using only **first trigger per input** and allowing to scan until the current `head`
- `num_records (b1, b5]` - **the correct answer**, but would require finding last `ExecuteTransform` of our derivative dataset instead of looking at triggers

src/domain/flow-system/src/entities/shared/batching_rule.rs
```diff
@@ -0,0 +1,134 @@
...
+                match &trigger.flow_result {
+                    FlowResult::Empty => {}
+                    FlowResult::DatasetUpdate(update) => {
+                        // Compute increment since the first trigger by this dataset.
+                        // Note: there might have been multiple updates since that time.
+                        // We are only recording the first trigger of particular dataset.
+                        let increment = dataset_changes_service
+                            .get_increment_since(&trigger.dataset_id, update.old_head.as_ref())
+                            .await
+                            .int_err()?;
```
sergiimk, line 80: If we are still allowing multiple `InputDatasetFlow` triggers per input - I don't see what stopping this loop from double / tripple-counting the same input.

src/adapter/graphql/src/queries/flows/flow_start_condition.rs
sergiimk: When we last spoke I saw two solutions:
- Store only one trigger per dataset and calculate all batching info from metadata of both derivative dataset and its inputs (using last `ExecuteTransfrom`)
- Store all triggers with `FlowResult::DatasetUpdate` inside of them and calculate batching based on stats **inside those events**

Current solution seems in between the two - it stores all triggers, but still evaluates batching based on inputs' metadata. 

Am I misunderstanding something?

sergiimk (changes requested):

src/domain/flow-system/src/entities/shared/batching_rule.rs
zaychenko-sergei: We are storing only 1 trigger per dataset

src/adapter/graphql/src/queries/flows/flow_start_condition.rs
zaychenko-sergei: 1) We are storing 1 trigger per dataset
2) The assumption is that flow system cannot miss any update in the middle of execution. If we disablee server for a while and worked directly with the storage, it will catch the updates up during initial startup phase, where it's scheduling unconditional update for all enabled derived datasets.

src/infra/core/src/dataset_changes_service_impl.rs
```diff
@@ -0,0 +1,219 @@
...
+            .await
+            .map_err(|e| match e {
+                GetRefError::Access(e) => GetIncrementError::Access(e),
+                GetRefError::NotFound(e) => GetIncrementError::RefNotFound(e),
+                GetRefError::Internal(e) => GetIncrementError::Internal(e),
+            })
+    }
+
+    // Scan updated dataset to detect statistics for task system result
+    async fn make_dataset_update_result_from_updated_pull(
```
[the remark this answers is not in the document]
zaychenko-sergei: That is why it is better not to wrtite comments at all. They get outdated and sit silently.

src/domain/core/src/services/dataset_changes_service.rs
```diff
@@ -0,0 +1,118 @@
...
+impl DatasetChangesService for DummyDatasetChangesService {
+    async fn get_increment_between(
+        &self,
+        _dataset_id: &DatasetID,
+        _old_head: Option<&Multihash>,
+        _new_head: &Multihash,
+    ) -> Result<DatasetIntervalIncrement, GetIncrementError> {
+        Ok(DatasetIntervalIncrement {
+            num_blocks: 1,
+            num_records: 12,
```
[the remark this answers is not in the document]
zaychenko-sergei: Mocks would require pollution of all tests with call expectations, while I just want some stub that fills the gap. Perhaps, I might want to declare numbers as constants and use those in the test.

src/domain/flow-system/src/entities/flow/flow_trigger.rs
```diff
@@ -37,6 +37,50 @@ impl FlowTrigger {
             panic!("Any trigger kind except Push unexpected")
         }
     }
+
+    /// Checks if new trigger is unique compared to the existing triggers
+    pub fn is_unique_vs(&self, existing_triggers: &[FlowTrigger]) -> bool {
+        // Try finding a similar existing trigger
+        for existing_trigger in existing_triggers {
+            match self {
```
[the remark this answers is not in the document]
zaychenko-sergei: The "Carthesian Product" like form is really compact. Btw, found I have to invert the short-circuit results.

src/infra/core/src/dataset_changes_service_impl.rs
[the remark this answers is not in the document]
zaychenko-sergei: It was moved from `ts` without update. Renamed

src/adapter/graphql/src/queries/flows/flow.rs
```diff
@@ -234,17 +255,36 @@ struct FlowDescriptionUpdateResult {
...
+                                update.old_head.as_ref(),
+                                &update.new_head,
+                            )
+                            .await
+                            .int_err()?;
+
+                        Ok(Some(Self {
+                            num_blocks: increment.num_blocks,
+                            num_records: increment.num_records,
+                        }))
```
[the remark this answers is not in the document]
zaychenko-sergei: Added. Not just flag, the updated datetime is also interesting.

src/domain/core/src/services/dataset_changes_service.rs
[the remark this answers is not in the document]
zaychenko-sergei: Okay, with some tricks, maybe not that ugly..

src/domain/flow-system/src/entities/shared/batching_rule.rs
sergiimk: Closing as it seems this problem was already solved

src/adapter/graphql/src/queries/flows/flow_start_condition.rs
sergiimk: Proposed on Slack to record the final `StartCondition` that made us decide to start the flow as part of the start event.

zaychenko-sergei: Decided to hold

src/infra/flow-system-inmem/src/services/flow/strategies/dataset_update_strategy.rs
```diff
@@ -0,0 +1,167 @@
...
+                }
+            }
+
+            Ok(())
+        } else {
+            unreachable!("Not expecting other types of flow keys than dataset");
+        }
+    }
+
+    async fn evaluate_batching_rule(
```
sergiimk, line 112: This functionality doesn't seem to be related at all to the method above, so it looks strange in the same trait. 

Why is it not a method in `FlowSystemImpl`? Do you expect it to have other implementations?

sergiimk (changes requested):

src/infra/flow-system-inmem/src/services/flow/flow_service_inmem.rs
```diff
@@ -261,29 +308,46 @@ impl FlowServiceInMemory {
...
-                    if throttling_boundary_time > trigger_time {
-                        flow.define_start_condition(
-                            self.time_source.now(),
-                            FlowStartCondition::Throttling(FlowStartConditionThrottling {
-                                interval: self.run_config.mandatory_throttling_period,
-                            }),
+                // Only merge unique triggers, ignore identical
+                if trigger.is_unique_vs(&flow.triggers) {
+                    flow.add_trigger(trigger_time, trigger).int_err()?;
+                }
```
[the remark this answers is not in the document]
zaychenko-sergei: `add_trigger_if_unique`

src/infra/flow-system-inmem/src/services/flow/strategies/flow_service_callbacks_facade.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use chrono::{DateTime, Utc};
+use kamu_core::InternalError;
+use kamu_flow_system::{BatchingRule, DatasetFlowType, FlowKey, FlowTrigger};
+use opendatafabric::DatasetID;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub(crate) trait FlowServiceCallbacksFacade: Send + Sync {
```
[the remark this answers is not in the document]
zaychenko-sergei: If you need comments, it means the interface is bad. Eliminated.

src/infra/flow-system-inmem/src/services/flow/strategies/dataset_update_strategy.rs
```diff
@@ -0,0 +1,167 @@
...
+            dependency_graph_service,
+        }
+    }
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+impl FlowTypeSupportStrategy for DatasetUpdateStrategy {
+    async fn on_flow_success(
```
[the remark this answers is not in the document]
zaychenko-sergei: Eliminated. Events won't work with private content, it cannot be added to catalogs.

src/infra/flow-system-inmem/src/services/flow/strategies/dataset_update_strategy.rs
zaychenko-sergei: Inlined with assertion

Issue #508, opened 2024-02-16
Support ODF repositories in `kamu search` command


sergiimk:
Currently `kamu search` command is implemented for LocalFS and S3 repositories in a very crude manner - by listing directories and filtering dataset aliases.

We should support ODF repositories (`odf+http(s)`) in this command as well so we could search datasets managed by `kamu-node`.

This issue related to #342 that will affect search functionality in remote repos, so we might want to use this as opportunity to replace the raw listing of directories with searching using read-only instances of `DatasetRepository`.

Issue #342, opened 2023-12-13
Unify file structure in repos and workspaces


sergiimk:
We have too many repository structures:
- Single-tenant alias-based
- Multi-tenant alias-based
- Multi-tenant DID-based

We should unify our repositories under **DID-based structure**.

**Additional Consideration:**
1) Consider where to store dataset aliases in different types of repositories (including local CLI workspace). Ideally we want `ID -> Alias, Alias -> ID` resolutions to be O(1), but at the same time:
- guarantee that aliases are unique
- our "index" is resistant to concurrent updates

2) Storing datasets by DID means it will no longer be possible to have same dataset under different aliases in CLI (or any other repo). This currently is allowed, but consider how to handle this after the change:
```
kamu pull s3://datasets.kamu.dev/odf/v2/contrib/com.cryptocompare.ohlcv.eth-usd --as a
kamu pull s3://datasets.kamu.dev/odf/v2/contrib/com.cryptocompare.ohlcv.eth-usd --as b
```

3) Consider what to do when we push datasets to [basic repositories](https://docs.kamu.dev/cli/collab/repositories/) (plain S3, Local FS), e.g. what the following operation does now and how it should work after this change:

```
kamu repo add myrepo /some/path
kamu push com.cryptocompare.ohlcv.eth-usd --to myrepo/com.cryptocompare.ohlcv.eth-usd
```

4) `SearchServiceImpl` is currently making hard assumptions about the remote repository structure and will need to be updated. See #508 for details.

5) We have some python scripts that push datasets from local (multi-tenant) workspace directly to S3. We should update them after these changes.

**Notes:**
- https://github.com/kamu-data/kamu-cli/pull/556/files#r1529599266

Issue #521, opened 2024-02-27
Smart transfer protocol is not used when pulling/pushing via repo alias


sergiimk:
### Describe the bug

Currently smart protocol is only used when pushing/pulling via explicit URL.

### Steps To Reproduce

Start a local API server.

Pull dataset from it using explicit URL:
```
kamu pull odf+http://locahost:8080/me/gps
```
Observe in logs:
```
Starting sync using Smart Transfer Protocol (Pull flow)
```

Now try:
```
kamu repo add repo odf+http://locahost:8080
kamu pull repo/me/gps
```

### Current Behavior

In logs you will see:
```
Starting sync using Simple Transfer Protocol
```

### Expected Behavior

Smart protocol is used in both cases

### kamu system info

```sh
0.160.0
```

PR #522, opened 2024-02-27
Support searching in ODF repos

sergiimk:
## Description

Closes: #508 

- Search service now supports `odf+http[s]` protocol
- As with other protocols the implementation is very crude - it's not using some established interfaces but reaches directly to GQL API
  - this is a technical debt and we will clean it up in future, once #342 will bring more understanding on how remote registries will be organized
- Search results now include more details (number of blocks, records, estimated size)
- Search command output updated to avoid boilerplate

- [x] Documentation update PR: N/A

### DIFF
--- a/src/infra/core/src/pull_service_impl.rs
+++ b/src/infra/core/src/pull_service_impl.rs
@@ -139,7 +139,7 @@ impl PullServiceImpl {
                 Some(
                     DatasetRefRemote::Alias(alias)
                     | DatasetRefRemote::Handle(DatasetHandleRemote { alias, .. }),
-                ) => DatasetAlias::new(alias.account_name.clone(), alias.dataset_name.clone()),
+                ) => DatasetAlias::new(None, alias.dataset_name.clone()),
                 Some(DatasetRefRemote::Url(url)) => DatasetAlias::new(
                     if self.dataset_repo.is_multi_tenant() {
                         match self.current_account_subject.as_ref() {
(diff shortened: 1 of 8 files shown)


src/infra/core/src/pull_service_impl.rs
```diff
@@ -139,7 +139,7 @@ impl PullServiceImpl {
                 Some(
                     DatasetRefRemote::Alias(alias)
                     | DatasetRefRemote::Handle(DatasetHandleRemote { alias, .. }),
-                ) => DatasetAlias::new(alias.account_name.clone(), alias.dataset_name.clone()),
+                ) => DatasetAlias::new(None, alias.dataset_name.clone()),
```
sergiimk, line 142: `kamu pull repo/account/dataset` was inferring destination alias as `account/dataset` which was of course confusing the single-tenant repo (panic). In case of a multi-tenant local repo I think it's still correct to default to current user (account=`None`)

src/infra/core/src/repos/dataset_factory_impl.rs
```diff
@@ -328,10 +328,20 @@ impl DatasetFactory for DatasetFactoryImpl {
...
+            "http" | "https" => {
                 let ds = Self::get_http(url, self.build_header_map(url), self.event_bus.clone());
                 Ok(Arc::new(ds))
             }
+            "odf+http" | "odf+https" => {
+                // TODO: PERF: Consider what speedups are possible in smart protocol
+                let http_url = Url::parse(url.as_str().strip_prefix("odf+").unwrap()).unwrap();
+                let ds = Self::get_http(
+                    &http_url,
+                    self.build_header_map(&http_url),
```
sergiimk, line 340: Per #521 `kamu pull repo/account/dataset` was resulting in using SimpleTP with `odf+http` protocol and the `build_header_map` here was crashing when accessing `.kamutokenstore` as it expected raw http/https URL. So I'm stripping off `odf+` schema here.

Issue #390, opened 2023-12-27
Redesign metadata traversal


sergiimk:
In many places to perform some operation we need to pull several things from metadata like:
- previous offset / checkpoint / watermark
- dataset vocabulary
- current active polling/push source

To minimize scanning these things are currently implemented by writing [complex iteration loops](https://github.com/kamu-data/kamu-cli/blob/master/src/infra/core/src/repos/metadata_chain_impl.rs#L198) that match events of interest in a single pass.

This is incredibly tedious, fragile (easy to result in a full-scan), and prevents any kind of traversal optimizations.

This ticket is to design a better system for pulling data out of the metadata chain that:
- hides the iteration logic
- minimizes the traversal depth
- enables under-the-hood optimizations (such as skip-lists) without having to modify all places that pull info from metadata

## Implementation
- [x] https://github.com/kamu-data/kamu-cli/pull/523
- [x] https://github.com/kamu-data/kamu-cli/pull/543

PR #523, opened 2024-02-27
Introduce MetadataChainVisitor's

s373r:
## Description

The purpose of this PR is to propose new entities by which the unification of the chain metadata traversal code will be performed.

All validators of the append block except `ValidateEventLogicalStructureVisitor` are implemented (it will be in the next PR).

Related issue: https://github.com/kamu-data/kamu-cli/issues/390

- [x] Unit-tests added


src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -720,14 +565,18 @@ where
...
-                .await?;
-            self.validate_append_watermark_is_monotonic(&block).await?;
-            self.validate_append_offsets_are_sequential(&block).await?;
+            let validators: BoxedVisitors = vec![
+                Box::<ValidateSeedBlockOrderVisitor>::default(),
+                Box::<ValidatePrevBlockExistsVisitor>::default(),
+                Box::<ValidateSequenceNumbersIntegrityVisitor>::default(),
+                Box::<ValidateSystemTimeIsMonotonicVisitor>::default(),
+                Box::<ValidateWatermarkIsMonotonicVisitor>::default(),
+                Box::<ValidateOffsetsAreSequentialVisitor>::default(),
```
sergiimk, line 574: Do you think it's better to box them or perhaps we could pass them all as `&[&mut dyn Visitor]`?

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -720,14 +565,18 @@ where
...
+                // TODO add ValidateEventLogicalStructureVisitor
+            ];
+
+            self.accept(&block, validators).await?;
```
sergiimk, line 578: We probably need two versions of `accept`:
- one that iterates over existing blocks (public method `accept(visitors)`)
- one that first feeds visitors one extra block that is about to be part of the chain (**private** method `validate(block, visitors)`

src/infra/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,73 @@
...
+        block_with_optional_hash: MetadataBlockWithOptionalHashRef,
+    ) -> Result<Decision, AppendError>;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub type BoxedVisitors = Vec<Box<dyn MetadataChainVisitor>>;
+
+#[async_trait::async_trait]
+pub trait MetadataChainVisitorHost {
```
sergiimk, line 65: What is the goal of this trait? Should it be just a method in existing `MetadataChain` trait?

src/infra/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,73 @@
...
+pub enum Decision {
+    Stop,
+    Next,
+    NextWithHash(Multihash),
+    NextOfType(MetadataBlockTypeFlags),
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub type MetadataBlockWithOptionalHashRef<'a> = (&'a MetadataBlock, Option<&'a Multihash>);
```
sergiimk, line 51: That's unfortunate... Should we calculate the hash of new block before validation just to get rid of this `Option`?

src/infra/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,73 @@
...
+
+pub trait MetadataChainVisitor: Sync + Send {
+    fn visit(
+        &mut self,
+        block_with_optional_hash: MetadataBlockWithOptionalHashRef,
+    ) -> Result<Decision, AppendError>;
```
sergiimk, line 57: Returning `AppendError` only makes sense for visitors used for validation.
We should implement at least one visitor from the "outside" that collects some data from the chain to understand what to do with the return type.

src/infra/core/src/repos/mod.rs
```diff
@@ -7,68 +7,52 @@
...
+mod dataset_impl;
+mod dataset_repository_helpers;
+mod dataset_repository_local_fs;
+mod dataset_repository_s3;
+mod metadata_block_repository_caching_inmem;
+mod metadata_block_repository_helpers;
+mod metadata_block_repository_impl;
+mod metadata_chain_impl;
+mod metadata_chain_visitor;
+mod metadata_chain_visitor_impl;
```
sergiimk, line 20: Let's call this module `metadata_chain_validators`, as we will not be putting any other visitors in there.

src/infra/core/src/repos/metadata_chain_visitor_impl.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Default)]
+pub struct ValidatePrevBlockExistsVisitor {
+    next_prev_block_hash: Option<Multihash>,
+}
+
+impl MetadataChainVisitor for ValidatePrevBlockExistsVisitor {
```
sergiimk, line 59: Perhaps this simply shouldn't be a visitor. This is something we can very easily check in `append` with a lot fewer lines of code.

src/infra/core/src/repos/metadata_chain_visitor_impl.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        Ok(Decision::Stop)
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Default)]
+pub struct ValidateSystemTimeIsMonotonicVisitor {
+    next_system_time: Option<DateTime<Utc>>,
```
sergiimk, line 140: Looks like many visitors have to start with some empty state and then initialize it with the first block...

I wonder if instead of:
```rust
self.accept_with_extra_block(block, [
  VisitorA::default(),
  VisitorB::default(),
])
```
it would be more ergonomic to use them as:
```rust
self.accept([
  VisitorA::new(&new_block),
  VisitorB::new(&new_block),
])
```
allowing visitors to initialize their state from first block and eliminate `Option`s.

sergiimk: This would eliminate the need for a special `accept` method that passes an extra block.

src/infra/core/src/repos/metadata_chain_visitor.rs
sergiimk: One in `writer.rs` in `with_metadata_state_scanned` could be a good candidate.

src/infra/core/src/repos/metadata_chain_visitor_impl.rs
```diff
@@ -0,0 +1,340 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+// Helpers
+///////////////////////////////////////////////////////////////////////////////
+
+pub struct MetadataChainVisitorBatchProcessor {}
+
+impl MetadataChainVisitorBatchProcessor {
+    pub fn get_next_decisions(
+        visitors: BoxedVisitors,
```
sergiimk, line 295: A bit strange that visitor vec is passed by value in and out

sergiimk: Oh, I see now - you are removing visitors that said stop from the vec...

I think the problem is more complex - I think visitors that returned `Decision::NextOfType(X)` should not be called for blocks that are not of type `X` - thus you'll need to store all visitors and decisions until all decisions become `Stop`.

sergiimk (changes requested):
This is a very good start! :+1:

src/infra/core/src/repos/mod.rs
s373r: Done

src/infra/core/src/repos/metadata_chain_visitor.rs
s373r: Done

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: Yes, I have endeavored to do so -- my past attempts have been unsuccessful. 

But now, I've been able to reduce the number of allocations, at least at this point.

src/infra/core/src/repos/metadata_chain_visitor.rs
s373r: I added this purely for interface separation and convenience.

If you think it's unnecessary, I can remove it, as it's not a critical thing for me. 
Should I remove it?

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: **Q:** could you please clarify -- can we dispense with the two methods of the Visitor itself? 

(same as https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1506383049)

src/infra/core/src/repos/metadata_chain_visitor_impl.rs
s373r: I think we can leave it in -- this logic is just for Validator.
1) it requests a block from the past
2) it validates :)

src/infra/core/src/repos/metadata_chain_visitor.rs
s373r: It will be done

src/infra/core/src/repos/metadata_chain_visitor_impl.rs
s373r: WIP

At this point, allocations still remain, I'm in the process of fixing them.

I think I will be able to remove some allocations.

s373r: All extra allocation have removed -- stack array has been used

src/infra/core/src/repos/metadata_chain_visitor.rs
s373r: Yes, while doing some thinking about implementation beyond Validators, realized that you are right and there is no need for this trait

Removed -- Done

src/infra/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use bitflags::bitflags;
+use kamu_core::{AppendError, HashedMetadataBlockRef};
+use opendatafabric::{MetadataBlock, MetadataEvent, Multihash};
+
+///////////////////////////////////////////////////////////////////////////////
+
+bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct MetadataBlockTypeFlags: u32 {
```
zaychenko-sergei, line 20: Note: unlike the block below, this one will not signal if it's no longer complete. Some comments that suggest updating the masks whenever we add new event types could be nice.

Or better, can we generate these bit masks from ODF makefiles side ?

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: It is likely that all allocations related to Visitors have been eliminated

src/infra/core/src/repos/metadata_chain_visitor.rs
s373r: @zaychenko-sergei , in general, with `impl From<&MetadataBlock> for MetadataBlockTypeFlags`, we definitely won't miss changes, but I agree with you -- it's better to move the Flags to the ODF land.

Do you mind if this is done in a follow-up PR?

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -720,14 +631,46 @@ where
...
         if opts.validation == AppendValidation::Full {
-            self.validate_append_prev_block_exists(&block).await?;
-            self.validate_append_sequence_numbers_integrity(&block)
-                .await?;
-            self.validate_append_seed_block_order(&block)?;
-            self.validate_append_system_time_is_monotonic(&block)
+            let hash = if let Some(b) = opts.precomputed_hash {
+                Cow::Borrowed(b)
+            } else {
+                let hash = self.meta_block_repo.get_block_hash(&block).int_err()?;
```
sergiimk, line 637: In my previous comments I gave you two ideas:
- **calculate the hash of the block being appended** to be able to express all validation in `visit((hash, block): ...)` functions of visitors without making `hash` on `Option` which is annoying to use (especially for non-validating visitors where it would never be `None`)
- or make **append validating visitors accept the first block in constructors**

You have implemented BOTH pre-hashing and accepting new block in `::new()`, which I think is redundant. Constructors of validating visitors already know that this initial block is special and don't need to require a hash for it.

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -720,14 +631,46 @@ where
...
+                (
+                    Decision::Stop,
+                    &mut ValidateSystemTimeIsMonotonicVisitor::new(hashed_block),
+                ),
+                (
+                    Decision::Stop,
+                    &mut ValidateWatermarkIsMonotonicVisitor::new(hashed_block),
+                ),
+                (
+                    Decision::Stop,
```
sergiimk, line 665: It's good that we got rid of the allocation per visitors, but I don't see why we should ruin ergonomics of the API by requiring users to pass `(Decision, &mut Visitor)` tuples.

One allocation for `Vec<Decision>::with_capacity(visitors.len())` is tolerable. If we wanted to avoid allocations completely (which is IMO overkill) - we could use a stack buffer and overflow to `Vec`.

The point is - let's keep this matter internal to `accept()` and not burden the API users.

sergiimk (changes requested):
I'd recommend:
- replace multiple `visit` methods with one + fallible `Validator::new()?`
- get rid of pre-hashing - as it seems unnecessary with `Validator::new(new_block)?` approach
- Hide `[Decision]` buffer into `accept()` method to simplify API
- Definitely implement at least one non-validating visitor

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,338 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+pub struct ValidateSeedBlockOrderVisitor<'a> {
+    initial_block: &'a MetadataBlock,
+}
+
+impl<'a> ValidateSeedBlockOrderVisitor<'a> {
+    pub fn new((_, block): HashedMetadataBlockRef<'a>) -> Self {
+        Self {
+            initial_block: block,
```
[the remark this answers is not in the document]
s373r:
```rust
...
pub fn new(block: &MetadataBlock) -> Result<Self, AppendError> {
...
```

That's exactly the solution I was thinking about this weekend -- it would really simplify it. 

Will be done

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,338 @@
...
+        (hash, _): HashedMetadataBlockRef,
+    ) -> Result<Decision, Self::VisitError> {
+        let Some(initial_prev_block_hash) = &self.initial_block.prev_block_hash else {
+            unreachable!()
+        };
+
+        if initial_prev_block_hash != hash {
+            return Err(
+                AppendValidationError::PrevBlockNotFound(BlockNotFoundError {
+                    hash: self.initial_hash.clone(),
```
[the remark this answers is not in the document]
s373r: Yes, you're right -- thank you for pointing that out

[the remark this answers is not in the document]
s373r: Yes, you're right -- thank you for pointing that out

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,338 @@
...
+pub struct ValidatePrevBlockExistsVisitor<'a> {
+    initial_hash: &'a Multihash,
+    initial_block: &'a MetadataBlock,
+}
+
+impl<'a> ValidatePrevBlockExistsVisitor<'a> {
+    pub fn new((hash, block): HashedMetadataBlockRef<'a>) -> Self {
+        Self {
+            initial_hash: hash,
+            initial_block: block,
```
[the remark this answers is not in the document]
s373r: You're right, there's no need for that -- we can store more specific data, especially by refs

Done

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+impl<'a> MetadataChainVisitor for ValidateSequenceNumbersIntegrityVisitor<'a> {
+    type VisitError = AppendError;
+
+    fn visit(&mut self) -> Result<Decision, Self::VisitError> {
+        let block = self.initial_block;
+
+        match &block.prev_block_hash {
+            Some(hash) => Ok(Decision::NextWithHash(hash.clone())),
```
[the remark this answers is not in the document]
s373r: Fixed

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: Yes, I introduced the hash calculation because I mistakenly thought in one of the validators that we needed to know the hash of the added block

For this reason, it turned out that the two approaches were combined.

Hashing will be removed

src/infra/core/src/repos/metadata_chain_visitor_impl.rs
s373r: API has been updated:
```rust
pub trait MetadataChainVisitor: Sync + Send {
    type VisitError: Error;

    fn visit(
        &mut self,
        hashed_block_ref: HashedMetadataBlockRef,
    ) -> Result<Decision, Self::VisitError>;
}
```

src/infra/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,83 @@
...
+    NextWithHash(Multihash),
+    NextOfType(MetadataBlockTypeFlags),
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub trait MetadataChainVisitor: Sync + Send {
+    type VisitError: Error;
+
+    fn visit(&mut self) -> Result<Decision, Self::VisitError>;
```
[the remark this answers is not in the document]
s373r: I agree, API has been simplified

src/infra/core/src/repos/metadata_chain_validators.rs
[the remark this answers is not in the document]
s373r: Done

The only difference is that the ctor also returns a `Decision`

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,338 @@
...
+        }
+    }
+}
+
+impl<'a> MetadataChainVisitor for ValidatePrevBlockExistsVisitor<'a> {
+    type VisitError = AppendError;
+
+    fn visit(&mut self) -> Result<Decision, Self::VisitError> {
+        match &self.initial_block.prev_block_hash {
+            Some(hash) => Ok(Decision::NextWithHash(hash.clone())),
```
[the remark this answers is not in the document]
s373r: `Decision::Next` was returned -- Done

src/infra/core/src/repos/metadata_chain_validators.rs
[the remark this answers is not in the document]
s373r: Done

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: Done

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: `MetadataChainImpl::accept()` interface has been simplified -- Done

src/infra/ingest-datafusion/src/visitor.rs
```diff
@@ -0,0 +1,254 @@
...
+            MetadataEvent::ExecuteTransform(_)
+            | MetadataEvent::SetTransform(_)
+            | MetadataEvent::SetAttachments(_)
+            | MetadataEvent::SetInfo(_)
+            | MetadataEvent::SetLicense(_) => {
+                unreachable!()
+            }
+        }
+
+        Ok(Decision::NextOfType(self.next_block_flags))
```
sergiimk, line 250: Will the `NextOfType` with no bits set be automatically treated as `Stop`?

src/infra/ingest-datafusion/src/visitor.rs
```diff
@@ -0,0 +1,254 @@
...
+        })
+    }
+
+    fn handle_set_data_schema(&mut self, e: &SetDataSchema) -> Result<(), ScanMetadataError> {
+        if self.maybe_schema.is_none() {
+            self.maybe_schema = Some(e.schema_as_arrow().int_err()?);
+        }
+
+        Ok(())
+    }
```
sergiimk, line 118: Since once we visit `SetDataSchema` you are un-setting the `Flag::SET_DATA_SCHEMA` - we should never see the older schemas again during iteration, so the `if self.maybe_schema.is_none()` check seems to be redundant.

Same applies to many other similar checks here.

sergiimk (changes requested):
Left a few suggestions, but overall I'm pretty happy with how visitors look now. I think they became very granular and readable.

I propose we get this baby merged and convert other places (from RFC) to visitors in separate PRs.

Cargo.toml
```diff
@@ -127,7 +127,6 @@ debug = 2
 [profile.release.package]
 opendatafabric = { debug = 1 }
 kamu = { debug = 1 }
-kamu-cli = { debug = 1 }
```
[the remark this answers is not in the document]
s373r: this is an automatic change that occurs when `cargo` dependencies are installed/uninstalled

--

I usually manually removed this change when it was suggested, but decided to leave it in this time. 

@sergiimk ,
Q: do you think this should be reverted?

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -146,6 +152,58 @@ pub trait MetadataChainExt: MetadataChain {
     fn iter_blocks_ref<'a>(&'a self, head: &'a BlockRef) -> DynMetadataStream<'a> {
         self.iter_blocks_interval_ref(head, None)
     }
+
+    async fn accept_interval<'a, E>(
```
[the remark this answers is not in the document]
s373r: Fair point. I mirrored the block iteration API, so there would be parity -- but, yes, I'll revisit it

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -15,10 +16,15 @@ use opendatafabric::{MetadataBlock, MetadataBlockTyped, MetadataEvent, Multihash
 use thiserror::Error;
 
 use super::metadata_stream::DynMetadataStream;
+use crate::entities::metadata_chain_visitor_facade::MetadataChainVisitorFacade;
 use crate::repos::{SetRefError as SetRefErrorRepo, *};
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+pub type VisitorsMutRef<'a, 'b, E> = &'a mut [&'b mut dyn MetadataChainVisitor<VisitError = E>];
```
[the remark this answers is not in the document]
s373r: We need the `mut` reference the slice, for the sake of the next fragment:
```rust
self.decisions.iter_mut().zip(self.visitors.iter_mut())
                              ^^^^^^^^^^^^^^^^^^^^^^^^
```

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,312 @@
...
+            None if block.sequence_number != 0 => Err(AppendValidationError::SequenceIntegrity(
+                SequenceIntegrityError {
+                    prev_block_hash: None,
+                    prev_block_sequence_number: None,
+                    next_block_sequence_number: block.sequence_number,
+                },
+            )
+            .into()),
+            None => Ok(Decision::Stop),
+        }?;
```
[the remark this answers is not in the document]
s373r: In the highlighted fragment, we have three code paths (next, `stop`, error), while in the proposed one we have only two (next, error). 

So, unfortunately, it's not going to work

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,312 @@
...
+        ))
+    }
+}
+
+impl MetadataChainVisitor for ValidateWatermarkIsMonotonicVisitor {
+    type VisitError = AppendError;
+
+    fn visit(&mut self, (_, block): HashedMetadataBlockRef) -> Result<Decision, Self::VisitError> {
+        let Some(data_steam_event) = block.event.as_data_stream_event() else {
+            return Ok(Decision::NextOfType(MetadataBlockTypeFlags::DATA_BLOCK));
```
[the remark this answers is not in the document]
s373r: Thought about that too, we can be a little more blunt in this case. 

Fixed

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
+
+impl<'a> MetadataChainVisitor for ValidateOffsetsAreSequentialVisitor<'a> {
+    type VisitError = AppendError;
+
+    fn visit(&mut self, (_, block): HashedMetadataBlockRef) -> Result<Decision, Self::VisitError> {
+        // Only check AddData and ExecuteTransform.
+        // SetWatermark is also considered a data stream event but does not carry the
+        // offsets.
```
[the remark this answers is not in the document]
s373r: Removed -- Done

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,312 @@
...
+        let Some(data_block) = block.as_data_stream_block() else {
+            return Ok(Decision::NextOfType(MetadataBlockTypeFlags::DATA_BLOCK));
```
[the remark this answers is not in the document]
s373r: Thought about that too, we can be a little more blunt in this case [2]

Fixed

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,312 @@
...
+                        new_data.offset_interval.start,
+                    ),
+                )
+                .into());
+            }
+
+            if new_data.offset_interval.end < new_data.offset_interval.start {
+                invalid_event!(self.initial_block_event.clone(), "Invalid offset interval",);
+            }
+        }
```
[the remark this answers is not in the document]
s373r: Indeed, you're right.

Fixed

src/infra/ingest-datafusion/src/visitor.rs
s373r: btw, no -- fixed it to an explicit stop. thank you for this find!

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        Ok(Decision::Stop)
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub struct ValidateSequenceNumbersIntegrityVisitor {
+    initial_sequence_number: u64,
```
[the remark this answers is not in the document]
s373r: Yes, unambiguously better -- indeed, `initial` is more likely to be perceived as the first, but in our case (we're going backwards), it creates confusion

Fixed

src/infra/ingest-datafusion/src/visitor.rs
s373r: makes sense

updated methods:
- `handle_set_data_schema()`
- `handle_set_vocab()`
- `handle_set_polling_source()`

remain the same:
- `handle_add_push_source()`

it seems to be more complex logic, so I decided to leave it as is

src/domain/core/src/entities/metadata_chain.rs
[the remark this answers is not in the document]
s373r: `accept_*`  API has been revisited

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -146,6 +152,58 @@ pub trait MetadataChainExt: MetadataChain {
...
+        ignore_missing_tail: bool,
+    ) -> Result<(), E>
+    where
+        E: Error + From<IterBlocksError>,
+    {
+        let mut decisions = vec![Decision::Next; visitors.len()];
+        let visitor_facade = MetadataChainVisitorFacade::new(&mut decisions, visitors);
+        let block_stream = self.iter_blocks_interval(head, tail, ignore_missing_tail);
+
+        accept_metadata_stream(visitor_facade, block_stream).await
```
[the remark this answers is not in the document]
s373r: >  I think accept should be implemented as a simple loop calling get_block.

Done

> I also don't see the point of MetadataChainVisitorFacade as a type either - seems like it could be just inlined here.

Inlined -- Done

src/domain/core/src/entities/metadata_chain_visitor_facade.rs
```diff
@@ -0,0 +1,80 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::error::Error;
+
+use crate::{Decision, HashedMetadataBlockRef, MetadataBlockTypeFlags, VisitorsMutRef};
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub type DecisionsMutRef<'a> = &'a mut [Decision];
```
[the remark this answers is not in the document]
s373r: Unactual -- the file was removed

src/infra/core/src/repos/metadata_chain_validators.rs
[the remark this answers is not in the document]
s373r: I was wrong -- it fits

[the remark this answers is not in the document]
s373r: Done

sergiimk (approved):
Added a few tweaks to satisfy my OCD. This looks ready to merge to me, but please wait for @zaychenko-sergei to sign-off.

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,79 @@
...
+            MetadataEvent::DisablePushSource(_) => Self::DISABLE_PUSH_SOURCE,
+            MetadataEvent::DisablePollingSource(_) => Self::DISABLE_POLLING_SOURCE,
+        }
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Decision {
```
zaychenko-sergei, line 64: Maybe, `MetadataVisitorDecision`? A very general name currently.

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,79 @@
...
+    Stop,
+    Next,
+    NextWithHash(Multihash),
+    NextOfType(MetadataBlockTypeFlags),
```
zaychenko-sergei, line 68: Plz write some comments explaining each decision type.

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub trait MetadataChainVisitor: Sync + Send {
```
zaychenko-sergei, line 73: I wonder if we would need an `async` version in future...

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -492,82 +413,24 @@ where
...
-                    (Some(prev_wm), Some(new_wm)) if prev_wm > new_wm => {
-                        return Err(AppendValidationError::WatermarkIsNotMonotonic.into())
-                    }
-                    _ => (),
-                }
-            }
-        }
-        Ok(())
-    }
+        let have_already_stopped = decisions.iter().all(|decision| *decision == Decision::Stop);
```
zaychenko-sergei, line 422: Minor: maybe holding a counter of active visitors among with `decisions` slice could reduce this comparison to 1 operation, on the other hand we would have to detect changes in `decisions` within the traversal code.

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -146,6 +147,97 @@ pub trait MetadataChainExt: MetadataChain {
...
+                    }
+                    Decision::NextOfType(type_flags) if type_flags.is_empty() => {
+                        stopped_visitors += 1;
+                    }
+                    Decision::Next => {
+                        *decision = visitor.visit(hashed_block_ref)?;
+                    }
+                    Decision::NextWithHash(requested_hash) => {
+                        if hash == *requested_hash {
+                            *decision = visitor.visit(hashed_block_ref)?;
```
zaychenko-sergei, line 222: So, we are quickly finishing if any of the visitors return an error. I wonder if there would be a need to collect multiple `Result` of visitors. Something like displaying multiple validation errors in 1 run.

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -146,6 +147,97 @@ pub trait MetadataChainExt: MetadataChain {
     fn iter_blocks_ref<'a>(&'a self, head: &'a BlockRef) -> DynMetadataStream<'a> {
         self.iter_blocks_interval_ref(head, None)
     }
+
+    async fn accept<E>(
```
zaychenko-sergei, line 151: Plz add comments for new methods. It seems all methods in this extension are commented

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -146,6 +147,97 @@ pub trait MetadataChainExt: MetadataChain {
...
+    {
+        let head_hash = self.resolve_ref(head).await.map_err(Into::into)?;
+
+        self.accept_by_hash(visitors, &head_hash).await
+    }
+
+    async fn accept_by_hash_with_decisions<E>(
+        &self,
+        decisions: &mut [Decision],
+        visitors: &mut [&mut dyn MetadataChainVisitor<Error = E>],
```
zaychenko-sergei, line 191: Since these two are so interconnected logically, shouldn't it be wrapped in a struct, something like `VisitingState` ?

zaychenko-sergei: If you decide to add a counter of stopped visitors, it would logically belong to this struct very naturally.

zaychenko-sergei (changes requested):
Good work, see some improvement suggestions and thoughts for future.

s373r:
For the record, I've decided to put the fixes for the comments in a follow-up PR. I'll provide cross-linking so that nothing gets lost.

СС @sergiimk @zaychenko-sergei

src/domain/core/src/repos/metadata_chain_visitor.rs
s373r: Moved to https://github.com/kamu-data/kamu-cli/pull/543/files#r1517779630

src/domain/core/src/repos/metadata_chain_visitor.rs
s373r: Moved to https://github.com/kamu-data/kamu-cli/pull/543/files#r1517819305

src/domain/core/src/repos/metadata_chain_visitor.rs
s373r: Moved to https://github.com/kamu-data/kamu-cli/pull/543#discussion_r1517779630

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: Moved to https://github.com/kamu-data/kamu-cli/pull/543#discussion_r1517879457

src/domain/core/src/entities/metadata_chain.rs
s373r: Moved to https://github.com/kamu-data/kamu-cli/pull/543#discussion_r1517882625

src/domain/core/src/entities/metadata_chain.rs
s373r: Moved to https://github.com/kamu-data/kamu-cli/pull/543#discussion_r1517884935

src/infra/core/src/repos/metadata_chain_visitor.rs
s373r: PR: https://github.com/open-data-fabric/open-data-fabric/pull/85

Issue #525, opened 2024-02-28
`kamu login` in user scope should not require a workspace


sergiimk:
### Describe the bug

Attempt to login in a user scope currently will fail outside of kamu workspace

### Steps To Reproduce

`kamu login --user node.demo.kamu.dev --access-token token`

### Current Behavior

`Error: Directory is not a kamu workspace`

### Expected Behavior

Login should be successful as the config file is placed under `~/.kamutokenstore` which does not require a workspace.

### kamu system info

```sh
0.161.0
```

Issue #524, opened 2024-02-27
`kamu verify` command crashes under verbose logging


sergiimk:
### Describe the bug

Bug with `unwrap()` causes verify command to crash

### Steps To Reproduce

1. Initialize `covid` example
2. Run `kamu verify covid19.british-columbia.case-details`
3. Now run `kamu -v verify covid19.british-columbia.case-details`

### Current Behavior

Crash:
```
thread 'main' panicked at src/app/cli/src/commands/verify_command.rs:152:33:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/1a648b397dedc98ada3dd3360f6d661ec2436c56/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/1a648b397dedc98ada3dd3360f6d661ec2436c56/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/1a648b397dedc98ada3dd3360f6d661ec2436c56/library/core/src/panicking.rs:144:5
   3: core::option::unwrap_failed
             at /rustc/1a648b397dedc98ada3dd3360f6d661ec2436c56/library/core/src/option.rs:1983:5
   4: core::option::Option<T>::unwrap
             at /rustc/1a648b397dedc98ada3dd3360f6d661ec2436c56/library/core/src/option.rs:932:21
   5: kamu_cli::commands::verify_command::VerifyCommand::verify::{{closure}}
             at /home/smikhtoniuk/Work/projects/kamu/kamu-cli/src/app/cli/src/commands/verify_command.rs:152:24
   6: <kamu_cli::commands::verify_command::VerifyCommand as kamu_cli::commands::Command>::run::{{closure}}
             at /home/smikhtoniuk/Work/projects/kamu/kamu-cli/src/app/cli/src/commands/verify_command.rs:255:40
```

### Expected Behavior

No crash

### kamu system info

```sh
0.161.0
```

PR #526, opened 2024-02-28
Select smart protocol when pushing/pullin via repo alias

sergiimk:
## Description

Closes: #521
Closes: #525 
Closes: #524 

- Sync service will first resolve all references to either local ref or a remote URL
- Remote references like `repo/account/dataset` will become URLs, so that later logic will correctly select the transfer protocol based on repo's scheme
- Also fixed a few minor issues along the way

- [ ] [CHANGELOG.md](./CHANGELOG.md) updated
- [ ] API changes are backwards-compatible
- [ ] Workspace layout changes include a migration
- [ ] Documentation update PR: <link or N/A>
- [ ] Dataset pipelines update scheduled if needed


src/infra/core/src/testing/minio_server.rs
```diff
@@ -113,7 +113,7 @@ impl LocalS3Server {
         let minio = MinioServer::new(tmp_dir.path(), access_key, secret_key).await;
 
         let url = Url::parse(&format!(
-            "s3+http://{}:{}/{}",
+            "s3+http://{}:{}/{}/",
```
sergiimk, line 116: This is needed only to make tests more succinct. As sync service normalizes URLs to always have the trailing slash adding it here allows direct comparisons between these URLs and those is all kinds of errors.

zaychenko-sergei (approved):
`SyncRef` is a nice idea

Issue #469, opened 2024-01-27
Make dataset names and aliases comparisons case-insensitive


sergiimk:
## Problem
- ODF restricts dataset name format to the grammar of domain names for compatibility with mechanisms like HTTP gateways and IPNS when dataset name appears in a subdomain. Subdomains are matched as case-insensitive.
- We likely don't want to allow datasets `fooBar` and `FooBar` and users `alice` and `alIce` to exist in our system due to risk of impersonation
- Finally the newly-introduced `DatasetRefPattern` also matches aliases as case-insensitive

## Proposal
- Spell out how case (in)sensitivity works in ODF spec
- Equality checks of `DatasetName` and `AccountName` should be made case-insensitive
- `DatasetRepository` implementations should be updated to case-insensitive comparisons when detecting name collisions
- Along the way we should revisit how equality is implemented for composite types like `DatasetRef`, e.g. cases like `DatasetRef::Alias("alice/foo") == DatasetRef::Handle("did:odf:abc", "alice/fool")`

khh7:
Hi! I'm looking for a good first issue to work on and thought this would be an interesting one. Can I work on this if possible?

sergiimk:
Hi @khh7, thanks for considering to contribute!

This issue should be a good one to start with and we'll be happy to guide you through the process.

You can start with `DEVELOPER.md` to set up your development environment and get the tests running locally.

Once you've done that and ready to start on the issue - please assign it to yourself, so that no one else picks it up simultaneously.

Feel free to reach out on Discord if you need help.

PR #537, opened 2024-03-05
Make dataset alias comparisons case-insensitive

rmn-boiko:
## Description

Closes: [Make dataset names and aliases comparisons case-insensitive](https://github.com/kamu-data/kamu-cli/issues/469)


src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -305,13 +317,19 @@ newtype_str!(
...
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
 newtype_str!(RepoName, Grammar::match_repo_name, RepoNameSerdeVisitor);
 
 ////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Clone, Eq, PartialOrd, Ord)]
```
zaychenko-sergei, line 332: If we are customizing `PartialEq`, shouldn't we also customize `PartialOrd` ? I assume it might be useful for correct ignore-case sorting.

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -437,6 +476,29 @@ impl DatasetAliasRemote {
...
+impl PartialEq<Self> for DatasetAliasRemote {
+    fn eq(&self, other: &Self) -> bool {
+        (self.repo_name == other.repo_name)
+            && ((self.account_name.is_some()
+                && other.account_name.is_some()
+                && self
+                    .account_name
+                    .as_ref()
+                    .unwrap()
+                    .lowercase_eq(other.account_name.as_ref().unwrap()))
```
zaychenko-sergei, line 488: Try new unstable syntax of `let_chains` Rust feature to implement this. We are already using it in some of our crates:

https://github.com/rust-lang/rfcs/pull/2497

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -257,7 +262,14 @@ impl DatasetAliasPattern {
     }
 
     pub fn matches(&self, dataset_handle: &DatasetHandle) -> bool {
-        (self.account_name.is_none() || self.account_name == dataset_handle.alias.account_name)
+        ((self.account_name.is_none() && dataset_handle.alias.account_name.is_none())
+            || (self.account_name.is_some()
+                && dataset_handle.alias.account_name.is_some()
+                && self
```
zaychenko-sergei, line 268: Same here, try let_chains

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -388,7 +427,7 @@ impl_serde!(DatasetAlias, DatasetAliasSerdeVisitor);
 
 ////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Clone, Eq, PartialOrd, Ord)]
```
zaychenko-sergei, line 430: Same here, I wonder if `PartialOrd` is necessary to be customized too.

zaychenko-sergei (changes requested):
1) I can see that you've touched the casing of accounts as well. On today account names are case-sensitive in all implementations of `AuthenticationProvider` trait. These would need to take into account that "Wasya" and "wasya" are the same accounts. Before you start digging into this direction, plz make a quick check if GitHub accounts are case-sensitive or not. It seems they are case-insensitive, but double check.

2) I think what is done in the repositories implementation is not enough. You need to make methods like `resolve_dataset_ref` to be case-insensitive as well. Note that in multi-tenant repos account names are used in datasets folder names, and in single-tenant local repo the dataset name is used as name of folder. Basically, the entire infrastructure of dataset repositories should be lowercasing all objects and references. This, btw, would fix some potential bugs on Windows platform, where file names are case-insensitive.

3) I'd like to see more tests. Repository changes for sure. Might be interesting to ensure our HTTP and GraphQL APIs are properly aligned with casing - in HTTP case you need to be careful with URL matching middlewares.

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -358,6 +376,27 @@ impl DatasetAlias {
...
+    fn eq(&self, other: &Self) -> bool {
+        ((self.account_name.is_none() && other.account_name.is_none())
+            || (self.account_name.is_some()
+                && other.account_name.is_some()
+                && self
+                    .account_name
+                    .as_ref()
+                    .unwrap()
+                    .lowercase_eq(other.account_name.as_ref().unwrap())))
+            && self.dataset_name.lowercase_eq(&other.dataset_name)
```
s373r, line 389: could you please use early returns to split conditions?  it'll make it easier to read

s373r (commented):
a minor comment added

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -221,6 +222,10 @@ impl DatasetName {
     pub fn into_local_ref(self) -> DatasetRef {
         DatasetRef::Alias(DatasetAlias::new(None, self))
     }
+
+    pub fn lowercase_eq(&self, other: &Self) -> bool {
+        self.to_lowercase() == other.to_lowercase()
```
sergiimk, line 227: 1) There is also [to_ascii_lowercase](https://doc.rust-lang.org/std/primitive.str.html#method.to_ascii_lowercase) that is probably much more performant than general unicode version.

2) Even `to_ascii_lowercase` returns a `String` i.e. it allocates per every call. I think this is too much overhead for our case, since we expect that most strings will already be lowercase. IMO it's a design error in std lib that `to_ascii_lowercase` does not return `Cow<str>`.

rmn-boiko: True, even if its return Strign it still keeps less characters and it will be more efficient

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -221,6 +222,10 @@ impl DatasetName {
     pub fn into_local_ref(self) -> DatasetRef {
         DatasetRef::Alias(DatasetAlias::new(None, self))
     }
+
+    pub fn lowercase_eq(&self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
rmn-boiko: Such as we use it only for `AccountName`, `DatasetName`, `RepoName`, `DatasetNamePattern` and in the future I believe we will use for `AccountNamePattern` and `RepoNamePattern` it is great to encapsulate this logic here

src/domain/opendatafabric/src/identity/dataset_identity.rs
rmn-boiko: Due to lastest changes this block was removed

src/domain/opendatafabric/src/identity/dataset_identity.rs
rmn-boiko: Due to lastest changes this block was removed

src/domain/opendatafabric/src/identity/dataset_identity.rs
rmn-boiko: Due to lastest changes I moved custom `PartialEq ` implementation to `newtype_istr` level

src/domain/opendatafabric/src/identity/dataset_identity.rs
rmn-boiko: Due to the lastest changes, I moved the custom PartialEq  implementation to `newtype_istr` level

src/app/cli/src/services/accounts/account_service.rs
```diff
@@ -120,17 +120,20 @@ impl AccountService {
         RelatedAccountIndication::new(target_account)
     }
 
-    fn find_account_info_impl(&self, account_name: &String) -> Option<AccountInfo> {
+    fn find_account_info_impl(&self, account_name: &AccountName) -> Option<AccountInfo> {
         // The account might be predefined in the configuration
-        self.predefined_accounts.get(account_name).cloned()
+        self.predefined_accounts.get(account_name.as_str()).cloned()
```
sergiimk, line 125: If `predefined_accounts` was keyed by `AccountName` instead of `String` would this avoid needing to do any `to_lowercase` conversions? i.e. it would "just work"?

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -104,18 +106,18 @@ macro_rules! impl_serde {
...
-macro_rules! newtype_str {
+macro_rules! newtype_istr {
     ($typ:ident, $parse:expr, $visitor:ident) => {
         #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
         pub struct $typ(Arc<str>);
 
         impl $typ {
             pub fn new_unchecked<S: AsRef<str> + ?Sized>(s: &S) -> Self {
-                Self(Arc::from(s.as_ref()))
+                Self(Arc::from(Self::into_lowercase(s.as_ref())))
```
sergiimk, line 120: There is a big difference between case-insensitive comparison and forcing all values to lowercase. If user wants to have a dataset called `myFancyDataset` - I think we should display it under this **exact** name, but it should be considered the same as `myfancydataset`.

> **Aside:** The first item on the ticket is to `Spell out how case (in)sensitivity works in ODF spec` - i.e. do a design in the form of a PR for ODF spec. This is where I hoped we would discuss such things - not during implementation.

I think the approach I propose is closer to how DNS works and how GitHub account names work.

CC @zaychenko-sergei what do you think?

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -257,7 +268,7 @@ impl DatasetAliasPattern {
     }
 
     pub fn matches(&self, dataset_handle: &DatasetHandle) -> bool {
-        (self.account_name.is_none() || self.account_name == dataset_handle.alias.account_name)
+        self.account_name == dataset_handle.alias.account_name
```
sergiimk, line 271: This seems incorrect. Shouldn't a pattern `foo` should match dataset with full alias `alice/foo`?

src/domain/opendatafabric/src/identity/dataset_identity.rs
sergiimk: My approach will of course would complicate `Eq, Ord, Hash` implementations, making them do conversions each time. But it shouldn't be a big deal for `Hash` and we don't do that many ordering comparisons to get a performance hit.

src/adapter/graphql/tests/tests/test_accounts.rs
```diff
@@ -89,6 +89,36 @@ async fn test_account_by_name() {
...
+                r#"
+                query {{
+                    accounts {{
+                        byName (name: "{}") {{
+                            accountName
+                        }}
+                    }}
+                }}
+                "#,
+                DEFAULT_ACCOUNT_NAME.to_ascii_uppercase(),
```
sergiimk, line 105: I think we need a test that:
- creates a dataset under `KaMu/DaTa` account and name
- fetches dataset from GQL as `kAmU/dAtA`
- and then checks that `alias` in response is still `KaMu/DaTa`

sergiimk (changes requested):

src/domain/opendatafabric/src/identity/dataset_identity.rs
sergiimk: I don't see a good use case where dataset would need a "display name" that is different from its name (symbolic ID) in any way other than the character case.

Accounts in github do have separate `login` (e.g. `sergiimk`) and `name` (e.g. `Sergii M`), but you only see the latter in user info - never when addressing a user. I wouldn't want us to have an account name, account display name, and then also account full name.

In general I was hoping to avoid enforcing any additional naming restrictions on account/dataset names other than making sure that they are valid DNS substrings and behave like those.

We could force to lowercase and push back the decision to support upper case names, but then we'll need to accept that users logging in via external OAuth providers like GH will have their account names lowercased, which is not ideal.

zaychenko-sergei:
Summary on the decisive discussion:
- a note on case sensitivity of name components was [added](https://github.com/open-data-fabric/open-data-fabric/commit/c9b498a765cf64f4effdf5309fc50bab2a4bd26b) explicitly to ODF protocol
- we should preserve original casing of the name components everywhere, but make case-insensitive lookups and case-insensitive name uniqueness checks, as well as ignore-case patterns
- note that `RepoName` component is also important, this wasn't mentioned before, as we've only touched `DatasetName` and `AccountName`.

src/app/cli/src/services/accounts/account_service.rs
rmn-boiko: definitely.
changed

src/domain/opendatafabric/src/identity/dataset_identity.rs
rmn-boiko: True, return it back

src/adapter/graphql/tests/tests/test_accounts.rs
rmn-boiko: Added tests

src/infra/core/tests/tests/repos/test_dataset_repository_shared.rs
```diff
@@ -61,6 +61,62 @@ pub async fn test_create_dataset(repo: &dyn DatasetRepository, account_name: Opt
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+pub async fn test_create_and_get_case_insensetive_dataset(
+    repo: &dyn DatasetRepository,
+    account_name: Option<AccountName>,
+) {
+    let dataset_alias_to_create =
```
sergiimk, line 68: Please extend this test to create two datasets `account/foo` and `ACCOUNT/BAR` and then try accessing them as `AcCoUnT/Foo` and `AcCoUnT/Boo`.

I have a suspicion that during the creation process we don't normalize the account name correctly and the datasets will be created under two separate account directories (i.e. we will end up with both `account` and `ACCOUNT` dirs)

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -517,7 +550,7 @@ impl DatasetStorageStrategy for DatasetSingleTenantStorageStrategy {
                 }
                 let dataset_name = DatasetName::try_from(&dataset_dir_entry.file_name()).int_err()?;
                 let dataset_alias = DatasetAlias::new(None, dataset_name);
-                match self.resolve_dataset_alias(&dataset_alias).await {
+                match self.attempt_resolve_dataset_alias(&dataset_alias).await {
```
sergiimk, line 553: Why do we call `resolve_dataset_alias` here to begin with?

We already have the alias from the fs path, we know that directory exists. I think we can directly call `attempt_resolving_summary_via_path` just to get the ID of the dataset to construct the handle, or your helper method `resolve_dataset_handle`.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -492,6 +492,39 @@ impl DatasetSingleTenantStorageStrategy {
                 }
             })
     }
+
+    async fn attempt_resolve_dataset_alias(
```
sergiimk, line 496: See my other comment - I think this method can be removed.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -546,15 +579,27 @@ impl DatasetStorageStrategy for DatasetSingleTenantStorageStrategy {
...
             return Err(ResolveDatasetError::NotFound(DatasetNotFoundError {
                 dataset_ref: dataset_alias.as_local_ref(),
             }));
         }
-        let summary = self
-            .attempt_resolving_summary_via_path(&dataset_path, dataset_alias)
-            .await?;
 
-        Ok(DatasetHandle::new(summary.id, dataset_alias.clone()))
+        self.resolve_dataset_handle(&dataset_path, dataset_alias)
```
sergiimk, line 601: When we are resolving an Alias to Handle - let's make sure that handle contains an alias in its "true form".

In other words:
- If a dataset was created as `Account/Foo`
- and we call `resolve` by `account/foo`
- let's make sure the handle has alias `Account/Foo`

Also please take a look at `create_dataset` method in this file and imagine a situation where:
- we are creating a new dataset as `Account/Foo`
- repo already contains dataset `account/bar`
- to me it seems like we're missing some step to normalize the `Account` into `account` before creating `account/Foo`

sergiimk (commented):
Things mostly look good, but I have doubts in `dataset_repository_local_fs.rs`. I suggested one extra test to see if my worries are correct.

Sorry you had to deal with all the ugly complexity of the local fs repo - I really hope it will soon go away after we stop storing datasets in `account/name` directories.

src/adapter/graphql/tests/tests/test_gql_data.rs
```diff
@@ -158,12 +165,76 @@ async fn test_dataset_schema_local_fs() {
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
+// #[test_group::group(engine, datafusion)]
+#[test_log::test(tokio::test)]
+async fn test_dataset_case_insensetive_schema_local_fs() {
```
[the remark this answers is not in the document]
rmn-boiko: Yeah, right, it's just testing that we will get schema with provided case insensitive params. 
Removed

src/adapter/graphql/tests/tests/test_gql_datasets.rs
```diff
@@ -92,9 +92,56 @@ async fn dataset_by_id() {
...
+                    datasets {
+                        byOwnerAndName(accountName: "kAmU", datasetName: "<name>") {
+                            name,
+                            owner { accountName },
+                        }
+                    }
+                }
+                "#
+            )
+            .replace("<name>", "FoO"),
```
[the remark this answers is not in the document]
rmn-boiko: Removed

src/infra/core/src/repos/dataset_repository_local_fs.rs
rmn-boiko: agree, removed

src/infra/core/src/repos/dataset_repository_local_fs.rs
rmn-boiko: Added

src/infra/core/tests/tests/repos/test_dataset_repository_shared.rs
rmn-boiko: Extended

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support `ArrowJson` schema output format in QGL API and CLI commands
 - New `kamu system compact <dataset>` command that compacts dataslices for the given dataset
+### Changed
+- Case insensetive comparisons of <dataset>s, <account>s and <repo>s
```
zaychenko-sergei, line 12: Spelling: sens**i**tive

zaychenko-sergei, line 12: We normally use angle brackets <> to indicate a placeholder for an object name, of certain object type.

src/domain/opendatafabric/src/identity/dataset_identity.rs
```diff
@@ -159,6 +170,31 @@ macro_rules! newtype_str {
...
+        impl PartialEq for $typ {
+            fn eq(&self, other: &$typ) -> bool {
+                self.eq_ignore_ascii_case(other)
+            }
+        }
+
+        impl Hash for $typ {
+            fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+                Self::into_lowercase(&self.0).hash(state);
+                Self::into_lowercase(&self.0).hash(state);
```
zaychenko-sergei, line 182: Identical statement twice?

src/infra/core/src/remote_repository_registry_impl.rs
```diff
@@ -36,6 +36,23 @@ impl RemoteRepositoryRegistryImpl {
         std::fs::create_dir_all(&repos_dir)?;
         Ok(Self::new(repos_dir))
     }
+
+    pub fn get_repository_file_path(&self, repo_name: &RepoName) -> Option<PathBuf> {
+        let file_path = self.repos_dir.join(repo_name);
+
+        if !file_path.exists() {
+            // run full scan to support case-insensetive matches
+            let all_repositories_stream = self.get_all_repositories();
```
zaychenko-sergei, line 45: Minor: perhaps we could disable this part on Windows, since its file system already is case-insensitive.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -611,6 +679,16 @@ impl DatasetStorageStrategy for DatasetSingleTenantStorageStrategy {
...
+
+    fn resolve_account_dir(
+        &self,
+        _account_name: &AccountName,
+    ) -> Result<(PathBuf, AccountName), ResolveDatasetError> {
+        Ok((
+            self.root.join(DEFAULT_ACCOUNT_NAME),
+            AccountName::new_unchecked(DEFAULT_ACCOUNT_NAME),
+        ))
+    }
```
zaychenko-sergei, line 691: There are no account directories in single-tenant mode, all datasets are stored by name. I have no idea where this path would point to. I assume we should not change directory layout in this ticket. The only place where it is used in the main repository is the creation of datasets:

```
            let (_, canonical_account_name) = self
                .storage_strategy
                .resolve_account_dir(account_name)
                .int_err()?;
```

and it does no even use the path. Can't we just create a dedicated operation to resolve account name only? The full form is only used by multi-tenant strategy internally, which does not require to declare this operation as a part of storage strategy trait.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -891,6 +973,39 @@ impl DatasetStorageStrategy for DatasetMultiTenantStorageStrategy {
...
     }
+
+    fn resolve_account_dir(
+        &self,
+        account_name: &AccountName,
+    ) -> Result<(PathBuf, AccountName), ResolveDatasetError> {
+        let account_dataset_dir_path = self.root.join(account_name);
+
+        if !account_dataset_dir_path.is_dir() {
+            let read_accout_dirs = std::fs::read_dir(self.root.as_path()).int_err()?;
```
zaychenko-sergei, line 984: accou**n**t

zaychenko-sergei (changes requested):

src/infra/core/src/remote_repository_registry_impl.rs
rmn-boiko: If I'm not mistaken it can be set up to be a case sensitive. Not sure it is safe to disable it

src/domain/opendatafabric/src/identity/dataset_identity.rs
rmn-boiko: Oh, huge fail. Removed

src/infra/core/src/repos/dataset_repository_local_fs.rs
rmn-boiko: It also usend in multitenant `resolve_dataset_alias` method. I'm not sure I can somehow split it

src/infra/core/src/remote_repository_registry_impl.rs
zaychenko-sergei: This was never possible before, but yes, I see they've added some capabilities in W10 and beyond to interract with WSL

src/infra/core/src/repos/dataset_repository_local_fs.rs
zaychenko-sergei: Managed to fix

zaychenko-sergei:
Problem:

image

I would expect that filter works without case sensitivity

zaychenko-sergei:
A similar problem 
image

zaychenko-sergei (approved):

PR #543, opened 2024-03-08
MetadataChainVisitor: continuing integration & address review comments

s373r:
## Description

The purpose of this PR is to continue the integration of the `MetadataChainExt::accept_*()` API (aka `MetadataChainVisitor`).

Also, addresses left comments on https://github.com/kamu-data/kamu-cli/pull/523

Test code will be updated in the next PR

---

Related issue: https://github.com/kamu-data/kamu-cli/issues/390

- [ ] Unit-tests added


src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -61,7 +61,7 @@ impl From<&MetadataBlock> for MetadataBlockTypeFlags {
 ///////////////////////////////////////////////////////////////////////////////
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Decision {
+pub enum MetadataVisitorDecision {
```
s373r, line 64: > Maybe, `MetadataVisitorDecision`? A very general name currently. (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517463178

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -61,10 +61,28 @@ impl From<&MetadataBlock> for MetadataBlockTypeFlags {
...
     NextWithHash(Multihash),
+    /// A request for previous blocks, of specific types, using flags.
+    ///
+    /// # Examples
+    /// ```
+    /// // Request for SetVocab block
+    /// return MetadataVisitorDecision::NextOfType(MetadataBlockTypeFlags::SET_VOCAB);
+    ///
+    /// // Request for data blocks (ADD_DATA || EXECUTE_TRANSFORM)
+    /// return MetadataVisitorDecision::NextOfType(MetadataBlockTypeFlags::DATA_BLOCK);
```
s373r, line 81: > Plz write some comments explaining each decision type. (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517464734

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -148,6 +148,9 @@ pub trait MetadataChainExt: MetadataChain {
         self.iter_blocks_interval_ref(head, None)
     }
 
+    /// A method of accepting Visitors ([MetadataChainVisitor]) that allows us
+    /// to go through the metadata chain once and, if desired,
+    /// bypassing blocks of no interest.
     async fn accept<E>(
```
s373r, line 154: > Plz add comments for new methods. It seems all methods in this extension are commented (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517495081

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -73,7 +91,10 @@ pub enum Decision {
 pub trait MetadataChainVisitor: Sync + Send {
     type Error: std::error::Error;
 
-    fn visit(&mut self, hashed_block_ref: HashedMetadataBlockRef) -> Result<Decision, Self::Error>;
+    fn visit(
+        &mut self,
+        hashed_block_ref: HashedMetadataBlockRef,
+    ) -> Result<MetadataVisitorDecision, Self::Error>;
 }
```
s373r, line 98: > I wonder if we would need an `async` version in future... (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517465608

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -415,11 +415,13 @@ where
...
         &'a self,
-        decisions: &'a mut [Decision],
+        decisions: &'a mut [MetadataVisitorDecision],
         visitors: &mut [&mut dyn MetadataChainVisitor<Error = AppendError>],
         prev_append_block_hash: Option<&'a Multihash>,
     ) -> Result<(), AppendError> {
-        let have_already_stopped = decisions.iter().all(|decision| *decision == Decision::Stop);
+        let have_already_stopped = decisions
+            .iter()
+            .all(|decision| *decision == MetadataVisitorDecision::Stop);
```
s373r, line 424: > Minor: maybe holding a counter of active visitors among with `decisions` slice could reduce this comparison to 1 operation, on the other hand we would have to detect changes in `decisions` within the traversal code. (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517474271

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -208,21 +221,21 @@ pub trait MetadataChainExt: MetadataChain {
...
-                    Decision::Next => {
+                    MetadataVisitorDecision::Next => {
                         *decision = visitor.visit(hashed_block_ref)?;
                     }
-                    Decision::NextWithHash(requested_hash) => {
+                    MetadataVisitorDecision::NextWithHash(requested_hash) => {
                         if hash == *requested_hash {
                             *decision = visitor.visit(hashed_block_ref)?;
                         }
                     }
```
s373r, line 237: > So, we are quickly finishing if any of the visitors return an error. I wonder if there would be a need to collect multiple `Result` of visitors. Something like displaying multiple validation errors in 1 run. (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517482067

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -185,15 +192,21 @@ pub trait MetadataChainExt: MetadataChain {
...
 
+    /// A lower-level method of accepting Visitors ([MetadataChainVisitor])
+    /// to specify their initial decisions.
+    ///
+    /// See also [Self::accept()]
     async fn accept_by_hash_with_decisions<E>(
         &self,
-        decisions: &mut [Decision],
+        decisions: &mut [MetadataVisitorDecision],
         visitors: &mut [&mut dyn MetadataChainVisitor<Error = E>],
```
s373r, line 202: > Since these two are so interconnected logically, shouldn't it be wrapped in a struct, something like `VisitingState` ? (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517496388
>
> If you decide to add a counter of stopped visitors, it would logically belong to this struct very naturally. (c) https://github.com/kamu-data/kamu-cli/pull/523#discussion_r1517497046

src/domain/core/src/repos/metadata_chain_visitor.rs
s373r: Renamed -- Done

src/domain/core/src/repos/metadata_chain_visitor.rs
s373r: Done

src/domain/core/src/entities/metadata_chain.rs
s373r: Done

src/domain/core/src/repos/metadata_chain_visitor.rs
s373r: All in all, I think if we need something like this, we can do it

src/domain/core/src/entities/metadata_chain.rs
s373r: I don't think we need it at the moment. We may need it later -- for example, in case of API processing. It's not hard to implement, but I think we can postpone it for now.

@zaychenko-sergei , what do you think?

src/domain/core/src/entities/metadata_chain.rs
s373r: I had a similar solution earlier (maybe with a less successful name than the one you suggested), but it was rejected.

I think we can leave it as it is, since these methods contain all the logic in themselves, and it's simple.

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: In this case, the validation of the earlier decisions (of Validators) is one-time and tentative -- I don't think we're going to get a significant advantage this time

src/domain/core/src/entities/metadata_chain.rs
sergiimk: The "VisitorFacade" you had before did not perform any additional logic or held any state, thus to me it was only obscuring what is going on in accept loop and I suggested removing it.

Space for creativity is big here, but let's be driven by real use cases and keep the solution no more complex than needed to satisfy them.

s373r: > The "VisitorFacade" you had before did not perform any additional logic or held any state, thus to me it was only obscuring what is going on in accept loop and I suggested removing it.

I meant that `VisitorFacade` would accept the "initial" decisions and `Visitors` as well as the proposed variant.

but yes, of course, it had no additional state

> Space for creativity is big here, but let's be driven by real use cases and keep the solution no more complex than needed to satisfy them.

:+1:

src/infra/core/src/dataset_changes_service_impl.rs
```diff
@@ -70,72 +72,93 @@ impl DatasetChangesServiceImpl {
...
-                        execute_transform.new_watermark
-                    }
-                    _ => None,
+        let mut data_block_analysis_visitor = GenericCallbackVisitor::new(
+            DataBlockAnalysisVisitorState {
+                num_blocks: 0,
+                num_records: 0,
+                updated_watermark: None,
+                latest_watermark: None,
+            },
```
zaychenko-sergei, line 98: Maybe auto-derive `Default`?

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -323,3 +328,457 @@ impl<'a> MetadataChainVisitor for ValidateOffsetsAreSequentialVisitor<'a> {
...
+            invalid_event!(e.clone(), "Transform queries must be normalized");
+        }
+
+        if transform.queries.is_none() || transform.queries.as_ref().unwrap().is_empty() {
+            invalid_event!(e.clone(), "Transform must have at least one query");
+        }
+
+        let queries = transform.queries.as_ref().unwrap();
+
+        if queries[queries.len() - 1].alias.is_some() {
```
zaychenko-sergei, line 374: Isn't there a `.last()` method?

zaychenko-sergei (commented):
The invocation code indeed becomes declarative, much more compact.

src/infra/core/src/dataset_changes_service_impl.rs
s373r: Done

src/infra/core/src/repos/metadata_chain_validators.rs
s373r: Updated -- Done

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -166,12 +162,14 @@ pub trait MetadataChainExt: MetadataChain {
     where
         E: Error + From<IterBlocksError>,
     {
-        let mut decisions = vec![Decision::Next; visitors.len()];
+        let mut decisions = vec![MetadataVisitorDecision::Next; visitors.len()];
```
sergiimk, line 165: Hm, I thought that the recent bug where visitor was crashing on unexpected event type showed us that `Decision::Next` is not a great default.

If a visitor only wants to handle events of certain type - the `Decision::Next` means that it can still receive events of any type during the first call, thus its logic has to be more complex to handle them...

Should we extend visitor trait with `initial_decision() -> MetadataVisitorDecision` function?

src/infra/core/src/query_service_impl.rs
```diff
@@ -357,27 +357,56 @@ impl KamuSchema {
...
+            last_records_to_consider: Option<u64>,
+        }
+
+        let mut slice_collector_visitor = GenericCallbackVisitor::new(
+            DataSliceCollectorVisitorState {
+                files: Vec::new(),
+                num_records: 0,
+                last_records_to_consider,
+            },
+            |state, (_, block)| {
```
sergiimk, line 372: I like this pattern!

I think we can make it even more convenient to use:
```rust
chain.reduce(
  state, // Initial state
  Decision::Next,  // Initial decision
  |state, hash, block| { // Flatten the tuple
    ...
  }
);
```
What do you think?

src/infra/core/src/repos/metadata_chain_impl.rs
```diff
@@ -55,371 +55,15 @@ where
         }
     }
 
-    async fn validate_append_event_logical_structure(
```
sergiimk, line 58: Oh wut, did we forget to delete this in the previous PR? I though we got rid of this monster already :)

sergiimk (commented):
Looks very promising! Left a few suggestions to improve usability even further.

Was mainly looking at design - didn't pay much attention to correctness.

src/adapter/graphql/src/mutations/dataset_metadata_mut.rs
```diff
@@ -71,7 +80,7 @@ impl DatasetMetadataMut {
         let dataset = self.get_dataset(ctx).await?;
 
         let old_attachments = self
-            .get_last_block_of_type::<odf::SetAttachments>(ctx)
+            .get_last_block_of_type(ctx, <SearchSetAttachmentsVisitor>::default())
```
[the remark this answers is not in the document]
s373r: Yes, it looks a bit unfamiliar and in my opinion too, but it's the only way to engage the default error type (`InternalError`).

For example:
```rust
pub type SearchSetVocabVisitor<E = InternalError> =
    SearchSingleTypedBlockVisitor<SetVocab, E, { Flag::SET_VOCAB.bits() }>;
```

src/domain/core/src/repos/metadata_chain_visitors.rs
```diff
@@ -0,0 +1,296 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+pub type SearchSetVocabVisitor<E = InternalError> =
+    SearchSingleTypedBlockVisitor<SetVocab, E, { Flag::SET_VOCAB.bits() }>;
+
+pub type SearchSeedVisitor<E = InternalError> =
+    SearchSingleTypedBlockVisitor<Seed, E, { Flag::SEED.bits() }>;
+
+pub type SearchSetPollingSourceVisitor<E = InternalError> =
+    SearchSingleTypedBlockVisitor<SetPollingSource, E, { Flag::SET_POLLING_SOURCE.bits() }>;
```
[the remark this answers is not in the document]
s373r: > e.g. a search for polling or push source should return None if that source was subsequently disabled.

Yes, I remember our discussions about this -- once, we will integrate the search for paired events.
In this case, we will also add a generalized Visitor with this requested logic.

src/infra/core/src/repos/metadata_chain_impl.rs
s373r: During the last PR, we tried hard not to notice this elephant in our virtual room :)

src/infra/core/src/query_service_impl.rs
s373r: I think it's a cool idea and should be implemented

src/domain/core/src/entities/metadata_chain.rs
s373r: > Should we extend visitor trait with initial_decision() -> MetadataVisitorDecision function?

Definitely -- it'll make our lives easier + code will be more declarative

src/adapter/graphql/src/mutations/dataset_metadata_mut.rs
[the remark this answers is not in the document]
sergiimk: What is the idea behind `<E = InternalError>`? 

When I'm using `SearchSetVocabVisitor` - the visitor is a concrete type - it should *tell me* what errors it can return... I don't understand how this error type can be overridable.

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -323,3 +341,462 @@ impl<'a> MetadataChainVisitor for ValidateOffsetsAreSequentialVisitor<'a> {
...
+                // Queries must be normalized
+                Self::validate_transform(appended_event, &e.transform)?;
+
+                Ok(Decision::Stop)
+            }
+            State::Stopped => Ok(Decision::Stop),
+        }
+    }
+
+    fn visit(&mut self, (_, block): HashedMetadataBlockRef) -> Result<Decision, Self::Error> {
```
sergiimk, line 734: Yikes! This visitor is a no-go. My eyes are bleeding.

The entire problem of `validate_append_event_logical_structure` was that there was a ton of rules it needed to enforce, but we couldn't separate it into multiple rules because there was no mechanism to run them in a **single pass over metadata**.

In other words the complexity of `validate_append_event_logical_structure` was exactly why we needed the visitor mechanism.

sergiimk: Roughly speaking, we could take every error message in `validate_append_event_logical_structure` and turn the conditions under which it is generated into a rule with its own visitor.

For example extracting visitors like:
- `ValidateEventIsNotEmptyOrNoOp`
- `ValidateDataSchemaIsSetBeforeAddingData`
- `ValidateCheckpointSequencing`
- `ValidateWatermarkSequencing`
- `ValidateTransformInputOrderIsStable`
- ...

src/infra/ingest-datafusion/src/visitor.rs
```diff
@@ -52,35 +53,32 @@ pub struct DataWriterDataFusionMetaDataStateVisitor<'a> {
...
-                maybe_source_event: None,
-                maybe_set_vocab: None,
-
-                maybe_prev_checkpoint: None,
-                maybe_prev_watermark: None,
-                maybe_prev_source_state: None,
-                maybe_prev_offset: None,
-            },
-        )
+    pub fn new(head: Multihash, maybe_source_name: Option<&'a str>) -> Self {
```
sergiimk, line 56: This visitor ended up very complex because there are at lest 5 different things that writer needs:
- last `AddData` event to connect next `AddData` event to
- names of parquet files from all `AddData` (for processing merge strategies)
- polling or push source
- data schema
- vocab

Therefore instead of having one very complex visitor you could express this as accepting 5 simple visitors that focus on one job each.

sergiimk (changes requested):

src/adapter/graphql/src/mutations/dataset_metadata_mut.rs
```diff
@@ -39,19 +44,23 @@ impl DatasetMetadataMut {
     }
 
     #[graphql(skip)]
-    async fn get_last_block_of_type<T: VariantOf<odf::MetadataEvent>>(
+    async fn get_last_block_of_type<T: VariantOf<odf::MetadataEvent>, const F: u32>(
```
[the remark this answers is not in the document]
s373r: `MetadataChain::accept_one()` was implemented to simplify the cases in which only one Visitor

Done

src/adapter/graphql/src/queries/datasets/dataset.rs
```diff
@@ -108,29 +107,33 @@ impl Dataset {
...
-        Ok(seed.system_time)
+
+        Ok(seed_block.system_time)
     }
 
     /// Creation time of the most recent metadata block in the chain
     async fn last_updated_at(&self, ctx: &Context<'_>) -> Result<DateTime<Utc>> {
         let dataset = self.get_dataset(ctx).await?;
-        let head = dataset
+        let mut visitor = <SearchNextBlockVisitor>::default();
```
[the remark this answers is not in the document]
s373r: Added `Dataset::get_block_by_ref()` and done

src/infra/core/src/query_service_impl.rs
s373r: Done

src/app/cli/src/commands/list_command.rs
```diff
@@ -258,19 +258,21 @@ impl Command for ListCommand {
             size.push(summary.data_size);
 
             if self.detail_level > 0 {
+                let mut data_block_visitor = <SearchDataBlocksVisitor>::next_data_block();
+                let mut next_block_visitor = <SearchNextBlockVisitor>::default();
```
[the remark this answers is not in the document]
s373r: Yes, you're right -- using the `reduce()` API looks better in this case

Done

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -7,73 +7,49 @@
...
+    NextOfType(MetadataEventTypeFlags),
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 
 pub trait MetadataChainVisitor: Sync + Send {
     type Error: std::error::Error;
 
-    fn visit(&mut self, hashed_block_ref: HashedMetadataBlockRef) -> Result<Decision, Self::Error>;
+    fn initial_decision(&self) -> Result<MetadataVisitorDecision, Self::Error>;
```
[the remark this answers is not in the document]
s373r: > Under what circumstances would `initial_decision()` return an error?

I originally thought this might serve as a simplification -- it worked for part of the cases, but yes, the examples you give show that it may not be the case

--

Now `initial_decision()` returns onlu `MetadataVisitorDecision`

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -12,50 +12,63 @@ use kamu_core::{
...
 
+    fn initial_decision(&self) -> Result<Decision, Self::Error> {
+        match (
+            self.is_seed_appended_block,
+            self.appended_prev_block_hash.is_some(),
+        ) {
+            (true, true) => Err(AppendValidationError::AppendingSeedBlockToNonEmptyChain.into()),
+            (false, false) => Err(AppendValidationError::FirstBlockMustBeSeed.into()),
+            (_, _) => Ok(Decision::Stop),
+        }
```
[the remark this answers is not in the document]
s373r: Done

src/infra/ingest-datafusion/src/visitor.rs
s373r: Explicitly pieces were extracted in individual Visitors -- Done

src/infra/core/src/repos/metadata_chain_validators.rs
s373r: The visitor was shredded to pieces

src/adapter/graphql/src/queries/datasets/dataset.rs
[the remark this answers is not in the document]
s373r: Done

src/domain/core/src/repos/metadata_chain_visitors.rs
```diff
@@ -0,0 +1,311 @@
...
+    NextDataBlock,
+    NextFilledNewWatermark,
+    NextFilledNewData,
+}
+
+pub struct SearchDataBlocksVisitor<E = InternalError> {
+    kind: SearchDataBlocksVisitorKind,
+    hashed_data_block: Option<(Multihash, MetadataBlockDataStream)>,
+    _phantom: PhantomData<E>,
+}
```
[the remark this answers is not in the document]
s373r: `SearchDataBlocksVisitor` removed, reduce API used -- Done

src/domain/core/src/repos/metadata_chain_visitors.rs
```diff
@@ -0,0 +1,311 @@
...
+    fn visit(&mut self, (hash, block): HashedMetadataBlockRef) -> Result<Decision, Self::Error> {
+        let Some(data_block) = block.as_data_stream_block() else {
+            unreachable!()
+        };
+
+        let has_data_block_found = match self.kind {
+            SearchDataBlocksVisitorKind::NextDataBlock => true,
+            SearchDataBlocksVisitorKind::NextFilledNewWatermark => {
+                data_block.event.new_watermark.is_some()
+            }
```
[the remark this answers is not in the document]
s373r: got it, thanks for your comment -- indeed, from a data schema point of view, the validation may have looked incorrect (although my goal was to move 1 to 1, past code):

image

I agree, when the code was separated into the Visitor it seemed like the logic was wrong (or redundant), although it was generally a workable option too

---

just thoughts, for the moment: `SearchDataBlocksVisitor` removed, reduce API used -- Done

src/adapter/graphql/src/mutations/dataset_metadata_mut.rs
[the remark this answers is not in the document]
s373r: > When I'm using SearchSetVocabVisitor - the visitor is a concrete type - it should tell me what errors it can return

This Visitor does not return any errors. Basically, we need this for calls with collections of Visitors (`accept_*()` methods). Thus, we can combine different translatable error types into one another.

--

Btw, After re-reading this comment, I tried removing `E` but could not compile in the end.

src/infra/core/src/repos/dataset_impl.rs
```diff
@@ -133,74 +133,80 @@ where
...
+        struct VisitorState {
+            requested_flags: Flag,
+            increment: UpdateSummaryIncrement,
+        }
 
-                    if let Some(checkpoint) = add_data.new_checkpoint {
-                        increment.seen_checkpoints_size += checkpoint.size;
+        const INITIAL_REQUESTED_FLAGS: Flag = Flag::SEED
+            .union(Flag::SET_TRANSFORM)
+            .union(Flag::DATA_BLOCK);
```
sergiimk, line 146: This is another case where I think `iter_blocks_interval` is simpler and as fast. Summaries need all data events so they won't be able to skip any blocks.

src/infra/core/src/repos/metadata_chain_validators.rs
```diff
@@ -323,3 +339,553 @@ impl<'a> MetadataChainVisitor for ValidateOffsetsAreSequentialVisitor<'a> {
...
+
+        Ok(Self {
+            appended_add_data,
+            prev_schema: None,
+            prev_add_data: None,
+            next_block_flags: Flag::SET_DATA_SCHEMA | Flag::ADD_DATA,
+        })
+    }
+
+    pub fn post_check(self) -> Result<(), AppendError> {
```
sergiimk, line 593: Should we make `finish() -> Result<(), E>` a part of the visitor interface with default empty implementation. This will ensure we will not forget to call these methods manually after `accpept` and that they will be called automatically in `accept_one` too.

src/infra/core/src/transform_service_impl.rs
```diff
@@ -204,60 +204,39 @@ impl TransformServiceImpl {
...
+            let mut execute_transform_visitor = <SearchExecuteTransformVisitor>::default();
+
+            dataset
+                .as_metadata_chain()
+                .accept_by_hash(
+                    &mut [
+                        &mut set_transform_visitor,
+                        &mut set_vocab_visitor,
+                        &mut set_data_schema_visitor,
+                        &mut execute_transform_visitor,
```
sergiimk, line 222: This looks amazing! :+1:

src/infra/core/src/transform_service_impl.rs
```diff
@@ -333,16 +312,29 @@ impl TransformServiceImpl {
...
-            .int_err()?;
-        Ok(last_data_block
-            .and_then(|(_, b)| b.event.last_offset())
+            .int_err()?
+            .as_metadata_chain()
+            .reduce(
+                None,
+                Decision::NextOfType(Flag::DATA_BLOCK),
+                |state, _, block| {
+                    let Some(data_block) = block.as_data_stream_block() else {
```
sergiimk, line 328: Given that data blocks are 99.9% of dataset the old version seemed simpler and as effective.

src/infra/core/src/transform_service_impl.rs
```diff
@@ -374,13 +366,30 @@ impl TransformServiceImpl {
...
+            .reduce_by_hash(
+                &last_unprocessed_block,
+                None,
+                Decision::NextOfType(Flag::DATA_BLOCK),
+                |state, hash, block| {
+                    if last_processed_block == Some(hash) {
+                        return Ok(Decision::Stop);
+                    }
+
+                    let Some(data_block) = block.as_data_stream_block() else {
```
sergiimk, line 383: ditto

src/infra/core/src/transform_service_impl.rs
```diff
@@ -465,16 +474,31 @@ impl TransformServiceImpl {
...
+                // TODO: This will not work with schema evolution
+                .reduce(
+                    None,
+                    Decision::NextOfType(Flag::DATA_BLOCK),
+                    |state, _, block| {
+                        let Some(data_block) = block.as_data_stream_block() else {
+                            unreachable!()
+                        };
+
+                        let Some(new_data) = data_block.event.new_data else {
```
sergiimk, line 490: I don't think this change improves anything. We might need a special skip list that skips over watermark-only data blocks, but this code will not be able to utilize it anyway, so might be better to leave the old simpler version.

src/infra/ingest-datafusion/src/writer.rs
```diff
@@ -851,16 +854,122 @@ impl DataWriterDataFusionBuilder {
...
+        }
 
-        let (decision, mut visitor) =
-            DataWriterDataFusionMetaDataStateVisitor::new(head.clone(), source_name);
+        let mut add_data_collection_visitor = GenericCallbackVisitor::new(
+            AddDataCollectionVisitorState {
+                data_slices: Vec::new(),
+                prev_checkpoint: None,
+                prev_watermark: None,
+                prev_offset: None,
```
sergiimk, line 897: Might be simpler to separate these into a visitor that gets last `AddDataBlock` - this way you won't need all the `is_none()` checks below.

sergiimk (changes requested):

src/app/cli/src/commands/list_command.rs
```diff
@@ -258,24 +258,46 @@ impl Command for ListCommand {
...
+                            let Some(data_block) = block.as_data_stream_block() else {
+                                return Ok(Decision::NextOfType(Flag::DATA_BLOCK));
+                            };
+
+                            state.last_watermark = data_block
+                                .event
+                                .new_watermark
+                                .map(DateTime::timestamp_micros);
+
+                            Ok::<Decision, InternalError>(Decision::Stop)
```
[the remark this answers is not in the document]
s373r: `reduce_*()` methods are separated into fallible & infallible variants -- Done

src/app/cli/src/commands/list_command.rs
```diff
@@ -258,24 +258,46 @@ impl Command for ListCommand {
...
+                    .reduce(
+                        State {
+                            num_blocks: 0,
+                            last_watermark: None,
+                        },
+                        Decision::Next,
+                        |state, _, block| {
+                            if state.num_blocks == 0 {
+                                // TODO: Should be precomputed
+                                state.num_blocks = block.sequence_number + 1;
```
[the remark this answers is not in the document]
s373r: Yeah, I probably overdid it at this point -- considering the case where the last block is the date block is unnecessary optimization, especially having different kinds of caches

Done

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -238,6 +232,64 @@ pub trait MetadataChainExt: MetadataChain {
...
+        V: MetadataChainVisitor<Error = E>,
+        E: Error + From<IterBlocksError>,
+    {
+        self.accept(&mut [&mut visitor]).await?;
+
+        Ok(visitor)
+    }
+
+    /// An auxiliary method that simplifies the work if only one custom Visitor
+    /// is used.
```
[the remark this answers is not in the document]
s373r: Eagle eyes! Fixed

src/domain/core/src/entities/metadata_chain.rs
```diff
@@ -238,6 +232,64 @@ pub trait MetadataChainExt: MetadataChain {
...
+    async fn reduce<S, F, E>(
+        &self,
+        state: S,
+        initial_decision: MetadataVisitorDecision,
+        callback: F,
+    ) -> Result<S, E>
+    where
+        S: Send + Sync,
```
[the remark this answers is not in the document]
s373r: > I suspect that this stems from MetadataChainVisitor being Send + Sync

Yes, you're right, that's the reason

> I don't see a case where we would access one visitor from multiple threads

in this case, the asynchrony is contributing to it -- since, I used mutable refs to a slice with Visitors, a future is not considered appropriate if the ref type inside the slice is not `Send`

We have several options:
1) we can leave it as is -- that's what I'd like to suggest
2) remove links and replace with owning a list of Visitors in the heap (`Vec`)
3) remove `Send` from `MetadataChainVisitor` и box futures that are waiting for `Send`

src/infra/core/src/dataset_changes_service_impl.rs
```diff
@@ -148,12 +174,21 @@ impl DatasetChangesServiceImpl {
...
-                    .filter_map_ok(|(_, b)| b.event.new_watermark)
-                    .try_first()
-                    .await
-                    .int_err()?;
+                    .reduce_by_hash(
+                        old_head,
+                        None,
+                        Decision::NextOfType(Flag::DATA_BLOCK),
+                        |state, _, block| {
+                            let Some(data_block) = block.as_data_stream_block() else {
```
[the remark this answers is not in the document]
s373r: Used `accept_one_by_hash()` with `SearchSingleDataBlockVisitor` -- Done

src/infra/core/src/repos/dataset_impl.rs
s373r: Reverted -- Done

src/infra/core/src/repos/metadata_chain_validators.rs
s373r: That's a great idea -- Done

src/infra/core/src/transform_service_impl.rs
s373r: I think you might like the new version better -- Done

src/infra/core/src/transform_service_impl.rs
s373r: I think you might like the new version better -- Done

src/infra/core/src/transform_service_impl.rs
s373r: I think you might like the new version better -- Done

src/infra/ingest-datafusion/src/writer.rs
```diff
@@ -851,16 +854,101 @@ impl DataWriterDataFusionBuilder {
             .resolve_ref(&self.block_ref)
             .await
             .int_err()?;
+        let mut seed_visitor = SearchSeedVisitor::create();
+        let mut set_vocab_visitor = SearchSetVocabVisitor::create();
+        let mut set_data_schema_visitor = SearchSetDataSchemaVisitor::create();
+        let mut prev_source_state_visitor = GenericCallbackVisitor::new(
+            None::<SourceState>,
```
sergiimk, line 861: Searching for last source state I think deserves a dedicated visitor type.

This is actually a very interesting edge case that we will have to keep in mind for skip lists.

sergiimk (changes requested):
Error handling is the only thing that I would like to agree on before merging - the rest is in a great shape :+1:

src/domain/core/src/repos/metadata_chain_visitors.rs
```diff
@@ -0,0 +1,402 @@
...
+    SetTransform,
+    SetVocab,
+    VariantOf,
+};
+
+use crate::{HashedMetadataBlockRef, MetadataChainVisitor, MetadataVisitorDecision as Decision};
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub struct SearchSetVocabVisitor<E = InternalError> {
```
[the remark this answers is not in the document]
sergiimk: Ohh, I think I finally get it ... it's because when we accept multiple visitors they ALL have to agree on a single error type. That's an interesting problem.

And I think I understand why you picked `InternalError` as the common denominator.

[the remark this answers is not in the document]
sergiimk: So if we have visitors A and B that **agree** on error types - below just works:
```rust
let visitor_a = VisitorA::new();
let visitor_b = VisitorB::new();
chain.accept(&mut[
  &mut visitor_a,
  &mut visitor_b,
]).await?;
````

If they **disagree** your current approach is to parametrize them with some common generic error type that their "true" error type converts into.

I think a better solution to this problem is (once again) composition:
```rust
let visitor_a = VisitorA::new().map_err(|e| ...);
let visitor_b = VisitorB::new().map_err(|e| ...);
chain.accept(&mut[
  &mut visitor_a,
  &mut visitor_b,
]).await?;
````
Where `map_err()` is implemented for `MetadataChainVisitor` trait and wraps visitors with a simple struct that converts its error.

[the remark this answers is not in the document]
Comment (sergiimk) [reply src/domain/core/src/repos/metadata_chain_visitors.rs:41]: Second problem I see is what error should visitors that aren't expected to fail return...

In future this might be a good case for never type (`!`) but its not yet stabilized for such general-purpose use...

We could introduce a `Never` error type specifically for the visitor case (that can be used but never created).

File: src/domain/core/src/entities/metadata_chain.rs [hunk 38]
```diff
@@ -236,8 +230,148 @@ pub trait MetadataChainExt: MetadataChain {
...
+    }
+
+    /// Method that searches for the last data block
+    async fn last_data_block(
+        &self,
+    ) -> Result<SearchSingleDataBlockVisitor<IterBlocksError>, IterBlocksError> {
+        self.accept_one(SearchSingleDataBlockVisitor::next()).await
+    }
+
+    /// Same as [Self::last_data_block()], but skipping date blocks that have no
```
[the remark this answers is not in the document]
s373r: Fixed

src/domain/core/src/repos/metadata_chain_visitors.rs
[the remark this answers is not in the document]
s373r: > In future this might be a good case for never type (!) but its not yet stabilized for such general-purpose use...

Yes, indeed -- I considered this option, but Rust in this case reduces the error type to the maximum, since it expects the never type everywhere. maybe I'm spoiled by the never type in Haskell, where the situation is the opposite (if I'm not confused).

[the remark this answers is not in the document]
s373r: > Ohh, I think I finally get it ... it's because when we accept multiple visitors they ALL have to agree on a single error type. That's an interesting problem.

Yeah, it's exactly like that -- mentioned it in a thread above (https://github.com/kamu-data/kamu-cli/pull/543#discussion_r1538388389)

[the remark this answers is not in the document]
s373r: > I think a better solution to this problem is (once again) composition:

Interesting point, I have my doubts about the error type output -- I'll have to take an extra portion of programmer's fuel (coffee) to implement that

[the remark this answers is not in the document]
s373r: Done

src/domain/core/src/repos/metadata_chain_visitors.rs
```diff
@@ -0,0 +1,404 @@
...
+    MetadataChainVisitor,
+    MetadataChainVisitorHolder,
+    MetadataChainVisitorHolderFactory,
+    MetadataVisitorDecision as Decision,
+};
+
+///////////////////////////////////////////////////////////////////////////////
+
+// TODO: use Infallible/!
+type VisitorError = IterBlocksError;
```
s373r, line 50: I played with `Infallible` (from stable) and `!` (from unstable) for a long time, in general I had a prototype, but it was very verbose, as in some (almost all situations), we had to duplicate types for method calls.

From the point of view of type purity, it's a compromise solution, I realize that -- in result, to make the final (caller) code simpler, I decided to use `IterBlocksError`.

src/domain/core/src/entities/metadata_chain.rs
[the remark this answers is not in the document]
s373r: `Sync` bound has been removed -- Done

sergiimk (approved):
Heck yess, this API is beautiful! :rocket: :ship:

rmn-boiko (approved):

PR #545, opened 2024-03-11
Simplify workspace

sergiimk:
I looked at some other projects' `Cargo.toml` and it doesn't look like we need to specify these versions explicitly.

This will reduce the amount of `diff` when we bump the release version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,32 +36,32 @@ resolver = "2"
 
 [workspace.dependencies]
 # Utils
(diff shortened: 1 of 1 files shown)


sergiimk:
Never mind, looks like `cargo-deny` is not happy about these as by just having `path = ''` they are treated as wildcard (`version = '*'`) dependencies.

I assume we would run into issues also if we decided to publish some of the crates to crates.io as it also forbids wildcards.

Issue #532, opened 2024-03-01
Pulling with `--all` flag in a multi-tenant workspace should only match current account datasets


sergiimk:
### Describe the bug

In multi-tenant local workspace (mostly used for demos and testing) executing `kamu --account x pull --all` will attempt to pull datasets of all users, not just the current one.

### Steps To Reproduce

1. `kamu init --multi-tenant`
2. `kamu --account x add dataset1.yaml`
3. kamu --account y add dataset2.yaml`
4. `kamu -account x pull --all`

### Current Behavior

Pull will successfully work for `dataset1` but then will fail for `dataset2` with permission error.

### Expected Behavior

`kamu --account x pull --all` flag will only pull datasets that account `x` has write access to

### kamu system info

```sh
0.162.1
```

PR #549, opened 2024-03-12
Fix pull dataset by account

rmn-boiko:
## Description

Closes: [Pulling with --all flag in a multi-tenant workspace should only match current account datasets](https://github.com/kamu-data/kamu-cli/issues/532)

- [x] Unit-tests added

### DIFF
--- a/src/app/cli/src/commands/pull_command.rs
+++ b/src/app/cli/src/commands/pull_command.rs
@@ -29,6 +29,7 @@ pub struct PullCommand {
     search_svc: Arc<dyn SearchService>,
     output_config: Arc<OutputConfig>,
     refs: Vec<DatasetRefAnyPattern>,
+    current_account_subject: Arc<CurrentAccountSubject>,
     all: bool,
     recursive: bool,
     fetch_uncacheable: bool,
@@ -44,6 +45,7 @@ impl PullCommand {
         search_svc: Arc<dyn SearchService>,
         output_config: Arc<OutputConfig>,
         refs: I,
+        current_account_subject: Arc<CurrentAccountSubject>,
         all: bool,
         recursive: bool,
         fetch_uncacheable: bool,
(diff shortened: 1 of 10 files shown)


src/app/cli/src/commands/pull_command.rs
```diff
@@ -104,11 +107,14 @@ impl PullCommand {
...
         &self,
         listener: Option<Arc<dyn PullMultiListener>>,
+        current_account_name: Option<AccountName>,
     ) -> Result<Vec<PullResponse>, CLIError> {
         let dataset_refs: Vec<_> = filter_datasets_by_any_pattern(
             self.dataset_repo.as_ref(),
             self.search_svc.clone(),
             self.refs.clone(),
+            current_account_name,
+            self.all,
```
zaychenko-sergei, line 117: I don't think that switch `-all` and list of references are a valid combination. If I use `-all`, I would expect to update all of my datasets without specifying any patterns/refs. If that is so, then it's quite weird to mix extraction of all account datasets and "-all" switch in 1 filtering algorithm.

rmn-boiko: We will not allow the use of both refs and `--all`. But as we are encapsulating a logic of fetching datasets to process, I decided it's a good place to handle this flag. And one more reason that `--all` will not mean like really all records in the case of multitenant workspace.

rmn-boiko: Agreed to bring changes and move `-all` flag handling back to pull implementation level

src/domain/core/src/services/pull_service.rs
```diff
@@ -124,6 +125,8 @@ pub struct PullMultiOptions {
     pub recursive: bool,
     /// Pull all known datasets
     pub all: bool,
+    /// Owner of datasets
+    pub current_account_name: AccountName,
```
zaychenko-sergei, line 129: Adding this parameter is unnecessary.
`PullServiceImpl` already injects `CurrentAccountSubject`, so no need to propagate this value.

PR #556, opened 2024-03-19
Aggressive S3 metadata caching

sergiimk:
## Description

Related to: https://github.com/kamu-data/kamu-node/issues/61

Adds two layers of caching:
- `ObjectRepositoryCachingLocalFS` - a read-through cache that stores objects in a local directory, which I use for metadata blocks
- `S3RegistryCache` - a singleton in-memory cache of datasets contained in S3 bucket

Together they turned some of my sample queries from taking 4.5s to taking just 9ms.

- [ ] Unit-tests added


sergiimk:
I will release API server based on this branch as I need good perf for tomorrow's demo.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -82,11 +82,25 @@ impl DatasetRepositoryLocalFs {
         ))
     }
 
+    fn build_dataset(layout: DatasetLayout, event_bus: Arc<EventBus>) -> Arc<dyn Dataset> {
```
sergiimk, line 85: Notice that I have severed ties between `DatasetRepository`s and `DatasetFactory`.

After making a bunch of changes to make `DatasetFactory` super configurable, I realized that `DatasetFactory` should be used only when talking to remote repositories, and `DatasetRepository` implementations should have independent low-level control over how they create dataset instances. Attempting to combine the two just results in configuration and dependency nightmare.

src/infra/core/src/repos/dataset_repository_s3.rs
```diff
@@ -415,3 +564,27 @@ impl DatasetRepository for DatasetRepositoryS3 {
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct S3RegistryCache {
+    state: Arc<Mutex<State>>,
```
sergiimk, line 569: This is far from elegant but I think we will factor it out into a proper registry component in future.
I'd rather save this refactoring until #342.

s373r (approved):

zaychenko-sergei (approved):

src/infra/core/src/repos/dataset_repository_s3.rs
zaychenko-sergei: After #342 , we are caching only the list of stored identifiers.

Issue #320, opened 2023-12-13
Compaction / squash operation (history-altering)


sergiimk:
Our periodically ingested datasets grow fast over time. The data size is NOT a problem here - it’s the number of files. Since our sync protocols operate on per-file basis with no parallelism syncs start to take a very long time once dataset reaches > 200 blocks which degrades user experience in our demos.

Currently we deal with this by periodically “defaulting” on history - deleting old datasets and creating new ones.

- In most cases we are throwing data away (non-reproducible APIs)
- The process is highly manual
- Reconstructing some dataset (e.g. from Ethereum) can take a long time

This task is to implement a squash / compaction solution that:

- merges all data slices in a dataset into a single one
- creates a new dataset with same initial metadata blocks and a single AddData block that links to latest checkpoint
  - in future we may support slicing data into several reasonably-sized blocks
- can be invoked as `kamu system compact <dataset>`
- it will effectively re-write history.

We can then automate the force-push of such datasets into storage to simplify the “default on history” procedure.

PR #557, opened 2024-03-19
Add compact CLI command logic

rmn-boiko:
## Description

Closes: [Compaction/squash operation](https://github.com/kamu-data/kamu-cli/issues/320)

- [x] Unit-tests added


src/domain/core/src/services/compact_service.rs
```diff
@@ -0,0 +1,208 @@
...
+use thiserror::Error;
+
+use crate::*;
+
+#[async_trait::async_trait]
+pub trait CompactService: Send + Sync {
+    async fn compact_dataset(
+        &self,
+        dataset_handle: &DatasetHandle,
+        dataset_dir_path: &Path,
```
zaychenko-sergei, line 23: Passing a path might be fine for local FS repository. What are you going to do with S3?
The repository actualy knows where it is located on the disk very well, I don't think you need to pass this info here.

src/domain/core/src/services/compact_service.rs
```diff
@@ -0,0 +1,208 @@
...
+        #[backtrace]
+        DatasetNotFoundError,
+    ),
+    #[error(transparent)]
+    RefNotFound(
+        #[from]
+        #[backtrace]
+        RefNotFoundError,
+    ),
+    #[error(transparent)]
```
zaychenko-sergei, line 46: It feels like this linearized error has no much sense. We should preserve layering here, except for cross-cutting concerns like security.

We asked a high level service to compact the given dataset. What could go wrong:
 - dataset does not exist `DatasetNotFoundError`
 - no access to read or write to this dataset `AccessError`
 - something internal, related to broken lower layers `InternalError`.
 
 The user of `compact_dataset` operates at higher level than the user of individual block operations..

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,289 @@
...
+        dataset_handle: &DatasetHandle,
+        dataset_dir_path: &Path,
+        multi_listener: Option<Arc<dyn CompactionMultiListener>>,
+    ) -> Result<(), CompactError> {
+        let listener = multi_listener
+            .and_then(|l| l.begin_compact(dataset_handle))
+            .unwrap_or(Arc::new(NullCompactionListener {}));
+        self.dataset_authorizer
+            .check_action_allowed(dataset_handle, domain::auth::DatasetAction::Read)
+            .await?;
```
zaychenko-sergei, line 242: Compaction is both read and write. Checking `Write` action should work.
Right now you have allowed anyone to compact public datasets.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,289 @@
...
+            .get_summary(GetSummaryOpts::default())
+            .await
+            .int_err()?
+            .kind;
+
+        if dataset_kind != DatasetKind::Root {
+            return Err(CompactError::InvalidDatasetKind(InvalidDatasetKindError {
+                dataset_name: dataset_handle.alias.dataset_name.clone(),
+            }));
+        }
```
zaychenko-sergei, line 259: Can you explain why compacting derived dataset is not allowed?

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        let compact_dir_path = self.create_compact_dir(dataset_dir_path)?;
```
zaychenko-sergei, line 261: Where would I create this in S3 on deployment?

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,289 @@
...
+        let new_file_path = self
+            .write_file(
+                chain_files_info.data_slice_file_urls.clone(),
+                &dataset_handle.alias.dataset_name,
+                &compact_dir_path,
+                listener.clone(),
+            )
+            .await?;
+
+        listener.begin_phase(CompactionPhase::ReplaceChainHead);
```
zaychenko-sergei, line 275: Weird name of the phase. I see replacement of data files, deleting old data files. Where is the head replaced?

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,289 @@
...
+        let mut block_stream = chain.iter_blocks_interval(&head, None, false);
+        let object_data_repo = dataset.as_data_repo();
+        let object_block_repo = chain.as_metadata_block_repository();
+        let mut last_dataslice_info: Option<LastDatasliceInfo> = None;
+
+        while let Some((block_hash, block)) = block_stream.try_next().await? {
+            block_file_urls.push(object_block_repo.get_internal_url(&block_hash).await.into());
+
+            match block.event {
+                MetadataEvent::AddData(add_data_event) => {
```
zaychenko-sergei, line 88: a) See 'filter_data_stream_blocks', however, be careful, as there are major iteration changes brought by DimaP in his "metadata chain visitor" activities.

b) Why I only see `AddData` for root datasets, and not see `ExecuteTransform` for derived datasets?

src/app/cli/src/commands/compact_command.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use crate::{CLIError, Command, CompactionMultiProgress, VerificationMultiProgress};
+
+pub struct CompactCommand {
+    dataset_repo: Arc<dyn DatasetRepository>,
+    verification_svc: Arc<dyn VerificationService>,
+    compact_svc: Arc<dyn CompactService>,
+    dataset_ref: DatasetRef,
+    dataset_dir_path: PathBuf,
+}
```
zaychenko-sergei, line 30: We discussed the importance of limiting maximum data chunks. 
Data could be bigger than the entire memory of the computer running it.
Note that deployment API services have quite limited memory, usually not more than 1 GB.

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,136 @@
...
+use futures::TryStreamExt;
+use kamu::domain::*;
+use kamu::testing::{DatasetTestHelper, MetadataFactory};
+use kamu::*;
+use kamu_core::{auth, CurrentAccountSubject};
+use opendatafabric::*;
+const FILE_DATA_ARRAY_SIZE: usize = 32;
+
+#[tokio::test]
+async fn test_dataset_compact() {
```
zaychenko-sergei, line 25: Must be tested with S3/Minio container as well.

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,136 @@
...
+        .add_value(CurrentAccountSubject::new_test())
+        .add::<auth::AlwaysHappyDatasetActionAuthorizer>()
+        .add_builder(
+            DatasetRepositoryLocalFs::builder()
+                .with_root(tempdir.path().join("datasets"))
+                .with_multi_tenant(false),
+        )
+        .bind::<dyn DatasetRepository, DatasetRepositoryLocalFs>()
+        .add::<CompactServiceImpl>()
+        .build();
```
zaychenko-sergei, line 44: Try separating catalog and other initialization code from the main test logic.
See a lot of examples of "harness" structs, where we prepare the initial state and all the plumbing, and keep test very readable at high level of abstraction.

zaychenko-sergei (changes requested):

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Oh, you are right, it was by mistake. In my mind it should be `Write` permission

src/app/cli/src/commands/compact_command.rs
zaychenko-sergei: One could say that our querying is also not prepared for such large files. I'll give another point - without chunks we are limiting transfer parallelism. I guess it would be much faster to have N parallel downloads of a few-megabyte files than 1 huge file.

src/infra/core/src/compact_service_impl.rs
rmn-boiko: We discussed in RFC that it is a little bit tricky to support it right now.

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Firstly we reset the head of the dataset chain and commit a new block.

src/infra/core/src/compact_service_impl.rs
rmn-boiko: It was my first solution, But such as I have to take all replaced blocks I decided to rewrite this part to iterate through all blocks to avoid additional iteration during cleaning up

src/infra/core/src/compact_service_impl.rs
sergiimk: I wonder if we could do compaction *within* the same dataset...

Compaction is essentially committing blocks with different data slices. If we support committing a block without modifying the HEAD - we could start essentially creating a branch and then switch `HEAD` to point to that branch as the very last step.

MetadataChain already supports appending blocks without modifying the `HEAD` reference, so we might just need to allow this in `Dataset::commit_*` helper functions too.

This way:
- we don't need any temporary directories
- we preserve original dataset intact until the very last point of updating the HEAD
- if compaction fails / is interrupted - we just have a bit of garbage left for GC to take care of

src/app/cli/src/commands/compact_command.rs
```diff
@@ -0,0 +1,111 @@
...
+            .await
+            .map_err(CLIError::failure)?;
+
+        if let Err(err) = self.verify_dataset(&dataset_handle).await {
+            eprintln!(
+                "{}",
+                console::style("Cannot perform compacting, dataset is invalid".to_string()).red()
+            );
+            return Err(err);
+        }
```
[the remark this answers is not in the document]
rmn-boiko: Probably, misunderstood this commend https://docs.google.com/document/d/1UCFUQ3V_muPbJuhl8dHFf3aRXlXL3yGbekbDwPlnaUA/edit?disco=AAABILoXam0. I was thinking that if only one block is corrupted, user can reset this block manually. And after run compaction. But if we merge corrupted block to fix it will be really dificult.

src/domain/core/src/services/compact_service.rs
rmn-boiko: Removed

src/app/cli/src/output/compact_progress.rs
```diff
@@ -0,0 +1,121 @@
...
+                console::style("Dataset compacted succesfully".to_owned()).green(),
+            ),
+        );
+    }
+
+    fn begin_phase(&self, phase: CompactionPhase) {
+        let message = match phase {
+            CompactionPhase::GatherChainInfo => "Gathering chain information",
+            CompactionPhase::ReadDatasliceFiles => "Reading dataslice files",
+            CompactionPhase::WriteDataasliceFiles => "Writing dataslice files",
```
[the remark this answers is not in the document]
rmn-boiko: Chaned

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Changed

src/app/cli/src/commands/compact_command.rs
rmn-boiko: Added

src/app/cli/src/cli_parser.rs
```diff
@@ -330,6 +330,29 @@ pub fn cli() -> Command {
...
+                        r#"
+                        This command commpact all files in the dataset into a few depends from max-slice-size.
+
+                        Take great care when compacting datasets. You will lose all history of metadata.
+
+                        **Examples:**
+
+                        Compact a local dataset:
+
+                            kamu compact my.dataset
```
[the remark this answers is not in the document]
rmn-boiko: Added new flag + changed command

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Great point, done

src/app/cli/src/cli_parser.rs
```diff
@@ -1244,6 +1244,39 @@ pub fn cli() -> Command {
...
+                                r#"
+                                This command commpact all files in the dataset into a few depends from max-slice-size.
+
+                                Take great care when compacting datasets. You will lose all history of metadata.
+
+                                **Examples:**
+
+                                Compact a local dataset:
+
+                                    kamu compact my.dataset
```
sergiimk, line 1277:
```
For datasets that get frequent small appends the number of data slices can grow over time and affect the performance of querying. This command allows to merge multiple small data slices into a few large files, which can be beneficial in terms of size from more compact encoding, and in query performance, as data engines will have to scan through far fewer file headers.

Compactions have two types: soft, and hard.

Soft compactions produce new files while leaving the old blocks intact. This allows for faster queries, while still preserving the accurate history of how dataset evolved over time.

Hard compactions rewrite the history of the dataset as if data was originally written in big batches. They allow to shrink the history of a dataset to just a few blocks, reclaim the space used by old data files, but at the expense of history loss. Hard compactions will rewrite the metadata chain, changing block hashes. Therefore they will **break all downstream datasets** that depend on them.

**Examples:**

Perform a history-altering hard compaction:

    kamu compact --hard my.dataset
```

src/app/cli/src/cli_parser.rs
```diff
@@ -1244,6 +1244,39 @@ pub fn cli() -> Command {
...
+                                Arg::new("max-slice-size")
+                                    .long("max-slice-size")
+                                    .value_parser(value_parser!(u64))
+                                    .default_value("0")
+                                    .value_name("SIZE")
+                                    .help("Maximum size of a single data slice file in bytes"),
+                                Arg::new("hard")
+                                    .long("hard")
+                                    .action(ArgAction::SetTrue)
+                                    .help("Run a hard dataset compaction"),
```
sergiimk, line 1265: `Perform 'hard' compaction that rewrites the history of a dataset`

src/app/cli/src/cli_parser.rs
rmn-boiko: Changed

src/app/cli/src/cli_parser.rs
rmn-boiko: Changed

src/domain/core/src/services/compact_service.rs
rmn-boiko: Changed to run dir

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,136 @@
...
+        )
+        .await
+        .unwrap();
+
+    let head = DatasetTestHelper::append_random_data(
+        dataset_repo.as_ref(),
+        root_dataset_alias.as_local_ref(),
+        FILE_DATA_ARRAY_SIZE,
+    )
+    .await;
```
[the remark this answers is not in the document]
rmn-boiko: Changed to real dataset
Still working on remote dataset testings

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+        let object_data_repo = dataset.as_data_repo();
+        let object_block_repo = chain.as_metadata_block_repository();
+
+        while let Some((block_hash, block)) = block_stream.try_next().await? {
+            block_file_urls.push(object_block_repo.get_internal_url(&block_hash).await);
+
+            match block.event {
+                MetadataEvent::AddData(add_data_event) => {
+                    if let Some(output_slice) = &add_data_event.new_data {
+                        new_head = block.prev_block_hash.clone();
```
zaychenko-sergei, line 110: I don't fully understand the role of this `new_head`, perhaps, the name does not match to it's purpose.
Note that the blocks are iterated in reverse direction - from the newest to oldes, so it maybe is an "old_head".
Looking at the algorithm, it will point to the last node before the very first `AddData`.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+                            batch_records += current_records;
+                        }
+
+                        if data_slice_batch_info.new_file_path.is_none() {
+                            data_slice_batch_info.new_file_path = Some(compact_dir_path.join(
+                                Self::get_random_operation_name_with_prefix(
+                                    format!("{dataset_name}-").as_str(),
+                                ),
+                            ));
+                        }
```
zaychenko-sergei, line 149: This looks unrelated to the gathering algorithm. This is already an operation assignment, post-processing de-facto. Maybe it should be done separately or event does not have to be stored. I think it should be generated in `write_file` and returned to use in `replace_data_files` instead of being part of this struct.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+        name.push_str(prefix);
+        name.extend(
+            rand::thread_rng()
+                .sample_iter(&Alphanumeric)
+                .take(10)
+                .map(char::from),
+        );
+        name
+    }
+}
```
zaychenko-sergei, line 311: I think we have similar algorithms scattered all over the code base. Maybe they deserve to be somehwere in utility crate?

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    fn remove_old_files(
+        &self,
+        chain_files_info: &ChainFilesInfo,
+        old_data_slices: &Vec<Url>,
+    ) -> Result<(), CompactError> {
+        for file_url in old_data_slices {
+            fs::remove_file(file_url.to_file_path().unwrap()).int_err()?;
+        }
```
zaychenko-sergei, line 284: 1) Will not work for S3. S3/Minio URLs would not convert to file paths.
2) This isn't removing the checkpoint files. You are not even considering them in entire algorithm.

In general, is this step even nevessary? I understand we've left some garbage, but we are not pointing to it. We have plans to implement Garbage Collector in the datasets, maybe it would be enough?

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    async fn replace_data_files(
+        &self,
+        dataset: Arc<dyn Dataset>,
+        chain_files_info: &ChainFilesInfo,
+    ) -> Result<(Vec<Url>, Multihash), CompactError> {
+        let chain = dataset.as_metadata_chain();
+
+        let mut new_head = chain_files_info.new_head.clone();
```
zaychenko-sergei, line 221: So that was a previous block hash before first `AddData`. Not a "new_head" from gathering point of view.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+                    },
+                )
+                .await
+                .int_err()?;
+            new_head = commit_result.new_head;
+
+            // Check is new commited block equals to old one
+            // It is possible when `max-slice-size` is smaller
+            // than size of one data slice
+            if data_slice_batch_info.data_slices_batch.len() == 1 {
```
zaychenko-sergei, line 255: It should not be okay. Just improve the gathering algorithm: if there is a single block that is larger than the limit, just make a batch of this single block unconditionally.

src/domain/core/src/entities/dataset.rs
```diff
@@ -98,6 +98,8 @@ pub struct CommitOpts<'a> {
     /// Whether to check for presence of linked objects like data and
     /// checkpoints in the respective repos
     pub check_object_refs: bool,
+    // Whether to reset head to new commited block
+    pub is_reset_head: bool,
```
zaychenko-sergei, line 102: is_reset_head => update_head

zaychenko-sergei (changes requested):

src/infra/core/tests/tests/test_compact_service_impl.rs
zaychenko-sergei: LFS vs S3 is a separate topic from local vs remote.
Regarding remote, if any side of the local or remote is compacted separately, it's no longer possible to sync. Although, as far as I understand, if compacting with the same limits is applied to both copies, this would produce identical results.

resources/cli-reference.md
```diff
@@ -1169,6 +1170,43 @@ Generate a platform token from a known secret for debugging
...
+* `<DATASET>` — Local dataset reference(s)
+
+**Options:**
+
+* `--max-slice-size <SIZE>` — Maximum size of a single data slice file in bytes
+
+  Default value: `1073741824`
+* `--max-slice-records <RECORDS>` — Maximum amount of records in a single data slice file
+
+  Default value: `10000`
```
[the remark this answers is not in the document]
rmn-boiko: 1000 records. We decided to add an additional field such as calculating the size of the parquet file before it was written pretty complex.

resources/cli-reference.md
```diff
@@ -1169,6 +1170,43 @@ Generate a platform token from a known secret for debugging
...
+
+**Usage:** `kamu system compact [OPTIONS] <dataset>...`
+
+**Arguments:**
+
+* `<DATASET>` — Local dataset reference(s)
+
+**Options:**
+
+* `--max-slice-size <SIZE>` — Maximum size of a single data slice file in bytes
```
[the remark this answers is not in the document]
rmn-boiko: I set 1Gb as default for `--max-slice-size` and 10000 records for `--max-slice-records`. Let me know if you have other values in mind.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+            return Err(CompactError::InvalidDatasetKind(InvalidDatasetKindError {
+                dataset_name: dataset_handle.alias.dataset_name.clone(),
+            }));
+        }
+
+        let compact_dir_path = self.create_run_compact_dir()?;
+
+        listener.begin_phase(CompactionPhase::GatherChainInfo);
+        let chain_files_info = self
+            .get_chain_files_info(
```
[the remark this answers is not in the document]
rmn-boiko: Changed

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,386 @@
...
+        // Declare mut values for result
+
+        let mut new_head: Option<Multihash> = None;
+        let mut block_file_urls: Vec<Url> = vec![];
+        let mut data_slice_batch_info: DataSliceBatchInfo = DataSliceBatchInfo {
+            prev_offset: None,
+            new_offset_interval: None,
+            data_slices_batch: vec![],
+            new_file_path: None,
+        };
```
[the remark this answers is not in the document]
rmn-boiko: Thanks, changed

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Yeah,  you are right this is exactly the last node before the very first `AddData`. Sorry for confusing name, In my mind it will be a new head after resetting old blocks. But `old_head` sounds better

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Agree, this is a tricky part that won't work correctly with remotes. @sergiimk @zaychenko-sergei What do you say if we currently remove this step and assign it as a separate issue??

src/domain/core/src/entities/dataset.rs
rmn-boiko: Changed

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Great point, changed

src/infra/core/src/compact_service_impl.rs
zaychenko-sergei: It is fine to skip it. We hope GC implementation will solve deleting all orphan files.

src/infra/core/src/compact_service_impl.rs
zaychenko-sergei: This wasn't a high priority, but I really meant using it everywhere in similar situations. This code is not trivial, and I think we copy paste it for container names, network names, ....

resources/cli-reference.md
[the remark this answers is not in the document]
zaychenko-sergei: I see the problem: the generated MD file is formatted in confusing way. There is an empty line between switch and it's default, but no empty line before the next switch. So I intuitively assumed the default applies to the switch below, not above. Plz check if you can correct the generation script.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,389 @@
...
+    ) -> Result<ChainFilesInfo, CompactError> {
+        // Declare mut values for result
+
+        let mut old_head: Option<Multihash> = None;
+        let mut block_file_urls: Vec<Url> = vec![];
+        let mut data_slice_batch_info: DataSliceBatchInfo = DataSliceBatchInfo::default();
+        let mut data_slice_batches: Vec<DataSliceBatchInfo> = vec![];
+        let (mut batch_size, mut new_end_offset_interval, mut batch_records): (u64, u64, u64) =
+            (0, 0, 0);
+
```
zaychenko-sergei, line 94: Minor: I think this can be compacted to `(0u64, 0u64, 0u64)` without type specification

src/infra/core/tests/tests/test_compact_service_impl.rs
zaychenko-sergei: On the other hand, it's the same for GIT. If someone force-pushed history rewrite change like `git rebase`, the other side has to use `git pull --rebase`, a normal procedure fails due to history divergence. We have "-force" switch in `kamu pull` and `kamu push` too.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,389 @@
...
+    // when data_slices_batch.len() == 1
+    // and will be used tp not rewriting such blocks
+    pub hash: Option<Multihash>,
+}
+
+struct ChainFilesInfo {
+    _block_file_urls: Vec<Url>,
+    old_head: Multihash,
+    data_slice_batches: Vec<DataSliceBatchInfo>,
+}
```
zaychenko-sergei, line 63: The algorithm does not consider checkpoints. We should leave the very latest checkpoint only and drop any checkpoints before. If there were no checkpoints, nothing to do.

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,355 @@
...
+            | 2 | b |
+            | 3 | c |
+            | 4 | d |
+            | 5 | e |
+            | 1 | a |
+            | 2 | b |
+            | 3 | c |
+            | 4 | d |
+            | 5 | e |
+            +---+---+
```
sergiimk, line 155: This test doesn't quite tell me that the ordering of blocks is preserved by the merge process - order is hard to tell when two data slices are identical :)

I think we have to write data properly, like in `test_push_ingest.rs` for us to have the `offset` column.

I would also recommend having another test that generates 100+ of small blocks and merges them into one, ensuring the order of `offset`s is preserved.

Let us know if you need help with that.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,389 @@
...
+        let mut current_head = chain_files_info.old_head.clone();
+        let mut old_data_slices: Vec<Url> = vec![];
+
+        for data_slice_batch_info in chain_files_info.data_slice_batches.iter().rev() {
+            if let Some(block_hash) = data_slice_batch_info.hash.as_ref() {
+                let block = chain.get_block(block_hash).await.int_err()?;
+
+                if let MetadataEvent::AddData(add_data_event) = block.event {
+                    let block_event = MetadataEvent::AddData(AddData {
+                        prev_checkpoint: None,
```
sergiimk, line 238: Ingest engines may actually produce checkpoints, so we need to preserve them. When merging two `AddData` blocks the resulting block's checkpoint should be one of the last block.

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Changed

src/domain/core/src/entities/dataset.rs
```diff
@@ -98,6 +98,8 @@ pub struct CommitOpts<'a> {
     /// Whether to check for presence of linked objects like data and
     /// checkpoints in the respective repos
     pub check_object_refs: bool,
+    // Whether to reset head to new commited block
+    pub update_head: bool,
```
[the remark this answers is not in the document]
rmn-boiko: Renamed

src/infra/core/src/repos/metadata_block_repository_caching_inmem.rs
```diff
@@ -176,4 +177,8 @@ where
 
         insert_result
     }
+
+    async fn get_internal_url(&self, hash: &Multihash) -> Url {
+        self.wrapped.get_internal_url(hash).await
+    }
```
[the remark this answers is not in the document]
rmn-boiko: Yeah, thanks, removed

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Added checkpoint handling

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Added checkpoint handling

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,389 @@
...
+                )
+                .await
+                .int_err()?;
+        }
+
+        Ok(())
+    }
+
+    fn create_run_compact_dir(&self) -> Result<PathBuf, CompactError> {
+        let compact_dir_path = self.run_info_dir.join("compact");
```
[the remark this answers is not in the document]
rmn-boiko: agree, changed

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,389 @@
...
+                    datafusion::execution::options::ParquetReadOptions {
+                        file_extension: "",
+                        ..Default::default()
+                    },
+                )
+                .await
+                .int_err()?;
+
+            data_slice_batch_info.new_file_path = Some(compact_dir_path.join(
+                get_random_operation_name_with_prefix(format!("{dataset_name}-").as_str()),
```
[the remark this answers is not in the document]
rmn-boiko: changed

src/infra/core/tests/tests/test_compact_service_impl.rs
rmn-boiko: Added new tests and in process found issue with sorting(Fixed as well)

src/app/cli/src/cli_parser.rs
```diff
@@ -330,6 +330,29 @@ pub fn cli() -> Command {
...
                     )),
+                Command::new("compact")
+                    .about("Compact a dataset")
+                    .args([
+                        Arg::new("dataset")
+                            .action(ArgAction::Append)
+                            .index(1)
+                            .required(true)
+                            .value_parser(value_parse_dataset_ref_local)
+                            .help("Local dataset reference(s)"),
```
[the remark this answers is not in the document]
rmn-boiko: For now we decided to not support multiple datasets or patterns

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Changed in all places

src/domain/core/src/entities/dataset.rs
```diff
@@ -61,6 +61,25 @@ pub trait Dataset: Send + Sync {
         checkpoint: Option<&OwnedFile>,
     ) -> Result<ExecuteTransform, InternalError>;
 
+    /// Helper function to prepare [AddDaat] event parameters
```
zaychenko-sergei, line 64: Typo: `AddDaat` => `AddData`.
`AddDaat` sounds very much like Dutch language :-)

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,378 @@
...
+
+pub struct CompactServiceImpl {
+    dataset_repo: Arc<dyn DatasetRepository>,
+    dataset_authorizer: Arc<dyn domain::auth::DatasetActionAuthorizer>,
+    time_source: Arc<dyn SystemTimeSource>,
+    run_info_dir: PathBuf,
+}
+
+#[derive(Debug, Default, Clone)]
+struct DataSliceBatchInfo {
```
zaychenko-sergei, line 49: Suggestions:
1) Make an enum to distinguish compacted vs non-compacted batch:

```
enum DataSliceBatch {
    CompactedBatch( DataSliceBatchInfo),
    SingleBlock(...)
}
```
This should reduce number of `Option` fields.

2) Simplify how `prev_*` and `new_*` vars are handled by grouping into child structs:
```

#[derive(Debug, Default, Clone)]
struct DataSliceBatchUpperBound {
    pub new_source_state: Option<SourceState>,
    pub new_checkpoint: Option<Checkpoint>,    
    pub new_end_offset: Option<u64>,
}

#[derive(Debug, Default, Clone)]
struct DataSliceBatchLowerBound {
    pub prev_offset: Option<u64>,
    pub prev_checkpoint: Option<Multihash>,
    pub prev_end_offset: Option<u64>,
}
```

src/domain/core/src/entities/dataset.rs
rmn-boiko: Just wanted bring some Dutch in project.
Fixed)

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,417 @@
...
+                DataSliceBatch::CompactedBatch(data_slice_batch_info) => {
+                    let add_data_file =
+                        OwnedFile::new(data_slice_batch_info.new_file_path.as_ref().unwrap());
+                    let new_offset_interval = OffsetInterval {
+                        start: data_slice_batch_info.lower_bound.start_offset,
+                        end: data_slice_batch_info.upper_bound.end_offset,
+                    };
+
+                    let (new_data, _) = dataset
+                        .prepare_objects(Some(new_offset_interval), Some(&add_data_file), None)
```
sergiimk, line 304: Why do you need this low-level access to `prepare/commit_objects`? Why can't we use the high-level `commit_add_data` function instead?

I would really like to avoid exposing such low-level functions in `Dataset` interface so would like to understand why this was necessary.

To me it looked like the compacted dataset could be constructed using same high-level `commit_add_data` / `commit_event` primitives and the only thing we were missing was the ability to skip updating the `head` ref until the whole process is done.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,417 @@
...
+            }
+            Ordering::Greater => data_slice_batches.push(DataSliceBatch::CompactedBatch(Box::new(
+                data_slice_batch_info,
+            ))),
+            _ => (),
+        };
+
+        Ok(ChainFilesInfo {
+            data_slice_batches,
+            offset_column: offset_column.unwrap_or("offset".to_owned()),
```
sergiimk, line 212: It's very dangerous to spread such hard-coded names like `"offset"` everywhere.

The correct way to is to use `DatasetVocabulary` type which knows about all the default constants:
```
// try to find vocab event
let set_vocab: Option<SetVocab> = ...;

// create a vocabulary from the event
let dataset_vocab: DatasetVocabulary = set_vocab.unwrap_or_default().into();

// Use field
dataset_vocab.offset_column
```

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,417 @@
...
+                        if data_slice_batch_info.upper_bound.new_source_state.is_none() {
+                            data_slice_batch_info.upper_bound.new_source_state =
+                                add_data_event.new_source_state;
+                        }
+                    }
+                }
+                MetadataEvent::SetVocab(set_vocab_event) => {
+                    offset_column = set_vocab_event.offset_column;
+                }
+                _ => continue,
```
sergiimk, line 196: Better to use `_ => ()` for a no-op, as `continue` will interrupt the flow of the loop which is not needed here.

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,417 @@
...
+                    .read_parquet(
+                        data_slice_batch_info.data_slices_batch.clone(),
+                        datafusion::execution::options::ParquetReadOptions {
+                            file_extension: "",
+                            ..Default::default()
+                        },
+                    )
+                    .await
+                    .int_err()?
+                    .sort(vec![col(offset_column).sort(true, false)])
```
sergiimk, line 237: Please leave a comment above this:
```
// TODO: PERF: Consider passing sort order hint to `read_parquet` to let DF now that the data is already pre-sorted
```

rmn-boiko: Added

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Changed

src/domain/core/src/services/compact_service.rs
```diff
@@ -0,0 +1,149 @@
...
+
+pub struct NullCompactionMultiListener;
+impl CompactionMultiListener for NullCompactionMultiListener {}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CompactionPhase {
+    GatherChainInfo,
+    MergeDataslices,
+    CommitNewBlocks,
+    CleanOldFiles,
```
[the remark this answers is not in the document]
rmn-boiko: Removed

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Yeah, I was searching for `SetVocab`, but did default unwrapping in wrong way
Changed

src/infra/core/src/compact_service_impl.rs
rmn-boiko: Refactored

src/domain/core/src/services/compact_service.rs
```diff
@@ -0,0 +1,149 @@
...
+
+#[async_trait::async_trait]
+pub trait CompactService: Send + Sync {
+    async fn compact_dataset(
+        &self,
+        dataset_handle: &DatasetHandle,
+        max_slice_size: u64,
+        max_slice_records: u64,
+        listener: Option<Arc<dyn CompactionMultiListener>>,
+    ) -> Result<(), CompactError>;
```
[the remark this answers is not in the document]
rmn-boiko: Changed

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,580 @@
...
+            ),
+        )
+        .await;
+
+    let new_blocks = harness
+        .get_dataset_blocks(&root_dataset_alias.as_local_ref())
+        .await;
+
+    // We compacted two data slices and blocks into one
+    assert_eq!(old_blocks.len() - 1, new_blocks.len());
```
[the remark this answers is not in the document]
rmn-boiko: Added

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,580 @@
...
+                Some(Arc::new(NullCompactionMultiListener {}))
+            )
+            .await,
+        Err(CompactError::InvalidDatasetKind(_)),
+    );
+}
+
+#[test_group::group(ingest, datafusion, compact)]
+#[tokio::test]
+async fn test_large_dataset_compact() {
```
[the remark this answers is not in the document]
rmn-boiko: Added supporting of all events between dataslices(and covered it with tests)

src/domain/core/src/services/compact_service.rs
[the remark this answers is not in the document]
sergiimk:
```rust
pub enum CompactResult {
    Finished,
    CleanUpRequired,
}
```
This is not a very informative result. It's more confusing than anything - does `CleanUpRequired` mean that the compaction didn't finish? I don't see `CleanUpRequired` returned anywhere.

I'd like to get at least a clear indication in response whether compaction:
- was performed (including info like old_head, new_head, old_num_blocks, new_num_blocks)
- was not required - head remains the same

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,778 @@
...
+    assert_eq!(old_blocks.len(), new_blocks.len());
+    let (_, last_old_block) = old_blocks.first().unwrap();
+    let (last_new_block_hash, last_new_block) = new_blocks.first().unwrap();
+    assert!(CompactTestHarness::assert_add_data_block_events(
+        &last_old_block.event,
+        &last_new_block.event
+    ));
+
+    harness
+        .commit_set_licence_block(&root_dataset_alias.as_local_ref(), last_new_block_hash)
```
sergiimk, line 300: This test is getting very hard to read as it's checking for so many things. Can we split it into several:
- happy path - merge two blocks
- respect row/size limits
- non-data block in the middle

I would also like to see a test where blocks that are not at the head of the dataset get merged, e.g. we have a chain like:
`(data 1 small) <- (data 2 small) <- (data 3 large)`
and the compaction results in:
`(data 1,2 medium) <- (data 3 large`

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,778 @@
...
+
+    // Shoud save original amount of blocks
+    // such as size is equal to max-slice-size
+    assert_eq!(old_blocks.len(), new_blocks.len());
+    let (_, last_old_block) = old_blocks.first().unwrap();
+    let (last_new_block_hash, last_new_block) = new_blocks.first().unwrap();
+    assert!(CompactTestHarness::assert_add_data_block_events(
+        &last_old_block.event,
+        &last_new_block.event
+    ));
```
sergiimk, line 297: Name of the function is a bit confusing... How about `assert_new_block_propagates_extra_info` (by extra info meaning checkpoint, watermark, and source state?

Would also recommend adding an assertion to check the offset range.

src/domain/core/src/utils/owned_file.rs
```diff
@@ -14,7 +14,7 @@ use std::path::{Path, PathBuf};
 /// Represents a file on local FS that should either be consumed using
 /// [`OwnedFile::into_inner`()] or will be cleaned up on drop. Is typically used
 /// to commit data and checkpoints generated by the engines into a dataset.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
sergiimk, line 17: Oh no - this is super dangerous!

This struct is meant to represent ownerwhip over a file. Notice that it implements `Drop` that will delete the file if it was not consumed by `into_inner()`.

So if you make this `Clone` all hell breaks loose - it's just like if `Box<T>` was clonable.

src/infra/core/tests/tests/test_compact_service_impl.rs
```diff
@@ -0,0 +1,778 @@
...
+             | 195    | 0  | 2050-01-01T12:00:00Z | 2020-04-08T00:00:00Z | B    | 2000       |
+             | 196    | 0  | 2050-01-01T12:00:00Z | 2020-04-08T00:00:00Z | A    | 1000       |
+             | 197    | 0  | 2050-01-01T12:00:00Z | 2020-04-09T00:00:00Z | B    | 2000       |
+             | 198    | 0  | 2050-01-01T12:00:00Z | 2020-04-09T00:00:00Z | A    | 1000       |
+             | 199    | 0  | 2050-01-01T12:00:00Z | 2020-04-10T00:00:00Z | B    | 2000       |
+             +--------+----+----------------------+----------------------+------+------------+
+            "#
+            ),
+        )
+        .await;
```
sergiimk, line 565: Can you please also add a call to `VerificationService` as the last step of this test to check consistency of the resulting dataset. I think over time we will add more validations to `VerificationService` such as checking that data offset ranges match the ones declared in metadata, that `op` sequences are valid and that events are sorted correctly... So this test will benefit from all those future checks.

src/infra/core/tests/tests/test_compact_service_impl.rs
rmn-boiko: agree, changed

src/infra/core/src/compact_service_impl.rs
```diff
@@ -0,0 +1,442 @@
...
+            old_head: old_head.unwrap(),
+        })
+    }
+
+    fn append_add_data_batch_to_chain_info(
+        data_slice_batches: &mut Vec<DataSliceBatch>,
+        hash: &Option<Multihash>,
+        data_slice_batch_info: &mut DataSliceBatchInfo,
+    ) -> bool {
+        match data_slice_batch_info.data_slices_batch.len().cmp(&1) {
```
zaychenko-sergei, line 228: Perhaps, the most advanced way to compare with 1 I've ever seen :-)

src/domain/core/src/utils/owned_file.rs
rmn-boiko: Yeah, it was stupid to add it. FIxed

src/domain/core/src/services/compact_service.rs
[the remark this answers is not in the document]
rmn-boiko: I was thinking that we have two important phases(first one all stages before cleaning up and after). If we get any error during the first phase we will return the error, which says us compaction wasn't successful and we did not change the provided dataset. But if we provide the stage of cleaning old files and we get here an error we should not return an error, because we already compacted the dataset but we will notify that cleaning is still required.

But, yeah it is nice to have such params.
Added

src/infra/core/tests/tests/test_compact_service_impl.rs
rmn-boiko: test added

src/infra/core/tests/tests/test_compact_service_impl.rs
rmn-boiko: Added

sergiimk (changes requested):
I pushed a change with some tweaks - please review.

I was hoping to do some minor changes and get this branch merged ... but after looking at tests - they didn't provide me enough confidence. Adding more tests, as I feared, revealed what seems to be a bug in watermark propagation. 

Please investigate the test failure.

PR #587, opened 2024-04-09
OData: allow queries by id/offset

sergiimk:
## Description

- OData API now supports querying by collection ID/key (e.g. `account/covid.cases(123)`)
- added env var to make the default number of entries per page configurable

- [x] Unit-tests added

### DIFF
--- a/src/adapter/odata/src/handler.rs
+++ b/src/adapter/odata/src/handler.rs
@@ -18,7 +18,7 @@
 
 use std::sync::Arc;
 
-use datafusion_odata::collection::QueryParamsRaw;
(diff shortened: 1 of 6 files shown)


src/adapter/odata/src/handler.rs
```diff
@@ -108,10 +109,24 @@ pub async fn odata_metadata_handler_common(
 pub async fn odata_collection_handler_common(
     axum::extract::Extension(catalog): axum::extract::Extension<Catalog>,
     account_name: Option<AccountName>,
-    dataset_name: DatasetName,
+    collection_addr: String,
```
sergiimk, line 112: Not the most elegant solution, but need to rush this change out for a client

Issue #540, opened 2024-03-06
Introduce repository of real Account IDs


zaychenko-sergei:
1. AuthenticationRepository:
    - AccountID should be a DID string like DatasetID
    - Repository: associate account's ID and name, resolve ID by name or name by ID
    - Repository: control uniqueness of visible account IDs and names
    - Implementations:
        - Tests: in-memory
        - CLI: persistence via `.kamuconfig`
        - Node: store in database (separate ticket):
            - ID
            - Name
            - Origin (CLI, GitHub, ..)
            - Registration date
            
2. AuthenticationService: find by AccountID operation:
    - look only in authentication repository!
    
3. CLI AccountService should use real AccountIDs:
    - predefined config:
        - store ID: for demo, generate DID from name
        - register ID in Authentication repository
    - login unknown:
        - try auto-creating, if not exists, generate random DID
        - saving name -> id association in Authentication repository
        
4. Managing IDs of CLI users
    - a custom DID for AccountID could be configured via `kamu config set` at any scope
        - it should be configured separately for each user:
            - "kamu" user in single-tenant mode
            - current OS user multi-tenant mode by default
            - the target user specified in switches
    - if not specified in advance, it should be auto-generated and stored behind the scenes at application startup when active user indication is resolved 
    - predefined CLI users will have some unique ID in config file, and it would be replicated in configs on startup
        
5. OAuthGithub should interact to extract real AccountID:
    - resolve by name, if exists
    - auto-creating, if not exists, generate random DID
    
6. CurrentAccountSubject should hold real account ID for logged users - to use in various services, i.e. in flow triggers.

7. Token store (kamu login) should store AccountID instead / in addition to name

8. GraphQL API adjustments:  
    - Account.byId operation
    - Account.from_account_name
    - Account.from_dataset_alias
    - Datasets.byAccountId operation
    - Flows: InitiatorFilterInput should use AccountID

9. Remove hard-coded dummy IDs:
    - image build scripts in `kamu-cli` - set fixed DIDs for users in config instead of faked 
    - Fisheries demo repo: set DIDs for users in config instead of faked
    - Flow system - remove from triggers, use real ID
    - various kinds of tests - use in-memory repo

Note - UI impact (separate tickets):
    - Apollo InMemoryCache: user IDs for `Account` or `AccountRef`.
    - Show ID on Profile page?
    - Search by users?

sergiimk:
- `AuthenticationRepository` -> `AccountRepository`?
[...]

Issue #566, opened 2024-03-25
Add login/password auth provider


sergiimk:
~~Currently dummy auth expects password to match the account name, but in some very basic auth cases we would like to be able to customize the password.~~

- Replace dummy provider with login+password provider that stores salted/hashed passwords in the database
- Allow defining these accounts as predefined in the config (as a hash?)
- `kamu` account will become one such predefined account in our public envs where we specify its password in the config and set `isAdmin: true` flag

zaychenko-sergei:
Motivation?

sergiimk:
@zaychenko-sergei we currently don't have "Organization" accounts like in GitHub, so when we want to set up an account for a company and give multiple people admin access to some datasets we don't have a good option but to use password-based auth.

I see it as just a step from transforming our "dummy" auth into a standalone auth provider.

zaychenko-sergei:
I think we should do normal database-backed implementation then.

sergiimk:
If we can get to db-backed auth within 2-3 weeks I think that's acceptable.

PR #611, opened 2024-04-23
540 introduce account repository

zaychenko-sergei:
## Description

Closes: #540, #566

Fisheries demo update: https://github.com/kamu-data/kamu-demos-2023-08-fisheries/pull/2

UI update: https://github.com/kamu-data/kamu-web-ui/pull/308

- [x] Unit-tests added


migrations/mysql/20240315194839_accounts_table.sql
```diff
@@ -1,11 +1,27 @@
...
+    avatar_url VARCHAR(1000),
+    registered_at TIMESTAMP(6) NOT NULL,
+    is_admin TINYINT NOT NULL
 );
 
-CREATE INDEX accounts_name_idx ON accounts (account_name);
+CREATE UNIQUE INDEX idx_accounts_name ON accounts(account_name);
+CREATE UNIQUE INDEX idx_accounts_email ON accounts(email);
+
+CREATE TABLE accounts_github(
```
sergiimk, line 16: Have you considered extending `accounts` table with:
- `provider` (e.g. "github")
- `provider_id` (varchar)

as opposed to having a separate table per provider?

I think as we add other auth providers like Google, Microsoft, LDAP etc. it would suck to create tables for each. Ideally adding another provider plugin would not affect DB migrations.

resources/schema.gql
```diff
@@ -34,17 +34,17 @@ type Account {
 	isAdmin: Boolean!
 }
 
+enum AccountCategory {
```
sergiimk, line 37: Don't really like the word "category" for this - in my head category is a many-to-many relationship like "paying" "free-trial" "non-paying" users.

Why did you decide to rename it?

src/domain/accounts/domain/src/entities/account.rs
```diff
@@ -0,0 +1,87 @@
...
+use serde::{Deserialize, Serialize};
+
+///////////////////////////////////////////////////////////////////////////////
+
+// TODO: have some length restrictions (0 < .. < limit)
+pub type AccountDisplayName = String;
+
+pub const DEFAULT_ACCOUNT_NAME_STR: &str = "kamu";
+pub const DEFAULT_AVATAR_URL: &str = "https://avatars.githubusercontent.com/u/50896974?s=200&v=4";
+pub const DEFAULT_EMAIL: &str = "kamu@example.com";
```
sergiimk, line 22: I really wish we got rid of these as per our discussion about not making `kamu` account special.

We could at least move them to layers where they make sense (CLI, tests) - don't think these belong in `domain`.

src/adapter/graphql/src/queries/accounts/account.rs
```diff
@@ -41,69 +45,86 @@ impl Account {
...
+            .find_account_by_name(&account_name)
+            .await?
+            .expect("Account must exist");
+
+        Ok(Self::new(account_id.into(), account_name.into()))
     }
 
     #[graphql(skip)]
-    pub(crate) fn from_dataset_alias(ctx: &Context<'_>, alias: &odf::DatasetAlias) -> Self {
+    pub(crate) async fn from_dataset_alias(
```
sergiimk, line 88: I feel like we should rethink this function. It seems dangerous to me...

I think on GQL level we shouldn't be having some "default" accounts.

src/domain/accounts/domain/src/entities/account.rs
```diff
@@ -0,0 +1,87 @@
...
+    User,
+    Organization,
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum AccountOrigin {
+    Cli,
+    Github(AccountOriginGithub),
```
sergiimk, line 61: Is "origin" different from "provider"? Do we need another concept?

src/domain/accounts/domain/src/sqlx/account_id_sqlx.rs
```diff
@@ -0,0 +1,46 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use opendatafabric::AccountID;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AccountIdSqlx(AccountID);
```
sergiimk, line 15: Curious why do this as newtype instead of implementing sqlx decode trait directly on `AccountID`?

If you are worried about coupling `AccountID` to `sqlx` - we could always put the trait impls under a feature flag.

migrations/mysql/20240315194839_accounts_table.sql
zaychenko-sergei: What makes you think there will be always a single identifier in every provider? Do you know how account reference looks in every possible protocol?

I kind of enjoy strict typing and database-level validation, which current structure introduces.

resources/schema.gql
zaychenko-sergei: I do not like the word "type" even more. Trying to find a better term.

src/domain/accounts/domain/src/entities/account.rs
zaychenko-sergei: There is a serious dependency on the  `DEFAULT_ACCOUNT_NAME` at the level of dataset repositories and OSO authorizer. Basically, we have to put something when there is no account name in a dataset alias.

src/adapter/graphql/src/queries/accounts/account.rs
zaychenko-sergei: Similarly, suggest a different way to resolve single-tenant dataset aliases.

src/domain/accounts/domain/src/entities/account.rs
zaychenko-sergei: I like your suggestion to call this "provider".

zaychenko-sergei: Renamed everywhere

src/domain/accounts/domain/src/sqlx/account_id_sqlx.rs
zaychenko-sergei: Plz tell me more about "feature flag". What I hated previously was a need to link `sqlx` crate in every using crate, which is logically very far from database.

resources/schema.gql
s373r: what about "kind"?

zaychenko-sergei: Reverting to "type". It seems many system use this "termless" term, including Github:
https://docs.github.com/en/get-started/learning-about-github/types-of-github-accounts

src/domain/accounts/domain/src/entities/account.rs
zaychenko-sergei: I got rid of email completely (not needed much) and avatar (moved to CLI).
But id & name are integral part of single-tenant alias resolutions.

src/domain/accounts/domain/src/sqlx/account_id_sqlx.rs
sergiimk: Here's an [example](https://github.com/kamu-data/kamu-cli/blob/755ec857fb77f420069b4b3e84dd65fea396ae48/src/domain/opendatafabric/src/dtos/dtos_extra.rs#L288) that enables some methods only if `arrow` feature is turned on.

Feature is declared [here](https://github.com/kamu-data/kamu-cli/blob/755ec857fb77f420069b4b3e84dd65fea396ae48/src/domain/opendatafabric/Cargo.toml#L26) and via `dep:arrow` it pulls in `arrow` as an [optional dependency](https://github.com/kamu-data/kamu-cli/blob/755ec857fb77f420069b4b3e84dd65fea396ae48/src/domain/opendatafabric/Cargo.toml#L62).

migrations/mysql/20240315194839_accounts_table.sql
sergiimk: I'd be really surprised to see a provider with multiple identifiers, but OK, let's see how this model fares.

src/domain/accounts/domain/src/entities/account.rs
sergiimk: `kamu` as the default account itself is already problematic - we don't want to make it look like `kamu` "owns" all datasets in ST mode. This account only makes sense on our own environments. In CLI its better to use the OS username, in ST api-server - to make it configurable.

I think the problem is that on dataset repo level - single-tenant means no account at all, but in auth and GQL we decided to have accounts in both configurations and now its on them to provide defaults. But they are not the right layer to decide what default account should be...

We can deal with this later, as it existed before this PR, but I think this IS a problem and we should ticket it up as tech debt.

src/adapter/graphql/src/queries/accounts/account.rs
sergiimk: yep, it's the same problem - will resolve this and let's discuss in the thread above

src/domain/accounts/domain/src/entities/account.rs
```diff
@@ -0,0 +1,85 @@
...
+    User,
+    Organization,
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum AccountProvider {
+    Cli,
+    Github(AccountProviderGithub),
```
sergiimk, line 59: (Related to my concern with having a dedicated `github_accounts` table)

Approaching providers this way means that adding a new provider is a domain-level change. A person adding a provider will have to add a migration and add an enum variant here etc. If they are adding a non-mainstream provider for their own org - their changes cannot be upstreamed and they are stuck maintaining a complete fork.

Such lack of extensibility is a sucky experience - I just went through it with RisingWave engine.

Ideally if I need to add a custom auth provider - I should be able to link `kamu` libraries and just throw my custom provider implementation into `Catalog`. No migrations, no patching code.

We already had auth providers identified via [string IDs](https://github.com/kamu-data/kamu-cli/blob/755ec857fb77f420069b4b3e84dd65fea396ae48/src/adapter/oauth/src/oauth_github.rs#L225), and I liked this model.

src/domain/accounts/domain/src/repos/account_repository.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    async fn get_account_by_id(
+        &self,
+        account_id: &AccountID,
+    ) -> Result<Account, GetAccountByIdError>;
+
+    async fn find_account_id_by_email(
+        &self,
+        email: &str,
+    ) -> Result<Option<AccountID>, FindAccountIdByEmailError>;
```
sergiimk, line 30: With multiple providers active same email may map to multiple accounts.

src/domain/accounts/domain/src/repos/account_repository.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    async fn find_account_id_by_github_id(
+        &self,
+        github_id: i64,
+    ) -> Result<Option<AccountID>, FindAccountIdByGithubIdError>;
```
sergiimk, line 35: The more I think about this - the more I want auth providers to be fully dynamic. If we support all providers that [Jupyter supports](https://z2jh.jupyter.org/en/stable/administrator/authentication.html#oauth2-based-authentication) this alone will put us at 8+ tables. If we add wallet auth for different blockchains - that will be even more.

src/domain/accounts/domain/src/services/authentication_provider.rs
```diff
@@ -27,20 +29,15 @@ pub trait AuthenticationProvider: Sync + Send {
...
+    ) -> Result<Account, InternalError>;
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 
 #[derive(Debug, Clone)]
 pub struct ProviderLoginResponse {
     pub provider_credentials_json: String,
-    pub account_info: AccountInfo,
+    pub account: Account,
```
sergiimk, line 40: Did not expect to see `Account` here. Providers should return info about accounts in the first-party system based on which account in kamu can be created. IMO they shouldn't be deciding what identity to assign to an account, or whether it's `is_admin`.

src/domain/accounts/domain/src/services/authentication_provider.rs
```diff
@@ -8,17 +8,19 @@
...
+use super::{InvalidCredentialsError, RejectedCredentialsError};
+use crate::Account;
 
 ///////////////////////////////////////////////////////////////////////////////
 
 #[async_trait::async_trait]
 pub trait AuthenticationProvider: Sync + Send {
     fn login_method(&self) -> &'static str;
 
+    fn get_predefined_accounts(&self) -> Vec<Account>;
```
sergiimk, line 22: Should predefined accounts be a feature of auth service, not a provider? If we are defining account in config - we will never reach out to external provider about it, so what is the point of having "GH" or "LDAP" predefined account.

src/domain/accounts/services/src/authentication_service_impl.rs
```diff
@@ -145,6 +145,35 @@ impl AuthenticationServiceImpl {
             },
         )
     }
+
+    async fn ensure_predefined_accounts_registration(&self) -> Result<(), InternalError> {
```
sergiimk, line 149: Can we keep the predefined accounts in memory?

This lazy init will bite us in the ass when we run multiple api servers. Also it only creates accounts in DB - if we remove some predefined account we'll need to clean up DB manually.

src/domain/accounts/domain/src/services/authentication_provider.rs
sergiimk: Actually, most logically to me seems to make predef accounts a feature of Login+Password provider.

src/domain/accounts/services/src/authentication_service_impl.rs
sergiimk: (suggested in a different commend to make predef accounts a concern of Login+Password provider)

src/domain/accounts/domain/src/repos/account_repository.rs
zaychenko-sergei: This will fail to register. We have a "unique" index constraint on email field.

src/domain/accounts/domain/src/entities/account.rs
zaychenko-sergei: Another possibility is to externalize additional provider-specific data, without coupling with the account. It might live in separate table of it's own with it's own repositoriy and associate some provider-specific data with `AccountID`.

src/domain/accounts/domain/src/repos/account_repository.rs
zaychenko-sergei: Mentioned above. I don't see a problem in having many strictly-typed tables, especially if we decouple a bit more. We could have a generic repository (find by ID, by name, email), and repositories for provider-specific data associated with accounts.

src/domain/accounts/domain/src/services/authentication_provider.rs
zaychenko-sergei: On today the configration mechanism is CLI specific. Node has a different mechanism,

src/domain/accounts/services/src/authentication_service_impl.rs
zaychenko-sergei: They must be stored in database. A predefined account may launch a flow, and their ID must be stored as a foreign key in flow data structures. Soon we would allow access keys, those also have to be associated with an account record in the database. Relaxing database-level integrity (i.e. store `AccountID` as a string without FK constraint) to support some accounts in memory only sounds like a very bad idea to me.

migrations/mysql/20240315194839_accounts_table.sql
zaychenko-sergei: 1) LDAP needs at least 2 fields, as far as I remember: user identifier + particular server address. There are also some groups, sort of namespaces.

2) Who said it will be limited to identifers. This design allows saving any number of provider-specific fields associated with an account. We might want to save other type of GitHub properties to customize the look of UI, if needed.

src/domain/accounts/domain/src/entities/account.rs
zaychenko-sergei: > I think the problem is that on dataset repo level - single-tenant means no account at all, but in auth and GQL we decided to have accounts in both configurations and now its on them to provide defaults. But they are not the right layer to decide what default account should be...

We can't let UI decide either, neither we can have separate UIs. This affects Angular URLs design a lot.

src/domain/accounts/domain/src/sqlx/account_id_sqlx.rs
zaychenko-sergei: Looks too rusty, but works.

src/domain/accounts/domain/src/entities/account.rs
zaychenko-sergei: I think I've demonstrated my approach with separate tables looks quite neat.

src/domain/accounts/domain/src/services/authentication_provider.rs
zaychenko-sergei: Moved predefined accounts to service as suggested.

src/domain/accounts/domain/src/services/authentication_provider.rs
zaychenko-sergei: Improved. The root cause was starting in GraphQL code

sergiimk (changes requested):

src/domain/accounts/domain/src/repos/account_repository.rs
zaychenko-sergei: Shouldn't this result in briding, thus, reusing accounts from same person?

src/adapter/graphql/src/queries/accounts/account.rs
```diff
@@ -41,69 +45,86 @@ impl Account {
...
-        Self::new(AccountID::from(odf::FAKE_ACCOUNT_ID), name.into())
+    pub(crate) async fn from_account_id(
+        ctx: &Context<'_>,
+        account_id: odf::AccountID,
+    ) -> Result<Self, InternalError> {
+        let authentication_service =
+            from_catalog::<dyn kamu_accounts::AuthenticationService>(ctx).unwrap();
+
+        let account_name = authentication_service
+            .find_account_name_by_id(&account_id)
```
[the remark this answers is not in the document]
zaychenko-sergei: Improved. Agree about GraphQL part, but still need a little find name->id for the flow system. Separated those clearly.

src/domain/accounts/services/src/authentication_service_impl.rs
```diff
@@ -175,19 +233,20 @@ impl AuthenticationService for AuthenticationServiceImpl {
...
+            account_id: provider_response.account_id,
+            account_name: provider_response.account_name,
         })
     }
 
     async fn account_info_by_token(
         &self,
         access_token: String,
-    ) -> Result<AccountInfo, GetAccountInfoError> {
+    ) -> Result<Account, GetAccountInfoError> {
```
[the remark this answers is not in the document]
zaychenko-sergei: Agree this is obsolete now

src/adapter/http/src/middleware/authentication_layer.rs
```diff
@@ -53,34 +59,32 @@ impl<Svc> AuthenticationMiddleware<Svc> {
...
             // TODO: Getting the full account info here is expensive while all we need is
             // the caller identity
             match authentication_service
                 .account_info_by_token(access_token.token)
                 .await
             {
-                Ok(account_info) => Ok(CurrentAccountSubject::logged(
-                    account_info.account_name,
-                    account_info.is_admin,
+                Ok(account) => Ok(CurrentAccountSubject::logged(
```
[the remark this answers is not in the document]
zaychenko-sergei: There are 2 consumers:
 - HTTP middleware
 - UI (when refreshing page and resolving current account).
 
 UI requires a full account model from token, so we are not winning anything.
 HTTP needs a little less: id, name and "is_admin" marker.
Even if we separate these 2 consumers to different APIs, HTTP needs more data then just ID, so a database trip is required, unless all the data is included into the token. I'm okay with including ID and "is_admin" into the token, but I don't think including account's name into token is now a good idea, as we will not be able to support renaming of accounts - this would invalidate the token.

src/adapter/oauth/src/oauth_github.rs
```diff
@@ -215,108 +173,146 @@ impl OAuthGithub {
...
                 })?;
 
         // Access GitHub OAuth Api
         let github_login_response = self
             .github_login_impl(github_login_credentials.code)
             .await?;
 
+        // Convert GitHub account to Kanu account
+        let account = self
+            .get_or_create_kamu_account_from_github_account(github_login_response.account_info)
```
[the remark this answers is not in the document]
zaychenko-sergei: The counter argument is that we cannot generalize the moment when account is created, as it is provider-specific.
Providers like GitHub, in future, similar OAuth-based providers, will be auto-creating accounts on first login attempts.
Unlike this, login/password provider would rather expect to use closed set of predefined accounts (maybe allowing explicit signup action in future).

src/domain/accounts/services/src/login_password_auth_provider.rs
```diff
@@ -0,0 +1,248 @@
...
+                    .await
+                    .map_err(|e| match e {
+                        GetAccountByIdError::Internal(e) => ProviderLoginError::Internal(e),
+                        GetAccountByIdError::NotFound(_) => unreachable!(),
+                    })?
+            }
+
+            // Not found => error, unless login unknown is enabled
+            Ok(None) => {
+                if self.allow_login_unknown {
```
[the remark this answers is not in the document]
zaychenko-sergei: It's something we had at the early stages of multi-tenant repositories support.
I aggree it's obsolete and can be removed now, no longer see practical uses.

src/adapter/http/src/middleware/authentication_layer.rs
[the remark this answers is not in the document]
sergiimk: Summary of Slack discussion:
- Kamu token was already changed to carry only account ID
- We agreed NOT to add additional information into the token like account name or `is_admin`
- New token structure breaks Jupyter demo login flows, to fix it we agreed to:
  - Change `kamu system generate-token` command to only accept account ID
    - This command will NOT be used any more in demo login flow
  - Extend `kamu login` to allow authenticating with first-party provider token (passing GH access token)
    - Will require to extend REST endpoint for auth
  - Simplify demo deployment to NOT propagate `JWT_SECRET` to user pods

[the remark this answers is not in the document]
zaychenko-sergei: Implemented

src/adapter/oauth/src/oauth_github.rs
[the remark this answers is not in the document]
zaychenko-sergei: Reorganized

Issue #578, opened 2024-04-08
Flows: Integrate SQLX repository


zaychenko-sergei:
1. Analyze SQLX prototype:
   - get understanding what SQLX is and how we use it in kamu architecture
   - read through development methodology changes in the guideline
   - install database-specific development tools recommended in the guideline
   - examine architectural changes made within Task System crates to store events in a database.
   
2. Refactor flow system crates by analogy with the changes in task system crates:
   - split out domain (pure model and traits) and service layer crates 
   - in-memory repository stays in a separate crate
   - implement shared test package and thoroughly cover flow event store operation with tests, cleanup using in-memory version
   - implement Postgres version of the flows repository and stabilize it with the shared test package
   - implement SQLite version of the flows repository and stabilize it with the shared test package

sergiimk:
We don't need 100% working persistance for tasks/flows at this time. If you are using this as a way to train implementing persistent repos with Sqlx before moving on to new functionality - that's fine. But let's not get stuck implementing perfect flow persistance - it's OK for us to have flow history as transient. Consider starting with flow configurations.

PR #612, opened 2024-04-24
Flows (configuration): Integrate SQLX -- repos implementation & tests

s373r:
## Description

Related issue: https://github.com/kamu-data/kamu-cli/issues/578

- [x] Unit-tests added

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4094,7 +4094,7 @@ dependencies = [
 
 [[package]]
 name = "kamu-flow-system-postgres"
-version = "0.176.3"
+version = "0.177.0"
 dependencies = [
  "async-stream",
  "async-trait",
@@ -4116,7 +4116,7 @@ dependencies = [
 
 [[package]]
 name = "kamu-flow-system-repo-tests"
-version = "0.176.3"
+version = "0.177.0"
 dependencies = [
  "chrono",
  "dill",
@@ -4161,7 +4161,7 @@ dependencies = [
 
 [[package]]
 name = "kamu-flow-system-sqlite"
-version = "0.176.3"
+version = "0.177.0"
 dependencies = [
  "async-stream",
  "async-trait",


.gitignore
```diff
@@ -1,11 +1,10 @@
...
 /kamu-core/target
-/kamu-core-test/target
 /kamu-cli/target
 /examples/**/.kamu
 /examples/**/.ipynb_checkpoints
 /images/demo/**/.kamu
 /images/demo/**/.ipynb_checkpoints
 .env
-/kamu.sqlite.db
+/kamu.sqlite.db*
```
s373r, line 10: sometimes temp `.db-shm` and `.db-wal` are present

migrations/postgres/20240328152300_task_events_table.sql
```diff
@@ -11,4 +11,4 @@ CREATE TABLE task_events (
 );
 
 CREATE INDEX task_events_task_id_idx ON task_events (task_id);
-CREATE INDEX task_events_dataset_id_idx On task_events(dataset_id);
\ No newline at end of file
+CREATE INDEX task_events_dataset_id_idx On task_events(dataset_id);
```
s373r, line 14: Q: just to double-check, in `task_events` table, we have nullable `dataset_id` field, is that expected?

s373r (commented):
Added questions

migrations/postgres/20240422100311_flow_configuration_events.sql
```diff
@@ -0,0 +1,26 @@
+CREATE SEQUENCE flow_configuration_event_id_seq AS BIGINT;
+
+CREATE TYPE system_flow_type AS ENUM ('gc');
+
+CREATE TABLE system_flow_configuration_events
+(
+    event_id         BIGINT PRIMARY KEY DEFAULT NEXTVAL('flow_configuration_event_id_seq'),
```
zaychenko-sergei, line 7: Separate sequence/nextval are not required here, simply use BIGSERIAL type.

migrations/postgres/20240422100311_flow_configuration_events.sql
```diff
@@ -0,0 +1,26 @@
...
+    event_type       VARCHAR(50)      NOT NULL,
+    event_time       TIMESTAMPTZ      NOT NULL,
+    event_payload    JSONB            NOT NULL
+);
+
+CREATE TYPE dataset_flow_type AS ENUM ('ingest', 'execute_transform', 'hard_compacting');
+
+CREATE TABLE dataset_flow_configuration_events
+(
+    event_id          BIGINT PRIMARY KEY DEFAULT NEXTVAL('flow_configuration_event_id_seq'),
```
zaychenko-sergei, line 18: Same here, use BIGSERIAL type

migrations/postgres/20240422100311_flow_configuration_events.sql
```diff
@@ -0,0 +1,26 @@
+CREATE SEQUENCE flow_configuration_event_id_seq AS BIGINT;
+
+CREATE TYPE system_flow_type AS ENUM ('gc');
+
+CREATE TABLE system_flow_configuration_events
```
zaychenko-sergei, line 5: Might be a little easier with just 1 table, but that is not critical.

migrations/postgres/20240422100311_flow_configuration_events.sql
```diff
@@ -0,0 +1,26 @@
...
+(
+    event_id         BIGINT PRIMARY KEY DEFAULT NEXTVAL('flow_configuration_event_id_seq'),
+    system_flow_type system_flow_type NOT NULL,
+    event_type       VARCHAR(50)      NOT NULL,
+    event_time       TIMESTAMPTZ      NOT NULL,
+    event_payload    JSONB            NOT NULL
+);
+
+CREATE TYPE dataset_flow_type AS ENUM ('ingest', 'execute_transform', 'hard_compacting');
```
zaychenko-sergei, line 14: Let's think of an example (maybe put it in comment) how we would extend this list if needed. We will likely be adding new types of flows in nearest months for sure.

I remember this wasn't trivial in Postgres. If I'm not mistaken, this would require creating new type and temporary columns. Or there was a solution that needs admin permissions and accesses `pg_*` system tables.

src/domain/flow-system/domain/src/entities/shared/batching_rule.rs
```diff
@@ -8,13 +8,16 @@
...
 use thiserror::Error;
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[serde_with::serde_as]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub struct BatchingRule {
     min_records_to_await: u64,
+    #[serde_as(as = "serde_with::DurationMicroSeconds<i64>")]
```
zaychenko-sergei, line 20: Why would we need microseconds? In UI we minimally allow minutes!

src/domain/flow-system/domain/src/entities/shared/schedule.rs
```diff
@@ -26,14 +28,18 @@ pub enum Schedule {
 
 /////////////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ScheduleTimeDelta {
+    #[serde_as(as = "serde_with::DurationMicroSeconds<i64>")]
```
zaychenko-sergei, line 34: Similarly, minimal amount allowed in UI is minutes.

src/infra/flow-system/postgres/src/flow_system_event_store_postgres.rs
```diff
@@ -0,0 +1,257 @@
...
+        let connection_mut = tr.connection_mut().await?;
+
+        let result = sqlx::query!(
+            r#"
+SELECT
+    CASE
+        WHEN is_called THEN last_value
+        ELSE 0
+    END AS count
+FROM flow_configuration_event_id_seq
```
zaychenko-sergei, line 214: That is not going to work if you delete any row, or sequence counter grows within fialed transactions (note, it's like a database-level atomic number, which does not chare about app transactions).

src/infra/flow-system/postgres/src/flow_system_event_store_postgres.rs
```diff
@@ -0,0 +1,257 @@
...
+            .await
+            .int_err()?;
+
+        if let Some(last_row) = rows.last() {
+            Ok(EventID::new(last_row.event_id))
+        } else {
+            let event_id = i64::try_from(self.len().await.int_err()?).int_err()?;
+
+            Ok(EventID::new(event_id))
+        }
```
zaychenko-sergei, line 200: I don't remember why I wrote this condition in tasks doain, and I don't understand what it is doing here as wel.
In partcular, I think INSERT will always return a good event id.

src/infra/flow-system/postgres/src/flow_system_event_store_postgres.rs
```diff
@@ -0,0 +1,257 @@
...
+    async fn list_all_dataset_ids(&self) -> FailableDatasetIDStream<'_> {
+        let mut tr = self.transaction.lock().await;
+
+        Box::pin(async_stream::stream! {
+            let connection_mut = tr.connection_mut().await?;
+
+            let mut query_stream = sqlx::query!(
+                r#"
+SELECT DISTINCT dataset_id
+FROM dataset_flow_configuration_events
```
zaychenko-sergei, line 240: I think you can safely filter a certain type of event, like "created" to optimize

src/infra/flow-system/sqlite/src/flow_system_event_store_sqlite.rs
```diff
@@ -0,0 +1,300 @@
...
+            .await
+            .int_err()?;
+
+        if let Some(last_row) = rows.last() {
+            Ok(EventID::new(last_row.event_id))
+        } else {
+            let event_id = i64::try_from(self.len().await.int_err()?).int_err()?;
+
+            Ok(EventID::new(event_id))
+        }
```
zaychenko-sergei, line 246: Same here, let's experiment if we really need this condition

src/infra/flow-system/sqlite/src/flow_system_event_store_sqlite.rs
```diff
@@ -0,0 +1,300 @@
...
+        let mut tr = self.transaction.lock().await;
+
+        Box::pin(async_stream::stream! {
+            let connection_mut = tr.connection_mut().await?;
+
+            let mut query_stream = sqlx::query!(
+                r#"
+SELECT DISTINCT dataset_id
+FROM dataset_flow_configuration_events
+ORDER BY event_id
```
zaychenko-sergei, line 283: As suggested in Postgres, consider narrowing the type of event.
Also, why do we need this order by?

src/infra/task-system/postgres/src/task_system_event_store_postgres.rs
```diff
@@ -106,13 +106,13 @@ impl EventStore<TaskState> for TaskSystemEventStorePostgres {
...
             .await
-            .unwrap();
+            .int_err()?;
 
         if let Some(last_row) = rows.last() {
             Ok(EventID::new(last_row.event_id))
         } else {
             Ok(EventID::new(
-                i64::try_from(self.len().await.unwrap()).unwrap(),
+                i64::try_from(self.len().await.int_err()?).int_err()?,
```
zaychenko-sergei, line 115: `len` returns `InternalError` already

zaychenko-sergei (changes requested):

migrations/postgres/20240328152300_task_events_table.sql
zaychenko-sergei: They are only stored in "created" events.

src/infra/flow-system/sqlite/src/flow_system_event_store_sqlite.rs
s373r: Removed -- added a check at the method contract level

Done

src/infra/task-system/postgres/src/task_system_event_store_postgres.rs
s373r: Not actual after https://github.com/kamu-data/kamu-cli/pull/612#discussion_r1579417590

src/domain/flow-system/domain/src/entities/shared/batching_rule.rs
s373r: I wanted to make a maximum match to the `Duration` type (nanoseconds inside) -- but, you're right, in terms of our business logic, it's unnecessary -- chose seconds as the minimum unit available

Done

migrations/postgres/20240422100311_flow_configuration_events.sql
s373r: I have a worry that using separate counters might add a potential problem in `save_events() -> EventID` (we have two tables), as a unique ID is expected in terms of the method contract

Do you think it's a false fear?

migrations/postgres/20240422100311_flow_configuration_events.sql
s373r: https://github.com/kamu-data/kamu-cli/pull/612#discussion_r1579461941

migrations/postgres/20240422100311_flow_configuration_events.sql
s373r: In general, as far as I understand the future usage, we will have 99% dataset of events, and for the sake of 1% of system events, it is not very desirable to make optional columns

migrations/postgres/20240422100311_flow_configuration_events.sql
s373r: In general, adding isn't the problem -- more of a challenge to remove.

Added a couple comments -- Done

src/infra/flow-system/sqlite/src/flow_system_event_store_sqlite.rs
s373r: > As suggested in Postgres, consider narrowing the type of event.

Done

> Also, why do we need this order by?

Generally only did it for tests -- but you're right, it's not necessary at all. Removed

migrations/postgres/20240422100311_flow_configuration_events.sql
zaychenko-sergei: 1) The external world will be communicating with flow configurations via flow key. Events are implementation detail within the domain. Since this is fully encapsulated, I don't see any issue.

2) If we imagine a function, where we need to list mixed types of flows (i.e., admin view), then maybe it's an argument for single table, rather than a single sequence.

migrations/postgres/20240422100311_flow_configuration_events.sql
zaychenko-sergei: We don't know yet if system flows will have any configuration at all. But I can imagine a scheduled GC, once a week i.e.

src/domain/flow-system/domain/src/entities/shared/batching_rule.rs
zaychenko-sergei: The default was microseconds, right?

src/infra/flow-system/sqlite/src/flow_system_event_store_sqlite.rs
zaychenko-sergei: Tests might want to sort loaded results in memory, that's fine.

src/infra/flow-system/postgres/src/flow_system_event_store_postgres.rs
s373r: Fair enough -- I may have been idealistic, since we only have events being appended. But in the real world, opt for a more stable solution (`COUNT(*)` function) 

Done

src/domain/flow-system/domain/src/entities/shared/batching_rule.rs
s373r: No, nanoseconds:
```rust
#[stable(feature = "duration", since = "1.3.0")]
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
#[cfg_attr(not(test), rustc_diagnostic_item = "Duration")]
pub struct Duration {
    secs: u64,
    nanos: Nanoseconds, // Always 0 <= nanos < NANOS_PER_SEC
}
```

migrations/postgres/20240422100311_flow_configuration_events.sql
zaychenko-sergei: How to rename?

src/infra/flow-system/inmem/src/flow_configuration/flow_configuration_event_store_inmem.rs
```diff
@@ -94,7 +94,16 @@ impl FlowConfigurationEventStore for FlowConfigurationEventStoreInMem {
...
 
-        let dataset_ids = self.inner.as_state().lock().unwrap().dataset_ids.clone();
+        let dataset_ids: Vec<_> = self
+            .inner
+            .as_state()
+            .lock()
+            .unwrap()
+            .dataset_ids
+            .iter()
+            .rev()
```
zaychenko-sergei: From my understanding, this function returns a "set", not  a "sequence", so there is no meaningful order.
I would suggest sorting in tests directly.

migrations/postgres/20240422100311_flow_configuration_events.sql
zaychenko-sergei: Let it be

src/infra/flow-system/postgres/src/flow_system_event_store_postgres.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        let result = sqlx::query!(
+            r#"
+SELECT MAX(event_count) as count
+FROM (SELECT COUNT(event_id) as event_count
+      FROM dataset_flow_configuration_events
+      UNION ALL
+      SELECT COUNT(event_id) as event_count
+      FROM dataset_flow_configuration_events) as counts;
+"#,
```
zaychenko-sergei, line 214: I would expect to see a sum of `COUNT(event_id) from 2 tables.

migrations/postgres/20240422100311_flow_configuration_events.sql
s373r: Added notes about that -- long story short: easy w/o any tricks

Done

src/infra/flow-system/postgres/src/flow_system_event_store_postgres.rs
s373r: Fixed

migrations/postgres/20240422100311_flow_configuration_events.sql
s373r: The existing unit-test has been updated to handle this kind of mistakes

zaychenko-sergei (approved):

PR #617, opened 2024-04-26
Json format varieties

sergiimk:
## Description

Related to: https://github.com/kamu-data/kamu-node/issues/60

**WHY:**
When encoding/decoding JSON the order of properties in objects can change depending on the serde implementation. This is annoying in restricted languages like Solidity where when I make a request `select a, b from ...` I would like to be 100% sure that `records[0]` gives me `a` and `records[1]` gives me `b`. Key-based access would cost a lot of gas.

**WHAT:**
- REST data APIs and CLI commands now support different variations of JSON representation, including:
  - `Array-of-Structures` e.g. `[{"c1": 1, "c2": 2}, {"c1": 3, "c2": 4}]` (default)
  - `Structure-of-Arrays` e.g. `{"c1": [1, 3], "c2": [2, 4]}`
  - `Array-of-Arrays` e.g. `[[1, 2], [3, 4]]`
- REST data APIs now also return schema of the result (pass `?schema=false` to switch it off)

- [x] Unit-tests added


sergiimk:
Sorry for jumping the gun, but I think this change is safe and well tested, so I'll merge as I need these changes in api-server.

Issue #368, opened 2023-12-14
MQTT support


sergiimk:
We want to make as easy as possible to ingest IoT data into ODF dataset.

MQTT is a widespread protocol and allowing `kamu` to act as a message broker would simplify data transfer.

This ticket is to:
- understand MQTT broker architecture - whether Kamu has to implement a broker API
- find a realistic IoT data source or emulate one
- setup MQTT server capable of listening to push messages or poll devices for data
- document findings

Note: on this stage we are intergest in MQTT purely as transport. Things like micro-batching and deduplication of events will be handled at a different protocol-agnostic layer.

# Tasks
- https://github.com/open-data-fabric/open-data-fabric/pull/88
- https://github.com/kamu-data/kamu-cli/pull/633
- https://github.com/kamu-data/kamu-web-ui/issues/313

sergiimk:
RisingWave has an undocumented [MQTT source](https://github.com/risingwavelabs/risingwave/tree/main/src/connector/src/source/mqtt) connector using [rumqttc](https://crates.io/crates/rumqttc) library.

PR #633, opened 2024-05-12
Add initial MQTT support

sergiimk:
## Description

Relates to: #368 

This PR allows to define a dataset like this:
```yaml
kind: DatasetSnapshot
version: 1
content:
  name: temp
  kind: Root
  metadata:
    - kind: SetPollingSource
      fetch:
        kind: mqtt
        host: test.mosquitto.org
        port: 1883
        topics:
          - path: dev.kamu.example.mqtt.temp
            qos: AtMostOnce
      read:
        kind: NdJson
        ...
```
In this case `kamu-cli` will act as an MQTT **client**, subscribing to some topic(s).

Upon `kamu pull` it will connect to the MQTT broker and either:
- get nothing and exit after some configurable timeout
- get "retained" MQTT event - e.g. last temperature measurement that broker stores and sends upon subscription
- get a history of events it missed if MQTT QoS level is == `AtLeastOnce` or `ExactlyOnce` which makes broker buffer events for some time when subscriber is offline (the behavior is highly broker-specific)

Technical debt:
- MQTT breaks polling/push source separation as it's neither - related to https://github.com/kamu-data/kamu-node/issues/57 we will likely replace these metadata events with standalone `Source` manifest
- MQTT client will auto-acknowledge the events, so if ingest fails later those events will be lost - I will convert to manual ACKs in later PR
- Server / always listening mode for MQTT will be in a separate PR
- Making `kamu-node` into an MQTT broker can be done later if needed - for now it will act as a client, so user has to have their own broker and it needs to be accessible to kamu

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Workspace layout and metadata: ❗
    - Relies on extended ODF schemas: https://github.com/open-data-fabric/open-data-fabric/pull/88 but change is backwards-compatible
    
  - [x] Configuration: ✅
    
  - [x] Container images: ❗
    - Added test dependency for `rumqttd` docker image, but it's very tiny and shouldn't impact builds
- [x] Documentation:
    
    - Documentation is part of the epic and will be done later
  
- [x] Downstream effects:
    
    - More work is needed to add server mode

### DIFF
--- a/src/infra/core/Cargo.toml
+++ b/src/infra/core/Cargo.toml
@@ -50,11 +50,22 @@ serde_yaml = "0.9"
 # Ingest
 # TODO: Using curl brings a lot of overhead including compiling and linking openssl
 # We should replace it with reqwest + a separate FTP client or drop FTP support in favor of container-based ingest.
-curl = { optional = true, version = "0.4", features = ["protocol-ftp", "static-curl", "static-ssl"] }
+curl = { optional = true, version = "0.4", features = [
+    "protocol-ftp",
+    "static-curl",
+    "static-ssl",
+] }
 curl-sys = { optional = true, version = "0.4" }
-flate2 = "1"  # GZip decoder
-reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "stream", "gzip", "brotli", "deflate"] }
+flate2 = "1" # GZip decoder
+reqwest = { version = "0.11", default-features = false, features = [
+    "rustls-tls",
+    "stream",
+    "gzip",
+    "brotli",
+    "deflate",
+] }
 ringbuf = "0.3"
+rumqttc = { version = "0.23" }
 zip = "0.6"
 
 # Data
@@ -69,7 +80,7 @@ aws-sdk-s3 = { version = "0.35" }
 aws-smithy-http = { version = "0.57", features = ["rt-tokio"] }
 aws-smithy-types = { version = "0.57" }
 aws-credential-types = { version = "0.57" }
-trust-dns-resolver = "0.23"  # TODO: Needed for DNSLink resolution with IPFS
+trust-dns-resolver = "0.23"                                     # TODO: Needed for DNSLink resolution with IPFS
 http = "0.2"
 
 # Utils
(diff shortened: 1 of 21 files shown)


zaychenko-sergei (approved):

src/infra/core/Cargo.toml
```diff
@@ -69,23 +80,23 @@ aws-sdk-s3 = { version = "0.35" }
 aws-smithy-http = { version = "0.57", features = ["rt-tokio"] }
 aws-smithy-types = { version = "0.57" }
 aws-credential-types = { version = "0.57" }
-trust-dns-resolver = "0.23"  # TODO: Needed for DNSLink resolution with IPFS
+trust-dns-resolver = "0.23"                                     # TODO: Needed for DNSLink resolution with IPFS
```
s373r, line 83: minor: is this formatting necessary?

src/infra/core/src/ingest/fetch_service.rs
```diff
@@ -722,6 +729,120 @@ impl FetchService {
...
+            .await
+            .int_err()?;
+
+        let mut fetched_bytes = 0;
+        let mut fetched_records = 0;
+        let mut file = std::fs::File::create(target_path).int_err()?;
+
+        // TODO: Reading from env vars is temporary - should be replaced by vars
+        // provider
+        let max_records: u64 = std::env::var(ODF_BATCH_SIZE)
```
s373r, line 778: please add participating env variables in `.env`

s373r (approved):
A classy first step into the bigger world of small devices!

Added minor comment & question

src/infra/core/src/ingest/fetch_service.rs
sergiimk: Didn't know that we have such a file - added.

src/infra/core/Cargo.toml
sergiimk: For some reason my `better-toml` auto-formatter thinks it is... reverted

sergiimk:
Documentation PR: https://github.com/kamu-data/kamu-docs/pull/22

PR #636, opened 2024-05-16
Initial database configuration

s373r:
## Description

Related: https://github.com/kamu-data/kamu-node/issues/68

- [x] Unit and integration tests added ❌
  - Since this PR changes the loading of the components themselves, I thought it was not worth adding unit-tests for that
  
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ❗
    - Added initial db configuration
    
  - [x] Container images: ✅
- [x] Documentation:
    
    - Added: Loading database components relying on CLI config
    - Fixed: Panic when creating a workplace with an existing config
    
    - The change is more of an internal one
  
- [x] Downstream effects:
    
    - Will be added later


images/kamu-base-with-data-mt/extra/.kamuconfig
```diff
@@ -2,7 +2,6 @@ kind: CLIConfig
 version: 1
 content:
   users:
-    allowLoginUnknown: true
```
s373r, line 5: outdated key

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- Loading database components relying on CLI config
+### Fixed
+- Panic when creating a workplace with an existing config
```
zaychenko-sergei, line 11: Workspace?

src/app/cli/src/app.rs
```diff
@@ -53,13 +46,20 @@ pub async fn run(
         std::env::set_var("RUST_BACKTRACE", "1");
     }
 
-    let workspace_svc = WorkspaceService::new(Arc::new(workspace_layout.clone()));
+    // Relevant for cases where we already have a config, but no workplace yet
+    // ("kamu init" call)
```
zaychenko-sergei, line 50: Why would we need the database in this scenario?

DEVELOPER.md
```diff
@@ -57,8 +57,12 @@ Prerequisites:
...
   * To keep all these cargo tools up-to-date use `cargo install-update -a`
 * Database tools (optional, unless modifying repositories is necessary):
-  * Install Postgres command line client `psql`: `sudo apt-get install -y postgresql-client`
-  * Install MariaDB command line client `mariadb`: `sudo apt install -y mariadb-client`
+  * Install Postgres command line client `psql`:
+    * deb: `sudo apt install -y postgresql-client`
+    * rpm: `sudo dnf install -y postgresql`
+  * Install MariaDB command line client `mariadb`: 
+    * deb: `sudo apt install -y mariadb-client`
+    * rpm: `sudo dnf install -y postgresql`
```
zaychenko-sergei, line 65: Does not look MariaDB to me

src/app/cli/src/app.rs
zaychenko-sergei: The workspace might not exist and this might not be the workspace creation command
There migth be still a global configuration user's home foder, I suppose, and it can be loaded.

src/app/cli/src/services/config/models.rs
```diff
@@ -309,6 +315,55 @@ impl Default for JupyterConfig {
...
 ////////////////////////////////////////////////////////////////////////////////////////
 
+#[skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "kind")]
+pub enum DatabaseConfig {
+    Sqlite(SqliteDatabaseConfig),
+    Remote(RemoteDatabaseConfig),
```
zaychenko-sergei, line 325: ClientServer would be a bit more correct term, but this is minor.

CHANGELOG.md
s373r: Fixed

src/app/cli/src/app.rs
s373r: > Why would we need the database in this scenario?

this change does not apply to the database

---

I fixed the `kamu init --multi-tenant` command in the case where a config with predefined users is present. 

Otherwise, I was getting the following assert triggered:

```rust
pub fn register_config_in_catalog(
    config: &config::CLIConfig,
    catalog_builder: &mut CatalogBuilder,
    multi_tenant_workspace: bool,
) {
    // ...
    if multi_tenant_workspace {
        // ...
    } else {
        if let Some(users) = &config.users {
            assert!(
                users.predefined.is_empty(),
                "There cannot be predefined users in a single-tenant workspace"
            );
        }
        // ...
    }
}
```

s373r: Since the comment left room for interpretations and questions, then it is not accurate. For this reason, I have updated it with more context

DEVELOPER.md
s373r: :) 

Fixed, thanks!

src/app/cli/src/services/config/models.rs
s373r: I wasn't happy with my naming proposition through, I like yours much better -- renamed

src/app/cli/src/services/config/models.rs
```diff
@@ -309,6 +315,55 @@ impl Default for JupyterConfig {
...
+#[serde(tag = "kind")]
+pub enum DatabaseConfig {
+    Sqlite(SqliteDatabaseConfig),
+    ClientServer(RemoteDatabaseConfig),
+}
+
+impl DatabaseConfig {
+    pub fn sample() -> Self {
+        Self::ClientServer(RemoteDatabaseConfig {
+            provider: DatabaseProvider::Postgres,
```
sergiimk, line 331: Having both `kind` and `provider` seems a bit redundant. Would it be better to have `kind: Postgres` and `kind: `MySql` as `DatabaseConfig` enum variants? They all can still reuse the same config struct.

src/app/cli/src/services/config/models.rs
```diff
@@ -309,6 +315,55 @@ impl Default for JupyterConfig {
 
 ////////////////////////////////////////////////////////////////////////////////////////
 
+#[skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "kind")]
+pub enum DatabaseConfig {
+    Sqlite(SqliteDatabaseConfig),
```
sergiimk, line 324: Why is there no `InMemory` option? I'm assuming all the rest are for testing purposes only and `kamu ui` / `kamu system api-server` will continue to use in-memory storage by default? We won't be having an sqlite db as part of our default workspace, won't we?

s373r: yes, that's a very relevant comment -- the current configuration assumes that if the database section is not specified, `InMemory` will be used.

It is really worth making this more explicit, while also leaving that the absence of the option will result in `InMemory`.

src/app/cli/src/services/config/models.rs
s373r: Done

src/app/cli/src/services/config/models.rs
s373r: Done

src/app/cli/src/app.rs
```diff
@@ -359,9 +358,8 @@ fn configure_database_components(
             .int_err()?;
 
             catalog_builder.add::<kamu_accounts_mysql::MySqlAccountRepository>();
-            // TODO: Task System MySQL version
 
-            Ok(())
+            todo!("Task System MySQL version");
```
s373r, line 362: please pay attention

zaychenko-sergei, line 362: flow system too

zaychenko-sergei (approved):

src/app/cli/src/app.rs
s373r: Fixed

Issue #630, opened 2024-05-09
Compaction flow: implement `keep_metadata_only` flag


rmn-boiko:
In scope of https://github.com/kamu-data/kamu-cli/issues/625 was added new `--keep-metadata-only` flag to `kamu system compact` command. 

Add this value to Compaction flow GraphQL API.
Requirements:

- It should be an optional `keep_metadata_only` parameter in manual trigger of `HARD_COMPACTING` flow.
- Extend compaction config with field `keep_metadata_only` by default `false`. Each flow on finished event should also store their config. 
    Store config snapshot in `FlowEventInitiated` event and extend `FlowState` with config field to be able to return it in 
    GQL response
- Compaction with this parameter should work for both root and derived dataset kinds.
    Such as for now we have validation where we check dataset kind for flow to trigger on the GraphQL level we may 
    extend this validation. If keep_metadata_only is `false` only root datasets can be compacted else if true both root and 
    derived path this check. 
- It should trigger the same flow for each downstream datasets, but only those that user owns / has permissions to reset
    This logic should be incapsulated in separate flow plugin where we will check permissions and required actions(we may 
    also move here logic with fetching such dependencies).
- Extend compaction config with new field `keep_metadata_only`, by default `false`
- When compaction was triggered manually save `keep_metadata_only` to config and reuse it(if no new value were provided) for the next compactions for this dataset after.

Test plan

- Extend GQL flow lost tests with new configuration snapshots.
- Add new tests for trigerring compaction with  `keep_metadata_only` and for setting new config value.
- Add tests to check triggering of all downstream dependencies.

PR #637, opened 2024-05-16
Add keep metadata only option to flow

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/630

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    

resources/schema.gql
```diff
@@ -192,6 +192,7 @@ type CommitResultSuccess implements CommitResult & UpdateReadmeResult {
 input CompactingConditionInput {
 	maxSliceSize: Int!
 	maxSliceRecords: Int!
+	isKeepMetadataOnly: Boolean!
```
zaychenko-sergei, line 195: why do we need this "is" prefix? it looks really weird

src/adapter/graphql/src/queries/flows/flow_config_snapshot.rs
```diff
@@ -0,0 +1,36 @@
...
+use kamu_flow_system::{ConfigSnapshot, Schedule};
+
+use crate::prelude::*;
+
+#[derive(SimpleObject)]
+pub struct FlowConfigurationSnapshot {
+    batching_rule: Option<FlowConfigurationBatching>,
+    schedule: Option<FlowConfigurationSchedule>,
+    compacting_rule: Option<FlowConfigurationCompacting>,
+}
```
zaychenko-sergei, line 19: Can it ever have more than 1 of these fields set to Some(_)? Looks more like an enum to me.
I think a case when all 3 are None is also impossible.

src/domain/flow-system/services/src/flow/active_configs_state.rs
```diff
@@ -101,6 +101,46 @@ impl ActiveConfigsState {
             .get(BorrowedFlowKeyDataset::new(dataset_id, flow_type).as_trait())
             .copied()
     }
+
+    pub fn get_merged_config_snapshot(
```
zaychenko-sergei, line 105: There is no merging happening here, just overrides.

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl FlowPermissionsPlugin for FlowPermissionsPluginImpl {
+    async fn get_account_downstream_dependencies(
+        &self,
+        dataset_id: &DatasetID,
+        account_id_maybe: Option<AccountID>,
+    ) -> Result<Vec<DatasetID>, InternalError> {
```
zaychenko-sergei, line 48: This thing is neither specifict to flow system in trait definition, nor in implementation.
I think with some more generic name, it should live in the core.

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
```diff
@@ -0,0 +1,83 @@
...
+            dataset_repo,
+            authentication_svc,
+            dependency_graph_service,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl FlowPermissionsPlugin for FlowPermissionsPluginImpl {
+    async fn get_account_downstream_dependencies(
```
zaychenko-sergei, line 44: Emphasize "ownership"

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -271,27 +273,31 @@ impl FlowServiceImpl {
                 trigger_time: start_time,
             }),
             FlowTriggerContext::Unconditional,
+            ConfigSnapshot::default(),
         )
         .await
     }
 
-    #[tracing::instrument(level = "trace", skip_all, fields(?flow_key, %flow_id))]
+    #[tracing::instrument(level = "trace", skip_all, fields(?flow))]
```
zaychenko-sergei, line 281: Full flow could be quite large. Try extracting fields

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -271,27 +273,31 @@ impl FlowServiceImpl {
...
+        flow: &Flow,
         flow_result: &FlowResult,
     ) -> Result<(), InternalError> {
-        if let FlowKey::Dataset(fk_dataset) = flow_key {
+        if let FlowKey::Dataset(fk_dataset) = &flow.flow_key {
             // Extract a list of downstream 1 level datasets
+            let flow_manual_trigger_initiator = flow
+                .try_get_manual_trigger()
+                .map(|trigger| trigger.initiator_account_id);
+
```
zaychenko-sergei, line 293: Why only manual trigger? This isn't a general rule for all possible flow types

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -271,27 +273,31 @@ impl FlowServiceImpl {
...
-                .get_downstream_dependencies(&fk_dataset.dataset_id)
-                .await
-                .int_err()?
-                .collect()
-                .await;
+                .permissions_plugin
+                .get_account_downstream_dependencies(
+                    &fk_dataset.dataset_id,
+                    flow_manual_trigger_initiator,
+                )
```
zaychenko-sergei, line 299: Note that this requirement to trigger only own accounts is specific for compacting, but is not the same for update flows - an update of root should trigger all derived, regardless of who owns them

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -308,24 +314,42 @@ impl FlowServiceImpl {
...
+                            .unwrap()
+                            .is_keep_metadata_only()
+                    {
+                        config_snapshot = flow.config_snapshot.clone();
+                        FlowKeyDataset::new(
+                            dependent_dataset_id.clone(),
+                            DatasetFlowType::HardCompacting,
+                        )
+                        .into()
+                    } else {
```
zaychenko-sergei, line 332: Batching rule was a good condition for updates, if we enabled automatic transforms.
I'm not seeing the same requirement for compacting case - the derived have to be reset regardless of whether they have an automatic update configured. 

This needs some review of responsibilities - what is checked by Flows System, and what is programmed for each different flow type.

src/domain/flow-system/domain/src/entities/shared/flow_run_rule.rs
```diff
@@ -0,0 +1,19 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use crate::*;
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct ConfigSnapshot {
```
zaychenko-sergei, line 15: Confusing name. `FlowConfigSnapshot` maybe. Dima is doing very different kind of configs, i.e.

src/domain/flow-system/domain/src/entities/shared/flow_run_rule.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub batching_rule: Option<BatchingRule>,
+    pub compacting_rule: Option<CompactingRule>,
+    pub schedule: Option<Schedule>,
```
zaychenko-sergei, line 18: Again, this should be an enum in my opinion

src/domain/flow-system/services/src/flow/active_configs_state.rs
zaychenko-sergei: The method is a bit unclear. Maybe we should return the configuration snapshot from the stored settings, rather than combining it with unrelated argument snapshot? Combining should better live in the service.

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
```diff
@@ -0,0 +1,83 @@
...
+        Self {
+            dataset_repo,
+            authentication_svc,
+            dependency_graph_service,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl FlowPermissionsPlugin for FlowPermissionsPluginImpl {
```
zaychenko-sergei, line 43: This isnt about permissions. We should decide on the affected dependent dataset, and it also depends on the flow type, which is not passed here.

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
```diff
@@ -0,0 +1,83 @@
...
+    async fn get_account_downstream_dependencies(
+        &self,
+        dataset_id: &DatasetID,
+        account_id_maybe: Option<AccountID>,
```
zaychenko-sergei, line 47: Account exists always, why Option?

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
```diff
@@ -0,0 +1,83 @@
...
+                    .dataset_repo
+                    .get_datasets_by_owner(&account_name)
+                    .map_ok(|dataset_handle| dataset_handle.id)
+                    .try_collect()
+                    .await
+                    .int_err()?;
+
+                let filtered_datasets: Vec<_> = dependent_dataset_ids
+                    .iter()
+                    .filter(|dataset_id| account_dataset_ids.contains(dataset_id))
```
zaychenko-sergei, line 75: N^2!!!

zaychenko-sergei: At least work with hashsets, and not vectors.
But it's better not to extract all account datasets, this could be very ineffecient.
Instead, can't we simply ask the core if the dependent dataset is owned by this account?

Also, consider if we can improve the graph of dependencies to allow filtering by owners there? Each dataset node might be associated with an ownin account.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -181,6 +180,12 @@ impl FlowServiceImpl {
     ) -> Result<(), InternalError> {
         match &flow_key {
             FlowKey::Dataset(dataset_flow_key) => {
+                self.state
+                    .lock()
+                    .unwrap()
+                    .active_configs
+                    .add_dataset_flow_config(dataset_flow_key, rule.clone());
+
```
zaychenko-sergei, line 188: This object's purpose was to work with automatic triggers only (keep only configs of enabled flow keys). You are giving it a different sense of a setting cache. But compacting is always manuall, and you can request the current config there instead of this.

src/adapter/graphql/src/queries/flows/flow_config_snapshot.rs
rmn-boiko: Basically for us important to know only one of these configs first my realization was with enum(similar to input for mutation) but I thought that it is possible that in the future we may have some combination of different rules per flow. But I'm totally fine to put here enum for now

src/domain/flow-system/services/src/flow/active_configs_state.rs
rmn-boiko: Changed

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: Changed

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: Changed

src/domain/flow-system/domain/src/entities/shared/flow_run_rule.rs
rmn-boiko: Changed

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
rmn-boiko: Modified

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
rmn-boiko: Changed for now
Will left ToDo and create issue with improvements

src/domain/flow-system/services/src/flow/flow_permissions_plugin_impl.rs
rmn-boiko: Changed method completely

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: I just move it to save configuration before event initialisation, due to now we are storing this snapshot in the Initiated event

src/adapter/graphql/src/mutations/flows_mut/dataset_flow_runs_mut.rs
```diff
@@ -168,3 +180,18 @@ impl CancelScheduledTasksSuccess {
...
+#[derive(SimpleObject)]
+#[graphql(complex)]
+pub struct FlowInvalidRunConfigurations {
+    pub error: String,
+}
+
+#[ComplexObject]
+impl FlowInvalidRunConfigurations {
+    pub async fn message(&self) -> String {
+        format!("Invalid flow provided: '{}'", self.error)
```
zaychenko-sergei, line 193: Maybe missing word "configuration"?

src/adapter/graphql/src/scalars/mod.rs
```diff
@@ -33,7 +33,7 @@ pub(crate) use dataset_endpoints::*;
 pub(crate) use dataset_id_name::*;
 pub(crate) use engine_desc::*;
 pub(crate) use event_id::*;
-pub(crate) use flow_configuration::*;
+pub use flow_configuration::*;
```
zaychenko-sergei, line 36: Seems this is not needed

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -737,7 +730,11 @@ impl FlowServiceImpl {
     }
 
     /// Creates task logical plan that corresponds to template
-    pub fn make_task_logical_plan(&self, flow_key: &FlowKey) -> LogicalPlan {
+    pub fn make_task_logical_plan(
+        &self,
+        flow_key: &FlowKey,
+        config_snapshot: &Option<FlowConfigurationSnapshot>,
```
zaychenko-sergei, line 736: Use `Option<&FlowConfigurationSnapshot>` instead

src/infra/core/src/dataset_ownership_service_inmem.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+#[component(pub)]
+#[interface(dyn DatasetOwnershipService)]
+#[interface(dyn AsyncEventHandler<DatasetEventCreated>)]
+#[interface(dyn AsyncEventHandler<DatasetEventDeleted>)]
+#[scope(Singleton)]
+impl DatasetOwnershipServiceInMemory {
```
zaychenko-sergei, line 44: Plz write some tests for this new service

zaychenko-sergei (approved):

PR #650, opened 2024-05-22
Persistence accounts: stabilize database components

s373r:
:warning: Changes from this PR are partially extracted into:
- https://github.com/kamu-data/kamu-cli/pull/660

---

## Description

### Fixed
- Stabilized startup using connection to databases
- Fixed potential crash when attempting to rollback a transaction if the connection fails to establish

- [x] Unit and integration tests added, ❌
  - No, but in a good way we need to cover CLI commands with unit-tests
  
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
  - [x] [Public documentation](https://github.com/kamu-data/kamu-docs/): ❌, not needed
  
- [x] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ❌, will be added later


src/app/cli/src/app.rs
```diff
@@ -337,6 +337,9 @@ fn configure_database_components(
     catalog_builder: &mut CatalogBuilder,
     db_configuration: &DatabaseConfiguration,
 ) -> Result<(), InternalError> {
+    // TODO: Remove after adding implementation of FlowEventStore for databases
+    catalog_builder.add::<kamu_flow_system_inmem::FlowEventStoreInMem>();
```
s373r, line 341: At first glance, we don't need this for persistence storage of accounts, but in general, it doesn't seem difficult to implement right now

src/app/cli/src/commands/system_api_server_run_command.rs
```diff
@@ -117,43 +115,52 @@ impl Command for APIServerRunCommand {
     async fn run(&mut self) -> Result<(), CLIError> {
         self.check_required_env_vars()?;
 
-        let access_token = self.get_access_token().await?;
-
-        // TODO: Cloning catalog is too expensive currently
```
s373r, line 122: at the moment, it's not actual -- `Catalog` is `Arc<State>`, so this comment removed

s373r (commented):
Added notes to reviewers

src/app/cli/src/commands/system_api_server_run_command.rs
```diff
@@ -117,43 +115,52 @@ impl Command for APIServerRunCommand {
...
+            move |[base_catalog, cli_catalog]| async move {
+                let access_token = self_ref.get_access_token(&base_catalog).await?;
+
+                let api_server = crate::explore::APIServer::new(
+                    base_catalog,
+                    &cli_catalog,
+                    self_ref.multi_tenant_workspace,
+                    self_ref.address,
+                    self_ref.port,
+                );
```
zaychenko-sergei, line 132: This would create a single transaction for the entire run of the API server, so we will never save anything to the database, as API server never stops.

src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -34,14 +34,26 @@ pub trait DatabaseTransactionManager: Send + Sync {
...
-    base_catalog: &Catalog,
+/// Enriching passed catalogs with [`TransactionRef`]
+///
+/// # Arguments
+///
+/// * `catalogs` - Catalogs in which `base_catalog` is expected to be the first
+///   one
+/// * `callback` - A callback that receives updated catalogs. If necessary, we
+///   can return a value from it
+pub async fn run_transactional<H, HFut, HFutResultT, const N: usize>(
```
zaychenko-sergei, line 45: I totally don't get the idea what you are trying to do here, and why do we need to send multiple catalog ???

zaychenko-sergei (changes requested):
The approach seems to be completely wrong - we cannot have a "global" transaction over the entire run of API/Web-UI server.

src/app/cli/src/app.rs
zaychenko-sergei: What is not going to be diffcult? A repo for flow system would be quite hard. There are many complex queries to implement.

src/app/cli/src/commands/system_api_server_run_command.rs
s373r: yes, I realized my mistake

src/app/cli/src/commands/ui_command.rs
```diff
@@ -61,49 +63,56 @@ impl UICommand {
...
-        if self.get_token {
-            tracing::info!(
-                token = %web_server.get_access_token(),
-                "Issued API server token"
-            );
-        }
+        // The borrow checker is happy if self is not explicitly used
+        let self_ref = &self;
+
+        run_transactional([&self.base_catalog], move |[base_catalog]| async move {
```
sergiimk, line 69: How is this passing compilation? I don't see `run_transactional` accepting an array...

zaychenko-sergei: Yes, this won't compile, and I guess, @s373r hasn't tried compiling in Web UI mode.

src/app/cli/src/explore/api_server.rs
```diff
@@ -95,6 +96,7 @@ impl APIServer {
                     )
                     .layer(Extension(base_catalog))
                     .layer(Extension(gql_schema))
+                    .layer(kamu_adapter_http::RunInDatabaseTransactionLayer::new())
```
zaychenko-sergei, line 99: It's probably okay for the start, but I think we would want more specific per-route application. Some HTTP routes likely do not not require database access.

zaychenko-sergei (commented):
We are missing some kind of a test that proves the transactions are emitted and isolating parallel changes

src/app/cli/src/commands/ui_command.rs
s373r: It was working, until I didn't get https://github.com/kamu-data/kamu-cli/pull/650#discussion_r1612147580.

I certainly tried compiling, before requesting a review for you -- the issue here is different in that after requesting the changes, I started making them, and didn't mark the PR status to draft.

Right now PR is in WIP state, so some things may not compile

src/app/cli/src/explore/api_server.rs
s373r: Yes, I agree with you -- but it's not yet a finished PR

src/adapter/http/src/middleware/authentication_layer.rs
```diff
@@ -115,7 +115,7 @@ where
     }
 
     fn call(&mut self, mut request: http::Request<Body>) -> Self::Future {
-        // Inspired by https://github.com/maxcountryman/axum-login/blob/main/axum-login/src/auth.rs
+        // Inspired by https://github.com/maxcountryman/axum-login/blob/5239b38b2698a3db3f92075b6ad430aea79c215a/axum-login/src/auth.rs
```
s373r, line 118: `axum-login` has been rewritten, so I've updated the commit links from the past

src/domain/flow-system/domain/src/services/flow/flow_service.rs
```diff
@@ -53,10 +53,10 @@ pub trait FlowService: Sync + Send {
     /// Returns initiators of flows associated with a given dataset
     /// ordered by creation time from newest to oldest.
     /// Applies specified filters/pagination
-    async fn list_all_flow_initiators_by_dataset(
+    async fn get_all_flow_initiators_by_dataset(
         &self,
         dataset_id: &DatasetID,
-    ) -> Result<FlowInitiatorListing, ListFlowsByDatasetError>;
+    ) -> Result<Vec<AccountID>, ListFlowsByDatasetError>;
```
s373r, line 59: The method signature was changed with the need to get away from `'a`.

I haven't found a way to do borrow from a temporarily existing transactional catalog. Previously the borrowing was taken from the `FlowServiceImpl::flow_event_store` field, but now the store is always retrieved dynamically.

TODO 
- [ ] Add a link to usage

s373r: CC @rmn-boiko

src/domain/task-system/services/src/task_executor_impl.rs
```diff
@@ -21,32 +22,29 @@ use kamu_core::{
 };
 use kamu_task_system::*;
 
-/////////////////////////////////////////////////////////////////////////////////////////
```
s373r, line 24: for some strange reason, we sometimes have these lines that are 89 characters long

src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -102,8 +127,9 @@ impl TransactionRef {
     pub fn into_maybe_transaction<DB: sqlx::Database>(
         self,
     ) -> Option<sqlx::Transaction<'static, DB>> {
-        let m = Arc::try_unwrap(self.inner).unwrap().into_inner();
```
s373r, line 105:
```rust
let m = Arc::try_unwrap(self.inner).unwrap().into_inner();
//                                  ^^^^^^^^
``` 

With a missing connection, was getting a crash here -- updated to see a readable error in cli logs

s373r (commented):
added notes for readers

src/app/cli/src/commands/ui_command.rs
s373r: Done (and double-checked)

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -1036,15 +1088,22 @@ impl FlowService for FlowServiceImpl {
...
+            |event_store: Arc<dyn FlowEventStore>| async move {
+                event_store
+                    .get_unique_flow_initiator_ids_by_dataset(dataset_id)
+                    .try_collect()
+                    .await
+            },
+        )
+        .await?;
+
+        Ok(initiator_ids)
```
s373r, line 1106: Example of use `FlowService::get_unique_flow_initiator_ids_by_dataset()`
(cross-linking comment)

src/domain/flow-system/domain/src/services/flow/flow_service.rs
s373r: Usage: https://github.com/kamu-data/kamu-cli/pull/650/files#r1619221031

src/domain/task-system/services/src/task_scheduler_impl.rs
```diff
@@ -101,26 +130,43 @@ impl TaskScheduler for TaskSchedulerImpl {
...
-            .await?;
+        let total_count = <DatabaseTransactionRunner>::run_transactional_with(
+            &self.catalog,
+            |event_store: Arc<dyn TaskSystemEventStore>| async move {
+                let total_count = event_store.get_count_tasks_by_dataset(dataset_id).await?;
 
-        let dataset_id = dataset_id.clone();
+                Ok(total_count)
+            },
+        )
```
zaychenko-sergei, line 140: I think this query might be obsolete.

src/domain/task-system/services/src/task_scheduler_impl.rs
```diff
@@ -101,26 +130,43 @@ impl TaskScheduler for TaskSchedulerImpl {
...
+                let task = DatabaseTransactionRunner::run_transactional_with(
+                    &self.catalog,
+                    |event_store: Arc<dyn TaskSystemEventStore>| async move {
+                        Task::load(task_id, event_store.as_ref()).await
+                    },
+                )
+                .await
+                .int_err()?;
+
                 yield Ok(task.into());
```
zaychenko-sergei, line 170: This looks very suspicious. Too many transactions just to iterate a list, and too many queries.
A simple possibility could be to load all tasks in one transactoins. This is fine if we are talking about few dozens.
But there might be more, so maybe we need chunks - load a chunk of tasks in 1 trip to the database, yield those while available in temporary collection, then proceed to the next chunk.

src/domain/task-system/services/src/task_executor_impl.rs
```diff
@@ -77,20 +75,25 @@ impl TaskExecutorImpl {
...
-                .int_err()?;
+            let mut task = DatabaseTransactionRunner::run_transactional_with(
+                &self.catalog,
+                |event_store: Arc<dyn TaskSystemEventStore>| async move {
+                    let task_id = self.task_sched.take().await.int_err()?;
+
+                    Task::load(task_id, event_store.as_ref()).await.int_err()
+                },
+            )
+            .await?;
```
zaychenko-sergei, line 93: `take` in scheduler is already using it's own transaction

src/domain/task-system/services/src/task_executor_impl.rs
```diff
@@ -167,21 +170,30 @@ impl TaskExecutor for TaskExecutorImpl {
...
-            task.update(self.event_store.as_ref()).await.int_err()?;
-            task.finish(self.time_source.now(), outcome.clone())
-                .int_err()?;
-            task.save(self.event_store.as_ref()).await.int_err()?;
+            let cloned_outcome = outcome.clone();
+            let task = DatabaseTransactionRunner::run_transactional_with(
+                &self.catalog,
+                |event_store: Arc<dyn TaskSystemEventStore>| async move {
+                    // Refresh the task in case it was updated concurrently (e.g. late cancellation)
+                    task.update(event_store.as_ref()).await.int_err()?;
```
zaychenko-sergei, line 184: What would happen in case of unexpected concurrent modification?

src/domain/task-system/services/src/task_scheduler_impl.rs
```diff
@@ -76,20 +80,45 @@ impl TaskScheduler for TaskSchedulerImpl {
...
+        let task = DatabaseTransactionRunner::run_transactional_with(
+            &self.catalog,
+            |event_store: Arc<dyn TaskSystemEventStore>| async move {
+                Task::load(task_id, event_store.as_ref()).await
+            },
+        )
+        .await?;
+
         Ok(task.into())
     }
```
zaychenko-sergei, line 92: We likely need a batch version of this.

src/utils/database-common/src/testing/fake_database_transaction_manager.rs
```diff
@@ -0,0 +1,56 @@
...
+// by the Apache License, Version 2.0.
+
+use internal_error::InternalError;
+use sqlx::SqlitePool;
+
+use crate::{DatabaseTransactionManager, TransactionRef};
+
+///////////////////////////////////////////////////////////////////////////////
+
+pub struct FakeDatabaseTransactionManager {
```
zaychenko-sergei, line 17: Why not to use `dill::Component` ?

src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -34,49 +34,74 @@ pub trait DatabaseTransactionManager: Send + Sync {
...
+        HFut: std::future::Future<Output = Result<HFutResultT, HFutResultE>> + Send,
+        HFutResultE: From<InternalError>,
+    {
+        Self::run_transactional(base_catalog, |updated_catalog| async move {
+            let event_store = updated_catalog.get_one::<Store>().int_err()?;
+
+            callback(event_store).await
+        })
+        .await
+    }
```
zaychenko-sergei, line 104: The original idea was to give the callback the entire catalog, as it might need a lot more than just 1 repository interface in a more complex scenario. This mightly be slightly simplifying a specific type of case.

src/domain/flow-system/services/src/flow_configuration/flow_configuration_service_impl.rs
```diff
@@ -34,17 +35,30 @@ pub struct FlowConfigurationServiceImpl {
...
+        flow_key: FlowKey,
+    ) -> Result<Option<FlowConfiguration>, TryLoadError<FlowConfigurationState>> {
+        DatabaseTransactionRunner::run_transactional_with(
+            &self.catalog,
+            |event_store: Arc<dyn FlowConfigurationEventStore>| async move {
+                FlowConfiguration::try_load(flow_key, event_store.as_ref()).await
+            },
+        )
+        .await
+    }
```
zaychenko-sergei, line 60: Almost all operations of `FlowConfigurationService` are called only from GraphQL, except `list_enabled_configurations`. Those queries already run in a transaction created by HTTP middleware.

I think a service like `FlowConfigurationService` should not be concerned of transactions management at all. I think transactions should be controlled by 2 types of places:
 - middleware for HTTP queries
 - actively running components with their own control loop (`TaskExecutor`, `FlowService`)
 
 
 ```
 
     pub async fn run(self) -> Result<(), InternalError> {
        tokio::select! {
            res = self.server => { res.int_err() },
            res = self.task_executor.run() => { res.int_err() },
            res = self.flow_service.run(self.time_source.now()) => { res.int_err() }
        }
    }
    
    ```

src/domain/task-system/services/src/task_scheduler_impl.rs
```diff
@@ -52,12 +50,18 @@ impl TaskSchedulerImpl {
 impl TaskScheduler for TaskSchedulerImpl {
     #[tracing::instrument(level = "info", skip_all, fields(?logical_plan))]
     async fn create_task(&self, logical_plan: LogicalPlan) -> Result<TaskState, CreateTaskError> {
-        let mut task = Task::new(
-            self.time_source.now(),
-            self.event_store.new_task_id().await?,
-            logical_plan,
-        );
-        task.save(self.event_store.as_ref()).await.int_err()?;
+        let task = <DatabaseTransactionRunner>::run_transactional_with(
```
zaychenko-sergei, line 53: See my concerns above, I don't think `TaskScheduler` is a right kind of component to control transactions.

src/domain/flow-system/services/src/flow_configuration/flow_configuration_service_impl.rs
zaychenko-sergei: `list_enabled_configurations` is only used at startup time in `FlowService`, which should be responsible for transaction management, as an active component.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -101,6 +102,44 @@ impl FlowServiceImpl {
...
+        DatabaseTransactionRunner::run_transactional_with(
+            &self.catalog,
+            |event_store: Arc<dyn FlowEventStore>| async move {
+                flow.save(event_store.as_ref()).await.int_err()?;
+
+                Ok(())
+            },
+        )
+        .await
+    }
```
zaychenko-sergei, line 125: I think we should control transactions in `FlowService` at a bit higher level.
1) Definitely a transaction on startup, which has to load initial configurations and do the planning.
2) 1 transaction per 1 call of `run_current_timeslot`. Scheduling is not really an expensive operation, we are loading a few flows and inserting tasks into a queue.
3) 1 transaction per event in `AsyncEventHandler`s. These are coming independently from `TaskExecutor`. Make sure events are only sent when task executor's own transaction is not open.

Double check if we have any long-running computations here in between, but I don't think so, as `FlowService` mostly coordinates execution.

Queries like `list_all_flows_by_dataset` are already runnning behind a transaction in HTTP middleware.

src/adapter/http/src/middleware/authentication_layer.rs
zaychenko-sergei: We might not care about this comment anyway :-)

src/utils/database-common/src/testing/fake_database_transaction_manager.rs
s373r: This requires registering `SqlitePool` as well -- I thought it best to leave knowledge of `SqlitePool` only at the `FakeDatabaseTransactionManager` level, and not take it higher to the `FakeDatabasePlugin` level

s373r: I suggest we leave it as it is

src/domain/flow-system/services/src/flow_configuration/flow_configuration_service_impl.rs
s373r: > I think a service like `FlowConfigurationService` should not be concerned of transactions management at all

Agree, done -- except for `AsyncEventHandler<DatasetEventDeleted>`

src/domain/task-system/services/src/task_executor_impl.rs
```diff
@@ -77,20 +72,32 @@ impl TaskExecutorImpl {
...
 #[async_trait::async_trait]
 impl TaskExecutor for TaskExecutorImpl {
     // TODO: Error and panic handling strategy
     async fn run(&self) -> Result<(), InternalError> {
         loop {
-            let task_id = self.task_sched.take().await.int_err()?;
-            let mut task = Task::load(task_id, self.event_store.as_ref())
-                .await
-                .int_err()?;
+            let mut task = DatabaseTransactionRunner::run_transactional(
```
sergiimk, line 82: Currently you are opening and closing transaction when loading a task and when saving a task, but what if task execution itself requires DB access?

I would expect the `run` loop to look more like this:
```rust
async fn run(&self) {
  loop {
    // Scheduler is a special **singleton** service - it cannot inject DB transactions and
    // thus should manage transactions on its own.
    let task_id = self.scheduler.take();

    run_transactional(|...| {
      let event_store = ...;
      let task = Task::load(event_store);

      // Run the task under the transaction
      Self::execute_task(task);

      // Store task state at the end of the transaction
      task.update(...);
      task.finish(...);
      task.save(event_store);
    });
  }
}
```
This way, assuming all task actions are transactional - we would get full atomicity.

sergiimk: The only tricky point I see are `publish_task_running` /  `publish_task_finished`. Our event bus is not yet transactional, so having them inside `run_transactional` would work. But once we make event bus use DB and transaction we will want a task to be marked "running" to become visible to all other users - thus be committed.

Let's ignore this error for now, and in future perhaps we will make `publish_task_running` a part of `scheduler.take()` - so that the task is marked "running" at the same time as it's associated with the specific executor.

src/domain/task-system/services/src/task_scheduler_impl.rs
```diff
@@ -52,11 +52,9 @@ impl TaskSchedulerImpl {
 impl TaskScheduler for TaskSchedulerImpl {
     #[tracing::instrument(level = "info", skip_all, fields(?logical_plan))]
     async fn create_task(&self, logical_plan: LogicalPlan) -> Result<TaskState, CreateTaskError> {
-        let mut task = Task::new(
-            self.time_source.now(),
-            self.event_store.new_task_id().await?,
-            logical_plan,
-        );
+        let new_task_id = self.event_store.new_task_id().await?;
+        let mut task = Task::new(self.time_source.now(), new_task_id, logical_plan);
```
sergiimk, line 56: `TaskSchedulerImpl` is a singleton - it is created once and lives until program terminates. It's made a singleton to maintain an in-memory task queue.

Because it's a singleton - all references it holds, e.g. `event_store`, will live forever too, but this won't work well if store needs a DB transaction.

I think since Scheduler is a "top level" component it should manage transactions manually and create event store in transactional scope when it's needed.

src/infra/accounts/mysql/tests/repos/test_mysql_account_repository.rs
```diff
@@ -7,67 +7,71 @@
...
-/////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(database, mysql)]
 #[test_log::test(sqlx::test(migrations = "../../../../migrations/mysql"))]
 async fn test_missing_account_not_found(mysql_pool: MySqlPool) {
     let harness = MySqlAccountRepositoryHarness::new(mysql_pool);
 
-    run_transactional(&harness.catalog, |catalog: Catalog| async move {
+    <DatabaseTransactionRunner>::run_transactional(&harness.catalog, |catalog| async move {
```
sergiimk, line 22: Why in some places we call it WITH and WITHOUT brackets? Are you sure these brackets are not an artifact of your IDE, as I can't see why they would be required?

sergiimk (changes requested):
Please poke me for final review - I'd like to be involved in this PR.

src/infra/accounts/mysql/tests/repos/test_mysql_account_repository.rs
s373r: > Are you sure these brackets are not an artifact of your IDE, as I can't see why they would be required?

yes, I'm sure

> Why in some places we call it WITH and WITHOUT brackets?

This is based on the default error type. I used a similar approach once, while we were rethinking the Visitors API

```rust
pub struct DatabaseTransactionRunner<HFutResultE = InternalError> {
    _phantom: PhantomData<HFutResultE>,
}
```

src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -34,49 +34,74 @@ pub trait DatabaseTransactionManager: Send + Sync {
...
-    // Commit or rollback transaction depending on the result
-    match result {
-        // In case everything succeeded, commit the transaction
-        Ok(_) => {
-            db_transaction_manager
-                .commit_transaction(transaction_ref)
-                .await?;
-            Ok(())
-        }
+pub struct DatabaseTransactionRunner<HFutResultE = InternalError> {
```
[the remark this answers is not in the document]
s373r: > Why this type needs to be generic over error? Why can't run_transactional which is just a static function be generic over it?

for some cases, we need an error more specific than an `InternalError` -- but, I'll revise the code, maybe it's not relevant anymore. thanks for this question

src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -34,49 +34,74 @@ pub trait DatabaseTransactionManager: Send + Sync {
...
+    _phantom: PhantomData<HFutResultE>,
+}
 
-        // Otherwise, do an explicit rollback
-        Err(e) => {
-            db_transaction_manager
-                .rollback_transaction(transaction_ref)
-                .await?;
-            Err(e)
+impl<HFutResultE> DatabaseTransactionRunner<HFutResultE> {
```
sergiimk, line 41: Can we make `DatabaseTransactionRunner` a dill component? This way a top-level service could inject it instead of a `catalog` and do:
```rust
self.tx_runner.transactional(|tx_catalog| { ... });
```
It's a small usability improvement, but may simplify testing top-level services.

As a step further in usability we could hide catalog entirely as such:
```rust
self.tx_runner.transactional(|my_service: Arc<dyn MyService>| my_service.do_stuff())
```
i.e. by closure argument type we can have runner extract the right component from the catalog and pass it into the transactional context.

s373r: very interesting proposition :thinking:

src/domain/task-system/services/src/task_executor_impl.rs
sergiimk: I guess another way to look at this problem is that a task may involve long-running data processing and we wouldn't like to hold the DB transaction open for the entire duration of the task.

Our task system may require separating the stages like:
- prepare - read-only DB access to gather info for the execution into a plan)
- execute - performing planned long-running work
- commit - applying all the updates to the DB

But for the time being holding a transaction open should not be too problematic.

I don't really see the alternative, as once we progressively start adding DB dependencies (e.g. managing dataset metadata and references in Postgres) - tasks will need transactional context.

What do you think @zaychenko-sergei ?

src/app/cli/src/app.rs
```diff
@@ -135,20 +135,28 @@ pub async fn run(
...
-            } else if command.needs_workspace() && workspace_svc.is_upgrade_needed()? {
-                Err(CLIError::usage_error_from(WorkspaceUpgradeRequired))
-            } else if current_account.is_explicit() && !is_multi_tenant_workspace {
-                Err(CLIError::usage_error_from(NotInMultiTenantWorkspace))
-            } else {
-                command.run().await
+    // Wrap command execution in a transaction
+    let result = DatabaseTransactionRunner::run_transactional(
+        &cli_catalog,
+        |cli_catalog_with_transaction| async move {
```
zaychenko-sergei, line 141: Why would this be necessary?
If a particular command, i.e. `kamu login`, requires it, the particular command should control the scope of the transaction.
Right now you are again, wrapping continous commands like `kamu ui` or `kamu system api-server` into a never ending transaction.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -105,19 +100,38 @@ impl FlowServiceImpl {
     async fn run_current_timeslot(
         &self,
         timeslot_time: DateTime<Utc>,
+        catalog_with_transaction: &Catalog,
```
zaychenko-sergei, line 103: This name inspires to suggest another composition-based type for clarity.
How about:

```
struct TransactionalCatalog {
    catalog: Catalog,
}

impl TransactionalCatalog {
    #[inline]
    pub get_catalog(&self) -> &Catalog { &self.catalog }
    
    #[inline]
    pub get_transaction(&self) -> ...
}

```

zaychenko-sergei: Right now we have to guess what makes this catalog special, and you tried to do this with a name, but I think we should have a more explicit modeling.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -105,19 +100,38 @@ impl FlowServiceImpl {
...
+
         let planned_flow_ids: Vec<_> = {
             let mut state = self.state.lock().unwrap();
             state.time_wheel.take_nearest_planned_flows()
         };
 
-        let mut planned_task_futures = Vec::new();
+        let mut planned_task_futures = Vec::with_capacity(planned_flow_ids.len());
         for planned_flow_id in planned_flow_ids {
+            let cloned_flow_event_store = flow_event_store.clone();
```
zaychenko-sergei, line 119: Why not to shadow names...

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -39,13 +40,11 @@ pub struct FlowServiceImpl {
...
     run_config: Arc<FlowServiceRunConfig>,
     event_bus: Arc<EventBus>,
-    flow_event_store: Arc<dyn FlowEventStore>,
     time_source: Arc<dyn SystemTimeSource>,
-    task_scheduler: Arc<dyn TaskScheduler>,
-    flow_configuration_service: Arc<dyn FlowConfigurationService>,
     dataset_changes_service: Arc<dyn DatasetChangesService>,
     dependency_graph_service: Arc<dyn DependencyGraphService>,
     dataset_ownership_service: Arc<dyn DatasetOwnershipService>,
+    catalog: Catalog,
```
zaychenko-sergei, line 47: This is a place we need to think over better - maybe later:
 - we have components that are transaction-independent (i.e. time source)
 - we have components that can only function with a transactional catalog (event store)
 - we have components that can control scope of the transaction (flow service).

But actually, some parts of flow service inject transaction-dependent components like event store, and now we are moving these from the component static state to function local vars / arguments. This indicates that we have an implicit logic boundary, and should try isolating sub-components to be able to use DI again (from transactional catalog). Those sub-components would be short-lived, with scope limited to the duration of 1 transaction.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -625,31 +676,39 @@ impl FlowServiceImpl {
         trigger: FlowTrigger,
         config_snapshot: Option<FlowConfigurationSnapshot>,
     ) -> Result<Flow, InternalError> {
+        let flow_event_store = self.catalog.get_one::<dyn FlowEventStore>().unwrap();
```
zaychenko-sergei, line 679: And in this place it's not very clear from which catalog we are taking the store from. If it's not from transactional catalog, will this even work?

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -1226,14 +1311,18 @@ impl FlowService for FlowServiceImpl {
         &self,
         flow_id: FlowID,
     ) -> Result<FlowState, CancelScheduledTasksError> {
-        let mut flow = Flow::load(flow_id, self.flow_event_store.as_ref()).await?;
+        let flow_event_store = self.catalog.get_one::<dyn FlowEventStore>().unwrap();
+        let task_scheduler = self.catalog.get_one::<dyn TaskScheduler>().unwrap();
```
zaychenko-sergei, line 1315: It's clearly starting to feel like we have 2 flow managers :-)
One is controlling transaction, another one is dependent on externally managed transactions.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -1331,43 +1431,68 @@ impl AsyncEventHandler<TaskEventFinished> for FlowServiceImpl {
...
-                && !flow_result.is_empty()
-            {
-                match flow.flow_key.get_type().success_followup_method() {
-                    FlowSuccessFollowupMethod::Ignore => {}
-                    FlowSuccessFollowupMethod::TriggerDependent => {
-                        self.enqueue_dependent_flows(finish_time, &flow, flow_result)
-                            .await?;
+            DatabaseTransactionRunner::run_transactional_with(
+                &self.catalog,
+                |flow_event_store: Arc<dyn FlowEventStore>| async move {
```
zaychenko-sergei, line 1461: What is the reasoning behind splitting this event into 2 transactions?

src/domain/flow-system/services/src/flow_configuration/flow_configuration_service_impl.rs
```diff
@@ -315,24 +339,28 @@ impl FlowConfigurationService for FlowConfigurationServiceImpl {
...
 
 #[async_trait::async_trait]
 impl AsyncEventHandler<DatasetEventDeleted> for FlowConfigurationServiceImpl {
     #[tracing::instrument(level = "debug", skip_all, fields(?event))]
     async fn handle(&self, event: &DatasetEventDeleted) -> Result<(), InternalError> {
+        let flow_configuration_event_store = self
+            .catalog
+            .get_one::<dyn FlowConfigurationEventStore>()
+            .unwrap();
+
```
zaychenko-sergei, line 359: I guess there is no transaction hiding here, and we should open one.

src/domain/task-system/services/src/task_executor_impl.rs
zaychenko-sergei: I think that execution of the tasks should clearly be outside the scope of the transactions that prepare and commit. We should assume that task execution is a heavyweight computational task, potentially running on other computer (future operator). If there will be another type of task that requires it's own transactions for different data, it would have to control it by itself, but I don't think that would be related to tasks management - tasks should be in saved state before that starts.

zaychenko-sergei: > Our event bus is not yet transactional, so having them inside run_transactional would work

Event bus will not be transactional. By definition, it's an asynchronous notification of an already saved state change. Rollback of the main change that caused the event should not be possible by the time the notification starts. So all event issuing should happen outside transactions. The handlers on their own may open new transactions to react to the change.

src/domain/task-system/services/src/task_scheduler_impl.rs
zaychenko-sergei: Scheduler is really a singleton, but it's basically an API to manage a queue, rather than an actively running component that should control transactions. Maybe it should receive the `event_store` reference as an argument - or transactional catalog.

src/domain/task-system/services/src/task_executor_impl.rs
sergiimk: > Event bus will not be transactional

We previously discussed two-tier event bus model:
- intra-process one, for decoupling components - this one carries events in memory and all subscribers can react to them in the scope of the original transaction
- inter-process one, for communication between services and instances of the same service - publishing events to this one can be made transactional, but instances react to them in independent transactions (eventually consistent)

src/app/cli/src/app.rs
s373r: > Right now you are again, wrapping continous commands like kamu ui or kamu system api-server into a never ending transaction.

oh, yes, you are right...

s373r:
The most important parts were extracted into https://github.com/kamu-data/kamu-cli/pull/660

This PR is no longer needed

Issue #644, opened 2024-05-21
Support hard compacting recovery with automatic re-transforms for derived datasets


zaychenko-sergei:
"Automatic handling of history resets during pull"

`kamu pull --reset-name-tbd-derivative-blah <derivative dataset>` - make flag part of the pull flow

This goes into `PullParams` and applies to all datasets matches by pattern / --recursive / --all
`PullService` will attempt to run deriv transform as usual
If `TransformService` returns `IntervalError` and our `reset-name-tbd-derivative-blah` flag was specified it will call `CompactService::compact_keep_metadata_only()` to reset the derivative dataset to empty state
`PullService`  will then try to repeat the transform.

Of course, give a better name to the switch, than here :-)

PR #651, opened 2024-05-27
Add reset derivative flag to pull command

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/644

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
### DIFF
--- a/resources/cli-reference.md
+++ b/resources/cli-reference.md
@@ -638,6 +638,7 @@ Pull new data into the datasets
 * `--no-alias` — Don't automatically add a remote push alias for this destination
 * `--set-watermark <TIME>` — Injects a manual watermark into the dataset to signify that no data is expected to arrive with event time that precedes it
 * `-f`, `--force` — Overwrite local version with remote, even if revisions have diverged
+* `--reset-derivatives-on-fail` — Run hard compacting of derivative dataset if transformation failed
 
 Pull is a multi-functional command that lets you update a local dataset. Depending on the parameters and the types of datasets involved it can be used to:
 - Run polling ingest to pull data into a root dataset from an external source
--- a/resources/cli-reference.md
+++ b/resources/cli-reference.md
@@ -638,7 +638,7 @@ Pull new data into the datasets
 * `--no-alias` — Don't automatically add a remote push alias for this destination
 * `--set-watermark <TIME>` — Injects a manual watermark into the dataset to signify that no data is expected to arrive with event time that precedes it
 * `-f`, `--force` — Overwrite local version with remote, even if revisions have diverged
-* `--reset-derivatives-on-diverged-input` — Run hard compacting of derivative dataset if transformation failed
+* `--reset-derivatives-on-diverged-input` — Run hard compacting of derivative dataset if transformation failed due to root dataset compaction
 
 Pull is a multi-functional command that lets you update a local dataset. Depending on the parameters and the types of datasets involved it can be used to:
 - Run polling ingest to pull data into a root dataset from an external source
--- a/resources/cli-reference.md
+++ b/resources/cli-reference.md
@@ -638,7 +638,7 @@ Pull new data into the datasets
 * `--no-alias` — Don't automatically add a remote push alias for this destination
 * `--set-watermark <TIME>` — Injects a manual watermark into the dataset to signify that no data is expected to arrive with event time that precedes it
 * `-f`, `--force` — Overwrite local version with remote, even if revisions have diverged
-* `--reset-derivatives-on-fail` — Run hard compacting of derivative dataset if transformation failed
+* `--reset-derivatives-on-diverged-input` — Run hard compacting of derivative dataset if transformation failed
 
 Pull is a multi-functional command that lets you update a local dataset. Depending on the parameters and the types of datasets involved it can be used to:
 - Run polling ingest to pull data into a root dataset from an external source
--- a/src/app/cli/src/commands/compact_command.rs
+++ b/src/app/cli/src/commands/compact_command.rs
@@ -150,7 +150,10 @@ impl Command for CompactCommand {
             Ok(())
         } else {
             Err(BatchError::new(
-                format!("Failed to compact {} dataset(s)", errors.len()),
+                format!(
+                    "Failed to compact {}/{total_results} dataset(s)",
+                    errors.len()
+                ),
                 errors,
             )
             .into())
--- a/src/app/cli/src/commands/compact_command.rs
+++ b/src/app/cli/src/commands/compact_command.rs
@@ -19,7 +19,7 @@ use kamu::domain::{
 };
 use opendatafabric::{DatasetHandle, DatasetRef};
 
-use crate::{CLIError, Command, CompactingMultiProgress, VerificationMultiProgress};
+use crate::{BatchError, CLIError, Command, CompactingMultiProgress, VerificationMultiProgress};
 
 pub struct CompactCommand {
     dataset_repo: Arc<dyn DatasetRepository>,
@@ -114,9 +114,10 @@ impl Command for CompactCommand {
             progress.draw();
         });
 
-        self.compacting_svc
-            .compact_dataset(
-                &dataset_handle,
+        let compacting_results = self
+            .compacting_svc
+            .compact_multi(
+                vec![dataset_handle.as_local_ref()],
                 CompactingOptions {
                     max_slice_size: Some(self.max_slice_size),
                     max_slice_records: Some(self.max_slice_records),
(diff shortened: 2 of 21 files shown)


src/infra/core/src/pull_service_impl.rs
```diff
@@ -419,20 +422,64 @@ impl PullServiceImpl {
...
+                        compacting_target_refs,
+                        CompactingOptions {
+                            keep_metadata_only: true,
+                            ..Default::default()
+                        },
+                        compacting_listener,
+                    )
+                    .await;
+
+                let retry_items: Vec<_> = std::iter::zip(batch, compacting_results)
```
sergiimk, line 465: If `transform_multi` is called on multiple datasets and only one of them fails with `InvalidInterval` then `batch` and `compacting_results` will have different lengths and `zip` will not do what we expect it to do here.

This will be the case with dataset graph like:
```
A -> B
C -> D
```

A and C will be pulled as roots first, then `transform_multi` will be called on B and D in one batch, and if A was compacted and D wasn't - you'll get into this case.

Please add a test for the case like this before fixing - it's a tricky one.

src/infra/core/tests/tests/test_pull_service_impl.rs
```diff
@@ -840,6 +843,43 @@ async fn test_set_watermark_rejects_on_derivative() {
     assert_eq!(harness.num_blocks(&dataset_alias).await, 1);
 }
 
+#[tokio::test]
+async fn test_compactin_on_transform_faile() {
```
s373r, line 847: `test_compactin_on_transform_failed()` (missed faile-`d`)

src/infra/core/tests/tests/test_pull_service_impl.rs
```diff
@@ -840,6 +843,43 @@ async fn test_set_watermark_rejects_on_derivative() {
...
+    let harnes_owerrides = PullTestHarnessResultOwerrides {
```
s373r, line 848: global comment: owerrides -> overrides

src/domain/core/src/services/transform_service.rs
```diff
@@ -133,6 +133,12 @@ pub enum TransformError {
         AccessError,
     ),
     #[error(transparent)]
+    InvalidInterval(
+        #[from]
+        #[backtrace]
+        InvalidIntervalError,
+    ),
```
s373r, line 140: Q: don't really understand the difference with `Internal` -- maybe use a more specific name?

src/infra/core/src/compacting_service_impl.rs
```diff
@@ -515,4 +516,42 @@ impl CompactingService for CompactingServiceImpl {
...
+    ) -> Vec<CompactingResponse> {
+        let dataset_handles: Vec<_> = match filter_datasets_by_local_pattern(
+            self.dataset_repo.as_ref(),
+            dataset_refs
+                .into_iter()
+                .map(DatasetRefPattern::Ref)
+                .collect(),
+        )
+        .try_collect()
+        .await
```
s373r, line 534: optional: extract `filter_datasets_by_local_pattern(...).try_collect().await` to a local let

src/infra/core/src/pull_service_impl.rs
```diff
@@ -419,20 +422,64 @@ impl PullServiceImpl {
...
+                        {
+                            return Some(pi.clone());
+                        }
+                        None
+                    })
+                    .collect();
+
+                return self
+                    .transform_multi(&retry_items, transform_listener, None, false)
+                    .await;
```
sergiimk, line 479: Similarly here, we need to return the combined result of datasets that succeeded the transform the first time and those that were reset and then transformed. Currently you will return only results for those that were reset.

The logic of combining results may get too complex... As an option - I'd consider making `reset_on_input_divergence` a parameter of the `TransformService` (as an options struct passed into `transform_multi`). This way it can handle divergence internally, and `PullService` will not need to deal with complexity of resetting, retrying, and merging results - retries can happen on per-dataset basis inside `TrnasformServiceImpl`.

What do you think @zaychenko-sergei ?

src/infra/core/src/compacting_service_impl.rs
```diff
@@ -515,4 +516,42 @@ impl CompactingService for CompactingServiceImpl {
...
+        {
+            Ok(matched_datasets) => matched_datasets,
+            Err(_) => return vec![],
+        };
```
s373r, line 538: optional: use let-else

src/domain/core/src/services/pull_service.rs
```diff
@@ -132,6 +136,9 @@ pub struct PullMultiOptions {
     pub ingest_options: PollingIngestOptions,
     /// Sync-specific options,
     pub sync_options: SyncOptions,
+    /// Run compacting of all derivative datasets without saving data
+    /// if transformation fails
+    pub reset_derivatives: bool,
```
s373r, line 141: Better to include in the parameter name some of the logic it carries -- we're not just doing `reset_derivatives`, but also running compacting

I don't know the best option to suggest right off the bat, but how about `run_save_compacting` (and just in the comment we will point out that save is not saving data, in case of failure)

src/infra/core/src/pull_service_impl.rs
```diff
@@ -419,20 +422,64 @@ impl PullServiceImpl {
...
+                            keep_metadata_only: true,
+                            ..Default::default()
+                        },
+                        compacting_listener,
+                    )
+                    .await;
+
+                let retry_items: Vec<_> = std::iter::zip(batch, compacting_results)
+                    .filter_map(|(pi, response)| {
+                        assert_eq!(pi.local_ref, response.dataset_ref);
```
s373r, line 467: re `pi`: I'd like to suggest another name, because only mathematical associations come to mind :)

src/app/cli/src/commands/compact_command.rs
```diff
@@ -114,22 +114,46 @@ impl Command for CompactCommand {
...
+            eprintln!(
+                "{}",
+                console::style(format!("{total_results} dataset(s) were compacted"))
+                    .green()
+                    .bold()
+            );
+            Ok(())
+        } else {
+            Err(BatchError::new(
+                format!("Failed to compact {} dataset(s)", errors.len()),
```
s373r, line 153: suggestion: how about adding the initial number of datasets?

`format!("Failed to compact {}/{dataset_count} dataset(s)", errors.len()),`

s373r (commented):
added minor comments

src/domain/core/src/services/transform_service.rs
zaychenko-sergei: "invalid interval" is very clear, there is no chain path from the end back to start.

src/infra/core/src/pull_service_impl.rs
zaychenko-sergei: I would have completely placed the entire algorithm of "error-compact-retry" into `TransformServiceImpl`. Pull is at a higher level of abstraction.

resources/cli-reference.md
```diff
@@ -638,6 +638,7 @@ Pull new data into the datasets
 * `--no-alias` — Don't automatically add a remote push alias for this destination
 * `--set-watermark <TIME>` — Injects a manual watermark into the dataset to signify that no data is expected to arrive with event time that precedes it
 * `-f`, `--force` — Overwrite local version with remote, even if revisions have diverged
+* `--reset-derivatives-on-fail` — Run hard compacting of derivative dataset if transformation failed
```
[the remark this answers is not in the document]
rmn-boiko: Changed

src/infra/core/src/pull_service_impl.rs
rmn-boiko: Such as we removed this logic from pull service to transform service where we work with single dataset it is not a case anymore

src/infra/core/src/pull_service_impl.rs
rmn-boiko: this logic was removed

src/infra/core/src/transform_service_impl.rs
```diff
@@ -697,9 +705,39 @@ impl TransformServiceImpl {
...
+                    CompactingOptions {
+                        keep_metadata_only: true,
+                        ..Default::default()
+                    },
+                    None,
+                )
+                .await
+                .int_err()?;
+
+            if let CompactingResult::Success { .. } = compacting_result {
```
zaychenko-sergei, line 727: What if result is not succesful?

src/infra/core/src/transform_service_impl.rs
```diff
@@ -763,10 +803,10 @@ impl TransformService for TransformServiceImpl {
             let f = match self.dataset_repo.resolve_dataset_ref(dataset_ref).await {
                 Ok(hdl) => {
                     let maybe_listener = multi_listener.begin_transform(&hdl);
-                    self.transform_impl(hdl.into(), maybe_listener)
+                    self.transform_impl(hdl.into(), options.clone(), maybe_listener)
                 }
                 // Relying on this call to fail to avoid boxing the futures
-                Err(_) => self.transform_impl(dataset_ref.clone(), None),
+                Err(_) => self.transform_impl(dataset_ref.clone(), options.clone(), None),
```
zaychenko-sergei, line 809: Do we need these clones? It seems they can be simply moved

src/infra/core/src/transform_service_impl.rs
rmn-boiko: If no we will return error as it is on the next line 
```if let Some(operation) = next_operation?```

sergiimk (commented):
Very nice, moving compaction logic into transform service simplified things a lot!

src/infra/core/src/transform_service_impl.rs
rmn-boiko: It is in loop. But I added Copy trait to make it looks cleaner

zaychenko-sergei (approved):

src/domain/core/src/services/transform_service.rs
s373r: oh, yes --  I read it earlier as `InvalidInter`-N-`al`

PR #660, opened 2024-06-04
Persistent accounts stabilization

s373r:
This PR changes are extracted from
- https://github.com/kamu-data/kamu-cli/pull/650

---

## Description

### Added
- Added E2E test infrastructure
  - Added necessary components for managed run -- for startup, operations, and shutdown
### Changed
- Hidden arguments and options are excluded from [the CLI reference](resources/cli-reference.md)
### Fixed
- Stabilized startup using connection to databases
  - Added HTTP middleware that wraps each request into a separate transaction 
  - Also added wrapping for some commands, in particular `kamu system generate-token`
  - The structure of services that required lazy access to databases was reorganized:
     - Extracted `PredefinedAccountsRegistrator` & `DatasetOwnershipServiceInMemoryStateInitializer` 
- Fixed potential crash when attempting to rollback a transaction if the connection fails to establish

- [x] Unit and integration tests added, ✅
  - E2E tests added
  
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
  - [x] [Public documentation](https://github.com/kamu-data/kamu-docs/): ❌, not needed
  
- [x] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ❌, will be added later


resources/cli-reference.md
```diff
@@ -306,8 +307,8 @@ Initialize an empty workspace in the current directory
 
 * `--exists-ok` — Don't return an error if workspace already exists
 * `--pull-images` — Only pull container images and exit
-* `--list-only` — List image names instead of pulling
-* `--multi-tenant` — Initialize a workspace for multiple tenants
+* `--list-only` — List image names instead of pulling (hidden)
+* `--multi-tenant` — Initialize a workspace for multiple tenants (hidden)
```
zaychenko-sergei, line 311: Offtopic: I wonder if content for hidden switches should be generated into the reference.

src/app/cli/src/app.rs
```diff
@@ -135,19 +135,38 @@ pub async fn run(
...
-                Err(CLIError::usage_error_from(WorkspaceUpgradeRequired))
-            } else if current_account.is_explicit() && !is_multi_tenant_workspace {
-                Err(CLIError::usage_error_from(NotInMultiTenantWorkspace))
-            } else {
-                command.run().await
+    // TODO: Extend Command trait with the corresponding property
+    let need_to_wrap_with_transaction = matches!(
+        matches.subcommand(),
+        Some(("login" | "add" | "delete" | "pull", _))
+    );
```
zaychenko-sergei, line 142: I think there is nothing to wait for, do this extension

src/domain/accounts/services/src/authentication_service_impl.rs
```diff
@@ -65,47 +64,34 @@ impl AuthenticationServiceImpl {
...
 
     fn resolve_authentication_provider(
         &self,
         login_method: &str,
-    ) -> Result<Arc<dyn AuthenticationProvider>, UnsupportedLoginMethodError> {
-        match self.authentication_providers_by_method.get(login_method) {
+        transactional_catalog: &TransactionalCatalog,
+    ) -> Result<Arc<dyn AuthenticationProvider>, LoginError> {
+        let authentication_providers =
+            self.get_authentication_providers_by_method(transactional_catalog);
```
zaychenko-sergei, line 87: It's not using `login_method`, so why calling it like this.

src/domain/accounts/services/src/authentication_service_impl.rs
```diff
@@ -198,182 +187,353 @@ impl AuthenticationServiceImpl {
...
+                    .keys()
+                    .copied()
+                    .collect();
+
+                // Stable order for tests
+                login_methods.sort_unstable();
+
+                Ok(login_methods)
+            })
+            .await
```
zaychenko-sergei, line 253: I don't fully understand why this requires a transaction

src/domain/accounts/services/src/login_password_auth_provider.rs
```diff
@@ -88,10 +88,14 @@ impl AuthenticationProvider for LoginPasswordAuthProvider {
         &self,
         login_credentials_json: String,
     ) -> Result<ProviderLoginResponse, ProviderLoginError> {
+        dbg!("@@@ login 1", &login_credentials_json);
```
zaychenko-sergei, line 91: If we want some logs to remain, use `tracing`

zaychenko-sergei:
What about Github Actions support for new testing type?

src/domain/accounts/services/src/authentication_service_impl.rs
```diff
@@ -32,22 +33,20 @@ use serde::{Deserialize, Serialize};
...
     predefined_accounts_config: Arc<PredefinedAccountsConfig>,
     time_source: Arc<dyn SystemTimeSource>,
     encoding_key: EncodingKey,
     decoding_key: DecodingKey,
-    authentication_providers_by_method: HashMap<&'static str, Arc<dyn AuthenticationProvider>>,
-    login_password_auth_provider: Option<Arc<LoginPasswordAuthProvider>>,
-    account_repository: Arc<dyn AccountRepository>,
+    transaction_runner: Arc<DatabaseTransactionRunner>,
     state: tokio::sync::Mutex<State>,
 }
```
zaychenko-sergei, line 51: A few things should be changed:

1) We are not strictly required to do lazy registration of predefined accounts. This could be moved to a separate "initializer" service that is only invoked from `app.py` (and it's `node` equivalent). The initialization code will be contained in it's own transaction.

2) If suggestion above works, we can get rid of `State` and `Singleton` status

3) I think at the end of refactoring, `AuthenticationServiceImpl` should not be responsible for controlling the scope of transaction. In case of web queries, there is middleware. In case of commands, there is a transaction in `app.py` around execution of commands.

resources/cli-reference.md
s373r: I've been thinking about that, too -- I think it's more of a no than a yes

src/domain/accounts/services/src/login_password_auth_provider.rs
s373r: oh, it wasn't supposed to be commited...

src/e2e/app/cli/postgres/tests/tests/test_auth.rs
```diff
@@ -0,0 +1,87 @@
...
+                "#
+            ),
+            host = db.get_host(),
+            user = db.get_username(),
+            password = db.get_username(), // It's intended: password is same as user for tests
+            database = db.get_database().unwrap(),
+        )
+        .as_str(),
+    )
+    .await;
```
zaychenko-sergei, line 42: Configuration/workspace code repeats, isolate it into harness

src/e2e/app/cli/postgres/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+          AND state = 'idle'
+          AND pid <> pg_backend_pid()   -- exclude the current connection
+          AND query = 'COMMIT'          -- last executed query
+          AND wait_event = 'ClientRead' -- awaiting new data
+        "#,
+        database
+    )
+    .fetch_one(&pg_pool)
+    .await
+    .unwrap();
```
zaychenko-sergei, line 54: I don't understand why is this necessary.

src/domain/accounts/services/src/authentication_service_impl.rs
s373r: > This could be moved to a separate "initializer" service

Added `PredefinedAccountsRegistrator` -- done 

> If suggestion above works, we can get rid of State and Singleton status
> ...
> I think at the end of refactoring, AuthenticationServiceImpl should not be responsible for controlling the scope of transaction`

Done

src/domain/accounts/services/src/authentication_service_impl.rs
s373r: `AuthenticationServiceImpl` no longer manages transactions

---

> I don't fully understand why this requires a transaction

But to answer the question: it was a necessity from a DI perspective

src/e2e/app/cli/postgres/tests/tests/test_auth.rs
s373r: Added `kamu_cli_e2e_test!()` macro

src/app/cli/src/app.rs
s373r: Unfortunately, since we need to know if there will be a transaction (!) even before a command is created, I had to do it with a separate function `cli_commands::command_needs_transaction()`

deny.toml
```diff
@@ -86,5 +86,5 @@ unsound = "warn"
 yanked = "deny"
 notice = "warn"
 ignore = [
+    "RUSTSEC-2023-0071"
```
zaychenko-sergei, line 89: A comment why would be nice

src/app/cli/src/app.rs
```diff
@@ -135,22 +142,64 @@ pub async fn run(
...
+                .get_one::<DatasetOwnershipServiceInMemoryStateInitializer>()
+                .map_err(CLIError::critical)?;
+
+            initializer
+                .eager_initialization()
+                .await
+                .map_err(CLIError::critical)
+        })
+        .await?;
+
```
zaychenko-sergei, line 171: Minor: let's make method from these

src/domain/accounts/services/src/predefined_accounts_registrator.rs
```diff
@@ -0,0 +1,97 @@
...
+            })?;
+
+        if account_config.provider == PROVIDER_PASSWORD
+            && let Some(login_password_auth_provider) =
+                self.maybe_login_password_auth_provider.as_ref()
+        {
+            login_password_auth_provider
+                .save_password(&account.account_name, account_config.get_password())
+                .await?;
+        }
```
zaychenko-sergei, line 91: Perhaps it should panic, if this dependency is undefined. Or we could always demand it instead of Option

deny.toml
s373r: It was a backport of a change from [another PR](https://github.com/kamu-data/kamu-cli/pull/665) (merged) but I've added context as to why the change was made

src/app/cli/src/app.rs
s373r: Done

src/domain/accounts/services/src/predefined_accounts_registrator.rs
s373r: `LoginPasswordAuthProvider` is now a required dependency -- done

zaychenko-sergei (approved):

src/app/cli/src/cli_commands.rs
```diff
@@ -580,6 +583,22 @@ pub fn get_command(
...
+    match arg_matches.subcommand() {
+        Some(("system", system_matches)) => match system_matches.subcommand() {
+            Some(("generate-token", _)) => Ok(true),
+            Some(_) => Ok(false),
+            None => Err(CommandInterpretationFailed.into()),
+        },
+        Some(_) => Ok(false),
+        None => Err(CommandInterpretationFailed.into()),
+    }
+}
```
sergiimk, line 596: This seems super fragile. What stops us from always having some transaction runner in the catalog (real DB or noop impl) and always run commands in transactional scope?

s373r: we have commands that take a long time to execute -- e.g.  API server

can't do that because the commit will never happen

-- 

The key point here is not the presence of the component itself in the Catalog (they are already present, after all), but the launch of the transaction

s373r: and yes, I was bitterly disappointed to discover that we need to know if a transaction is required BEFORE creating a Command

src/e2e/app/cli/common/src/e2e_hartness.rs
```diff
@@ -0,0 +1,174 @@
+// Copyright Kamu Data, Inc. and contributors. All rights reserved.
```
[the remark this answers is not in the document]
s373r: Fixed via 
- https://github.com/kamu-data/kamu-cli/pull/667

src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -102,8 +135,9 @@ impl TransactionRef {
     pub fn into_maybe_transaction<DB: sqlx::Database>(
         self,
     ) -> Option<sqlx::Transaction<'static, DB>> {
-        let m = Arc::try_unwrap(self.inner).unwrap().into_inner();
-        m.maybe_transaction
+        Arc::try_unwrap(self.inner)
+            .ok()
+            .and_then(|m| m.into_inner().maybe_transaction)
```
sergiimk, line 140: This seems very dangerous - we `unwrap()` the result of extracting from `Arc` to ensure that `into_maybe_transaction` is called ONLY when there are no components holding it as a reference somewhere - i.e. that transaction is not "stuck" in some singleton. Without the `unwrap()` this situation will go unnoticed.

Or am I reading this wrong?

s373r: You read that correctly -- yes, that change needs to be removed

PR #673, opened 2024-06-17
Add an E2E test: test_rest_api_request_dataset_tail()

s373r:
## Description

### Added
- Added an E2E test group for REST API
### Changed
- MySQL E2E tests are turned off
- The `<owner/dataset>/tail` REST API endpoint will:
  - return `404 Not Found` on not found datasets
  - return `202 No Content` on empty datasets
- Temporarily disabled tests for OData
### Fixed
- Stabilized query handling in case of database usage; affected:
  - The `<owner/dataset>/tail` REST API
  - The `/query` REST API
  - `kamu system tail` command
  - `kamu sql` command
- Fixed memory leak when working with DataFusion

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
  - [x] [Public documentation](https://github.com/kamu-data/kamu-docs/): ❌ -- not needed
  
- [x] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ❌ -- will be added later

### DIFF
--- a/src/adapter/http/src/data/query_handler.rs
+++ b/src/adapter/http/src/data/query_handler.rs
@@ -56,7 +56,6 @@ pub async fn dataset_query_handler_post(
 
     // Apply pagination limits
     let df = res
-        .df_with_ctx
         .df
         .limit(
             usize::try_from(body.skip).unwrap(),
--- a/src/adapter/http/src/data/query_handler.rs
+++ b/src/adapter/http/src/data/query_handler.rs
@@ -56,6 +56,7 @@ pub async fn dataset_query_handler_post(
 
     // Apply pagination limits
     let df = res
+        .df_with_ctx
         .df
         .limit(
             usize::try_from(body.skip).unwrap(),
@@ -64,15 +65,10 @@ pub async fn dataset_query_handler_post(
         .int_err()
         .api_err()?;
 
-    let res = QueryResponse {
-        df,
-        state: res.state,
-    };
-
-    let arrow_schema = res.df.schema().inner().clone();
+    let arrow_schema = df.schema().inner().clone();
 
     let schema = if body.include_schema {
-        Some(serialize_schema(res.df.schema(), body.schema_format).api_err()?)
+        Some(serialize_schema(df.schema(), body.schema_format).api_err()?)
     } else {
         None
     };
@@ -83,7 +79,7 @@ pub async fn dataset_query_handler_post(
         None
     };
 
-    let record_batches = res.df.collect().await.int_err().api_err()?;
+    let record_batches = df.collect().await.int_err().api_err()?;
     let json = serialize_data(&record_batches, body.data_format).api_err()?;
     let data = serde_json::value::RawValue::from_string(json).unwrap();
 
--- a/src/adapter/odata/tests/tests/test_handlers.rs
+++ b/src/adapter/odata/tests/tests/test_handlers.rs
@@ -134,7 +134,6 @@ async fn test_metadata_handler() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
-#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
 #[test_log::test(tokio::test)]
 async fn test_collection_handler() {
     let harness = TestHarness::new();
@@ -231,7 +230,6 @@ async fn test_collection_handler() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
-#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
 #[test_log::test(tokio::test)]
 async fn test_collection_handler_by_id() {
     let harness = TestHarness::new();
@@ -289,7 +287,6 @@ async fn test_collection_handler_by_id() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
-#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
 #[test_log::test(tokio::test)]
 async fn test_collection_handler_by_id_not_found() {
     let harness = TestHarness::new();
--- a/src/adapter/odata/tests/tests/test_handlers.rs
+++ b/src/adapter/odata/tests/tests/test_handlers.rs
@@ -134,6 +134,7 @@ async fn test_metadata_handler() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
+#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
 #[test_log::test(tokio::test)]
 async fn test_collection_handler() {
     let harness = TestHarness::new();
@@ -230,6 +231,7 @@ async fn test_collection_handler() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
+#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
 #[test_log::test(tokio::test)]
 async fn test_collection_handler_by_id() {
     let harness = TestHarness::new();
@@ -287,6 +289,7 @@ async fn test_collection_handler_by_id() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
+#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
 #[test_log::test(tokio::test)]
 async fn test_collection_handler_by_id_not_found() {
     let harness = TestHarness::new();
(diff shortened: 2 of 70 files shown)


src/adapter/http/src/data/query_handler.rs
```diff
@@ -64,15 +65,10 @@ pub async fn dataset_query_handler_post(
         .int_err()
         .api_err()?;
 
-    let res = QueryResponse {
-        df,
-        state: res.state,
-    };
-
-    let arrow_schema = res.df.schema().inner().clone();
```
s373r, line 72: We have no problem removing that variable, right?

src/e2e/app/cli/repo-tests/src/test_rest_api.rs
```diff
@@ -62,17 +92,49 @@ pub async fn test_rest_api_request_dataset_tail(kamu_api_server_client: KamuApiS
...
...2024-06-18T15:00:00Z","match_time":"2000-01-01T00:00:00Z","match_id":1,"player_id":"Alice","score":100},{"offset":1,"op":0,"system_time":"2024-06-18T15:00:00Z","match_time":"2000-01-01T00:00:00Z","match_id":1,"player_id":"Bob","score":80}],"schema":"{\"fields\":[{\"name\":\"offset\",\"data_type\":\"Int64\",\"nullable\":true,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}},{\"name\":\"op\",\"data_type\":\"Int32\",\"nullable\":false,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}},{\"name\":\"system_time\",\"data_type\":{\"Timestamp\":[\"Millisecond\",\"UTC\"]},\"nullable\":false,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}},{\"name\":\"match_time\",\"data_type\":{\"Timestamp\":[\"Millisecond\",\"UTC\"]},\"nullable\":true,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}},{\"name\":\"match_id\",\"data_type\":\"Int64\",\"nullable\":true,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}},{\"name\":\"player_id\",\"data_type\":\"Utf8\",\"nullable\":true,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}},{\"name\":\"score\",\"data_type\":\"Int64\",\"nullable\":true,\"dict_id\":0,\"dict_is_ordered\":false,\"metadata\":{}}],\"metadata\":{}}"}"#),
```
sergiimk: Such strings are very hard to read and maintain. Why not use a similar text vs `serde_json::Value` for `expected_response_body` so you could write the expected response using `json!` macro?

Consider also setting `"includeSchema": false` request parameter if you don't want to deal with this messy string. I imagine ensuring schema is exactly that is not the point of this test.

src/domain/core/src/services/query_service.rs
```diff
@@ -73,13 +76,29 @@ pub trait QueryService: Send + Sync {
...
 }
 
-///////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+/// Dataframe and its context. Without context, we cannot perform operations on
+/// the dataframe
+#[derive(Clone)]
+pub struct DataFrameWithContext {
+    pub ctx: SessionContext,
```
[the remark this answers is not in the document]
sergiimk: Is this because you make `KamuShema` hold a weak reference to state and need `SessionContext` to outlive it?

src/infra/core/src/query/mod.rs
```diff
@@ -74,7 +76,14 @@ pub(crate) struct KamuSchema {
...
 struct KamuSchemaImpl {
-    session_context: SessionContext,
+    session_config: SessionConfig,
+    table_options: TableOptions,
+    // To avoid creating an Arc cyclic reference[1], with SessionContext, we are forced to use
+    // weak references instead.
+    //
+    // [1] SessionContext -> KamuCatalog(CatalogProvider) -> KamuSchema -> KamuSchemaImpl
+    //     -> SchemaCache -> KamuTable -> SessionContext
+    session_state: Weak<RwLock<SessionState>>,
```
sergiimk, line 86: Amazing job at figuring this out!
I didn't think about circular references and imagine this was super hard to find.

I think the actual solution can be simplifed to possibly avoid shared state completely. I will think about it more and suggest some options.

src/adapter/odata/tests/tests/test_handlers.rs
```diff
@@ -134,6 +134,7 @@ async fn test_metadata_handler() {
 /////////////////////////////////////////////////////////////////////////////////////////
 
 #[test_group::group(engine, datafusion)]
+#[ignore = "Enable after migration of ODataCollectionContext to DataFrameWithContext"]
```
sergiimk, line 137: So OData adapter has the same problem? Does the upstream library need to be updated or you simply want to confirm the fix in `QueryService` before fixing `ODataCollectionContext`?

src/domain/core/src/services/query_service.rs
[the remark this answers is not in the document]
s373r: `DataFrame` cannot be used without a context present -- previously dataframe could be used without explicit context because it leaked (never released)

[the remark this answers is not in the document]
s373r: > Secondly, If we do have to pass SessionContext around I would prefer it to become a part of QueryResponse structure

It looks like I was trying to accomplish this with `DataFrameWithContext` structure, but apparently I didn't pick the wrong name.

Q: Does your comment relate more to naming, or more to the proposed composition?

src/domain/core/src/services/query_service.rs
```diff
@@ -207,15 +238,15 @@ pub enum QueryError {
...
+////////////////////////////////////////////////////////////////////////////////
 
 #[derive(Error, Clone, PartialEq, Eq, Debug)]
 #[error("Dataset schema is not yet available: {dataset_ref}")]
 pub struct DatasetSchemaNotAvailableError {
     pub dataset_ref: DatasetRef,
 }
 
-/////////////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
s373r: > Can we make whatever stylistic changes we agree on in one go as a separate PR

Yes, for sure -- thought about it too, but couldn't resist.... sorry

> I don't want to see this kind of noise span multiple PRs for months.

Will be done after this PR is merged

src/adapter/odata/tests/tests/test_handlers.rs
s373r: > So OData adapter has the same problem?

Yes, indirectly -- once we remove the cyclic reference, `DataFrame` that OData is processing with stops working

> Does the upstream library need to be updated or you simply want to confirm the fix in QueryService before fixing ODataCollectionContext?

Exactly -- as soon as this PR merges, I'll immediately update the OData dependency and activate those tests back

src/infra/core/src/query/mod.rs
s373r: > I think the actual solution can be simplifed to possibly avoid shared state completely. I will think about it more and suggest some options.

please note that a `DataFrame` cannot exist without a `SessionContext`. Or in other words, `SessionContext` can (should) exist longer than `DataFrame`

src/e2e/app/cli/common/src/e2e_harness.rs
```diff
@@ -9,23 +9,57 @@
...
+            is_multi_tenant: true,
+            env_vars: Some(vec![
+                ("KAMU_AUTH_GITHUB_CLIENT_ID".into(), "1".into()),
+                ("KAMU_AUTH_GITHUB_CLIENT_SECRET".into(), "2".into()),
+            ]),
+            ..Default::default()
+        }
+    }
+
+    pub fn freeze_system_time_with(date_time: DateTime<Utc>) -> Self {
```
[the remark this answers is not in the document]
s373r: > hould this be a "builder" interface instead?

that's a great idea -- I'll update it now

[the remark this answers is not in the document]
s373r: Done

src/e2e/app/cli/repo-tests/src/test_rest_api.rs
s373r: > Consider also setting "includeSchema": false request parameter if you don't want to deal with this messy string. I imagine ensuring schema is exactly that is not the point of this test.

A most excellent comment -- grateful for it

s373r: Done

src/e2e/app/cli/common/src/kamu_api_server_client.rs
```diff
@@ -8,13 +8,28 @@
...
+pub enum RequestBody {
+    JSON(serde_json::Value),
+    NDJSON(String),
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+pub enum ExpectedResponseBody {
+    JSON(String),
+    PLAIN(String),
```
sergiimk, line 28: Why all caps names? Won't clippy complain about these?

s373r: > warning: variant `json` should have an upper camel case name

Interestingly, I discovered by accident that rustc allows upper case names for enum elements as well

s373r: Fixed

sergiimk (approved):

PR #675, opened 2024-06-20
Newtypes for common paths

sergiimk:
Simplifies dependency injection configuration by wrapping run/cache directories into new types.

This way they don't have to be passed into builders using e.g. `with_run_dir(..)` calls, and thus most types can be added without going through a builder interface, which in its turn avoids the need for binding them to interfaces explicitly.

This removes a ton of boilerplate.


rmn-boiko (approved):

s373r (approved):
:+1: 

Minor: add an entry to the changelog

sergiimk:
> 👍
> 
> Minor: add an entry to the changelog

This is not a user-facing change, so no point in logging it there.

PR #680, opened 2024-06-20
Avoid referencing SessionState in KamuTable

sergiimk:
In this PR:
- We avoid the need for having `SessionState` in `KamuTable`
  - This is possible because `KamuTable` only needs the state to call `resolve_schema()` that may need to infer the schema from files, bun in our case we already know exactly what the schema is
- This in turn avoids the need to have a strong reference for `SessionContext` somewhere and thus the need for `DataFrameWithContext` struct, so all `QueryService` API changes can be reverted


s373r (approved):
Wow! Super, without the weak refs the code looks much better!

PR #685, opened 2024-06-26
Fix dividing lines

s373r:
## Description

Brings all dividing lines to the same size

- [x] Unit and integration tests added ❌ not needed
  
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation: 
    
  - [x] [Changelog](./CHANGELOG.md): 
    
- [x] Downstream effects: 
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ❌ not needed


scripts/lint-dividing-lines.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if grep -Ernxo -e "/{4,79}" -e "/{81,}" ./src/; then
```
sergiimk, line 5: Have you considered making this a test in `kamu-repo-tools`, just like the one that checks for license headers? This is nice an concise but I generally prefer to avoid shell scripts due to slight differences in utils version and portability.

s373r: No problem, will be done

zaychenko-sergei, line 5: Why 80? This sounds like a setting from the times we had 15'' monitors.
120 would be more realistic.

s373r: Done

s373r: 120 used as a base value -- done

s373r: Added https://github.com/kamu-data/kamu-cli/pull/685/files#diff-ac447473973f0a9697d6262d62ac648b6563dfffdb827e2bd7531dee2f0b8932

s373r: Added https://github.com/kamu-data/kamu-cli/pull/685/files#diff-ac447473973f0a9697d6262d62ac648b6563dfffdb827e2bd7531dee2f0b8932

zaychenko-sergei (approved):

scripts/lint-dividing-lines.sh
sergiimk: Rust [style guide](https://doc.rust-lang.org/nightly/style-guide/) says:

`The maximum width for a line is 100 characters`

I wouldn't care if its 77 or 89 as long as its consistent everywhere, but wouldn't setting it above the limit we use for `rustfmt` lead to problems?

PR #690, opened 2024-06-28
Modify revoke GQL response according to design

rmn-boiko:


zaychenko-sergei (approved):
Looks good, but you need to sync already

PR #714, opened 2024-07-16
Upgrade rustc and remove some git dependencies

sergiimk:
Serarately updating `rustc` to be able to merge before the datafusion blocker is resolved


Cargo.toml
```diff
@@ -194,7 +194,7 @@ wildcard_imports = "allow"
 
 
 # clippy::cargo
-cargo = { level = "warn", priority = 0 }
+cargo = { level = "warn", priority = -1 }
```
sergiimk, line 197: @s373r I couldn't find any documentation about these priority levels. Clippy gives me an error about conflicting priorities with `priority = 0` and recommends `-1` but I actually have no idea what this means :)

s373r: @sergiimk , Interesting, it looks like a warning was added when crossing rule groups, resulting in a higher priority (lower value) -- judging from the source code, if no priority is explicitly specified, then 0 is set

so yeah, it'll work.

Anyway, thanks for mentioning it

PR #715, opened 2024-07-16
Fix odf's sqlx feature

sergiimk:
If you run `cargo test -p opendatafabric` on master it will fail because optional `sqlx` feature was added incorrectly.


src/domain/datasets/domain/Cargo.toml
```diff
@@ -25,7 +25,9 @@ doctest = false
 database-common = { workspace = true }
 internal-error = { workspace = true }
 opendatafabric = { workspace = true, default-features = true, features = [
-    "sqlx",
+    "sqlx-mysql",
```
sergiimk, line 28: Why do we depend on sqlx from domain-level crate? I don't even see tests in this crate so its seems that it's not needed even as a `dev-dependency`
@rmn-boiko @zaychenko-sergei

PR #733, opened 2024-07-19
Fix special chars in db string

rmn-boiko:
Fix special characterss in database string in migration script

### DIFF
--- a/images/sqlx-cli-with-migrations/scripts/fetch-db-secret.sh
+++ b/images/sqlx-cli-with-migrations/scripts/fetch-db-secret.sh
@@ -18,4 +18,23 @@ SECRET_STRING=$(aws secretsmanager get-secret-value --secret-id $SECRET_NAME --r
 USERNAME=$(echo $SECRET_STRING | jq -r .username)
 PASSWORD=$(echo $SECRET_STRING | jq -r .password)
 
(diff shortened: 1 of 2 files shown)


images/sqlx-cli-with-migrations/scripts/fetch-db-secret.sh
```diff
@@ -18,4 +18,23 @@ SECRET_STRING=$(aws secretsmanager get-secret-value --secret-id $SECRET_NAME --r
...
-export DB_CONNECTION_STRING="${DB_PROVIDER}://${USERNAME}:${PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_NAME}"
+urlencode() {
+    local string="${1}"
+    local encoded=""
+    local special_chars="!#$&'()*+,/:;=?@[]"
+
+    for (( i = 0; i < ${#string}; i++ )); do
+        local c="${string:i:1}"
+        if [[ $special_chars == *"$c"* ]]; then
+            encoded+=$(printf %s "$c" | jq -sRr @uri)
```
sergiimk, line 29: that's a cool trick to use `jq`

but I don't understand why you need to do this one character at a time, why not feed the entire string into it?
```sh
$ printf %s 'pa$$word' | jq -sRr @uri
pa%24%24word
```

rmn-boiko: It was my first idea. But during my tests it goes wrong(but now I double checked and it works perfectly). Probably issue with late work). Changed

zaychenko-sergei (approved):

Issue #703, opened 2024-07-04
Support persistency/transactions for flow configuration data


zaychenko-sergei:
Currently we are using in-memory repository for flow configuration data.
There is a database implementation, but we did not finish debugging it with transactions.
Fix the behavior of transactions, and switch to database version when such a configuration is active.

PR #741, opened 2024-07-26
Experiment/outbox pattern

zaychenko-sergei:
## Description

Closes: #703.

Reliable transaction-based internal cross-domain message passing component (`MessageOutbox`), replacing `EventBus`:
 - Metadata-driven producer/consumer annotations
 - Immediate and transaction-backed message delivery
 - Background transactional message processor, respecting client idempotence

Persistent storage for flow configuration events.

Supplementary refactoring:
- Introduced use case layer, encapsulating authorization checks and action validations, for first 6 basic dataset scenarios
   (creating, creating from snapshot, deleting, renaming, committing an event, syncing a batch of events),
- Separated `DatasetRepository` on read-only and read-write parts
- Isolated `time-source` library
- Decoupled repositories from dependency graph.
- Clarified dataset find/get interfaces in dataset repository
- Unified naming of repository structs, files, harnesses

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/utils/messaging-outbox/src/consumers/message_consumers_mediator_utils.rs
```diff
@@ -0,0 +1,180 @@
...
+        ConsumerFilter::AllConsumers => all_consumers_for::<TMessage>(catalog),
+        ConsumerFilter::BestEffortConsumers => best_effort_consumers_for::<TMessage>(catalog),
+        ConsumerFilter::SelectedConsumer(consumer_name) => {
+            particular_consumers_for::<TMessage>(catalog, consumer_name)
+        }
+    };
+
+    let consumption_tasks = consumers
+        .into_iter()
+        .map(|consumer| (consumer, message.clone()));
```
sergiimk, line 40: Can we pass message by ref to avoid cloning it per consumer?

src/domain/flow-system/services/src/messages/flow_message_consumer_mediator.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    async fn consume_message<'a>(
+        &self,
+        catalog: &Catalog,
+        producer_name: &str,
+        consumer_filter: ConsumerFilter<'a>,
+        content_json: serde_json::Value,
+    ) -> Result<(), InternalError> {
+        match producer_name {
+            MESSAGE_PRODUCER_KAMU_FLOW_CONFIGURATION_SERVICE => {
```
sergiimk, line 41: I want to get rid of mediators so bad. Their logic is so repetitive now ... there has to be a way to generalize it.

sergiimk: I'm imagining something like this:

```rust
catalog_builder
  .add(outbox::Dispatcher::for::<TaskServiceMessage>(MESSAGE_PRODUCER_TASK_SERVICE))
  .add(outbox::Dispatcher::for::<FlowServiceMessage>(MESSAGE_PRODUCER_FLOW_SERVICE))
```

i.e. there is a generic implementation of a "dispatcher" that deserialized JSON into the right type and sends it off to the typed handlers.

And `Dispatcher::for` returns a builder with producer ID contained as metadata, so that outbox service could select the correct dispatcher implementation.

src/domain/flow-system/services/src/messages/flow_message_producers.rs
```diff
@@ -0,0 +1,17 @@
...
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub const MESSAGE_PRODUCER_KAMU_FLOW_CONFIGURATION_SERVICE: &str =
```
sergiimk, line 12: Nitpic: These long constants names make my head heart. Perhaps instead of giving them unique names we could utilizes the module namespaces?

e.g. 
```rust
kamu_flow_system::domain::services::flow::MESSAGE_PRODUCER_ID
kamu_flow_system::domain::services::flow_configuration::MESSAGE_PRODUCER_ID
```

src/utils/messaging-outbox/src/services/implementation/outbox_dispatching_impl.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Outbox for OutboxDispatchingImpl {
+    #[tracing::instrument(level = "debug", skip_all, fields(producer_name, content_json))]
+    async fn post_message_as_json(
+        &self,
+        producer_name: &str,
+        content_json: &serde_json::Value,
+    ) -> Result<(), InternalError> {
+        if self.durable_producers.contains(producer_name) {
+            self.transactional_outbox
+                .post_message_as_json(producer_name, content_json)
```
sergiimk, line 88: Looks like durability is currently a **consumer-side** parameter (determined by `MessageConsumerMeta`, but shouldn't it be **producer-side**?

Message is either a domain event or a push notification, right?

The nature of the message should be determined from the very beginning by the producer, so that all domain events were saved into DB even if no one is consuming them yet.

Whether a message is a domain event or a notification should not depend on a current set of consumers. After all there might be always zero consumers if a publisher is extracted into a standalone service.

src/utils/messaging-outbox/src/services/implementation/outbox_immediate_impl.rs
```diff
@@ -0,0 +1,66 @@
...
+        content_json: &serde_json::Value,
+    ) -> Result<(), InternalError> {
+        let maybe_consumers_mediator = self.mediators_by_producers.get(producer_name);
+        if let Some(consumers_mediator) = maybe_consumers_mediator {
+            consumers_mediator
+                .consume_message(
+                    &self.catalog,
+                    producer_name,
+                    self.consumer_filter,
+                    content_json.clone(),
```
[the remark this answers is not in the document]
zaychenko-sergei: Changed partially. We still need `Value` to serialize/deserialize to JSONB in the databases, but yes, some cloning got reduced.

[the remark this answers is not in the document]
zaychenko-sergei: Practically, no sense. Okay, I removed 1 `Value` clone, but added `Value` -> `String` conversion instead, which is allocating resources anyway. We want to store events as JSONB in the database, and string is not a suitable type for this goal.

src/utils/messaging-outbox/src/consumers/message_consumers_mediator_utils.rs
zaychenko-sergei: yes

src/domain/flow-system/services/src/messages/flow_message_producers.rs
zaychenko-sergei: We are going to have "glob reexport" issues this way, besides, I do not see value in this change, It's looking roughly identical.

src/utils/messaging-outbox/src/services/implementation/outbox_dispatching_impl.rs
zaychenko-sergei: I don't agree, currently there are messages that get delivered via both channels, i.e. dataset updates. Overall, this is highly debatable until we get rid of most in-memory implementations.

src/domain/flow-system/services/src/messages/flow_message_consumer_mediator.rs
zaychenko-sergei: Couldn't solve it with a builder, the solution is a bit more verbose, but still, mediators are gone.

src/utils/messaging-outbox/src/services/implementation/outbox_dispatching_impl.rs
sergiimk: I thought we agreed to pursue an architecture where:
- services can choose between "pure event sourcing" and "state + domain events" design under the hood
- no matter what is primary, events or state, the (public) state is reconcilable via events
- external services synchronize their state via having access to a complete history of such events

I don't see how consumer-side durability fits into this picture. If we write a new service - it has to start producing the outbox events from the start - otherwise when consumers for these events appear they will not have any event history to "catch up" through and will have to have some custom state initialization procedure.

I don't understand how to reason about a system where events are "sometimes persistent" depending on who is listening to them.

PR #742, opened 2024-07-26
Fix transaction issue in ingest flow

rmn-boiko:
## Description

Closes: #issue

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/domain/datasets/domain/src/services/dataset_key_value_service.rs
```diff
@@ -0,0 +1,23 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait DatasetKeyValueService: Sync + Send {
+    async fn get_dataset_env_var_value_by_key<'a>(
+        &self,
+        dataset_env_var_key: &str,
+        dataset_env_vars: &'a [DatasetEnvVar],
+    ) -> Result<Secret<String>, GetDatasetEnvVarError>;
```
zaychenko-sergei, line 22: Not all of those are secrets, so maybe an enum in result could be a little nicer.

zaychenko-sergei, line 22: As you renamed service, change the method name and error name too.

src/domain/datasets/services/src/dataset_key_value_service_impl.rs
```diff
@@ -0,0 +1,73 @@
...
+#[async_trait::async_trait]
+impl DatasetKeyValueService for DatasetKeyValueServiceImpl {
+    async fn get_dataset_env_var_value_by_key<'a>(
+        &self,
+        dataset_env_var_key: &str,
+        dataset_env_vars: &'a [DatasetEnvVar],
+    ) -> Result<Secret<String>, GetDatasetEnvVarError> {
+        if let Some(existing_dataset_env_var) = dataset_env_vars
+            .iter()
+            .find(|dataset_env_var| dataset_env_var.key == dataset_env_var_key)
```
zaychenko-sergei, line 59: Hash map plz

src/domain/task-system/services/src/task_executor_impl.rs
```diff
@@ -99,9 +102,34 @@ impl TaskExecutor for TaskExecutorImpl {
 
             let outcome = match &task.logical_plan {
                 LogicalPlan::UpdateDataset(upd) => {
+                    let dataset_env_vars = DatabaseTransactionRunner::new(self.catalog.clone())
+                        .transactional_with(
+                            |dataset_env_vars_svc: Arc<dyn DatasetEnvVarService>| async move {
+                                let dataset_env_vars = dataset_env_vars_svc
+                                    .get_all_dataset_env_vars_by_dataset_id(
```
zaychenko-sergei, line 109: This code is growing too big, let's split handlers for each type of logical plan into separate methods for the beginning.

src/domain/datasets/domain/src/services/dataset_key_value_service.rs
rmn-boiko: Just a small note. In static implementation where we fetch from env vars we cannot find is it secret or not so I put here as always secret value

src/domain/task-system/services/Cargo.toml
```diff
@@ -22,9 +22,11 @@ doctest = false
 
 
 [dependencies]
+database-common = { workspace = true }
 event-bus = { workspace = true }
 opendatafabric = { workspace = true }
 kamu-core = { workspace = true }
+kamu-datasets-services = { workspace = true }
```
zaychenko-sergei, line 29: Why do we have to link to service implementation crate and not the domain crate?

zaychenko-sergei: I only see imports like 

`use kamu_datasets_services::domain::{DatasetEnvVar, DatasetEnvVarService};`

rmn-boiko: great catch. Thanks for pointing. Probably VScode autoimport did that and I missed to doublecheck. FIxed and double checked

zaychenko-sergei (approved):

Issue #720, opened 2024-07-18
Private Datasets: Implement ReBAC-related data persistent storage + `DatasetEntryRepository` (in-mem variants)


s373r:
The purpose of this ticket is to implement repositories (+ domain services) required for Private Repositories to work.

<details>
  <summary>Scheme proposed in RFC</summary>

```
table datasets:
- id: odf::DatasetID (PK)
- owner_account_id: odf::AccountID (FK to accounts.id)
- alias: String

table rebac_properties:
- id (PK)
- entity_type: String
- entity_id: did::odf
- property_name: String
- property_value: String
// 1, dataset, <did...dataset1>, allows_anonymous_read, true
// 2, dataset, <did...dataset1>, allows_public_read, true

table rebac_relationships:
- id (PK)
- subject_entity_type: String
- subject_entity_id: did::odf
- relationship: String
- object_entity_type: String
- object_entity_id: did::odf
// 1, account, <did...account1>, Reader, dataset, <did...dataset1>
// 2, account, <did...account2>, Editor, dataset, <did...dataset1>
```
  
</details>

In other words, needs to be introduced:
- `DatasetEntryRepository` trait 
  - (and its implementation for in-mem variant).
- `RebacRepository` trait 
  - (and its implementation for in-mem variant).
- `RebacService` trait 
  - (and its implementation) to provide a more convenient concrete interface to work with the ReBAC repo.

---

Based on [the "4.9. Database Structure" RFC chapter](https://docs.google.com/document/d/1bycQpRD015YyKNZD1atRScC2uChVWStAE70UsdZjX0g/edit#heading=h.r6b925568h6e)

PR #750, opened 2024-07-31
Implement ReBAC-related data persistent storage + `DatasetEntryRepository` (in-mem variants)

s373r:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/720

### Added
- Private Datasets, preparation work:
  - Added in-mem implementation of ReBAC repository
  - Added in-mem implementation of `DatasetEntryRepository`

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ❌ will be added after


src/domain/auth-rebac/domain/src/services/rebac_service.rs
```diff
@@ -0,0 +1,94 @@
...
+    PropertyName,
+    PropertyValue,
+    Relation,
+    UpsertEntityPropertyError,
+};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait RebacService: Send + Sync {
```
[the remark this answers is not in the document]
s373r: I conceived of this service to hide the more general repository interface into a more specific of the service.

For example, to avoid constantly passing `entity_type: EntityType, entity_id: EntityId`, for dataset -- to eliminate the possibility (at the interface level) of passing `EntityType::Account` and `EntityId` of dataset (`dataset_id`)

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,194 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// TODO: derives
+
+pub enum EntityType {
+    Dataset,
+    Account,
+}
+
+pub type EntityId = DidOdf;
```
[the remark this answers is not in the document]
s373r: Done

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,194 @@
...
+        entity_id: &EntityId,
+    ) -> Result<Vec<EntityRelation>, GetEntityRelationsError>;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct EntityRelation {
+    pub relation: Relation,
+    pub entity_type: EntityType,
+    pub entity_id: EntityId,
```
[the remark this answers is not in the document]
s373r: Argument structure and naming updated to addressing this comment + improve readability

Fixed

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,194 @@
...
+        subject_entity_id: &EntityId,
+        relationship: &Relation,
+        object_entity_type: EntityType,
+        object_entity_id: &EntityId,
+    ) -> Result<(), DeleteEntitiesRelationError>;
+
+    async fn get_entity_relations(
+        &self,
+        entity_type: EntityType,
+        entity_id: &EntityId,
```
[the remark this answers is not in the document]
s373r: Done

src/domain/datasets/domain/src/repos/dataset_repository.rs
```diff
@@ -0,0 +1,97 @@
...
+use internal_error::InternalError;
+use opendatafabric::{AccountID, DatasetAlias, DatasetID};
+use thiserror::Error;
+
+use crate::Dataset;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait DatasetRepository: Send + Sync {
```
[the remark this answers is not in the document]
s373r: Yes, it was an open question that I wanted to address in the review
Thanks for the detailed comment -- the name seems appropriate to me, I decided to use it

Fixed

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,241 @@
...
+pub enum DeleteEntitiesRelationError {
+    #[error(transparent)]
+    NotFound(EntitiesRelationNotFoundError),
+
+    #[error(transparent)]
+    Internal(InternalError),
+}
+
+#[derive(Error, Debug)]
+#[error("Entities relation not found: {self}")]
```
sergiimk, line 221: Hm, how does this work? I thought `#[error("Blah")]` implements the `Display` trait. By having `{self}` in the error message are you recursively calling display on the error. Will this crash with stack overflow?

src/domain/auth-rebac/domain/src/services/rebac_service.rs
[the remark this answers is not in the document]
sergiimk: Ah I see now. That makes sense.

Although I wonder whether we will still need access to the generic interface to connect this info to OSO.

How OSO integration will work is where I lack clarity the most, so it might be worth as one of your next steps to provide stub implementation of these services and create a test that connects data to the OSO ruleset.

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: Oh, you're right...

> Will this crash with stack overflow?

Unequivocally

s373r: Fixed

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
```diff
@@ -0,0 +1,129 @@
...
+        owner_id: &AccountID,
+    ) -> Result<Vec<DatasetEntry>, GetDatasetEntryError>;
+
+    async fn save_dataset_entry(&self, dataset: &DatasetEntry)
+        -> Result<(), SaveDatasetEntryError>;
+
+    async fn update_dataset_entry_alias(
+        &self,
+        dataset_id: &DatasetID,
+        new_alias: &DatasetAlias,
```
s373r, line 36: While writing tests, caught myself asking a question:

**Q:** Domain question: we can change only the dataset name in the alias using the API, right?

If so, I'll update the implementation (a minor thing)

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
```diff
@@ -0,0 +1,129 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Error, Debug)]
+pub enum UpdateDatasetEntryAliasError {
+    #[error(transparent)]
+    NotFound(#[from] DatasetEntryNotFoundError),
+
+    #[error(transparent)]
+    SameAlias(#[from] DatasetEntryAliasSameError),
```
s373r, line 96: **Q:** Domain question: do we need to check this at the interface/type/contract level?

s373r (commented):
Added 2 domain-related questions

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
s373r: Which should be explicitly added, as comment of this PR:

`DatasetEntryRepository::update_dataset_entry_alias()` как раз будет отвечать за ...

> naming/renaming of datasets

.

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
s373r: > As per my other comment, dataset names are scoped to owner account, so I think this should not speak about aliases, but rather DatasetName collisions.

yes, for sure

> I'm not sure what you mean by "interface/type/contract level". For naming/renaming to be atomic this will need to be handled on data repository level and an informative error will be necessary to give reason to the caller.

Yes, that was my question -- at what level of the interface should this error be triggered. Or even, more precisely, whether this error is needed in general.

But anyway, I got the answer, thanks!

--

next time, I'll phrase such general questions more clearly

s373r: PS. for the record, the error is left as is

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
s373r: Updated

src/infra/auth-rebac/inmem/src/repos/rebac_repository_inmem.rs
```diff
@@ -0,0 +1,495 @@
...
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// EntityHasher
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+struct EntityHasher {}
+
+impl EntityHasher {
```
sergiimk, line 439: This is surprisingly complex...

If I'm not mistaking, unless this is a perfect hash - you have a possibility of collisions that go undetected.

Why not simply use `HashMap<Entity<'static>, ...>`?

Honestly, even storing the whole table as `Vec<(Entity, Relation, Entity)>` with linear search complexity would be fine for the first in-mem implementation.

sergiimk: I might be wrong, but my impression is that using `Cow` in `Entity` type costed you a whole bunch of time that could be better spent elsewhere e.g. derisking OSO integration. 

I like idiomatic Rust, but this kind of optimizations can always be done later when there's a need.

src/infra/datasets/inmem/src/repos/dateset_entry_repository_inmem.rs
```diff
@@ -0,0 +1,164 @@
...
+            .fold(vec![], |mut acc, dataset| {
+                if dataset.owner_id == *owner_id {
+                    acc.push(dataset.clone());
+                }
+
+                acc
+            });
+
+        if dataset_entries.is_empty() {
+            return Err(DatasetEntryNotFoundError::ByOwnerId(owner_id.clone()).into());
```
sergiimk, line 99: This is unnecessarily strict - just return an empty Vec and let user decide if this is a normal situation or an error.

Cargo.lock
```diff
@@ -2796,7 +2796,7 @@ dependencies = [
  "aws-credential-types",
  "aws-sdk-secretsmanager",
  "chrono",
- "dill",
+ "dill 0.8.1",
```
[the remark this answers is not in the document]
s373r: > Check your dill version - this package should not be duplicated

Yes, I tried starting the integration of the new version -- will revert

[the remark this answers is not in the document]
s373r: Fixed

src/domain/auth-rebac/domain/src/services/rebac_service.rs
```diff
@@ -0,0 +1,89 @@
...
+    // Account
+    async fn set_account_property(
+        &self,
+        account_id: &AccountID,
+        property: &Property,
```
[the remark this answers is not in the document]
s373r: A good point, will do

src/domain/core/src/repos/metadata_chain_visitor.rs
```diff
@@ -43,6 +43,9 @@ pub enum MetadataVisitorDecision {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// TODO: PERF: MetadataChainVisitor::visit(): use Cow<HashedMetadataBlock>.
```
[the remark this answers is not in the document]
s373r: My thought was, if we see in `MetadataChainExt::accept_by_interval()` that there is only one visitor, we could pass the block and hash (`(hash, block)`) by value, instead of by reference as now (`(&hash, &block)`).

That way, if only one visitor is used, we could avoid cloning -- this only applies to those who save blocks (e.g. `SearchSingleDataBlockVisitor`).

--

If the idea seems appropriate to you after this comment, I'll update the comment to (add context) to make it clearer

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    async fn get_dataset_entries_by_owner_id(
+        &self,
+        owner_id: &AccountID,
+    ) -> Result<Vec<DatasetEntry>, GetDatasetEntryError>;
+
+    async fn save_dataset_entry(&self, dataset: &DatasetEntry)
+        -> Result<(), SaveDatasetEntryError>;
+
+    async fn update_dataset_entry_alias(
```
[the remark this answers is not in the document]
s373r: Renamed -- Done

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
```diff
@@ -0,0 +1,129 @@
...
+    Internal(InternalError),
+}
+
+#[derive(Error, Debug)]
+pub enum DatasetEntryNotFoundError {
+    #[error("Dataset entry not found by dataset_id: '{0}'")]
+    ByDatasetId(DatasetID),
+
+    #[error("Datasets entry not found by owner_id: '{0}'")]
+    ByOwnerId(AccountID),
```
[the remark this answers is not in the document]
s373r: In principle, yes -- in general, the collection type itself can handle saying it's empty as well

Done

src/infra/datasets/inmem/src/repos/dateset_entry_repository_inmem.rs
s373r: https://github.com/kamu-data/kamu-cli/pull/750#discussion_r1711417572

In short: Done

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
```diff
@@ -0,0 +1,129 @@
...
+    #[error(transparent)]
+    SameAlias(#[from] DatasetEntryAliasSameError),
+
+    #[error(transparent)]
+    Internal(InternalError),
+}
+
+#[derive(Error, Debug)]
+#[error("Dataset entry with dataset_id '{dataset_id}' same alias '{dataset_alias}' update attempt")]
+pub struct DatasetEntryAliasSameError {
```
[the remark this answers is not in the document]
sergiimk: what about name collisions within one owner?

Cargo.lock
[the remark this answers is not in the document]
sergiimk: It's always better to pull in new versions on master and then rebase your branch.
a) you're doing a favor to everyone else
b) you won't have to deal with merge conflicts in cargo.lock

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,494 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Error, Debug)]
+pub enum DeleteEntityPropertyError {
+    #[error(transparent)]
+    EntityNotFound(EntityNotFoundError),
+
+    #[error(transparent)]
+    PropertyNotFound(EntityPropertyNotFoundError),
```
zaychenko-sergei, line 250: Make sure these errors are needed. This might be an overkill as we prefer idempotent handlers.

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,494 @@
...
+        Self::NotFound(entity.into())
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Error, Debug)]
+pub enum InsertEntitiesRelationError {
+    #[error(transparent)]
+    Duplicate(InsertEntitiesRelationDuplicateError),
```
zaychenko-sergei, line 319: Again, isn't this against idempotence?

src/domain/datasets/domain/src/entities/dataset_entry.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use opendatafabric::{AccountID, DatasetID, DatasetName};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DatasetEntry {
+    pub id: DatasetID,
+    pub owner_id: AccountID,
+    pub name: DatasetName,
```
zaychenko-sergei, line 18: Why do we need to store the name here? It does not look like it's used in lookups.

zaychenko-sergei (changes requested):
1) It's unclear how new domain interacts with existing dataset events: creation & deletion in particular. We might also need a rename event, if storing entity name is a must.

2) If we have this first ReBAC repository with dataset->owner entries, do we still need `DatasetOwnershipRepository` ?

src/domain/datasets/domain/src/repos/dataset_entry_repository.rs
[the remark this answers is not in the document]
s373r: fair point -- the error returned

Done

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -0,0 +1,206 @@
...
+    ) -> Result<HashMap<PropertyName, PropertyValue>, GetEntityPropertiesError> {
+        let account_id = account_id.as_did_str().to_stack_string();
+        let account_entity = Entity::new_account(account_id.as_str());
+
+        let properties = self
+            .rebac_repo
+            .get_entity_properties(&account_entity)
+            .await?;
+
+        Ok(represent_properties_as_a_map(properties))
```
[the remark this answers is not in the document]
s373r: > I would avoid making assumptions about what data structure user needs 

very good comment, thank you -- indeed, that's exactly what I did

> Personally I would consider getting rid of `Property` type completely and return `Vec<(PropertyName, PropertyValue)>` from here and from repository

Done

src/domain/auth-rebac/domain/src/services/rebac_service.rs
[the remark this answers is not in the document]
s373r: Done

Cargo.lock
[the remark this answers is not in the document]
s373r: makes sense

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,438 @@
...
+impl<'a> From<EntityWithRelation<'a>> for Entity<'a> {
+    fn from(v: EntityWithRelation<'a>) -> Self {
+        Self {
+            entity_type: v.entity_type,
+            entity_id: v.entity_id,
+        }
+    }
+}
+
+#[reuse(entity)]
```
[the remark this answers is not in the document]
s373r: Done

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,438 @@
...
+            property_name,
+        })
+    }
+}
+
+#[derive(Error, Debug)]
+#[error("Entity not found: entity_type='{entity_type:?}', entity_id='{entity_id}'")]
+pub struct EntityNotFoundError {
+    pub entity_type: EntityType,
+    pub entity_id: String,
```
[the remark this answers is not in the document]
s373r: Thanks for the tip and vector! Realized using `Entity::into_owned()` method

PS. Since `Clone` derive also adds the `ToOwned` implementation, it came out cleaner this way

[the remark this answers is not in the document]
s373r: Fixed

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,438 @@
...
+}
+
+#[derive(Error, Debug)]
+#[error(
+    "Entity property not found: entity_type='{entity_type:?}', entity_id='{entity_id}', \
+     property_name='{property_name:?}'"
+)]
+pub struct EntityPropertyNotFoundError {
+    pub entity_type: EntityType,
+    pub entity_id: String,
```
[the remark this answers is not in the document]
s373r: Fixed

src/domain/auth-rebac/domain/src/services/rebac_service.rs
```diff
@@ -0,0 +1,89 @@
...
+    async fn get_dataset_properties(
+        &self,
+        dataset_id: &DatasetID,
+    ) -> Result<HashMap<PropertyName, PropertyValue>, GetEntityPropertiesError>;
+
+    // Relations
+    async fn insert_account_dataset_relation(
+        &self,
+        account_id: &AccountID,
+        relationship: Relation,
```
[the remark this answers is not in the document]
s373r: Done

src/infra/auth-rebac/inmem/src/repos/rebac_repository_inmem.rs
s373r: > If I'm not mistaking, unless this is a perfect hash - you have a possibility of collisions that go undetected.

I used the default hasher used in `HashSet` (which is based on `HashMap`)

> Why not simply use `HashMap<Entity<'static>, ...>`?

I wanted to get away from the cloning that occurs even at times when lookups -- but, yeah, it was hard to read...

> Honestly, even storing the whole table as Vec<(Entity, Relation, Entity)> with linear search complexity would be fine for the first in-mem implementation.

It will also be linear time for the search, but used `HashSet<T>`. And in general, simplified the implementation, without programmatically created indexes.

> I might be wrong, but my impression is that using `Cow` in `Entity` type costed you a whole bunch

No, `Cow<T>` has nothing to do with it -- I rather wanted to get away from O(N), as well as storing duplicate records in indexes. This implementation didn't take much time.

Still, it's hard to deny that it turned out to be more difficult for the reader than I wanted it to be.

> I like idiomatic Rust, but this kind of optimizations can always be done later when there's a need.

I tried to keep the interfaces also ideomatic, but with simpler implementation

-- 

Fixed

src/infra/auth-rebac/inmem/src/repos/rebac_repository_inmem.rs
```diff
@@ -0,0 +1,495 @@
...
+    index_subject_entity: RowsIndex,
+    index_subject_entity_object_type: RowsIndex,
+    index_subject_entity_object_entity: RowsIndex,
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct RebacRepositoryInMem {
+    entities_properties_map: Arc<RwLock<EntitiesPropertiesMap>>,
+    entities_relations_state: Arc<RwLock<EntitiesRelationsState>>,
```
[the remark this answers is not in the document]
s373r: Fixed

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: > Make sure these errors are needed

For repositories, we usually use errors like "not found" and "duplicate"

> This might be an overkill as we prefer idempotent handlers.

But for `RebacService` the errors will be updated

s373r: `UnsetEntityPropertyError` & `InsertRelationError` added -- Done

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: Replied in another thread: https://github.com/kamu-data/kamu-cli/pull/750#discussion_r1713847812

s373r:
Replies for https://github.com/kamu-data/kamu-cli/pull/750#pullrequestreview-2232251440

> 1. It's unclear how new domain interacts with existing dataset events: creation & deletion in particular. We might also need a rename event, if storing entity name is a must.

As part of this epic, the following events will be added:
- `DatasetOwnerChanged`
- `DatasetNameChanged`

Cross-linking: ["4.1. General" 6) in RFC](https://docs.google.com/document/d/1bycQpRD015YyKNZD1atRScC2uChVWStAE70UsdZjX0g/edit#heading=h.azirovg4csdq)

> If we have this first ReBAC repository with dataset->owner entries, do we still need `DatasetOwnershipRepository` ?

I'm planning to use `RebacService` as a dependency for `DatasetOwnershipServiceInMemory`

--

I'd like to add both things not within the scope of this PR

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,438 @@
...
+        Self::new(EntityType::Account, entity_id, relation)
+    }
+
+    pub fn new_dataset(entity_id: impl Into<Cow<'a, str>>, relation: Relation) -> Self {
+        Self::new(EntityType::Dataset, entity_id, relation)
+    }
+}
+
+pub type ObjectEntity<'a> = Entity<'a>;
+pub type ObjectEntityWithRelation<'a> = EntityWithRelation<'a>;
```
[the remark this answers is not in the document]
s373r: Subjective improvement in readability, e.g. for

```rust
...
    async fn get_subject_entity_relations(
        &self,
        subject_entity: &Entity,
    ) -> Result<Vec<ObjectEntityWithRelation>, SubjectEntityRelationsError>;
...
```

src/domain/datasets/domain/src/entities/dataset_entry.rs
s373r: > Why do we need to store the name here?

Yes, we need it

> It does not look like it's used in lookups.

Added `DatasetEntryRepository::get_dataset_entry()` method -- Fixed

src/domain/core/src/repos/metadata_chain_visitor.rs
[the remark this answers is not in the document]
sergiimk: Let's not spend time on this micro-optimization until we have a clear indication that it's a performance problem

src/infra/auth-rebac/inmem/src/repos/rebac_repository_inmem.rs
sergiimk: > I used the default hasher used in HashSet (which is based on HashMap)

When you do a lookup in a hash map it first looks up a value by hash and then does `Eq` test to check that the right key was found and not just a key with the same hash.

My point was that by using pre-hashed values instead of `Entity<'static>` as keys the `Eq` test would no longer guarantee correct functionality under collisions.

Just making sure we understood one another.

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,378 @@
...
+
+impl<'a> Entity<'a> {
+    pub fn new(entity_type: EntityType, entity_id: impl Into<Cow<'a, str>>) -> Self {
+        Self {
+            entity_type,
+            entity_id: entity_id.into(),
+        }
+    }
+
+    pub fn new_account(entity_id: impl Into<Cow<'a, str>>) -> Self {
```
zaychenko-sergei, line 99: Plz move all entity & relation structs into "entities" module, how we do it in all domains.

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    pub fn new_dataset(entity_id: impl Into<Cow<'a, str>>, relation: Relation) -> Self {
+        let dataset_entity = Entity::new_dataset(entity_id);
+
+        Self::new(dataset_entity, relation)
+    }
+}
+
+pub type ObjectEntity<'a> = Entity<'a>;
+pub type ObjectEntityWithRelation<'a> = EntityWithRelation<'a>;
```
zaychenko-sergei, line 142: What's the value of these typedefs?

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -0,0 +1,378 @@
...
+
+impl From<DatasetPropertyName> for PropertyName {
+    fn from(value: DatasetPropertyName) -> PropertyName {
+        PropertyName::Dataset(value)
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum AccountPropertyName {
+    // TBA
```
zaychenko-sergei, line 190: This could be "admin" indication

zaychenko-sergei (approved):
My extra requests are minor

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: Moved -- Done

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: Since these type-aliases are causing a second question, I decided to remove them

Removed

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: Yes, agree -- but for now, it's a feature of the accounts domain

Added a task in order to make this change:
- https://github.com/kamu-data/kamu-cli/issues/766

Issue #752, opened 2024-08-01
Extend Ingest flow GQL api with fetch uncachable flag


rmn-boiko:
In kamu-cli we have such a thing as `Uncacheable source` - if you ingest data from a REST endpoint that does not provide `ETag` or `LastUpdated` headers, i.e. there is no quick way to check if data changed or not - the source will be skipped as uncacheable upon subsequent pulls unless you specify `--fetch-uncacheable` flag.

But our ingest flow is leaking such configuration. Add new flag in schedule config `fetch_uncacheable` and pass it to ingest options

Issue #515, opened 2024-02-20
GraphQL: document all properties of the schema objects


zaychenko-sergei:
Currently, some elements of the output schema are documented like this:

```

type Account {
	"""
	Unique and stable identifier of this account
	"""
	id: AccountID!
	"""
	Symbolic account name
	"""
	accountName: AccountName!
	"""
	Account name to display
	"""
	displayName: AccountDisplayName!
	"""
	Account type
	"""
	accountType: AccountType!
	"""
	Avatar URL
	"""
	avatarUrl: String
	"""
	Indicates the administrator status
	"""
	isAdmin: Boolean!
}
```

which is auto-derived from Rust source code - documentation comments:

```

    /// Unique and stable identifier of this account
    async fn id(&self) -> &AccountID {
        &self.account_id
    }

    /// Symbolic account name
    async fn account_name(&self) -> &AccountName {
        &self.account_name
    }

    /// Account name to display
    async fn display_name(&self, ctx: &Context<'_>) -> Result<AccountDisplayName> {
        let full_account_info = self.get_full_account_info(ctx).await?;

        Ok(AccountDisplayName::from(
            full_account_info.display_name.clone(),
        ))
    }
```

Signifficant part of the GraphQL API is not documented at all.
This ticket assumes adding meaningful documentation comments to all items of the GraphQL API.

PR #756, opened 2024-08-05
Add uncacheable flag to ingest flow

rmn-boiko:
## Description

Closes: #752

- [x] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

resources/schema.gql
```diff
@@ -1119,6 +1130,7 @@ input FlowRunConfiguration @oneOf {
 	schedule: ScheduleInput
 	batching: BatchingConditionInput
 	compaction: CompactionConditionInput
+	ingest: IngestConditionInput
```
sergiimk, line 1133: You replaced `schedule` in `FlowConfiguration` output enum with `ingest`, but kept both in this one. Is this intentional?

resources/schema.gql
```diff
@@ -1204,6 +1216,11 @@ type GetFlowSuccess implements GetFlowResult {
 }
 
 
+input IngestConditionInput {
+	fetchUncacheable: Boolean!
```
sergiimk, line 1220: Let's document the fields

src/domain/core/src/services/pull_service.rs
```diff
@@ -172,17 +172,25 @@ pub trait PullMultiListener: Send + Sync {
...
     Updated {
         old_head: Option<Multihash>,
         new_head: Multihash,
     },
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub enum UpToDateResult {
+    IngestUpToDate { uncacheable: bool },
+    UpToDate,
```
sergiimk, line 185: I think it will be better to change this type to something like:
```rust
enum PullResultUpToDate {
  PollingIngest(PollingInsgestResultUpToDate),
  PushIngest(PushInsgestResultUpToDate),
  Transform(TransformResultUpToDate),
  Sync(SyncResultUpToDate),
}
```
This will require creating structs like `PollingInsgestResultUpToDate` in the lower-level services, but when we in future add more fields to them we will not have to touch the pull service - they be nicely wrapped already.

Also note the name `PullResultUpToDate` - when creating structs for enum variants I try to always use `EnumNameVariantName` convention, so this will be more consistent with the existing code.

src/domain/task-system/domain/src/entities/task_status.rs
```diff
@@ -57,9 +57,23 @@ pub struct TaskUpdateDatasetResult {
...
+            PullResult::Updated { .. } => TaskUpdateDatasetResult {
+                pull_result: pull_result.clone(),
+            },
+            PullResult::UpToDate(up_to_date_result) => match up_to_date_result {
+                UpToDateResult::UpToDate => TaskUpdateDatasetResult {
+                    pull_result: pull_result.clone(),
+                },
+                UpToDateResult::IngestUpToDate { uncacheable } => TaskUpdateDatasetResult {
+                    pull_result: PullResult::UpToDate(UpToDateResult::IngestUpToDate {
+                        uncacheable: *uncacheable && fetch_uncacheable,
```
sergiimk, line 72: I don't understand why we need this logic. Task should wrap result from the lower level service, but here we are changing the meaning of `uncacheable` flag based on whether `fetch_uncacheable` parameter was specified.

rmn-boiko: We have the same logic in the pull command. It looks reasonable to indicate it was flag provided or not to return have we tried to fetch_uncacheable flag or not.

sergiimk: In the `PullCommand` this logic is present to not complain/warn about uncacheable sources if `--fetch-uncacheable` was already specified - no point to hint user about this flag if they are already using it.

But take into account that `PullCommand` is a UI/presentation layer - its logic is about how to present things that happen in the core most informatively to the user.

In your case we are in the core domain layer. Its job is not to decide what to show to the user but to provide detailed information to the upper layers about what actually happened under the hood.

When you apply this logic - this looks to me as changing the meaning of what `uncacheable` flag means.

I think the best solution is leave this filtering to UI assuming UI will have access to both:
- the task/flow result - that tells whether the source is uncacheable or not
- and the input parameter - whether `fetch-uncachable` was specified

resources/schema.gql
```diff
@@ -552,7 +552,7 @@ type DatasetFlowConfigs {
 }
 
 type DatasetFlowConfigsMut {
-	setConfigSchedule(datasetFlowType: DatasetFlowType!, paused: Boolean!, schedule: ScheduleInput!): SetFlowConfigResult!
+	setConfigIngest(datasetFlowType: DatasetFlowType!, paused: Boolean!, ingest: IngestConditionInput!): SetFlowConfigResult!
```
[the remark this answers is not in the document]
zaychenko-sergei: This certainly breaks the compatibility, but we should not double the costs of our development too just because we want to be cleaner than the Pope.

resources/schema.gql
```diff
@@ -1827,7 +1850,7 @@ type TasksMut {
 	Schedules a task to update the specified dataset by performing polling
 	ingest or a derivative transformation
 	"""
-	createUpdateDatasetTask(datasetId: DatasetID!): Task!
+	createUpdateDatasetTask(datasetId: DatasetID!, fetchUncacheable: Boolean): Task!
```
zaychenko-sergei, line 1853: This wasn't worth it, I think we might remove this API entirely. UI will not allow direct tasks management anyway.

src/domain/flow-system/domain/src/entities/flow_configuration/flow_configuration_rule.rs
```diff
@@ -18,6 +18,7 @@ pub enum FlowConfigurationRule {
     Schedule(Schedule),
     BatchingRule(BatchingRule),
     CompactionRule(CompactionRule),
+    IngestRule(IngestRule),
```
zaychenko-sergei, line 21: Why do we need previous `Schedule` branch ?

src/domain/flow-system/domain/src/entities/flow_configuration/flow_configuration_rule.rs
```diff
@@ -18,6 +18,7 @@ pub enum FlowConfigurationRule {
     Schedule(Schedule),
     BatchingRule(BatchingRule),
```
zaychenko-sergei, line 19: If "schedule" becomes "ingest", then "batching" should probably become "transform".

src/domain/datasets/services/src/dataset_env_var_service_static_impl.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[component(pub)]
+#[interface(dyn DatasetEnvVarService)]
+impl DatasetEnvVarServiceStaticImpl {
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
```
zaychenko-sergei, line 40: I don't think changes in this file belong to this PR.

resources/schema.gql
```diff
@@ -1024,12 +1029,18 @@ type FlowDescriptionSystemGC {
...
+union FlowDescriptionUpdateResult = FlowDescriptionUpdateResultUpToDate | FlowDescriptionUpdateResultSuccess
+
+type FlowDescriptionUpdateResultSuccess {
 	numBlocks: Int!
 	numRecords: Int!
 	updatedWatermark: DateTime
 }
 
+type FlowDescriptionUpdateResultUpToDate {
+	uncacheable: Boolean!
```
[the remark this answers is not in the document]
zaychenko-sergei: #515

resources/schema.gql
zaychenko-sergei: #515

src/domain/flow-system/domain/src/entities/flow_configuration/flow_configuration_rule.rs
rmn-boiko: It is still using for system flow

resources/schema.gql
rmn-boiko: Yeah, in GQL I removed it

Issue #739, opened 2024-07-26
Expose dataset reset method via GQL Api


rmn-boiko:
We have a command `kamu reset <dataset> <hash>` which resets the state of the dataset to a defined state. New flow should ba available for both root and derived dataset kinds.

Extend existing HardCompaction enum with new `Reset` type to trigger reset method. 
  `<hash>` param should be optional and reset to seed block if it was not provided.
  `<recursive>` param. If it is set to true trigger metadata only compacting for each derived dataset.

PR #758, opened 2024-08-07
Add reset dataset flow

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/739

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/domain/flow-system/services/tests/tests/utils/flow_harness_shared.rs
```diff
@@ -181,9 +181,8 @@ impl FlowHarness {
         }
     }
 
-    pub async fn create_root_dataset(&self, dataset_alias: DatasetAlias) -> DatasetID {
-        let result = self
-            .dataset_repo
+    pub async fn create_root_dataset(&self, dataset_alias: DatasetAlias) -> CreateDatasetResult {
```
s373r, line 184: Optional: leave `create_root_dataset()` method as it is, but add a new one (to be used inside `create_root_dataset()` as well) that returns `CreateDatasetResult`.

Motivation, this way we can take away the pressure on the reader from ....
```rust
        .await
        .dataset_handle
        .id;
```
... since, only in one (if I counted correctly) case is `CreateDatasetResult` used.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -757,22 +765,49 @@ impl FlowServiceImpl {
...
                         max_slice_records,
                         keep_metadata_only,
-                    })
+                    }))
+                }
+                DatasetFlowType::Reset => {
+                    let new_head_hash = if let Some(config_rule) = config_snapshot
+                        && let FlowConfigurationSnapshot::Reset(reset_rule) = config_rule
+                    {
+                        Some(reset_rule.new_head_hash.clone())
```
s373r, line 779: **Q:** What if I, as an API user, specify as a new head a hash that is already a head for a dataset?

maybe in that case, we could not perform the action chain (flow -> task -> write files), but immediately reply to the user that "hash is already a head"?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- New `Reset` flow which can be triggered manually for `Root` and `Derived` datasets
```
s373r, line 9: Please indicate that this applies to GraphQL (CHANGELOG.md reader, user, ideally should immediately understand how to use the new added functionality OR understand where to go to find out more details)

s373r (commented):
Looks good! Added one question and a couple of minor comments

src/domain/flow-system/services/tests/tests/utils/flow_harness_shared.rs
rmn-boiko: Changed

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -757,22 +765,49 @@ impl FlowServiceImpl {
...
+                            .await
+                            .int_err()?;
+                        let dataset = self
+                            .dataset_repo
+                            .get_dataset(&dataset_handle.as_local_ref())
+                            .await
+                            .int_err()?;
+                        dataset
+                            .as_metadata_chain()
+                            .try_get_ref(&BlockRef::Head)
```
sergiimk, line 793: This `else` confuses me.

The condition is:
```rust
if let Some(config_rule) = config_snapshot
  && let FlowConfigurationSnapshot::Reset(reset_rule) = config_rule
```
so:
- if a configuration was provided - **when wouldn't it?**
- and configuration is for the correct type of the rule - **why don't we immediately fail the flow if it isn't?**
- then we reset to the block that is already a `head` - **why do all these extra steps to perform a no-op?**

src/domain/task-system/domain/src/entities/logical_plan.rs
```diff
@@ -74,6 +77,15 @@ pub struct HardCompactionDataset {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+/// A task to perform the resetting of a dataset
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ResetDataset {
+    pub dataset_id: DatasetID,
+    pub new_head_hash: Multihash,
```
sergiimk, line 84: Reset is a destructive operation that can lead to data loss.

We better be 100% sure that we are only loosing data that the user wanted to loose and **not one bit more**.

Imagine a situation where:
- User creates a reset flow
- For whatever reason the flow system is stuck for a significant amount of time
- During this time user's IoT device continues to write data and get HTTP 200 responses that data was durably stored
- We fix the flow system and it happily resets the dataset and all the new data that arrived

We should be making **optimistic concurrency control** a part of every destructive flow.

So I think we should add `old_head_hash` parameter to the task, so that it aborts if it sees that dataset has changed since the decision was made to reset the dataset.

The handling of this parameter likely belongs all the way down in `ResetService`, so that in future we could make it fully atomic when `head` pointer is stored in a database.

And I think this parameter should originate from the API caller - only this way we can be 100% sure that we never reset something that user cared about.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- New `Reset` flow in GQL Api which can be triggered manually for `Root` and `Derived` datasets
```
sergiimk, line 9: `Derived` => `Derivative`

sergiimk (changes requested):
A little worrying how much code it takes to add a flow/task that just call one simple method. We'll need to work to improve this in future.

Overall changes are clear and well tested. My major suggestion is to add OCC which should be a very simple change.

src/domain/task-system/domain/src/entities/logical_plan.rs
rmn-boiko: Yeah. It is actually a great improvement. Thanks

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: Yeah, it is logical mistake. It should be seed block as default

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: Done

src/domain/flow-system/services/src/flow/flow_service_impl.rs
sergiimk: I'd recommend to move "seed block as default" logic all the way up to API layer, as it's mostly a user experience thing that we might reconsider in future - it doesn't belong in the core.

src/adapter/graphql/src/mutations/flows_mut/flows_mut_utils.rs
```diff
@@ -98,6 +98,7 @@ pub(crate) async fn ensure_flow_preconditions(
     ctx: &Context<'_>,
     dataset_handle: &DatasetHandle,
     dataset_flow_type: DatasetFlowType,
+    flow_run_configuration: &Option<FlowRunConfiguration>,
```
zaychenko-sergei, line 101: Use `Option<&T>` instead

src/adapter/graphql/src/mutations/flows_mut/flows_mut_utils.rs
```diff
@@ -128,6 +129,46 @@ pub(crate) async fn ensure_flow_preconditions(
...
+                let dataset_repo = from_catalog::<dyn kamu_core::DatasetRepository>(ctx).unwrap();
+
+                let dataset = dataset_repo
+                    .get_dataset(&dataset_handle.as_local_ref())
+                    .await
+                    .int_err()?;
+                let new_head_block_result = dataset
+                    .as_metadata_chain()
+                    .get_block(&reset_configuration.new_head_hash)
+                    .await;
```
zaychenko-sergei, line 145: Method became too large, split it on parts for each flow type.

src/adapter/graphql/src/queries/flows/flow.rs
```diff
@@ -368,3 +385,30 @@ impl FlowDescriptionDatasetHardCompactionResult {
...
+}
+
+impl FlowDescriptionResetResult {
+    fn from_maybe_flow_outcome(maybe_outcome: Option<&fs::FlowOutcome>) -> Option<Self> {
+        if let Some(outcome) = maybe_outcome {
+            match outcome {
+                fs::FlowOutcome::Success(result) => match result {
+                    fs::FlowResult::Empty
+                    | fs::FlowResult::DatasetCompact(_)
+                    | fs::FlowResult::DatasetUpdate(_) => None,
```
zaychenko-sergei, line 401: Shouldn't other types of flow be `unreachable!` ?

src/adapter/graphql/src/mutations/flows_mut/flows_mut_utils.rs
zaychenko-sergei, line 145: We do not actually need to read the block to check if it exists. Not a large read, but still.

There is such a method:

```
#[async_trait]
pub trait MetadataBlockRepository: Send + Sync {
    async fn contains_block(&self, hash: &Multihash) -> Result<bool, ContainsBlockError>;
    ...
}
```

We might need to bring it up through `MetadataChain` interface.

src/domain/flow-system/domain/src/entities/flow/flow_outcome.rs
```diff
@@ -55,13 +58,19 @@ impl FlowResult {
...
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FlowRootDatasetCompactedError {
     pub dataset_id: DatasetID,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FlowResetNewHeadHashNotFoundError {
+    pub head_hash: Multihash,
```
zaychenko-sergei, line 71: "_hash" is more than clear from the data type

src/domain/flow-system/services/src/flow/flow_service_impl.rs
```diff
@@ -44,6 +48,7 @@ pub struct FlowServiceImpl {
     task_scheduler: Arc<dyn TaskScheduler>,
     flow_configuration_service: Arc<dyn FlowConfigurationService>,
     dataset_changes_service: Arc<dyn DatasetChangesService>,
+    dataset_repo: Arc<dyn DatasetRepository>,
```
zaychenko-sergei, line 51: I'm very much against this dependency, if we can avoid it.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
zaychenko-sergei: I agree, and this would allow to revert new dependency + `async` modifier on this method.

src/domain/task-system/domain/src/entities/task_status.rs
```diff
@@ -49,13 +49,18 @@ impl TaskOutcome {
...
     UpdateDatasetResult(TaskUpdateDatasetResult),
+    ResetDatasetResult(TaskResetDatasetResult),
     CompactionDatasetResult(TaskCompactionDatasetResult),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct TaskUpdateDatasetResult {
     pub pull_result: PullResult,
 }
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
zaychenko-sergei, line 60: Blank line missing

zaychenko-sergei (changes requested):
For the first time we have a flow that is dependent on the state of the HEAD.
It might not be a part of this ticket, but, I think we need an event for change of HEAD.
A potential handler in flow/task domains could scan the active queue, and cancel the affected flows/tasks.

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: Sounds great, and issue with tests stuck will not be reproducable

src/domain/flow-system/services/src/flow/flow_service_impl.rs
rmn-boiko: Yeah, it will be removed and used in reset service level

src/adapter/graphql/src/mutations/flows_mut/flows_mut_utils.rs
rmn-boiko: Almost all this logic moved to another levels

zaychenko-sergei (approved):

Issue #727, opened 2024-07-19
Private Datasets: Update use case: Creating a dataset


s373r:
The purpose of this ticket is to add the ability to create a Private Dataset.

In order to accomplish this, we need to update two code paths:

**1) API**

In the GQL API, it is suggested to `add public: Boolean!` argument to `DatasetsMut::createEmpty()` & `DatasetsMut::createFromSnapshot()`.
If the argument is `true`, then the dataset will be considered a Publicly Available Dataset.

**2) CLI (only multi-tenant)**

`kamu add` command: add the hidden `--public` option. So by default (in case of no option), we have Private Datasets.

---

Based on [the "4.2. Use case: Creating a Dataset" RFC chapter](https://docs.google.com/document/d/1bycQpRD015YyKNZD1atRScC2uChVWStAE70UsdZjX0g/edit#heading=h.tkjtbiikv8a9)

PR #761, opened 2024-08-08
Private Datasets: Update use case: Creating a dataset

s373r:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/727

### Added
- Private Datasets, ReBAC integration: 
  - ReBAC properties update based `DatasetLifecycleMessage`'s:
  - `kamu add`: added hidden `--visibility private|public` argument, assumed to be used in multi-tenant case
  - GQL: `DatasetsMut`:
    - `createEmpty()`: added optional `datasetVisibility` argument
    - `createFromSnapshot()`: added optional `datasetVisibility` argument

- [x] Unit and integration tests added ✅
  
- [x] Compatibility:
    
  - [x] Network APIs: :warning: New optional parameters of GQL methods have been added, which will not affect the frontend operation
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ✅ added :arrow_down: 
    - https://github.com/kamu-data/kamu-node/pull/116


src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -336,12 +320,21 @@ impl DatasetRepository for DatasetRepositoryLocalFs {
     async fn create_dataset_from_snapshot(
         &self,
         snapshot: DatasetSnapshot,
+        publicly_available: bool,
```
sergiimk, line 323: Let's introduce `CreateDatasetOpts` structure right away. I think public/private is just the first out of many parameters to come (e.g. storage backend selection), so its better to add `CreateDatasetOpts::default()` in all the bazillion places now and avoid such changes later.

src/adapter/graphql/src/mutations/datasets_mut.rs
```diff
@@ -109,7 +109,11 @@ impl DatasetsMut {
     ) -> Result<CreateDatasetFromSnapshotResult> {
         let dataset_repo = from_catalog::<dyn domain::DatasetRepository>(ctx).unwrap();
 
-        let result = match dataset_repo.create_dataset_from_snapshot(snapshot).await {
+        // TODO: read param from input
```
sergiimk, line 112: you can create an optional parameter (defaulting to private) for compatibility

consider also using an input struct here to anticipate many other creation options to come

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
```diff
@@ -1572,19 +1576,24 @@ impl FlowConfigHarness {
...
                     .push_event(MetadataFactory::set_polling_source().build())
                     .build(),
+                publicly_available,
             )
             .await
             .unwrap()
     }
 
     async fn create_derived_dataset(&self) -> CreateDatasetResult {
+        let publicly_available = true;
```
sergiimk, line 1595: Should we have a `Visibility::Public` enum to avoid the problem of unnamed bool literals?

src/app/cli/src/cli_parser.rs
```diff
@@ -106,6 +107,10 @@ pub fn cli() -> Command {
                             .action(ArgAction::Append)
                             .index(1)
                             .help("Dataset manifest reference(s) (path, or URL)"),
+                        cli_arguments::add::PUBLIC.to_clap_arg_long()
+                            .hide(true)
+                            .action(ArgAction::SetTrue)
```
sergiimk, line 112: How about using `--visibility public` instead of a flag?

If we ever add more kinds of visibility (`external` / `public:rw`?)  - it would allow easily extending without breaking the interface.

src/app/cli/src/cli_arguments.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::ops::Deref;
+
+use clap::Arg;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub mod add {
+    use super::CliArgument;
+
+    pub const PUBLIC: CliArgument = CliArgument::new("public");
```
sergiimk, line 19: This seems a lot more typing than it's worth. I suggest to stick to the existing style for now.

I think in the near future we should instead invest the time to switch to macro-based clap configuration - I already [experimented with it](https://github.com/kamu-data/kamu-node/blob/master/src/app/oracle-provider/src/config.rs) in the oracle provider.

It's a lot more concise and type-safe.

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -53,22 +53,6 @@ impl DatasetRepositoryLocalFs {
         }
     }
 
-    pub fn create(
```
s373r, line 56: Unused

src/infra/core/src/repos/dataset_repository_local_fs.rs
s373r: Yes, it's better to really use that pattern -- absolutely agree

I added the option structure, but within the use cases (I decided to move it there, since after the introduction of usecase as a separate entity, it is better to have additional logic there)

src/app/cli/src/cli_arguments.rs
s373r: Ok, with a sore heart I will return string hardcoded constants :)

s373r: Done

src/adapter/graphql/src/mutations/datasets_mut.rs
s373r: Since all the tests add the in-memory version of the service, I decided to make the default version with `PubliclyAvailable` value.

What do you think?

src/adapter/graphql/tests/tests/test_gql_dataset_flow_configs.rs
s373r: `Default::default()` is now used in all tests

s373r (commented):
Add notes for readers

src/adapter/graphql/src/mutations/datasets_mut.rs
```diff
@@ -94,23 +106,40 @@ impl DatasetsMut {
...
         let create_from_snapshot =
             from_catalog::<dyn domain::CreateDatasetFromSnapshotUseCase>(ctx).unwrap();
 
-        let result = match create_from_snapshot.execute(snapshot).await {
+        let create_options = CreateDatasetFromSnapshotUseCaseOptions {
+            dataset_visibility: if dataset_visibility {
+                DatasetVisibility::PubliclyAvailable
+            } else {
+                DatasetVisibility::Private
+            },
```
zaychenko-sergei, line 136: Minor: might look nicer if encapsulated within the enum itself

src/app/cli/src/cli_value_parser.rs
```diff
@@ -113,3 +132,17 @@ pub(crate) fn value_parse_url(url_str: &str) -> Result<Url, String> {
         }
     }
 }
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub(crate) fn value_parse_dataset_visibility(value: &str) -> Result<DatasetVisibility, String> {
+    let res = match value {
+        "private" => DatasetVisibility::Private,
+        "public" => DatasetVisibility::PubliclyAvailable,
```
zaychenko-sergei, line 141: Again, this unlikely is a responsibility of CLI app, could be enum's implementation detail or some implemented conversion trait

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -14,59 +14,60 @@ use crate::{Entity, EntityType, EntityWithRelation, PropertyName, PropertyValue,
...
         property_name: PropertyName,
     ) -> Result<(), DeleteEntityPropertyError>;
 
-    async fn get_entity_properties(
+    async fn get_entity_properties<'a, 'b>(
         &self,
-        entity: &Entity,
-    ) -> Result<Vec<(PropertyName, PropertyValue)>, GetEntityPropertiesError>;
+        entity: &Entity<'a>,
+    ) -> Result<Vec<(PropertyName, PropertyValue<'b>)>, GetEntityPropertiesError>;
```
zaychenko-sergei, line 38: Try getting rid of these lifetimes, it looks like garbage

src/domain/core/src/use_cases/create_dataset_use_case.rs
```diff
@@ -15,6 +15,10 @@ use crate::{CreateDatasetError, CreateDatasetResult};
 
 #[async_trait::async_trait]
 pub trait CreateDatasetUseCase: Send + Sync {
+    // TODO: Add "options" that specify the visibility dataset
+    //
+    //       Based on: Private Datasets: Update use case: Pushing a dataset
+    //                 https://github.com/kamu-data/kamu-cli/issues/728
```
zaychenko-sergei, line 21: Why is this different from snapshot? I think you should share a common structure

src/domain/core/src/use_cases/create_dataset_from_snapshot_use_case.rs
```diff
@@ -18,7 +18,48 @@ pub trait CreateDatasetFromSnapshotUseCase: Send + Sync {
     async fn execute(
         &self,
         snapshot: DatasetSnapshot,
+        options: &CreateDatasetFromSnapshotUseCaseOptions,
```
zaychenko-sergei, line 21: What is the point of passing by ref? Most of the time you are using temporary objects, like `Default::default()`

src/infra/core/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
```diff
@@ -30,29 +34,69 @@ use crate::DatasetRepositoryWriter;
...
+            self.dataset_repo_writer
+                .delete_dataset(&create_dataset_result.dataset_handle)
+                .await
+                // Return a deletion error if there is one
+                .map_int_err(CreateDatasetFromSnapshotError::Internal)?;
+
+            // On successful deletion, return the property setting error
+            return Err(set_rebac_property_err);
+        }
+
```
zaychenko-sergei, line 88: I think there might be a better approach now using the Outbox.
1) Make uniqueness validation before doing anything. I don't think it's possible to break invairants of ReBac repository like name uniqueiness, if the same check has already passed in the storage repository.

2) Create a dataset in the storage part (disk or S3). If it fails, the entire scenario fails.

3) There already is a message issued into outbox about a created dataset,. It can be extended with extra attributes like publicity, if necessary.

4) Re-Bac part may subscribe to event and create a necessary record. If this process fails from technical issues, like DB connectivity, outbox will ensure retrial until transaction passes.

It would be really beneficial of `core` domain was complete unaware of `rebac` domain.

src/infra/core/Cargo.toml
```diff
@@ -32,15 +32,16 @@ query-extensions-json = ["dep:datafusion-functions-json"]
...
 [dependencies]
 # Kamu
-internal-error = { workspace = true }
 container-runtime = { workspace = true }
-kamu-data-utils = { workspace = true }
-opendatafabric = { workspace = true }
+internal-error = { workspace = true }
 kamu-accounts = { workspace = true }
-kamu-datasets = { workspace = true }
+kamu-auth-rebac = { workspace = true }
```
zaychenko-sergei, line 38: Try not to insert this dependency, and similar in dev version.

src/infra/core/Cargo.toml
```diff
@@ -32,15 +32,16 @@ query-extensions-json = ["dep:datafusion-functions-json"]
...
 kamu-core = { workspace = true }
+kamu-data-utils = { workspace = true }
+kamu-datasets = { workspace = true }
```
zaychenko-sergei, line 41: Similarly, can we avoid depending on this domain?

src/adapter/graphql/src/mutations/datasets_mut.rs
s373r: Yes, indeed -- there's no point in having a flag as an argument if we already have enumeration

Fixed

src/app/cli/src/cli_value_parser.rs
s373r: In this case, I disagree: this is validation of values coming from user input, which refer specifically to CLI parameters

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: > Due to the fact that mockall macros play with lifetimes, leading to impenetrable compilation errors, I had to raise the derive here so that the lifetimes are exactly matched up

(c) https://github.com/kamu-data/kamu-cli/pull/761#discussion_r1724684785

--

Alternatively, I can replace the mock-impl with a real one specifically for my test case -- I'll think about what's best to do, because I'm not sure

src/domain/core/src/use_cases/create_dataset_from_snapshot_use_case.rs
s373r: Indeed -- especially since these objects may not even be merely clonable, but copyable

Fixed

src/infra/core/Cargo.toml
s373r: > kamu-datasets = { workspace = true }

is the `kamu_datasets::{DatasetEnvVar, DatasetKeyValueService};` dependencies for `FetchService` -- I haven't touched this area

what I did: sorting the imports

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: Since the processing of the added dataset has been moved outside of `CreateDatasetFromSnapshotUseCaseImpl`, there is no need for `MockRebacRepository` anymore

Fixed

src/infra/core/Cargo.toml
s373r: Fixed

src/domain/core/src/use_cases/create_dataset_use_case.rs
s373r: Yeah, there's no point in putting it off until this ticket

Done

src/domain/core/src/messages/core_message_types.rs
```diff
@@ -74,3 +82,11 @@ pub struct DatasetLifecycleMessageDeleted {
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DatasetLifecycleMessageRenamed {
+    pub dataset_id: DatasetID,
+    pub new_name: DatasetName,
```
sergiimk: Consider including the old name and the owner account ID+name in the message. The place that performs the renaming will likely have all this info on hand, and including it in the message both simplifies diagnostics and is likely to make the life of consumers much easier and avoid additional lookup.

src/domain/core/src/messages/core_message_types.rs
```diff
@@ -48,6 +55,7 @@ impl Message for DatasetLifecycleMessage {}
 pub struct DatasetLifecycleMessageCreated {
     pub dataset_id: DatasetID,
     pub owner_account_id: AccountID,
+    pub dataset_visibility: DatasetVisibility,
```
sergiimk, line 58: Why don't we include the name of new dataset and the name of the owner account?

src/domain/core/src/messages/core_message_types.rs
sergiimk: Weird, don't know why github was showing me old version of the code for review. This message is gone now, so I will resolve the comment.

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
```diff
@@ -115,6 +120,25 @@ pub struct EntityPropertyNotFoundError {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Error, Debug)]
+pub enum DeleteEntityPropertiesError {
+    #[error(transparent)]
+    NotFound(EntityNotFoundError),
```
sergiimk, line 126: I would not expect to get an error if there are no properties to delete. Let's make APIs idempotent.

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -30,19 +32,76 @@ use kamu_auth_rebac::{
...
+    pub async fn handle_dataset_lifecycle_deleted_message(
+        &self,
+        message: &DatasetLifecycleMessageDeleted,
+    ) -> Result<(), InternalError> {
+        self.delete_dataset_properties(&message.dataset_id)
+            .await
+            .map_err(|err| match err {
+                DeleteEntityPropertiesError::NotFound(e) => e.int_err(),
+                DeleteEntityPropertiesError::Internal(e) => e,
+            })
```
sergiimk, line 104: Can be replaced with `.await.int_err()`.

I think having `DeleteEntityPropertiesError` in the error chain provides valuable information, which you are loosing by extracting the internal error out of the enum.

src/domain/core/src/entities/dataset.rs
```diff
@@ -255,3 +256,30 @@ impl From<CheckpointRef> for Option<OwnedFile> {
...
     }
 }
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Copy, Clone, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub enum DatasetVisibility {
+    #[default]
+    Private,
+    PubliclyAvailable,
```
sergiimk, line 266: Why `PubliclyAvailable` instead of a concise `Public`? I don't see any ambiguity that a longer name would help to resolve.

Also you could add serde tags to allow deserializing these fields from all variations `Pulbic / public / Private / private` and reuse these in CLI.

When we move towards declarative configurations I wouldn't want to spell out `PubliclyAvailable` in the configs.

src/domain/core/src/use_cases/create_dataset_use_case.rs
```diff
@@ -19,7 +19,24 @@ pub trait CreateDatasetUseCase: Send + Sync {
...
+#[derive(Debug, Copy, Clone)]
+pub struct CreateDatasetUseCaseOptions {
+    pub dataset_visibility: DatasetVisibility,
+}
+
+// Used primarily for tests
+impl Default for CreateDatasetUseCaseOptions {
+    fn default() -> Self {
+        Self {
+            dataset_visibility: DatasetVisibility::PubliclyAvailable,
```
sergiimk, line 37: I don't think that convenience of writing tests should dictate what defaults we use.

Our defaults should either:
- favor security by default
- be non-opinionated (i.e. `visibility: None`) letting the service itself decide what default visibility is

The latter option would let us provides the defaults on DI / config level when constructing the service.

src/adapter/graphql/tests/tests/test_gql_account_flow_configs.rs
```diff
@@ -694,7 +696,9 @@ impl FlowConfigHarness {
             .bind::<dyn PollingIngestService, MockPollingIngestService>()
             .add::<DatasetOwnershipServiceInMemory>()
             .add::<DatasetOwnershipServiceInMemoryStateInitializer>()
-            .add::<DatabaseTransactionRunner>();
+            .add::<DatabaseTransactionRunner>()
+            .add::<InMemoryRebacRepository>()
+            .add::<RebacServiceImpl>();
```
zaychenko-sergei, line 701: Do we still need these dependencies in unrelated tests?

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -30,19 +32,76 @@ use kamu_auth_rebac::{
...
+        rebac_repo: Arc<dyn RebacRepository>,
+        dataset_repo_reader: Arc<dyn DatasetRepository>,
+    ) -> Self {
+        Self {
+            rebac_repo,
+            dataset_repo_reader,
+        }
+    }
+
+    pub async fn handle_dataset_lifecycle_created_message(
```
zaychenko-sergei, line 81: Why public?

src/infra/core/tests/tests/use_cases/test_create_dataset_from_snapshot_use_case.rs
```diff
@@ -87,51 +111,180 @@ async fn test_create_derived_dataset_fron_snapshot() {
...
+        let private_create_res = harness.use_case.execute(snapshot, options).await.unwrap();
+
+        assert_matches!(
+            harness
+                .rebac_service
+                .get_dataset_properties(&private_create_res.dataset_handle.id)
+                .await,
+            Ok(props)
+                if props == [PropertyName::dataset_allows_public_read(false)]
+        );
```
zaychenko-sergei, line 180: I don't think use case test is a right place to test rebac logic.
It would be fine to check that the visibility arguments sent to use case on the input are represented in the delivered outbox messages.

The use case, which we are testing here, does not care if Rebac Service exists!

src/infra/core/Cargo.toml
```diff
@@ -151,6 +151,9 @@ libc = "0.2" # For getting uid:gid
 database-common = { workspace = true }
 kamu-accounts-inmem = { workspace = true }
 kamu-accounts-services = { workspace = true }
+kamu-auth-rebac = { workspace = true, default-features = false }
+kamu-auth-rebac-inmem = { workspace = true }
+kamu-auth-rebac-services = { workspace = true }
```
zaychenko-sergei, line 156: These dependencies are unexpected in core tests too.

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -276,20 +294,23 @@ impl MessageConsumerT<DatasetLifecycleMessage> for RebacServiceImpl {
         _: &Catalog,
         message: &DatasetLifecycleMessage,
     ) -> Result<(), InternalError> {
+        let is_single_tenant_workspace = !self.dataset_repo_reader.is_multi_tenant();
+
+        if is_single_tenant_workspace {
```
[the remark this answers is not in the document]
zaychenko-sergei: Yes, this fact is known at initialization time, and does not have to be re-queried on each event.

src/domain/core/src/messages/core_message_types.rs
zaychenko-sergei: It would likely be necessary at the very next step, where we integrate `DatasetEntry` and `MessagingOutbox`.

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: The error type for `RebacService::delete_dataset_properties()` and `RebacService::delete_account_dataset_relation()` has been updated

Done

src/domain/auth-rebac/services/src/rebac_service_impl.rs
s373r: Important comment, I'll pay more attention to things like this, thanks!

Fixed

src/domain/auth-rebac/services/src/rebac_service_impl.rs
s373r: I agree, it's unnecessary

Fixed

src/domain/auth-rebac/services/src/rebac_service_impl.rs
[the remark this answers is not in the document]
s373r: `MultiTenantRebacDatasetLifecycleMessageConsumer` extracted

Done

src/domain/core/src/entities/dataset.rs
s373r: > Why PubliclyAvailable instead of a concise Public? I don't see any ambiguity that a longer name would help to resolve.

Changed -- Done

> Also you could add serde tags to allow deserializing these fields from all variations Pulbic / public / Private / private and reuse these in CLI

Reused -- Done

src/domain/core/src/use_cases/create_dataset_use_case.rs
s373r: > favor security by default

I chose this way because it seemed to me more suitable in the case of use case parameters

--

Fixed

src/infra/core/tests/tests/use_cases/test_create_dataset_from_snapshot_use_case.rs
s373r: > I don't think use case test is a right place to test rebac logic.

Tests moved to `RebacServiceImpl` area -- Done

src/infra/core/Cargo.toml
s373r: Removed -- Done

src/adapter/graphql/tests/tests/test_gql_account_flow_configs.rs
s373r: Nope -- thank you for noticing!

Removed

src/domain/core/src/messages/core_message_types.rs
s373r: > Why don't we include the name of new dataset ...

I was planning to add a dataset name to the message as part of the dataset ownership transfer implementation

> ... and the name of the owner account?

I don't see a case where we can use it yet, but no problem, it could be added

src/domain/auth-rebac/domain/src/repos/rebac_repository.rs
s373r: NB. for the repository, the type remains the same -- idempotency is provided by the service

src/domain/core/src/messages/core_message_types.rs
zaychenko-sergei, line 58: Not critical, but we need to have a general solution for migration.
There is a database with these events already in Europort environment.
At this stage we can still manually reset some tables, but in general we would need a migration path.
In this particular case, it might live with some reasonable deserialization default.

zaychenko-sergei (approved):

src/adapter/graphql/src/mutations/datasets_mut.rs
s373r: Not actual: Private is now used by default

PR #762, opened 2024-08-08
Fix panic for execute transform flow

rmn-boiko:
Fix panic for execute transform flow

### DIFF
--- a/src/app/cli/src/app.rs
+++ b/src/app/cli/src/app.rs
@@ -580,8 +580,8 @@ pub fn register_config_in_catalog(
     let dataset_env_vars_config = config.dataset_env_vars.as_ref().unwrap();
     match dataset_env_vars_config.mode.as_ref().unwrap() {
         DatasetEnvVarsType::Static => {
-            catalog_builder.add::<kamu_datasets_services::DatasetKeyValueServiceStaticImpl>();
-            catalog_builder.add::<kamu_datasets_services::DatasetEnvVarServiceStaticImpl>();
+            catalog_builder.add::<kamu_datasets_services::DatasetKeyValueServiceSysEnv>();
+            catalog_builder.add::<kamu_datasets_services::DatasetEnvVarServiceNull>();
         }
         DatasetEnvVarsType::Storage => {
             assert!(
--- a/src/app/cli/src/app.rs
+++ b/src/app/cli/src/app.rs
@@ -361,8 +361,6 @@ pub fn configure_base_catalog(
 
     b.add::<kamu_accounts_services::LoginPasswordAuthProvider>();
 
-    b.add::<kamu_datasets_services::DatasetEnvVarServiceImpl>();
-
     // No GitHub login possible for single-tenant workspace
     if multi_tenant_workspace {
         if is_e2e_testing {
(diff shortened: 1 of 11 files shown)


src/domain/datasets/services/src/dataset_env_var_service_static_impl.rs
```diff
@@ -0,0 +1,93 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct DatasetEnvVarServiceStaticImpl {}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[component(pub)]
+#[interface(dyn DatasetEnvVarService)]
+impl DatasetEnvVarServiceStaticImpl {
+    #[allow(clippy::needless_pass_by_value)]
```
sergiimk, line 35: linter suppression seems unnecessary here?

src/domain/datasets/services/src/dataset_env_var_service_static_impl.rs
```diff
@@ -0,0 +1,93 @@
...
+    GetDatasetEnvVarError,
+    ModifyDatasetEnvVarError,
+    SaveDatasetEnvVarError,
+};
+use opendatafabric::DatasetID;
+use uuid::Uuid;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct DatasetEnvVarServiceStaticImpl {}
```
sergiimk, line 28: Gotta say that I really dislike the name `Static` - it doesn't tell me what this implementation actually does.

You already have `DatasetKeyValueServiceStaticImpl` and this is a real implementation that fetches variables from system env vars.

In case of this service - it looks more like `Null` implementation, so I would advise naming it `DatasetEnvVarServiceNull` and consider renaming `DatasetKeyValueServiceStaticImpl` to `DatasetKeyValueServiceSysEnv`.

sergiimk: Note that we use suffix `Impl` only when there is exactly one implementation of a service, e.g. `IngestServiceImpl`. When there are multiple we add only the variety suffix without `Impl` e.g. `ObjectRepositoryS3`. 

It's a nitpick, but it's nice to keep the code consistent.

src/app/cli/src/app.rs
```diff
@@ -582,7 +580,8 @@ pub fn register_config_in_catalog(
     let dataset_env_vars_config = config.dataset_env_vars.as_ref().unwrap();
     match dataset_env_vars_config.mode.as_ref().unwrap() {
         DatasetEnvVarsType::Static => {
-            catalog_builder.add::<kamu_datasets_services::DatasetKeyValueServiceStaticImpl>()
+            catalog_builder.add::<kamu_datasets_services::DatasetKeyValueServiceStaticImpl>();
+            catalog_builder.add::<kamu_datasets_services::DatasetEnvVarServiceStaticImpl>();
```
sergiimk, line 584: I assume that a similar change needs to be done in `kamu-node` as well?

sergiimk (approved):
Feel free to merge and release once you follow up on comments

src/app/cli/src/app.rs
rmn-boiko: Yes, thats, correct

PR #764, opened 2024-08-12
Change dataset env var config

rmn-boiko:
Change dataset env var config


src/adapter/graphql/src/utils.rs
```diff
@@ -114,10 +114,10 @@ pub(crate) async fn get_task(
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 pub(crate) fn ensure_dataset_env_vars_mode(ctx: &Context<'_>) -> Result<(), GqlError> {
```
s373r, line 116: what about `ensure_dataset_env_vars_enabled()`?

rmn-boiko: Yeah, good catch. Thanks

src/domain/datasets/domain/src/entities/dataset_env_var.rs
```diff
@@ -232,43 +232,46 @@ impl std::error::Error for AesGcmError {}
...
+        }
+        false
+    }
 }
 
 impl Default for DatasetEnvVarsConfig {
     fn default() -> Self {
         Self {
-            mode: Some(DatasetEnvVarsType::Static),
+            enabled: Some(false),
```
sergiimk, line 274: If we default to `Some(false)` then the feature is disabled by default and we will never warn a user that it was not configured?

rmn-boiko: yeah, changed

zaychenko-sergei (approved):

PR #765, opened 2024-08-12
Add RecursiveCompaction flag to GQL api reset flow

rmn-boiko:
Add `recursive` flag to GQL api reset flow

### DIFF
--- a/resources/schema.gql
+++ b/resources/schema.gql
@@ -959,7 +959,6 @@ type FlowConfigurationCompactionRule {
 type FlowConfigurationReset {
 	newHeadHash: Multihash
 	oldHeadHash: Multihash
-	recursiveCompaction: Boolean!
 	recursive: Boolean!
 }
 
@@ -1562,7 +1561,6 @@ type RequestHeader {
 input ResetConditionInput {
 	newHeadHash: Multihash
 	oldHeadHash: Multihash
-	recursiveCompaction: Boolean!
 	recursive: Boolean!
 }
 
--- a/resources/schema.gql
+++ b/resources/schema.gql
@@ -958,8 +958,9 @@ type FlowConfigurationCompactionRule {
 
 type FlowConfigurationReset {
 	newHeadHash: Multihash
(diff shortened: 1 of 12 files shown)


resources/schema.gql
```diff
@@ -1558,8 +1568,13 @@ type RequestHeader {
 }
 
 input ResetConditionInput {
-	newHeadHash: Multihash
-	oldHeadHash: Multihash
+	mode: ResetPropagationMode!
```
zaychenko-sergei, line 1571: Maybe call it "propagationMode"

resources/schema.gql
```diff
@@ -958,7 +958,17 @@ type FlowConfigurationCompactionRule {
 
 type FlowConfigurationReset {
 	newHeadHash: Multihash
-	oldHeadHash: Multihash!
+	oldHeadHash: Multihash
+	recursive: Boolean!
+}
```
zaychenko-sergei, line 963: This is used only as a view model, but why do we make a difference? I think it should reflect the propagation mode too.

zaychenko-sergei (changes requested):
I think we should refine the names in this enum:

```

enum DownstreamDependencyTriggerType {
    TriggerAllEnabled,
    TriggerOwnUnconditionally,
    Empty,
}
```

First flow is used to trigger update of dependent datasets.
Second flow is used to trigger metadata-only compacting of dependent datasets.
So the names should indicate this clearly.

resources/schema.gql
```diff
@@ -958,7 +958,17 @@ type FlowConfigurationCompactionRule {
...
+
+input FlowConfigurationResetCustom {
+	newHeadHash: Multihash!
+	oldHeadHash: Multihash
+}
+
+input FlowConfigurationResetToSeed {
+	oldHeadHash: Multihash
```
zaychenko-sergei, line 971: If this field is shared, maybe move it up to the struct?

zaychenko-sergei (approved):

Issue #718, opened 2024-07-18
E2E: Flaky test errors for MySQL implementation


s373r:
### Describe the bug

We have periodic errors when working with E2E tests for MySQL. On CI it is enough (at the moment) to restart the test when errors occur, but in local development it is sometimes not enough.

The purpose of this ticket is to investigate the causes of deadlocks and how to fix them.

### Steps To Reproduce

```sh
make test-e2e
```

### Current Behavior

Sometimes, our team members get an error even on re-runs

### Expected Behavior

No issues

### kamu system info

```sh
latest
```

### Logs

```
[...]

PR #767, opened 2024-08-13
E2E: modify a workaround for MySQL tests

s373r:
## Description

Related issue: https://github.com/kamu-data/kamu-cli/issues/718

### Fixed
- E2E: modify a workaround for MySQL tests

- [x] Unit and integration tests added ❌ this change for the tests themselves
  
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ❌ only this repo

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -2,10 +2,24 @@
 fail-fast = false
 
 [test-groups]
+mysql = { max-threads = 1 } # Explanation below
 setup = { max-threads = 1 }  # Serialize the setup steps
 containerized = { max-threads = 8 }  # Don't use too much memory
 engine = { max-threads = 2 }  # Engine tests are very memory-hungry
 database = { max-threads = 8 }  # Don't use too much memory
+
+# TODO: Delete this workaround when this PR is merged:
+#       - Fix: nextest cleanup race condition by bonega
+#         https://github.com/launchbadge/sqlx/pull/3334
+#
+# NOTE: There is an incompatibility between nextest and sqlx:
+#       - nextest implies multiprocessing,
+#       - while sqlx has a lock on cleanup within the current process
+#         (https://github.com/launchbadge/sqlx/pull/2640#issuecomment-1659455042).
+[[profile.default.overrides]]
+filter = "test(::mysql::)"
+test-group = "mysql"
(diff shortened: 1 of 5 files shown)


dmitriy-borzenko:
+

src/app/cli/src/app.rs
```diff
@@ -677,6 +677,7 @@ fn configure_logging(
             .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
             .with_writer(std::io::stderr)
             .pretty()
+            .with_ansi(!no_color_output)
```
dmitriy-borzenko, line 680: Good

.config/nextest.toml
```diff
@@ -2,11 +2,25 @@
...
+# TODO: Delete this workaround when this PR is merged:
+#       - Fix: nextest cleanup race condition by bonega
+#         https://github.com/launchbadge/sqlx/pull/3334
+#
+# NOTE: There is an incompatibility between nextest and sqlx:
+#       - nextest implies multiprocessing,
+#       - while sqlx has a lock on cleanup within the current process
+#         (https://github.com/launchbadge/sqlx/pull/2640#issuecomment-1659455042).
+[[profile.default.overrides]]
+filter = "test(::mysql::)"
```
zaychenko-sergei, line 20: This filter would cover both unit tests and e2e tests. We only have problem with e2e tests. Can we improve tagging/filter to narrow the selection?

s373r: > We only have problem with e2e tests

Not really -- the two groups of tests affect each other. It has to do with the sqlx per-test-binary cleanup, which only shows up on E2E because they take longer to run.

> Can we improve tagging/filter to narrow the selection?

But I'll double-check that possibility

s373r: I double-checked, this solution doesn't work for us -- since the problem is not with the e2e tests themselves in particular, but with sqlx-per-binary (test binary) cleanup

zaychenko-sergei (approved):

Issue #511, opened 2024-02-18
Improve Smart Transfer Protocol error handling


sergiimk:
### Describe the bug

Errors that happen during websocket phase of smart transfer protocol are hard to diagnose both on the client and the server side - seems that some crucial error handling steps are missing.

### Steps To Reproduce

1. `kamu login platform.demo.kamu.dev`
2. auth via GitHub
3. create any dataset (in my case called `gps`)
4. `kamu push gps --to "odf+https://api.demo.kamu.dev/sergiimk/gps"`

### Current Behavior

Client side:
```
kamu push gps --to "odf+https://api.demo.kamu.dev/sergiimk/gps"     
  Failed to sync dataset to repository (gps > odf+https://api.demo.kamu.dev/sergiimk/gps)                         Error: Failed to push 1 dataset(s)

Summary of individual errors:

> Failed to push gps to odf+https://api.demo.kamu.dev/sergiimk/gps:
  0: Internal error
  1: read_error: WebSocket protocol error: Connection reset without closing handshake, push_phase: MetadataRequest)
```

API Server logs:
```
[2024-02-18T00:00:02.963Z] DEBUG: kamu-api-server/7 on kamu-api-server-5559c5b5db-xjqjg: [REQUEST - START] (line=109,method=GET,target=tower_http::trace::make_span,uri=/sergiimk/gps/refs/head,version=HTTP/1.1)
    file: /cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-http-0.4.4/src/trace/make_span.rs
[2024-02-18T00:00:02.963Z] DEBUG: kamu-api-server/7 on kamu-api-server-5559c5b5db-xjqjg: [REQUEST - EVENT] started processing request (line=80,method=GET,target=tower_http::trace::on_request,uri=/sergiimk/gps/refs/head,version=HTTP/1.1)
    file: /cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-http-0.4.4/src/trace/on_request.rs
[2024-02-18T00:00:03.568Z]  WARN: kamu-api-server/7 on kamu-api-server-5559c5b5db-xjqjg: [REQUEST - EVENT] Dataset not found: DatasetNotFoundError { dataset_ref: Alias(DatasetAlias { account_name: Some(AccountName("sergiimk")), dataset_name: DatasetName("gps") }) } (line=149,method=GET,uri=/sergiimk/gps/refs/head,version=HTTP/1.1)
[...]

PR #782, opened 2024-08-20
Improve smart transfer protocol errors

rmn-boiko:
## Description

⚠️ The next changes will break the compatibility with current envs. Merge it only with the following release+deploy procedure.

Closes: https://github.com/kamu-data/kamu-cli/issues/511

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/adapter/http/src/smart_protocol/errors.rs
```diff
@@ -183,6 +184,33 @@ pub enum PushServerError {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Debug, Serialize)]
+pub struct ServerInternalError {
```
sergiimk, line 188: I don't see where this type is used

src/adapter/http/src/smart_protocol/errors.rs
```diff
@@ -183,6 +184,33 @@ pub enum PushServerError {
...
+
+impl ServerInternalError {
+    pub fn new(msg: &str) -> Self {
+        Self {
+            details: msg.to_string(),
+        }
+    }
+}
+
+impl fmt::Display for ServerInternalError {
```
s373r, line 200: what about `#[error("{details}")]`? derive for `ServerInternalError`?

src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
```diff
@@ -199,7 +199,23 @@ impl AxumServerPushProtocolInstance {
             {
                 Ok(head) => Some(head),
                 Err(GetRefError::NotFound(_)) => None,
-                Err(e) => return Err(PushServerError::Internal(e.int_err())),
+                Err(e) => {
+                    let int_err = e.int_err();
+                    axum_write_close_payload::<DatasetPushResponse>(
+                        &mut self.socket,
+                        Err(DatasetPushRequestError::Internal(DatasetInternalError {
+                            error_message: int_err.reason(),
```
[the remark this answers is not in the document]
rmn-boiko: Agree, for internal errors we will return `Server error` with phase when it happend

src/adapter/http/src/smart_protocol/axum_server_pull_protocol.rs
```diff
@@ -48,7 +48,7 @@ impl AxumServerPullProtocolInstance {
         let pull_request = match self.handle_pull_request_initiation().await {
             Ok(pull_request) => pull_request,
             Err(e) => {
-                tracing::debug!("Pull process aborted with error: {}", e);
+                tracing::error!("Pull process aborted with error: {}", e);
```
sergiimk, line 51: If I'm reading this correctly, in case of this and similar errors below we will:
- Log only top-level error message (which could be a non-informative `Internal error`) without a backtrace
- `return` out of the function which will cause socket to close abruptly, not providing any information to the client

src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
```diff
@@ -74,7 +74,20 @@ impl AxumServerPushProtocolInstance {
...
+                            phase: int_err.phase.clone(),
+                            error_message: "Internal error".to_string(),
+                        })),
+                    )
+                    .await
+                    {
+                        tracing::error!("Failed to send error to client with error: {}", write_err);
+                    };
+                }
+                tracing::error!("Push process aborted with error: {}", e);
```
sergiimk, line 90: It's best to avoid formatting strings into error messages and use structured approach instead. This will let us easier filter logs by error types.

We should follow the same format as in [other APIs](https://github.com/kamu-data/kamu-cli/blob/90dd02d8fd37d68a99cfac625f5cf41f431b5b75/src/utils/http-common/src/api_error.rs#L140-L145):
```rust
tracing::error!(
  error = ?err,
  error_msg = %err,
  "Push process aborted with error",
);
```

The code above is not ideal, but it would be consistent with other places, and via `?err` it will include the full debug representation of the error, so that we'll see the whole error chain and backtraces.

src/adapter/http/src/middleware/headers.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub type BearerHeader = TypedHeader<Authorization<Bearer>>;
+pub type VersionHeaderTyped = TypedHeader<VersionHeader>;
+
+#[derive(Debug)]
+pub struct VersionHeader(pub i32);
+
+impl Header for VersionHeader {
+    fn name() -> &'static HeaderName {
+        static NAME: HeaderName = HeaderName::from_static("x-websocket-version");
```
sergiimk, line 23: Very nice use of typed headers!

`x-websocket-version` sounds like we're versioning the WS protocol itself (which has its own `Sec-WebSocket-Version` header)

maybe let's name it `x-odf-smtp-version`?

sergiimk: And mb change the type name to `OdfSmtpVersion` to make it clear what we're versioning

src/adapter/http/src/smart_protocol/messages.rs
```diff
@@ -12,6 +12,13 @@ use opendatafabric::{DatasetID, Multihash};
 use serde::{Deserialize, Serialize};
 use url::Url;
 
+use super::phases::TransferPhase;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub const SMART_TRANSFER_PROTOCOL_CLIENT_VERSION: i32 = 1;
+pub const SMART_TRANSFER_PROTOCOL_SERVER_VERSION: i32 = 1;
```
sergiimk, line 20: Why do we need to two separate versions for client and server?

src/adapter/http/src/simple_protocol/handlers.rs
```diff
@@ -273,4 +280,8 @@ fn get_base_dataset_url(
     base_url_rest.join(path_string.as_str()).unwrap()
 }
 
+fn ensure_version_compatibility(client_version: i32) -> bool {
```
sergiimk, line 283: Could return `Result<(), ApiError>` instead of `bool`.

Also consider making this a method of `OdfSmtpVersion` header - this will make it easier to find.

src/adapter/http/src/smart_protocol/axum_server_pull_protocol.rs
```diff
@@ -45,40 +46,70 @@ impl AxumServerPullProtocolInstance {
...
+        // we wait for client acknowledgment. And to handle
+        // custom clients without logic of close acknowledgment
+        // we will give 5 secs timeout
+        let timeout_duration = Duration::from_secs(5);
+        tokio::select! {
+            _ = wait_for_close(&mut self.socket) => {}
+            _ = tokio::time::sleep(timeout_duration) => {
+                tracing::debug!("Timeout reached, closing connection");
             }
         };
```
sergiimk, line 94: A better way to express this might be via [tokio::time::timeout](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html)

src/adapter/http/src/smart_protocol/axum_server_pull_protocol.rs
```diff
@@ -45,40 +46,70 @@ impl AxumServerPullProtocolInstance {
...
-            Ok(pull_request) => pull_request,
-            Err(e) => {
-                tracing::debug!("Pull process aborted with error: {}", e);
-                return;
+        match self.pull_main_flow().await {
+            Ok(_) => {
+                tracing::debug!("Pull process success");
+            }
+            Err(e) => match &e {
+                PullServerError::Internal(int_err) => {
```
sergiimk, line 54: You can un-nest this match as:
```rust
Err(e @ PullServerError::Internal(_)) => ...
Err(e) => ...
```

src/adapter/http/src/smart_protocol/axum_server_pull_protocol.rs
```diff
@@ -110,7 +141,7 @@ impl AxumServerPullProtocolInstance {
         let head = metadata_chain
             .resolve_ref(&BlockRef::Head)
             .await
-            .map_err(|e| PullServerError::Internal(e.int_err()))?;
+            .protocol_int_err(TransferPhase::Pull(PullPhase::InitialRequest))?;
```
sergiimk, line 144: Nice!

To make this even better you can change the signature:
```rust
fn protocol_int_err(phase: Into<TransferPhase>)
```
so you could call it as `.protocol_int_err(PullPhase::InitialRequest)` without needing wrap into enum manually.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Many examples were simplified due to ident normalization changes
 ### Fixed
 - Crash in `kamu login` command on 5XX server responses
+- The push smart protocol now delivers internal errors to the client
```
sergiimk, line 44: Question for @zaychenko-sergei:

Should we merge and deploy this change (which I assume will include deploying a new version of Jupyter, as we're breaking the protocol), or should we put it on a `smtp-improvements` branch and accumulate further improvements there to be merged and deployed together in a single batch.

Do you think some other improvements that we planned will need breaking protocol changes?

src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
```diff
@@ -75,9 +78,54 @@ impl AxumServerPushProtocolInstance {
             Ok(_) => {
                 tracing::debug!("Push process success");
             }
-            Err(e) => {
-                tracing::debug!("Push process aborted with error: {}", e);
-            }
+            Err(e) => match &e {
+                PushServerError::Internal(int_err) => {
```
sergiimk, line 82: same here

src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
```diff
@@ -140,7 +193,7 @@ impl AxumServerPushProtocolInstance {
                     Ok(create_result) => self.dataset = Some(create_result.dataset),
                     Err(err) => {
                         if let CreateDatasetError::RefCollision(err) = &err {
-                            axum_write_close_payload::<DatasetPushObjectsTransferResponse>(
+                            axum_write_payload::<DatasetPushObjectsTransferResponse>(
```
sergiimk, line 196: In the pull flow it looks like we no longer write errors manually into the channel or do complex error mapping - we just use `?` and the handler in `serve` takes care of the rest.

I wonder why this didn't happen for push flow.

src/adapter/http/src/smart_protocol/ws_tungstenite_client.rs
```diff
@@ -765,6 +796,7 @@ impl SmartTransferProtocolClient for WsSmartTransferProtocolClient {
                 return Err(SyncError::Internal(e.int_err()));
             }
         };
+        println!("@@ finish push");
```
sergiimk, line 799: debug code

src/adapter/http/tests/tests/test_protocol_dataset_helpers.rs
```diff
@@ -120,6 +122,10 @@ async fn test_object_url_local_fs() {
             physical_hash == test_case.data_slice_object().physical_hash &&
             download_from_url == test_case.dataset_url.join(format!("data/{}", physical_hash.as_multibase()).as_str()).unwrap() &&
             download_from_headers == vec![
+                messages::HeaderRow {
+                    name: VersionHeader::name().to_string(),
```
sergiimk, line 126: Let's add a test for version compatibility check

sergiimk: Also would like to see a test that simulates an internal error on the server and checks that client gets the message

src/adapter/http/src/smart_protocol/messages.rs
rmn-boiko: Yeah, agree, currently we don't need two separate versions

src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
rmn-boiko: Yeah, changed

src/adapter/http/tests/tests/test_protocol_dataset_helpers.rs
rmn-boiko: added

zaychenko-sergei (approved):

CHANGELOG.md
zaychenko-sergei: We are not winning much from feature branch complications.

PR #784, opened 2024-08-20
Normalize column names on read

sergiimk:
## Description

See for context: https://github.com/apache/datafusion/issues/7460

To me this looks like Postgres behavior is somewhere in between case-sensitive and case-insensitive. This causes very unpleasant issues in our ingest as all fields containing uppercase letters have to be quoted in SQL queries, merge strategies, and vocab columns.

The proposed solution lowercases the column names on read (when possible without creating collisions), which is consistent with DataFusion's [`enable_ident_normalization`](https://datafusion.apache.org/user-guide/configs.html) config option which is on by default.

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ❌ - This may break some examples - I will do thorough testing before merging
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
### DIFF
--- a/src/infra/core/src/ingest/push_ingest_service_impl.rs
+++ b/src/infra/core/src/ingest/push_ingest_service_impl.rs
@@ -368,6 +368,7 @@ impl PushIngestServiceImpl {
             .await?;
 
         let df = reader.read(input_data_path).await?;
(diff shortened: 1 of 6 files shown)


sergiimk:
Closing in favor of #785

Issue #753, opened 2024-08-01
Private Datasets: use SQLite storage by default for multi-tenant


s373r:
Depend on https://github.com/kamu-data/kamu-cli/issues/754

---

Currently, we default to in-memory data storage (`*InMem` repositories).

The purpose of this ticket is to change the default storage to SQLite (unless otherwise specified).

---

Based on [the "4.1. General" RFC chapter](https://docs.google.com/document/d/1bycQpRD015YyKNZD1atRScC2uChVWStAE70UsdZjX0g/edit#heading=h.azirovg4csdq)

PR #787, opened 2024-08-21
Feature/753 private datasets use sq lite storage by default for multi tenant

s373r:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/753

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/app/cli/build.rs
```diff
@@ -18,5 +18,8 @@ fn main() -> Result<(), Box<dyn Error>> {
         .all_rustc()
         .all_cargo()
         .emit()?;
+
+    println!("cargo:rerun-if-changed=migrations");
```
zaychenko-sergei, line 22: ?

src/app/cli/src/app.rs
```diff
@@ -444,6 +473,12 @@ pub fn configure_cli_catalog(
 
 #[tracing::instrument(level = "info", skip_all)]
 async fn initialize_components(cli_catalog: &Catalog) -> Result<(), CLIError> {
+    // Pre-initialization of the singleton is necessary to avoid a situation where,
+    // during the execution of short commands (e.g., `kamu add`),
+    // the service instance is created with a catalog that holds a transaction,
+    // which could potentially lead to a transaction leak.
+    let _flow_service = cli_catalog.get_one::<kamu_flow_system_services::FlowServiceImpl>()?;
+
```
zaychenko-sergei, line 481: I think we can do better.
There are several dependencies like
  - flow service
  - flow configuration service
  - task scheduler
  - task executor
  - ...
which are never activated via CLI commands other than API server or embedded UI.

I think what could be done is to split the catalog with server-only dependencies and CLI base dependencies.
Server could chain the base CLI dependencies catalog.
Once that is  possible, there should not be a problem with the flow service in your scenario.

src/app/cli/src/database.rs
```diff
@@ -229,3 +232,24 @@ fn init_database_password_provider(b: &mut CatalogBuilder, raw_db_config: &Datab
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+static SQLITE_MIGRATOR: Migrator = sqlx::migrate!("../../../migrations/sqlite");
```
zaychenko-sergei, line 236: Is this embedding migration files into the executable?

src/app/cli/src/app.rs
```diff
@@ -156,10 +167,13 @@ pub async fn run(
         let base_catalog = base_catalog_builder.build();
 
         // Database requires extra actions:
-        let final_base_catalog = if let Some(db_config) = config.database {
+        let final_base_catalog = if let Some(db_config) = database_config {
             // Connect database and obtain a connection pool
             let catalog_with_pool = connect_database_initially(&base_catalog).await?;
 
+            // Apply migrations to the connected database
+            apply_migrations(&catalog_with_pool).await;
```
zaychenko-sergei, line 175: Maybe this should be hidden in database plugin?

src/app/cli/src/services/config/models.rs
```diff
@@ -606,6 +606,12 @@ impl DatabaseConfig {
             port: Some(DatabaseProvider::Postgres.default_port()),
         })
     }
+
+    pub fn sqlite_database() -> Self {
+        Self::Sqlite(SqliteDatabaseConfig {
+            database_path: "kamu.sqlite.db".into(),
```
zaychenko-sergei, line 612: If we use this defualt path in CLI, it would place the file in the working directory.
I think in CLI mode we need to place the database within the `.kamu` workspace folder.

src/app/cli/build.rs
s373r: Updated this piece to add descriptions of what was going on

Done

src/app/cli/src/database.rs
s373r: Yes

src/app/cli/src/services/config/models.rs
s373r: Yes, fair comment -- I originally did that, but ran into issues and decided to postpone. 

Now I realize that this was the right direction and should be implemented that way. Will do

src/app/cli/src/app.rs
s373r: > I think what could be done is to split the catalog with server-only dependencies and CLI base dependencies.

This is a very cool suggestion. This way, in case of sending Outbox Messages, we won't have "unexpected" consumers.

src/app/cli/src/app.rs
s373r: Yeah, it's going to be the best place

Moved; Done

src/app/cli/src/services/config/models.rs
s373r: Done, with commentary: the temporary directory is used for cases where we don't have a workspace yet, followed by a database move

src/app/cli/src/app.rs
```diff
@@ -654,6 +730,58 @@ pub fn register_config_in_catalog(
     ));
 }
 
+fn get_app_database_config(
+    config: &config::CLIConfig,
+    multi_tenant_workspace: bool,
+    init_command: bool,
+) -> AppDatabaseConfig {
```
zaychenko-sergei, line 737: Maybe move to `database.rs`, since it concentrates everything about the databases.

src/app/cli/src/explore/web_ui_server.rs
```diff
@@ -59,7 +59,7 @@ struct WebUILoginInstructions {
 struct WebUIFeatureFlags {
     enable_logout: bool,
     enable_scheduling: bool,
-    enable_dataset_env_vars_managment: bool,
+    enable_dataset_env_vars_management: bool,
```
zaychenko-sergei, line 62: Don't just change it here, make a PR in Web UI repo.

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
```diff
@@ -77,32 +88,47 @@ impl OutboxTransactionalProcessor {
...
+        self.run_with_mode(RunMode::SingleIterationOnly).await
+    }
+
+    #[tracing::instrument(level = "debug", skip_all)]
+    async fn run_with_mode(&self, mode: RunMode) -> Result<(), InternalError> {
         // Trace current routes
         self.debug_message_routes();
 
         // Make sure consumption records represent the routes
         self.init_consumption_records().await?;
```
zaychenko-sergei, line 111: I think I've moved these lines to new `pre_run` part. Careful when merging.

src/app/cli/src/app.rs
```diff
@@ -87,13 +99,16 @@ pub async fn run(workspace_layout: WorkspaceLayout, args: cli::Cli) -> Result<()
 
     prepare_run_dir(&workspace_layout.run_info_dir);
 
-    let maybe_db_connection_settings = config
-        .database
-        .clone()
+    let is_init_command = maybe_init_multi_tenant_flag.is_some();
+    let app_database_config =
+        get_app_database_config(&config, is_multi_tenant_workspace, is_init_command);
+    let (database_config, maybe_sqlite_db_temp_dir) = app_database_config.into_inner();
```
zaychenko-sergei, line 105: Minor: mentining SQlite at this particular place looks like too low-level detail. Can we hide it better?

zaychenko-sergei (changes requested):
```
pub async fn run(workspace_layout: WorkspaceLayout, args: cli::Cli) -> Result<(), CLIError> {
```

The main run function is becoming a monster with every merge of "important mechanism".
I'd like to see some suggestions how do we decompose it and improve abstractions.

src/app/cli/src/app.rs
s373r: Done

src/app/cli/src/explore/web_ui_server.rs
s373r: Extracted into a dedicated issue: https://github.com/kamu-data/kamu-cli/issues/841

s373r: The change was reverted

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
s373r: Thanks for the comment, I've taken into account

src/app/cli/src/app.rs
s373r, line 1: >  `pub async fn run(workspace_layout: WorkspaceLayout, args: cli::Cli) -> Result<(), CLIError> {`
>
> The main run function is becoming a monster with every merge of "important mechanism".
> I'd like to see some suggestions how do we decompose it and improve abstractions.

(c) https://github.com/kamu-data/kamu-cli/pull/787#pullrequestreview-2290754122

The request has been allocated to a separate ticket:
- https://github.com/kamu-data/kamu-cli/issues/845

src/app/cli/src/app.rs
s373r: Agree. Did some refactoring to partially hide the extra context sticking out

Done

src/app/cli/src/explore/api_server.rs
```diff
@@ -191,7 +191,9 @@ impl APIServer {
     pub async fn pre_run(&self) -> Result<(), InternalError> {
         self.task_executor.pre_run().await?;
         self.flow_executor.pre_run(self.time_source.now()).await?;
-        self.outbox_processor.pre_run().await?;
+        // NOTE: We don't need `self.outbox_executor.pre_run()` here,
+        //       since it has already been made.
+
```
zaychenko-sergei, line 196: Then maybe TE and FE `pre_run` would move to `initialize_server_components` ?

src/app/cli/src/app.rs
```diff
@@ -173,10 +187,28 @@ pub async fn run(workspace_layout: WorkspaceLayout, args: cli::Cli) -> Result<()
...
+        (
+            guards,
+            final_base_catalog,
+            cli_catalog,
+            maybe_server_catalog,
+            output_config,
+        )
     };
 
+    initialize_base_components(&base_catalog).await?;
```
zaychenko-sergei, line 210: I've recently added logging to the original call.
There's no error printed anywhere, if this fails, i.e. due to database connectivity issue.

zaychenko-sergei (approved):

src/app/cli/src/explore/api_server.rs
s373r: That's a good idea!

s373r: Done

src/app/cli/src/app.rs
s373r: I've also added logging for this method as well -- at this point, it's a common problem since logging is initiated far from immediately in this (`run()`) method

This is scheduled to be fixed in
- https://github.com/kamu-data/kamu-cli/issues/845

Issue #790, opened 2024-08-22
Automatically coerce `event_time` columns to timestamps


sergiimk:
When ingesting a CSV/JSON file and not specifying the schema, dataset that already contains `event_time` column will fail with error:
```
Event time column 'event_time' should be either Date or Timestamp, but found: Utf8
```
As part of the inference process it would be reasonable to try to coerce the column into a timestamp automatically.

PR #791, opened 2024-08-23
Auto rename and coerce columns on ingest

sergiimk:
## Description

Closes: #790

If user does not declare `read` schema or `preprocess` step (which is always the case when ingesting data from a file upload) we apply the following new inference rules:
- if `event_time` column is present - we will try to coerce it into a timestamp
  - strings will be parsed as RFC3339 date-times
  - integers will be treated as UNIX timestamps in seconds
- if any of the columns conflicts with system column - it will get renamed with an underscore prefix

While I like being strict about names and types, I think this should improve the overall user experience.

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
    will add a paragraph to public docs if this gets merged
  
- [x] Downstream effects:
    
### DIFF
--- a/src/adapter/http/src/data/ingest_handler.rs
+++ b/src/adapter/http/src/data/ingest_handler.rs
@@ -117,6 +117,7 @@ pub async fn dataset_ingest_handler(
                 media_type: arguments.media_type,
                 source_event_time,
                 auto_create_push_source: is_ingest_from_upload,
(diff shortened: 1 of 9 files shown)


zaychenko-sergei (approved):

zaychenko-sergei:
Changelog missing

PR #794, opened 2024-08-27
Set schema early

sergiimk:
## Description

Problem scenario:
- I created datasets for each of our GH repo that track who gave it a "star"
- I then combined them all into one "community" derivative dataset
- Because some of our repos are mostly internal - they may not have any stars
- This causes a problem - currently if the source of a root dataset does not produce any data - the `SetDataSchema` event is not written - and datasets without schema will block derivative transformations from executing

In this PR:
- I change push/pull ingest to define schema as soon as possible, even if input data is empty using the `read` stage schema
- I changed DF engine to produce Parquet files even if output DataFrame is empty - this way I can propagate output schema to kamu
- kamu now handles engines propagating empty Parquet files to define derivative dataset schema, even if the dataset stays empty
- I eliminate the hack where one of the old Parquet files was passed as a "Schema carrier" to the engine - instead kamu will always write an empty Parquet file using schema from `SetDataSchema`

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ❌
    Requires new DF engine. I will consider updating Flink, Spark, RW too
- [x] Documentation:
    
- [x] Downstream effects:
    
### DIFF
--- a/src/domain/core/src/entities/engine.rs
+++ b/src/domain/core/src/entities/engine.rs
@@ -123,10 +123,10 @@ pub struct TransformRequestInputExt {
     /// `(prevOffset, newOffset]` interval of data records that will be
     /// considered in this transaction.
     pub new_offset: Option<u64>,
+    /// Arrow schema of the slices
+    pub schema: SchemaRef,
     /// List of data files that will be read
     pub data_slices: Vec<Multihash>,
-    /// TODO: replace with actual schema
-    pub schema_slice: Multihash,
     /// TODO: remove?
     pub explicit_watermarks: Vec<Watermark>,
 }
@@ -137,9 +137,13 @@ pub struct TransformResponseExt {
     pub new_offset_interval: Option<OffsetInterval>,
     /// Watermark advanced by the transaction, if any
     pub new_watermark: Option<DateTime<Utc>>,
+    /// Schema of the output
+    /// TODO: This field should be made required once all engines are updated
+    pub output_schema: Option<SchemaRef>,
     /// New checkpoint written by the engine, if any
     pub new_checkpoint: Option<OwnedFile>,
-    /// Data produced by the operation, if any
+    /// Data produced by the operation, if any. Must be `None` if offset
+    /// interval is empty.
     pub new_data: Option<OwnedFile>,
 }
 
--- a/src/infra/core/src/engine/engine_odf.rs
+++ b/src/infra/core/src/engine/engine_odf.rs
@@ -180,28 +180,49 @@ impl ODFEngine {
         new_data_path: PathBuf,
         new_checkpoint_path: PathBuf,
     ) -> Result<TransformResponseExt, EngineError> {
-        let new_data = if engine_response.new_offset_interval.is_some() {
-            if !new_data_path.exists() {
-                return Err(EngineError::contract_error(
-                    "Engine did not write a response data file",
-                    Vec::new(),
-                ));
-            }
+        let (new_data, output_schema) = if new_data_path.exists() {
             if new_data_path.is_symlink() || !new_data_path.is_file() {
                 return Err(EngineError::contract_error(
                     "Engine wrote data not as a plain file",
                     Vec::new(),
                 ));
             }
-            Some(OwnedFile::new(new_data_path))
+
+            // Read output schema
+            let output_schema =
+                datafusion::parquet::arrow::arrow_reader::ArrowReaderMetadata::load(
+                    &std::fs::File::open(&new_data_path).int_err()?,
+                    Default::default(),
+                )
+                .int_err()?
+                .schema()
+                .clone();
+
+            let new_data = OwnedFile::new(new_data_path);
+
+            // Clean up the data file if it's empty - it was produced only to provide us the
+            // schema
+            let new_data = if engine_response.new_offset_interval.is_none() {
+                None
+            } else {
+                Some(new_data)
+            };
+
+            (new_data, Some(output_schema))
         } else {
-            if new_data_path.exists() {
+            if engine_response.new_offset_interval.is_some() {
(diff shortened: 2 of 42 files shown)


src/domain/core/src/entities/engine.rs
```diff
@@ -137,9 +137,13 @@ pub struct TransformResponseExt {
     pub new_offset_interval: Option<OffsetInterval>,
     /// Watermark advanced by the transaction, if any
     pub new_watermark: Option<DateTime<Utc>>,
+    /// Schema of the output
+    /// TODO: This field should be made required once all engines are updated
+    pub schema: Option<SchemaRef>,
```
zaychenko-sergei, line 142: Let's be more explicit about the naming. The comment is saying "schema of the output", so maybe the field could be named `output_schema` to be expressive.

src/infra/core/src/testing/dataset_data_helper.rs
```diff
@@ -93,6 +110,18 @@ impl DatasetDataHelper {
             .unwrap()
     }
 
+    pub async fn get_last_schema(&self) -> SchemaRef {
```
zaychenko-sergei, line 113: Maybe "latest", "relevant".

src/infra/ingest-datafusion/src/writer.rs
```diff
@@ -721,98 +721,117 @@ impl DataWriter for DataWriterDataFusion {
...
-                            block_ref: &self.block_ref,
-                            system_time: Some(staged.system_time),
-                            prev_block_hash: Some(Some(&self.meta.head)),
-                            check_object_refs: false,
-                            update_block_ref: true,
-                        },
-                    )
-                    .await?;
+        // Commit schema changes if any
+        if let Some(new_schema) = staged.new_schema {
```
zaychenko-sergei, line 760: I don't think I understood what is the scenario when schema evolves suddenly?
Previously we discussed this is a complex topic of dataset data part migrations..

sergiimk: This is not about schema evolution yet (will clarify the comment).

This is purely about writing `SetDataSchema` event even if we don't have anything to write into `AddData` yet for datasets that stay empty for a long time.

src/domain/core/src/entities/engine.rs
sergiimk: done

src/infra/core/src/testing/dataset_data_helper.rs
sergiimk: renamed to `get_latest_set_schema` to refer to the event type more clearly (not to be confused with retrieving a schema from Parquet file)

src/infra/core/src/ingest/fetch_service.rs
```diff
@@ -523,9 +523,13 @@ impl FetchService {
 
         let has_more = new_has_more_data_path.exists();
 
-        if target_path.metadata().int_err()?.len() == 0
-            && prev_source_state == source_state.as_ref()
-        {
+        if source_state.is_some() && source_state.as_ref() == prev_source_state {
```
s373r, line 526: Minor: we can remove `source_state.is_some()` (the condition branch), since `as_ref()` already contains this branch

src/infra/core/src/testing/dataset_data_helper.rs
```diff
@@ -29,6 +32,20 @@ impl DatasetDataHelper {
...
+            .dataset
+            .as_metadata_chain()
+            .iter_blocks()
+            .filter_data_stream_blocks()
+            .filter_map_ok(|(_, b)| b.event.new_data.map(|_| 1))
+            .try_collect()
+            .await
+            .unwrap();
+
+        v.iter().sum()
```
s373r, line 46: `tokio_stream::stream_ext::StreamExt` have `fold()` and maybe more importantly `try_fold()`.

I think this might be a better solution, especially for IoT cases where we can have a potentially huge number of blocks (hence a huge allocated vector)

sergiimk: you got me, I got annoyed with trying to make `try_fold` work and just because this is a test code wrote a very lazy solution - revisited now

src/infra/core/src/ingest/polling_ingest_service_impl.rs
```diff
@@ -540,6 +536,28 @@ impl PollingIngestServiceImpl {
             )
             .await?;
 
+        if !input_data_path.exists() || input_data_path.metadata().int_err()?.len() == 0 {
```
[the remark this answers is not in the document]
sergiimk: done

src/infra/core/src/ingest/fetch_service.rs
sergiimk: The idea here is: we consider the source "up-to-date" if its state did not advance (e.g. ETH block number stayed the same).

We can have a situation of "uncacheable" source that doesn't have a state - in this case empty output data should be considered an actual valid output - `empty source  != up-to-date`.

So `is_some()` is important to ensure that `source_state` and `prev_source_state` are not `None`.

src/infra/core/src/engine/engine_odf.rs
```diff
@@ -180,28 +180,49 @@ impl ODFEngine {
...
+                    "Engine did not write a response data file",
                     Vec::new(),
                 ));
             }
-            None
+
+            tracing::warn!(
+                "Engine didn't write a data file when output slice is empty - in future this will \
+                 be considered an error"
+            );
```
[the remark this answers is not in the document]
sergiimk: I created tickets in engine repos and will chip away at them later. Currently making datafusion ingest + transform work is sufficient to unblock work on social data pipelines.

PR #795, opened 2024-08-27
Release 0.198.0 (focus on the "address RUSTSEC-2024-0363")

s373r:
### Changed
- If a polling/push source does not declare a `read` schema or a `preprocess` step (which is the case when ingesting data from a file upload) we apply the following new inference rules:
  - If `event_time` column is present - we will try to coerce it into a timestamp:
    - strings will be parsed as RFC3339 date-times
    - integers will be treated as UNIX timestamps in seconds
  - Columns with names that conflict with system columns will get renamed
- All tests related to databases use the `database_transactional_test` macro
- Some skipped tests will now also be run
- Access token with duplicate names can be created if such name exists but was revoked (now for MySQL as well)
- :warning: Updated `sqlx` crate to address [RUSTSEC-2024-0363](https://rustsec.org/advisories/RUSTSEC-2024-0363)
### Fixed
- Derivative transform crash when input datasets have `AddData` events but don't have any Parquet files yet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,9 +141,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.27"
+version = "0.1.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b515e82c8468ddb6ff8db21c78a5997442f113fd8471fd5b2261b2602dd0c67"
+checksum = "bb07629a5d0645d29f68d2fb6f4d0cf15c89ec0965be915f303967180929743f"
 dependencies = [
  "num_enum",
  "strum 0.26.3",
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -8933,14 +8934,14 @@ checksum = "01b2e185515564f15375f593fb966b5718bc624ba77fe49fa4616ad619690554"
...
  "proc-macro2",
  "quote",
- "syn 2.0.75",
+ "syn 2.0.76",
 ]
 
 [[package]]
 name = "sqlx"
-version = "0.8.0"
+version = "0.8.1"
```
s373r, line 8942: `sqlx@0.8.1` contains the fix

rmn-boiko (approved):
Just to notice if you are going to release `kamu-node` don't forget to remove `"RUSTSEC-2024-0363"` from deny.toml as well

PR #796, opened 2024-08-28
Chore/fetch service and dto cleanup

sergiimk:
In this PR:
- I'm removing dynamic trait wrappers that we've been generating for years for ODF DTOs
  - the original idea was to use them for wrapping Flatbuffers to read metadata without any copying, but they are likely not the right approach
  - they are constantly getting in the way of IDE autocomplete
- I'm breaking apart `FetchService` into multiple files per source type
  - This is primarily done to eliminate compilation warnings when compiling with some features disabled
  - But also I'm preparing to make fetch source registration dynamic, replacing one monolithic service with multiple via DI


src/infra/core/src/ingest/fetch_service/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+
+    source_config: Arc<SourceConfig>,
+    http_source_config: Arc<HttpSourceConfig>,
+
+    #[cfg_attr(not(feature = "ingest-evm"), allow(dead_code))]
+    eth_source_config: Arc<EthereumSourceConfig>,
+
+    #[cfg_attr(not(feature = "ingest-mqtt"), allow(dead_code))]
+    mqtt_source_config: Arc<MqttSourceConfig>,
+}
```
zaychenko-sergei, line 59: I definitely do not like when `mod.rs` contains anything but module or macro declarations.

src/infra/core/src/ingest/fetch_service/evm.rs
```diff
@@ -0,0 +1,300 @@
...
+use internal_error::{ErrorIntoInternal, ResultIntoInternal};
+use kamu_core::*;
+use kamu_datasets::DatasetEnvVar;
+use opendatafabric::*;
+
+use super::*;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+impl FetchService {
```
s373r, line 24: Note: Interesting approach, with the implementation of different methods, in different modules

src/infra/core/src/ingest/fetch_service/evm.rs
```diff
@@ -0,0 +1,300 @@
...
+    // `DataFrame::execute_stream()` should not only yield data batches, but also
+    // provide access to the state of the underlying stream for us to know how
+    // far in the block range we have scanned. Since currently we can't access
+    // the state of streaming - we can't interrupt the stream to resume later.
+    // The implementation therefore uses DataFusion only to analyze SQL and
+    // convert the WHERE clause into a filter, and then calls ETH RPC directly
+    // to scan through block ranges.
+    //
+    // TODO: Account for re-orgs
+    #[cfg(feature = "ingest-evm")]
```
s373r, line 38: We can remove this conditional compilation since the module itself is hidden behind the feature gate (in `mod.rs`)

src/infra/core/src/ingest/fetch_service/ftp.rs
```diff
@@ -0,0 +1,111 @@
...
+use internal_error::*;
+use kamu_core::*;
+use url::Url;
+
+use super::*;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+impl FetchService {
+    #[cfg(feature = "ingest-ftp")]
```
s373r, line 23: ditto

s373r (commented):
Added minor comments

src/infra/core/src/ingest/fetch_service/evm.rs
sergiimk: yep, that was my original intention but I guess I got distracted

src/infra/core/src/ingest/fetch_service/mod.rs
sergiimk: As we discussed, I'm also not a fan of code in `mod.rs`, and did this to allow other files to access private functions and types in `mod`.

I refactored to move all main logic from `mod.rs` into `fetch_service/core.rs` and made private methods `pub(super)` (first time using such modifier) to make them accessible to sibling modules.

src/infra/core/src/ingest/fetch_service/evm.rs
sergiimk: I would not recommend this pattern. This is just a transitional step towards separating these into different `Source` types.

I think multiple `impl` blocks are justified only in `opendatafabric` where we add extra functionality to code-generated types.

PR #797, opened 2024-08-28
Use dataset env vars in container sources

sergiimk:
Note: this is based off my refactoring branch

Initially I was just going to support templating in containerized source's env vars, but I realized that the container fetch was not converted to use the new vars service and was remaining a severe security threat as it continued to pull data from system env.

The fix took me longer than expected as I was fighting the API.

Will leave some inline comments to explain.

### DIFF
--- a/src/domain/datasets/domain/src/services/dataset_key_value_service.rs
+++ b/src/domain/datasets/domain/src/services/dataset_key_value_service.rs
@@ -16,12 +16,11 @@ use crate::{DatasetEnvVar, DatasetEnvVarNotFoundError, DatasetEnvVarValue};
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#[async_trait::async_trait]
 pub trait DatasetKeyValueService: Sync + Send {
-    async fn find_dataset_env_var_value_by_key<'a>(
+    fn find_dataset_env_var_value_by_key(
         &self,
         dataset_env_var_key: &str,
-        dataset_env_vars: &'a HashMap<String, DatasetEnvVar>,
+        dataset_env_vars: &HashMap<String, DatasetEnvVar>,
     ) -> Result<DatasetEnvVarValue, FindDatasetEnvVarError>;
 }
 
@@ -30,10 +29,10 @@ pub trait DatasetKeyValueService: Sync + Send {
 #[derive(Error, Debug)]
 pub enum FindDatasetEnvVarError {
     #[error(transparent)]
-    NotFound(DatasetEnvVarNotFoundError),
+    NotFound(#[from] DatasetEnvVarNotFoundError),
 
     #[error(transparent)]
-    Internal(InternalError),
+    Internal(#[from] InternalError),
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/domain/datasets/services/src/dataset_key_value_service_sys_env.rs
+++ b/src/domain/datasets/services/src/dataset_key_value_service_sys_env.rs
@@ -10,9 +10,10 @@
 use std::collections::HashMap;
 
 use dill::*;
-use internal_error::ErrorIntoInternal;
+use internal_error::*;
 use kamu_datasets::{
     DatasetEnvVar,
+    DatasetEnvVarNotFoundError,
     DatasetEnvVarValue,
     DatasetKeyValueService,
     FindDatasetEnvVarError,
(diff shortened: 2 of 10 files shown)


src/domain/datasets/domain/src/services/dataset_key_value_service.rs
```diff
@@ -16,12 +16,11 @@ use crate::{DatasetEnvVar, DatasetEnvVarNotFoundError, DatasetEnvVarValue};
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#[async_trait::async_trait]
 pub trait DatasetKeyValueService: Sync + Send {
-    async fn find_dataset_env_var_value_by_key<'a>(
```
sergiimk, line 21: If API assumes that all vars are extracted and then decrypted on-demand then this method should not be async.

When I see a low-level method that templates a URL string turn `async` - this is a clear alarm bell that we did not separate logic from infrastructure correctly.

src/domain/datasets/services/src/dataset_key_value_service_sys_env.rs
```diff
@@ -35,15 +36,28 @@ impl DatasetKeyValueServiceSysEnv {
...
 impl DatasetKeyValueService for DatasetKeyValueServiceSysEnv {
-    async fn find_dataset_env_var_value_by_key<'a>(
+    fn find_dataset_env_var_value_by_key(
         &self,
         dataset_env_var_key: &str,
-        _dataset_env_vars: &'a HashMap<String, DatasetEnvVar>,
+        dataset_env_vars: &HashMap<String, DatasetEnvVar>,
     ) -> Result<DatasetEnvVarValue, FindDatasetEnvVarError> {
+        // Search explicitly provided vars first
+        if let Some(var) = dataset_env_vars.get(dataset_env_var_key) {
```
sergiimk, line 46: This is just to simplify testing. I was hoping for a simpler way to pass predefined vars into the `FetchService`, but this was the simplest way without a major refactoring.

src/infra/core/src/ingest/fetch_service/container.rs
```diff
@@ -89,13 +90,13 @@ impl FetchService {
         if let Some(env_vars) = &fetch.env {
             for EnvVar { name, value } in env_vars {
                 let value = if let Some(value) = value {
-                    Cow::from(value)
+                    self.template_string(value, dataset_env_vars)?
                 } else {
-                    // TODO: This is insecure: passthrough envvars
-                    let value =
-                        std::env::var(name).map_err(|_| IngestParameterNotFound::new(name))?;
```
sergiimk, line 96: Goodbye kubernetes secrets

src/infra/core/src/ingest/fetch_service/mod.rs
```diff
@@ -235,11 +233,11 @@ impl FetchService {
 
                 if let Some(cenv) = re_env.captures(ctpl.get(1).unwrap().as_str().trim()) {
                     let env_name = cenv.get(1).unwrap().as_str();
+
                     let dataset_env_var_secret_value = self
                         .dataset_key_value_svc
-                        .find_dataset_env_var_value_by_key(env_name, dataset_env_vars)
-                        .await
-                        .int_err()?;
```
sergiimk, line 242: This service even had an error defined for this case already - `IngestParameterNotFound` and it was regressed into internal error :(

src/domain/datasets/services/src/dataset_key_value_service_sys_env.rs
```diff
@@ -35,15 +36,28 @@ impl DatasetKeyValueServiceSysEnv {
...
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#[async_trait::async_trait]
 impl DatasetKeyValueService for DatasetKeyValueServiceSysEnv {
-    async fn find_dataset_env_var_value_by_key<'a>(
+    fn find_dataset_env_var_value_by_key(
         &self,
         dataset_env_var_key: &str,
-        _dataset_env_vars: &'a HashMap<String, DatasetEnvVar>,
+        dataset_env_vars: &HashMap<String, DatasetEnvVar>,
```
sergiimk, line 43: I found the code of this service extremely verbose. If we are in `DatasetKeyValueService` and in method called `find_dataset_env_var_value_by_key` instead of `dataset_env_var_key` we could call parameter `key` without any ambiguity.

src/infra/core/tests/tests/ingest/test_fetch.rs
```diff
@@ -823,6 +823,77 @@ async fn test_fetch_container_ok() {
...
+            &Utc::now(),
+            &HashMap::from([(
+                "KAMU_TEST_PROPAGATED".to_owned(),
+                kamu_datasets::DatasetEnvVar::new(
+                    "KAMU_TEST_PROPAGATED",
+                    Utc::now(),
+                    &kamu_datasets::DatasetEnvVarValue::Regular("foobar".to_owned()),
+                    &DatasetID::new_seeded_ed25519(b"doesnt-matter"),
+                    "",
+                )
```
sergiimk, line 870: Having this `HashMap` in parameters feels to me as if we're passing around a part of (what should've been an) internal state of the `DatasetEnvVars`.

zaychenko-sergei (approved):

src/domain/datasets/services/src/dataset_key_value_service_sys_env.rs
```diff
@@ -35,15 +36,28 @@ impl DatasetKeyValueServiceSysEnv {
...
     ) -> Result<DatasetEnvVarValue, FindDatasetEnvVarError> {
+        // Search explicitly provided vars first
+        if let Some(var) = dataset_env_vars.get(dataset_env_var_key) {
+            return Ok(DatasetEnvVarValue::Regular(
+                var.get_non_secret_value()
+                    .expect("Unexpected encrypted value"),
+            ));
+        }
+
+        // Fall back to system environment
```
rmn-boiko, line 53: Not sure it is safe to put fallback logic where we fetch env from env vars.
As we discussed we shouldn't allow API to work with secrets as env vars. Let's say API will require some envs like `PORT` etc and in the dataset, we will have a similar secret. If the user forgets to set such a value it will be read from env vars and it will be hard to identify

sergiimk: This change is not adding a fallback to read from sys env, but rather adding an option to read values from expliciltly-defined variables in the HashMap before fetching from sys env.

The `DatasetKeyValueServiceImpl` is the one that should be used in `api-server` and that one should never fall back to sys env.

PR #808, opened 2024-08-30
CLI refactor to clap derive macro

sergiimk:
## Description

Our CLI parsing was verbose and fragile. This change makes it type-safe.

This refactoring was long and it sucked, but had to bite the bullet as we would suffer more with the upcoming IaC changes that affect CLI significantly.

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    

src/app/cli/src/cli_commands.rs
```diff
@@ -7,39 +7,36 @@
...
-            submatches.get_one("name").cloned(),
-            submatches.get_flag("recursive"),
-            submatches.get_flag("replace"),
-            submatches.get_flag("stdin"),
+            c.manifest,
+            c.name,
+            c.recursive,
+            c.replace,
+            c.stdin,
             // Safety: This argument has a default value
```
s373r, line 35: Minor: outdated comment

src/app/cli/src/cli_commands.rs
```diff
@@ -574,45 +488,34 @@ pub fn get_command(
...
 }
 
-pub fn command_needs_transaction(arg_matches: &clap::ArgMatches) -> Result<bool, CLIError> {
-    match arg_matches.subcommand() {
-        Some(("system", system_matches)) => match system_matches.subcommand() {
-            Some(("generate-token", _)) => Ok(true),
-            Some(_) => Ok(false),
-            None => Err(CommandInterpretationFailed.into()),
-        },
-        Some(("delete", _)) => Ok(true),
```
s373r, line 613: We need a transaction for deletion: saving the Outbox messages requires a transaction, in case of running with databases. But I don't mind if you delete in this PR.

In https://github.com/kamu-data/kamu-cli/pull/787/, I add processing of the Outbox messages (as well as the transaction for `add`), I can also return back `delete`.

src/app/cli/src/output/output_config.rs
```diff
@@ -26,6 +26,18 @@ pub struct OutputConfig {
     pub trace_file: Option<PathBuf>,
 }
 
+impl Default for OutputConfig {
+    fn default() -> Self {
+        Self {
+            quiet: false,
+            verbosity_level: 0,
+            is_tty: false,
```
s373r, line 34: Q: should we use true as the default value?

s373r: While I can see that this is not a new change, but a legacy -- still, it is interesting to ask

src/app/cli/tests/tests/test_generate_cli_markdown.rs
```diff
@@ -107,14 +108,21 @@ fn build_command_markdown(
                 continue;
             }
 
+            let aliases = subcommand.get_visible_aliases().collect::<Vec<_>>();
```
s373r, line 111: Note: cool stuff, I like it

src/app/cli/src/output/output_config.rs
sergiimk: `OutputConfig::default()` is currently used in just one test, and we probably never want to have fancy TTY output in tests.

src/app/cli/src/cli_commands.rs
sergiimk: Thanks for noticing - removing delete was not intentional. Fixed.

src/app/cli/src/cli_commands.rs
sergiimk: fixed

Issue #724, opened 2024-07-19
Private Datasets: GQL API: Ability to change dataset visibility


s373r:
```gql
"""
Types
"""
input PrivateDatasetVisibilityInput {
	dummy: String
}

input PublicDatasetVisibilityInput {
	anonymousAvailable: Boolean!
}

input DatasetVisibilityInput @oneOf {
	private: PrivateDatasetVisibilityInput
	public: PublicDatasetVisibilityInput
}

"""
Methods
"""
DatasetMut::setVisibility(visibility: DatasetVisibilityInput!): SetDatasetPropertyResultSuccess!
```

---

Based on [the "4.5. Use case: Changing dataset visibility" RFC chapter](https://docs.google.com/document/d/1bycQpRD015YyKNZD1atRScC2uChVWStAE70UsdZjX0g/edit?pli=1#heading=h.9ruhckqn8hyp)

---

- [ ] Update `test_smart_push_visibility()` test

s373r:
Done via https://github.com/kamu-data/kamu-cli/pull/814

PR #814, opened 2024-09-04
Private Datasets: GQL API: Ability to change dataset visibility

s373r:
## Description

Closes: #724 #875 #876

- Speed up project build time by removing unused dependencies which were not detected by automated tools
- Private Datasets:
  - Deletion of previously created ReBAC-properties and reindexing
  - OSO: updating the schema to use identifiers instead of names
  - OSO: added resource storage for access speed
  - E2E: Using the correct account in multi-tenant mode
    - And also the possibility of set it up
  - GQL, added `DatasetMut.setVisibility()` to be able to update ReBAC properties


src/adapter/auth-oso-rebac/src/oso_dataset_authorizer.rs
```diff
@@ -8,67 +8,96 @@
...
+    #[expect(clippy::needless_pass_by_value)]
     pub fn new(
         kamu_auth_oso: Arc<KamuAuthOso>,
         current_account_subject: Arc<CurrentAccountSubject>,
+        oso_resource_holder: Arc<OsoResourceServiceInMem>,
     ) -> Self {
         Self {
             oso: kamu_auth_oso.oso.clone(),
             current_account_subject,
+            oso_resource_service: oso_resource_holder,
```
zaychenko-sergei, line 47: Naming difference intentional?

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
```diff
@@ -0,0 +1,124 @@
...
+    )]
+    async fn run_initialization(&self) -> Result<(), InternalError> {
+        use futures::TryStreamExt;
+
+        let dataset_entries = self
+            .dataset_entry_repository
+            .get_dataset_entries(PaginationOpts::all())
+            .await
+            .try_collect::<Vec<_>>()
+            .await?;
```
zaychenko-sergei, line 78: I think this needs to be split on pages. Don't load all datasets at once, some day we will have tens of thousands.

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
```diff
@@ -0,0 +1,124 @@
...
+        let accounts = self.account_repository.get_accounts().await.int_err()?;
+        let accounts_map = accounts
+            .into_iter()
+            .fold(HashMap::new(), |mut acc, account| {
+                acc.insert(account.id.clone(), account);
+                acc
+            });
```
zaychenko-sergei, line 85: I would avoid loading all accounts in general and rather load them on-demand with small batches. I think there is a high probability there will be many accounts that do not pocess any of datasets, and just query existing.

src/adapter/auth-oso-rebac/src/oso_resource_service_inmem.rs
```diff
@@ -0,0 +1,222 @@
...
+struct State {
+    user_actor_map: HashMap<EntityId, UserActor>,
+    dataset_resource_map: HashMap<EntityId, DatasetResource>,
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct OsoResourceServiceInMem {
+    state: RwLock<State>,
+}
```
zaychenko-sergei, line 50: Is this really a service? Looks like an in-mem repository to me.

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
zaychenko-sergei: Such orphan users, which are not owning any datasets, and neither are given access to private datasets as viewers/editors, could be mimiced without storing.

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        let mut dataset_resources = Vec::with_capacity(dataset_entries.len());
+
+        for dataset_entry in dataset_entries {
+            let properties = self
+                .rebac_service
+                .get_dataset_properties(&dataset_entry.id)
+                .await
+                .int_err()?;
```
zaychenko-sergei, line 94: I don't see where do we access ReBac relations?

src/adapter/graphql/src/mutations/dataset_mut/dataset_mut_utils.rs
```diff
@@ -0,0 +1,36 @@
...
+pub(crate) async fn ensure_account_owns_dataset(
+    ctx: &Context<'_>,
+    dataset_handle: &odf::DatasetHandle,
+) -> Result<()> {
+    let dataset_ownership_service = from_catalog::<dyn DatasetOwnershipService>(ctx).unwrap();
+    let logged_account = utils::get_logged_account(ctx);
+
+    let not_owner = !dataset_ownership_service
+        .is_dataset_owned_by(&dataset_handle.id, &logged_account.account_id)
+        .await?;
```
zaychenko-sergei, line 27: I was hopping we get rid of the `DatasetOwnershipService` completely via ReBAC queries, but I see it's still popular for some reason.

src/domain/accounts/domain/src/repos/account_repository.rs
```diff
@@ -21,6 +21,9 @@ use crate::Account;
 pub trait AccountRepository: Send + Sync {
     async fn create_account(&self, account: &Account) -> Result<(), CreateAccountError>;
 
+    // TODO: Private Datasets: tests & stream version
+    async fn get_accounts(&self) -> Result<Vec<Account>, GetAccountsError>;
+
```
zaychenko-sergei, line 26: Overall: having a routine that returns all accounts on the platform is quite an indication of a suboptimal design from performance point of view.

Definitely this should be a stream, rather than a vector. Don't load too much into memory, we recently discussed this.

Also, consider paging from the beginning.

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
s373r: I plan to integrate this after this PR

resources/schema.gql
```diff
@@ -730,6 +738,11 @@ type DatasetPermissions {
 	canSchedule: Boolean!
 }
 
+type DatasetProperties {
+	allowsAnonymousRead: Boolean!
+	allowsPublicRead: Boolean!
+}
+
```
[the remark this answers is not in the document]
s373r: This is an assumption made earlier about UI needs. But I'm getting ahead of things. This will be deleted

src/adapter/graphql/src/mutations/dataset_mut/dataset_mut_utils.rs
s373r: I was planning on doing it a little later

resources/schema.gql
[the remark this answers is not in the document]
s373r: Done f947bdcb62b5782e8e9c17c241d3057e7ff36e4e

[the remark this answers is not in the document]
s373r: Deleted

CHANGELOG.md
```diff
@@ -14,6 +14,15 @@ Recommendation: for ease of reading, use the following order:
 ## [Unreleased]
 ### Added
 - Console warning when deleting datasets which are out of sync with their push remotes
+### Changed
```
s373r, line 17: Note: I plan to update this list, prior to merging into the feature branch

s373r (commented):
Add comments for readers to the second iteration of the review

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
s373r: No more `oso_resource_service_initializator.rs` as properties are now loaded vectorially and page-by-page.

It turns out to be fixed, but in a different way

src/adapter/auth-oso-rebac/src/oso_resource_service_initializator.rs
s373r: Replaced with user actor cache

Done

src/adapter/auth-oso-rebac/src/oso_resource_service_inmem.rs
s373r: `OsoResourceServiceInMem` has been renamed `OsoResourceServiceImpl` and its implementation has become more service-like

src/domain/accounts/domain/src/repos/account_repository.rs
s373r: A variant of page-by-page loading into the stream is proposed

src/adapter/auth-oso-rebac/src/oso_dataset_authorizer.rs
```diff
@@ -0,0 +1,240 @@
...
+            .get_multiple_dataset_resources(&dataset_ids)
+            .await
+            .int_err()?;
+        let mut dataset_handle_id_mapping =
+            dataset_handles
+                .into_iter()
+                .fold(HashMap::new(), |mut acc, hdl| {
+                    acc.insert(hdl.id.clone(), hdl);
+                    acc
+                });
```
zaychenko-sergei, line 153: Like this syntax

src/adapter/auth-oso-rebac/src/oso_dataset_authorizer.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        for (dataset_id, dataset_resource) in dataset_resources_resolution.resolved_resources {
+            let is_allowed = self
+                .kamu_auth_oso
+                .is_allowed(user_actor.clone(), action, dataset_resource)
```
zaychenko-sergei, line 158: What is the overhead now within OSO, when we heave pre-loaded resource objects?
Is this rebuilding Polar? Or just evaluating the resource on a prebuilt structure?

src/adapter/graphql/src/queries/search.rs
```diff
@@ -32,13 +33,19 @@ impl Search {
...
+            dyn domain::auth::DatasetActionAuthorizer
+        );
 
         let page = page.unwrap_or(0);
         let per_page = per_page.unwrap_or(Self::DEFAULT_RESULTS_PER_PAGE);
 
+        // TODO: Private Datasets: PERF: find a way to narrow down the number of records
+        //       to filter, e.g.:
+        //       - Anonymous: get all the public
+        //       - Logged: all available based on ReBAC
```
zaychenko-sergei, line 48: Meaning by having a relation between this account and dataset. "Any relation" is a quick query, it does not need to rewrite Polar as SQL. Polar then would refine the criteria.

src/domain/core/src/auth/dataset_action_authorizer.rs
```diff
@@ -23,13 +23,14 @@ use crate::AccessError;
 pub trait DatasetActionAuthorizer: Sync + Send {
     async fn check_action_allowed(
         &self,
-        dataset_handle: &DatasetHandle,
+        // TODO: Private Datasets: just odf::DatasetID? here and below
```
zaychenko-sergei, line 26: Absolutely. Previously we queried the dataset repository, now we don't

src/infra/datasets/postgres/src/repos/postgres_dataset_entry_repository.rs
```diff
@@ -95,7 +98,7 @@ impl DatasetEntryRepository for PostgresDatasetEntryRepository {
                     dataset_name as name,
                     created_at   as "created_at: _"
                 FROM dataset_entries
-                ORDER BY dataset_name ASC
+                ORDER BY created_at ASC
```
zaychenko-sergei, line 101: Hmm, why like this?
Probably the order will need to be parameterizable later.

src/utils/database-common/src/entities.rs
```diff
@@ -15,4 +21,97 @@ pub struct PaginationOpts {
...
+
+impl EntityPageStreamer {
+    pub fn new(start_offset: usize, page_limit: usize) -> Self {
+        Self {
+            start_offset,
+            page_limit,
+        }
+    }
+
+    pub fn into_stream<'a, Entity, Args, HInitArgs, HInitArgsFut, HListing, HListingFut>(
```
zaychenko-sergei, line 66: Cool!

src/domain/datasets/services/src/dataset_entry_service_impl.rs
```diff
@@ -59,11 +62,13 @@ pub struct DatasetEntryServiceImpl {
...
     current_account_subject: Arc<CurrentAccountSubject>,
     tenancy_config: Arc<TenancyConfig>,
-    accounts_cache: Arc<Mutex<AccountsCache>>,
+    accounts_cache: Arc<RwLock<AccountsCache>>,
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// TODO: Private Datasets: Q: DatasetEntryServiceImpl is not singleton,
+//                            but has a cache -- is that intentional?
```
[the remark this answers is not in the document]
zaychenko-sergei: Yes, this is intentional. It's a cache, where the current HTTP request or other kind of parent process defines the scope. The data it caches is not valid outside the transaction anyway. The cache avoids multiple re-queries within the same scenario, which is typical for authors, esp. if there is just 1 user (single-tenant case). 

In my new researches I extend this cache with more data about datasets too.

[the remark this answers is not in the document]
s373r: Got it, thanks for the detailed reply!

src/adapter/auth-oso-rebac/src/oso_dataset_authorizer.rs
s373r: And in the release build, `acc` copying (callback <---> fold internal) will be eliminated (optimized)

src/adapter/auth-oso-rebac/src/oso_dataset_authorizer.rs
s373r: Ideally, I should have taken measurements before answering...

But what I can say right now: 
- At the query level: 
    - We parse Polar every request
- At the dataset iteration level: 
    - OSO-provided (`PolarValue`) actor/resource is used for evaluating

---

On a high-level reading, it follows that the logical step is to make OSO a singleton so we don't have to do parsing every time. But the devil is in the details: internally, OSO creates cached copies for each object (even if it's the same object -- copies are created for each call).

So, for now, this change (OSO as singleton) goes to the class of optional optimizations.

src/infra/datasets/postgres/src/repos/postgres_dataset_entry_repository.rs
s373r: was based on this logic: 

- Let's imagine that a new `dataset_entry` was added during streaming
- When sorting by `created_at`, the new entry will be guaranteed to be on the last page, without affecting the previous ones in any way
- When sorting by name, a new entry may be inserted in the middle of an already issued page, creating a shift and duplicates

src/adapter/graphql/src/queries/search.rs
s373r: NB. TODO updated

src/app/cli/src/app.rs
```diff
@@ -128,17 +128,22 @@ pub async fn run(workspace_layout: WorkspaceLayout, args: cli::Cli) -> Result<()
...
 
-        if workspace_svc.is_in_workspace() {
-            base_catalog_builder.add::<DatasetEntryIndexer>();
-        }
+        base_catalog_builder.add_builder(
+            kamu_datasets_services::DatasetEntryIndexer::builder()
+                .with_is_in_workspace(workspace_svc.is_in_workspace()),
+        );
+        // The indexer has no other interfaces
+        base_catalog_builder
```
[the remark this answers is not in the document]
s373r: Yes, since other entities in the `InitOnStartup` context depend on it -- so I had to add a pass inside `is_in_workspace`.

But you're right, it should be more explicit -- added comment

src/domain/core/src/auth/dataset_action_authorizer.rs
s373r: To be implemented in a follow-up PR

src/infra/datasets/postgres/src/repos/postgres_dataset_entry_repository.rs
zaychenko-sergei: I understand the phantom reads are possible, but have you noticed this actually?

s373r: No, I've seen it in practice -- more of an attempt to avoid potential problems

PR #816, opened 2024-09-07
Implement Batch Query Commitements RFC

sergiimk:
## Description

Closes: https://github.com/kamu-data/kamu-node/issues/125
Closes: https://github.com/kamu-data/kamu-node/issues/126

Implements response signing and a new commitment scheme [according to the RFC](https://docs.google.com/document/d/1JIuI5dd6aSYiBZpXZOmv5DGS4eFKNHQ3jSsfSsDxL00/edit).

Note that the old response format is preserved for compatibility via `V1 / V2` enum trick. If user request in the new format (e.g. by including `?include=proof`) only then they will get a response in the new format. This will allow us to transition to the new format progressively, without breaking the oracles for example.

Also introduces a new `{dataset}/metadata` endpoint that allows getting dataset schema, license, info etc.

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    Extra steps taken to preserve the compatibility
    New config options added, but they have defaults
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Documentation:
    
    Will be done separately
- [x] Downstream effects:
    
    DI update is needed and new configuration needs to be exposed - will be preparing a PR


src/app/cli/src/services/config/models.rs
```diff
@@ -24,54 +25,59 @@ use url::Url;
 #[derive(Debug, Clone, Merge, Serialize, Deserialize)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct CLIConfig {
+    /// Database connection configuration
+    pub database: Option<DatabaseConfig>,
```
sergiimk, line 29: Sorry for the diff - sorted fields alphabetically

src/adapter/http/tests/tests/test_data_query.rs
```diff
@@ -276,6 +285,208 @@ async fn test_data_query_handler_full() {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[test_group::group(engine, datafusion)]
+#[test_log::test(tokio::test)]
+async fn test_data_query_handler_v2() {
```
sergiimk, line 290: This test explains it all

src/adapter/http/src/data/metadata_handler.rs
```diff
@@ -0,0 +1,230 @@
...
+        .then(vis::SearchSetVocabVisitor::new);
+
+    let mut visitors: [&mut dyn MetadataChainVisitor<Error = vis::Infallible>; 6] = [
+        &mut attachments_visitor,
+        &mut info_visitor,
+        &mut license_visitor,
+        &mut schema_visitor,
+        &mut seed_visitor,
+        &mut vocab_visitor,
+    ];
```
sergiimk, line 171: FYI @s373r I'm super happy about this trick:
- Notice that all response elements are optional - so if user did not request `seed` or `schema` we ideally don't want to search for this event
- So I made all visitors above optional
- I then implemented `MetadataChainVisitor` for `Option<T> where T: MetadataChainVisitor`
- If the `Option` is `None` - the visitor will immediately `Stop`
- If `Some` - it delegates the calls to inner visitor

This is about as concise as it gets :)

s373r: @sergiimk , :100: 

Awesome idea with the optional visitors -- looking at the changes in the other PR noticed this, very cool!

PR #827, opened 2024-09-13
Integrate prometheus metrics

sergiimk:
Closes https://github.com/kamu-data/kamu-node/issues/130

- Moving `observability` crate from `kamu-node` down into `kamu-cli`
- Improved DI <> Prometheus integration
- Switched `kamu system api-server` to use same tracing middlewares as `kamu-node` and expose health and metrics endpoints
- Added `outbox_failed_consumers` metric

Example:
```
$ kamu -vv --metrics system api-server --http-port 8899

$ xh GET "http://localhost:8899/system/metrics"

HTTP/1.1 200 OK
Content-Length: 855
Content-Type: text/plain; charset=utf-8
Date: Fri, 13 Sep 2024 01:59:15 GMT

# HELP outbox_failed_consumers Number of consumers that are in the failed state and have stopped consuming messages
# TYPE outbox_failed_consumers gauge
outbox_failed_consumers{consumer="dev.kamu.domain.flow-system.FlowConfigurationService",producer="dev.kamu.domain.core.services.DatasetService"} 0
outbox_failed_consumers{consumer="dev.kamu.domain.flow-system.FlowExecutor",producer="dev.kamu.domain.core.services.DatasetService"} 0
outbox_failed_consumers{consumer="dev.kamu.domain.flow-system.FlowExecutor",producer="dev.kamu.domain.flow-system.FlowConfigurationService"} 0
outbox_failed_consumers{consumer="dev.kamu.domain.flow-system.FlowExecutor",producer="dev.kamu.domain.task-system.TaskExecutor"} 0
outbox_failed_consumers{consumer="dev.kamu.domain.flow-system.FlowTimeWheelService",producer="dev.kamu.domain.flow-system.FlowProgressService"} 0
```

I also added `--metrics` flag that will dump prometheus metrics into a file after command execution. It's not useful currently, but just thought it might be cool feature in future that would show us e.g. how many metadata blocks was read, how long hashing took etc.

### DIFF
--- a/src/utils/database-common/src/transactions/db_transaction_manager.rs
+++ b/src/utils/database-common/src/transactions/db_transaction_manager.rs
@@ -44,7 +44,7 @@ impl DatabaseTransactionRunner {
         Self { catalog }
     }
 
-    #[tracing::instrument(level = "info", skip_all)]
+    #[tracing::instrument(level = "debug", skip_all)]
     pub async fn transactional<H, HFut, HFutResultT, HFutResultE>(
         &self,
         callback: H,
@@ -74,7 +74,7 @@ impl DatabaseTransactionRunner {
                 .build();
 
             callback(catalog_with_transaction)
-                .instrument(tracing::info_span!("Transaction Body"))
+                .instrument(tracing::debug_span!("Transaction Body"))
                 .await
         };
 
@@ -82,9 +82,10 @@ impl DatabaseTransactionRunner {
         match result {
             // In case everything succeeded, commit the transaction
             Ok(res) => {
+                tracing::debug!("Transaction COMMIT");
+
                 db_transaction_manager
                     .commit_transaction(transaction_ref)
-                    .instrument(tracing::info_span!("Transaction COMMIT"))
                     .await?;
 
                 Ok(res)
@@ -92,10 +93,12 @@ impl DatabaseTransactionRunner {
 
             // Otherwise, do an explicit rollback
             Err(e) => {
+                tracing::warn!("Transaction ROLLBACK");
+
                 db_transaction_manager
                     .rollback_transaction(transaction_ref)
-                    .instrument(tracing::error_span!("Transaction ROLLBACK"))
                     .await?;
+
                 Err(e)
             }
         }
--- a/src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
+++ b/src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
@@ -13,6 +13,7 @@ use std::sync::{Arc, Mutex};
 use database_common_macros::{transactional_method, transactional_method1};
 use dill::{component, scope, Catalog, Singleton};
 use internal_error::{InternalError, ResultIntoInternal};
+use observability::metrics::MetricsProvider;
 
 use crate::*;
 
@@ -23,6 +24,7 @@ pub struct OutboxTransactionalProcessor {
     config: Arc<OutboxConfig>,
     routes_static_info: Arc<RoutesStaticInfo>,
     producer_relay_jobs: Vec<ProducerRelayJob>,
+    metrics: Arc<OutboxTransactionalProcessorMetrics>,
 }
 
 #[component(pub)]
@@ -32,11 +34,14 @@ impl OutboxTransactionalProcessor {
         catalog: Catalog,
         config: Arc<OutboxConfig>,
         message_dispatchers_by_producers: Vec<Arc<dyn MessageDispatcher>>,
+        metrics: Arc<OutboxTransactionalProcessorMetrics>,
     ) -> Self {
         let routes_static_info = Arc::new(Self::make_static_routes_info(
             &catalog,
             message_dispatchers_by_producers,
         ));
+
+        metrics.init(&routes_static_info.consumers_by_producers);
 
         let mut producer_relay_jobs = Vec::new();
         for (producer_name, consumer_names) in &routes_static_info.consumers_by_producers {
@@ -46,6 +51,7 @@ impl OutboxTransactionalProcessor {
                 routes_static_info.clone(),
                 producer_name.clone(),
                 consumer_names.clone(),
+                metrics.clone(),
             ));
         }
 
@@ -54,6 +60,7 @@ impl OutboxTransactionalProcessor {
             config,
             routes_static_info,
             producer_relay_jobs,
+            metrics,
         }
     }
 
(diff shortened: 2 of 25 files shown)


src/utils/database-common/src/transactions/db_transaction_manager.rs
```diff
@@ -73,29 +71,30 @@ impl DatabaseTransactionRunner {
                 .add_value(transaction_ref.clone())
                 .build();
 
-            callback(catalog_with_transaction)
-                .instrument(tracing::info_span!("Transaction Body"))
-                .await
+            callback(catalog_with_transaction).await
```
zaychenko-sergei, line 74: At least keep the event that transaction has started separately from the body statements.
I've seen that `sqlx` only has it's own events for `commit` phase, but not for `begin`.

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
```diff
@@ -288,13 +292,23 @@ impl ProducerRelayJob {
...
         consumer_names: Vec<String>,
+        metrics: Arc<OutboxTransactionalProcessorMetrics>,
     ) -> Self {
+        // Initialize labeled metrics so they show up in the output early
+        for consumer_name in &consumer_names {
+            metrics
+                .failed_consumers
+                .with_label_values(&[&producer_name, consumer_name])
+                .set(0);
+        }
```
zaychenko-sergei, line 303: This is coupling the service to metrics library directly. 

I think we should have a higher level API to cover it. 
I.e.:
    `metrics.track_routing_path(&producer_name, consumer_name)`
and then
    `metrics.routing_failed(&producer_name, consumer_name)`.
    
The metrics itself should be a trait, not a struct, so that we can unit test this applying mocking or dummy implementation.

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
```diff
@@ -509,3 +530,39 @@ impl ProducerRelayJob {
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug)]
+
+pub struct OutboxTransactionalProcessorMetrics {
+    pub failed_consumers: prometheus::IntGaugeVec,
+}
+
```
zaychenko-sergei, line 539: This feels like `infra` layer to me. A trait could have been fine at service level, but a particular implementation that couples directly to `prometheus`, is not supposed to be here.

I understand `Prometheus` is de-facto a metrics standard, but there definitely are some alternatives, including OTEL collector itself via [OTEL SDK](https://docs.rs/opentelemetry/latest/opentelemetry/metrics/index.html).

zaychenko-sergei (changes requested):
1. Changelog wasn't updated yet.
2. A test is missing to see how metrics are activated (might be a separate branch).

In general, we need to decide on the typical workflow with the metrics from the code coupling and testability points of view.

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
```diff
@@ -509,3 +530,39 @@ impl ProducerRelayJob {
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug)]
+
+pub struct OutboxTransactionalProcessorMetrics {
+    pub failed_consumers: prometheus::IntGaugeVec,
```
zaychenko-sergei, line 537: For my understanding, why using _gauge_ and not a _counter_?
Looking at definitions in Prometheus docs:

>A _counter_ is a cumulative metric that represents a single monotonically increasing counter whose value **can only increase or be reset to zero on restart**.

> A _gauge_ is a metric that represents a single numerical value that can arbitrarily go up and down.

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
sergiimk: I spent a long time thinking about what is the best separation between our code and the metrics library. I started with the idea of isolation behind traits, as you suggested, but eventually changed my mind:

- `prometheus` is a very lightweight library - exposing it even in `domain` layer (like we already do with much larger DataFusion / Arrow) would have negligible impact on compile times / dependency bloat
- The metrics primitives (gauges, counters, histograms) are low level - e.g. an `IntGauge` is just a wrapper around atomic integer. These objects can be used in complete isolation from the collector part of the library, so objects like my `OutboxTransactionalProcessorMetrics` can be used directly in tests, we can assert on values of the metrics, without any need for mocking traits
- The chances that we will switch from Prometheus to something else are near-zero
- Metrics are not a part of crate-to-crate communications interface. Service that provides metrics will implement them themselves, test them, but other crates will not be ever injecting or reading them. So switching from one library to another will only impact individual services and registration - it will not have any ripple effects in APIs.

So while large part of me wants a complete separation between core domain and dependencies, I think in this case it will not be very practical. Just like it would likely be a waste of time wrapping `DataFrame`s into our own traits, I think wrapping primitives like counters and gauges will have more overhead than value.

I'm open to changing my mind on this, but suggest that we start with this simpler one and refactor towards traits later if there is a real need. If we decide on traits later - this should be a very local non-impactful change.

src/utils/database-common/src/transactions/db_transaction_manager.rs
sergiimk: Transaction manager spams A LOT currently, so I wanted to trim it.

I see the value of knowing when transaction starts and ends, and whether it commits or rolls back, but I didn't understand the value of tracing the "body" separately.

What information do you think it gives us? Is it how long it took to open and commit transaction or something else?

My feeling was that such fine-grain tracing is crossing into "profiling" territory and maybe we should use metrics and histograms for fine timings instead.

zaychenko-sergei: We might want to see that opening and closing transaction takes a lot of time. This might be an indication of heavy load on the connection pool, and it might need to be adjusted (more connections, tweaking timeouts, ..)

With Sqlite we might wait longer, i.e., as there is just 1 connection allowed in the pool.

src/utils/messaging-outbox/src/services/outbox_transactional_processor.rs
sergiimk: I added a few more metrics just to give them a good try:
- counter `outbox_messages_processed_total{producer,consumer}`
- gauge `outbox_messages_pending_total{producer,consumer}` - this one measures the lag behing latest ID and the last processed ID - also a great alert candidate

src/utils/database-common/src/transactions/db_transaction_manager.rs
sergiimk: restored the span for now

PR #832, opened 2024-09-17
Replaced flow time wheel with queries to flow event store

zaychenko-sergei:
## Description

Got rid of `FlowTimeWheelService`, and it's integration via Outbox.
Replaced with:
 - recording `FlowEventScheduledForActivation` event (previously, placement moment into the time wheel)
 - maintaining`schedule_for_activation_at` time field in denormalized `flows` Postgres/SQLite tables
 - using `schedule_for_activation_at` for queries of flows to process within the main flow executor loop
 - in-memory implementation replaced with mirrored lookup tables between `FlowID` and `schedule_for_activation_at` time.
 - extra unit tests for new operations in flow events store.

- [ ] Compatibility:
  - [x] Network APIs: ✅
    UI will need to handle new `FlowEventEnqueued` event
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
  - [x] Database schema: ✅
     Structural migration is provided, but `flows`, `flow_events` and outbox tables must be reset
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/domain/flow-system/domain/src/aggregates/flow/flow.rs
```diff
@@ -80,6 +80,21 @@ impl Flow {
         }
     }
 
+    /// Indicate flow in enqueued for activation at particular time
+    pub fn enqueue(
+        &mut self,
+        now: DateTime<Utc>,
+        enqueued_for: DateTime<Utc>,
```
sergiimk, line 87: Very minor: When I first read `enqueued_for` - it sounded like a duration to me.

I really liked how you described it as "activation" in the above comment, so mb `activation_time` would be a better name for this field.

My mental model is still that we are not "queueing" flows - we are determining the appropriate time when they should run. Queue-based logic is for tasks. So perhaps event could be named `Scheduled`. Thus we will be _**scheduling** a flow for **activation** at time T_.

src/domain/flow-system/domain/src/entities/flow/flow_state.rs
```diff
@@ -143,6 +135,19 @@ impl Projection for FlowState {
                             Ok(FlowState { triggers, ..s })
                         }
                     }
+                    E::Enqueued(FlowEventEnqueued { enqueued_for, .. }) => {
+                        if s.outcome.is_some() || s.timing.awaiting_executor_since.is_some() {
+                            Err(ProjectionError::new(Some(s), event))
+                        } else {
+                            Ok(FlowState {
+                                timing: FlowTimingRecords {
+                                    enqueued_for: Some(enqueued_for),
```
sergiimk, line 144: Does this need to be reset somewhere? E.g. if flow is cancelled do we want to clear this field in the projected state?

src/infra/flow-system/postgres/src/postgres_flow_event_store.rs
```diff
@@ -488,6 +496,61 @@ impl FlowEventStore for PostgresFlowEventStore {
...
+    ) -> Result<Vec<FlowID>, InternalError> {
+        let mut tr = self.transaction.lock().await;
+
+        let connection_mut = tr.connection_mut().await?;
+        let flow_ids = sqlx::query!(
+            r#"
+            SELECT f.flow_id as flow_id
+                FROM flows f
+                WHERE
+                    f.enqueued_for = $1 AND
```
sergiimk, line 540: Hm, do we really want to filter by equality here? I expected to see a range query.

If the service was down for a while and then was started, don't we want to grab all `waiting` flows in one request and kick off their tasks, instead of doing potentially tens of thousands individual requests hopping a few milliseconds at a time?

sergiimk (changes requested):

src/infra/flow-system/postgres/src/postgres_flow_event_store.rs
zaychenko-sergei: We are intentionally rounding activation times very carefully within the flow system, so this comparison is quite valid.
This was already necessary to support repeatable test patterns.

src/domain/flow-system/domain/src/entities/flow/flow_state.rs
zaychenko-sergei: No, we don't need to. This is always used in combination with status, and later remains for history.
We don't clear timing records too, like when the task was scheduled.

src/domain/flow-system/domain/src/aggregates/flow/flow.rs
zaychenko-sergei: Did a big rename over the system to completely destroy term "enqueue" within the flow system.

PR #833, opened 2024-09-18
Upgrade to datafusion 42

sergiimk:
This branch tries to get DF 42 upgrade into a consistent state, but we might not want to merge it even if tests pass, because:
- `arrow-flight` is using newer `tonic` than `opendatafabric`, leading to duplication
- `object_store` is using new `http` stack
- `datafusion-cli` is using v1 of `aws-sdk`, leading to massive duplication


zaychenko-sergei:
Merged in [kamu-cli#716](https://github.com/kamu-data/kamu-cli/pull/716)

Issue #450, opened 2024-01-18
Poor first push user experience


sergiimk:
### Describe the bug

Pushing a new dataset into ODF repository is hard!

### Steps To Reproduce

I just created a new dataset in my local workspace. To push it to ODF repo I currently need:
- `kamu login -s https://platform.demo.kamu.dev`
- `kamu push my.dataset --to "odf+https://api.demo.kamu.dev/sergiimk/my.dataset"`

### Current Behavior

Note that:
1. I need to figure out the backend hostname myself - I cannot push to the same domain I logged in to
2. I need to specify `odf+http` protocol which most people don't know about
3. I need to provide my username and target dataset name in URL

### Expected Behavior

Ideally the following interactions "just work":
```
kamu login platform.demo.kamu.dev  # defaults to https scheme

# If only one known repository
kamu push my.dataset

# To disambiguate between multiple repositories
kamu push my.dataset --to platform.demo.kamu.dev
```

In other words:
- `kamu login` command should automatically register a repo alias: `kamu repo add <frontend domain> <backend url>`
- `kamu push my.dataset` should automatically pick a repo is there is only one
- `kamu push my.dataset --to platform.demo.kamu.dev` should understand that remote is using ODF protocol and construct a proper push URL for me

### kamu system info

```sh
-
```

sergiimk:
This was auto-closed by the merge of CLI changes. Will re-open this until:
- Node is released with new endpoints
- New Jupyter image is release
- All components are deployed to demo env and tested

sergiimk:
Following issues identified during the acceptance testing:
- https://github.com/kamu-data/kamu-cli/issues/946
- https://github.com/kamu-data/kamu-cli/issues/947

sergiimk:
Conducted acceptance testing and all works as expected!
Closing as done :partying_face: :champagne:

PR #859, opened 2024-09-26
Improve login command ux

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/450

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/adapter/http/src/data/account_handler.rs
```diff
@@ -0,0 +1,97 @@
...
+    AuthenticationService,
+    CurrentAccountSubject,
+};
+use opendatafabric::{AccountID, AccountName};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Response {
```
s373r, line 38: What about `AccountResponse`?

src/adapter/http/src/data/account_handler.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub id: AccountID,
+    pub account_name: AccountName,
+    pub email: Option<String>,
+    pub display_name: AccountDisplayName,
+    pub account_type: AccountType,
+    pub avatar_url: Option<String>,
+    pub registered_at: DateTime<Utc>,
+    // TODO: ReBAC: absorb the `is_admin` attribute from the Accounts domain
+    //       https://github.com/kamu-data/kamu-cli/issues/766
```
s373r, line 47: not needed for the response struct

src/adapter/http/src/data/dataset_info_handler.rs
```diff
@@ -0,0 +1,69 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[transactional_handler]
+pub async fn dataset_info_handler(
+    Extension(catalog): Extension<Catalog>,
+    Path(dataset_id): Path<DatasetID>,
+) -> Result<Json<Response>, ApiError> {
+    let response = get_dataset_by_id(&catalog, &dataset_id).await?;
+    tracing::debug!(?response, "Get workspace info response");
```
s373r, line 45: datataset?

src/adapter/http/src/data/dataset_info_handler.rs
```diff
@@ -0,0 +1,69 @@
...
+use dill::Catalog;
+use http_common::*;
+use kamu_core::{DatasetRepository, GetDatasetError};
+use opendatafabric::{AccountName, DatasetID, DatasetName};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Response {
```
s373r, line 31: What about `DatasetInfoResponse`?

src/adapter/http/src/data/workspace_info_handler.rs
```diff
@@ -0,0 +1,51 @@
...
+use axum::response::Json;
+use dill::Catalog;
+use http_common::*;
+use kamu_core::DatasetRepository;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Response {
```
s373r, line 29: What about `WorkspaceInfoResponse`?

src/adapter/http/tests/tests/test_account_info.rs
```diff
@@ -0,0 +1,108 @@
...
+struct AccountInfoHarness {
+    #[allow(dead_code)]
+    run_info_dir: tempfile::TempDir,
+    root_url: url::Url,
+    server_harness: ServerSideLocalFsHarness,
+}
+
+impl AccountInfoHarness {
+    async fn new(is_multi_tenant: bool) -> Self {
+        // TODO: Need access to these from harness level
```
s373r, line 83: Q: Is it actual?

src/app/cli/src/services/odf_server/access_token_registry_service.rs
```diff
@@ -63,98 +63,82 @@ impl AccessTokenRegistryService {
...
-                true
-            } else {
-                match &c.frontend_url {
-                    Some(frontend_url) => frontend_url == odf_server_url,
-                    _ => false,
+        for registry_ptr in [&self.workspace_registry, &self.user_registry] {
+            let registry = registry_ptr
+                .lock()
+                .expect("Could not lock access tokens registry");
+            if let Some(server_record) = registry.iter().find(|c| {
```
s373r, line 72: Please extract a variable with the maybe search result to improve readability

src/app/cli/src/services/odf_server/access_token_registry_service.rs
```diff
@@ -63,98 +63,82 @@ impl AccessTokenRegistryService {
...
-                    access_token: ac.clone(),
-                })
-        } else {
-            None
+        for registry_ptr in [&self.workspace_registry, &self.user_registry] {
+            let registry = registry_ptr
+                .lock()
+                .expect("Could not lock access tokens registry");
+
+            if let Some(server_record) = registry.iter().find(|c| match &c.frontend_url {
```
s373r, line 103: ditto re extract a variable

src/infra/core/src/push_service_impl.rs
```diff
@@ -10,13 +10,13 @@
...
 
 use dill::*;
-use internal_error::ResultIntoInternal;
 use kamu_core::*;
 use opendatafabric::*;
 
 pub struct PushServiceImpl {
     dataset_repo: Arc<dyn DatasetRepository>,
     remote_alias_reg: Arc<dyn RemoteAliasesRegistry>,
+    remote_alias_resolver: Arc<dyn RemoteAliasResolver>,
```
s373r, line 19: please add this PR to the `release-queue` board

src/infra/core/src/push_service_impl.rs
```diff
@@ -257,9 +189,12 @@ impl PushService for PushServiceImpl {
             for request in &initial_requests {
                 if let PushRequest {
                     local_ref: Some(local_ref),
-                    remote_ref: Some(remote_ref),
+                    remote_ref: Some(_transfer_ref @ TransferDatasetRef::RemoteRef(remote_ref)),
```
s373r, line 192: Probable we do not need `_transfer_ref @`

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,305 @@
...
+            .remote_alias_reg
+            .get_remote_aliases(&local_handle.as_local_ref())
+            .await
+            .int_err()?;
+
+        let mut push_aliases: Vec<_> = remote_aliases.get_by_kind(remote_alias_kind).collect();
+
+        match push_aliases.len() {
+            0 => Ok(None),
+            1 => Ok(Some(push_aliases.remove(0).clone())),
```
s373r, line 63: Q: what about `push_aliases[0]`?

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    fn combine_remote_alias(
+        &self,
+        repo_url: &Url,
+        account_name_maybe: Option<odf::AccountName>,
+        dataset_name: &odf::DatasetName,
+    ) -> Result<odf::DatasetRefRemote, InternalError> {
+        let mut res_url = repo_url.clone().as_odf_protocol().int_err()?;
+        if let Some(account_name) = account_name_maybe {
+            res_url.path_segments_mut().unwrap().push(&account_name);
```
s373r, line 76: `res_url.path_segments_mut().unwrap()` can be extracted to a local mut ref to simplify code a bit

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,305 @@
...
+        remote_alias_kind: RemoteAliasKind,
+    ) -> Result<odf::DatasetRefRemote, ResolveAliasError> {
+        let repo_name: odf::RepoName;
+        let mut account_name = None;
+        let mut dataset_name = None;
+
+        if let Some(transfer_dataset_ref) = &transfer_dataset_ref_maybe {
+            match transfer_dataset_ref {
+                odf::TransferDatasetRef::RemoteRef(
+                    _dataset_ref_remote @ DatasetRefRemote::Alias(dataset_alias_remote),
```
s373r, line 119: Looks like we do not need `_dataset_ref_remote @` anymore

s373r (commented):
Add comments & questions

src/infra/core/src/remote_alias_resolver_impl.rs
rmn-boiko: It was copied, but agreed it looks better

src/infra/core/src/push_service_impl.rs
rmn-boiko: Added

src/adapter/http/src/data/account_handler.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub registered_at: DateTime<Utc>,
+    pub is_admin: bool,
+    pub provider: String,
+    pub provider_identity_key: String,
+}
+
+impl From<Account> for AccountResponse {
+    fn from(value: Account) -> Self {
+        Self {
+            id: value.id,
```
sergiimk, line 54: In the situation where you want to map one DTO to another, ideally reflecting the structure 1-to-1 you could use the deconstructing pattern:
```rust
let Account { id, account_name } = account;
Self { id, account_name }
```
This way whenever a new field is added to the `domain::Account` type this code will stop compiling.

Your current code will compile fine without forcing you to consider if the new field should be also exposed in REST API.

src/adapter/http/src/data/query_handler.rs
```diff
@@ -143,7 +143,7 @@ pub(crate) async fn query_handler_post_v2(
             },
         })
     } else {
-        Err(ApiError::not_implemented(ResponseSigningNotCongigured))?
+        Err(ApiError::not_implemented(ResponseSigningNotConfigured))?
```
sergiimk, line 146: thanks!

src/adapter/http/src/data/mod.rs
```diff
@@ -7,12 +7,15 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+mod account_handler;
```
sergiimk, line 10: I think all three of these new handlers don't belong in `data` group. I created it for raw data queries and commitments. Let's create some other group, perhaps called `general`?

src/adapter/http/src/data/workspace_info_handler.rs
```diff
@@ -0,0 +1,51 @@
...
+use axum::response::Json;
+use dill::Catalog;
+use http_common::*;
+use kamu_core::DatasetRepository;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct WorkspaceInfoResponse {
```
sergiimk, line 29: Let's call this `NodeInfo`. The definition of "workspace" as I see it is a directory on a local file system that stores datasets and other ODF objects managed by an ODF node (such as `kamu-cli`).

src/adapter/http/src/data/router.rs
```diff
@@ -29,6 +29,18 @@ pub fn root_router() -> axum::Router {
             "/verify",
             axum::routing::post(super::verify_handler::verify_handler),
         )
+        .route(
+            "/workspace/info",
```
sergiimk, line 33: Let's use `/info` as this will be returning very fundamental information about the ODF node

src/adapter/http/src/data/router.rs
```diff
@@ -29,6 +29,18 @@ pub fn root_router() -> axum::Router {
...
+            axum::routing::get(super::workspace_info_handler::workspace_info_handler),
+        )
+        .route(
+            "/me",
```
sergiimk, line 37: perhaps `/accounts/me` is better?

Note: I'm preparing a proposal to improve endpoint schemes, so we will revisit soon.

src/app/cli/src/cli_value_parser.rs
```diff
@@ -74,6 +74,18 @@ pub(crate) fn dataset_ref_remote(s: &str) -> Result<odf::DatasetRefRemote, Strin
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub(crate) fn transfer_dataset_ref_remote(s: &str) -> Result<odf::TransferDatasetRef, String> {
+    match odf::TransferDatasetRef::from_str(s) {
+        Ok(push_dataset_ref) => Ok(push_dataset_ref),
+        Err(_) => Err("Remote reference should be in form: `did:odf:...` or \
```
sergiimk, line 80: `TransferDatasetRef` is used in `--to` argument, but does `did:odf:...` make sense in this position?

From our conversation I though in `--to` we will only allow:
- `Url`
- `RepoName`
- `RepoName/DatasetName` (defaults to current account)
- `RepoName/AccountName/DatasetName`

Also perhaps `DatasetPushTarget` is a better name for this type.

src/domain/core/src/services/push_service.rs
```diff
@@ -40,7 +40,7 @@ pub trait PushService: Send + Sync {
 #[derive(Debug, Clone)]
 pub struct PushRequest {
     pub local_ref: Option<DatasetRef>,
-    pub remote_ref: Option<DatasetRefRemote>,
+    pub remote_ref: Option<TransferDatasetRef>,
```
sergiimk, line 43: Hmm, it's now strange to see this type here. 

If we want `remote_ref` to represent a *resolved* push destination, i.e. **where exactly** the push is going to, then perhaps it should be a struct containing a Url, and optional repo, account, and dataset name?

And perhaps this what `RemoteAliasResolver::resolve_remote_alias` should be returning?

What do you think?

src/domain/core/src/services/remote_aliases_registry.rs
```diff
@@ -46,3 +46,82 @@ impl From<GetDatasetError> for GetAliasesError {
...
     }
 }
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// RemoteAliasResolver
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait]
+pub trait RemoteAliasResolver: Send + Sync {
+    async fn resolve_remote_alias(
```
sergiimk, line 56: Please document the API

sergiimk (changes requested):
Did a quick pass. Structurally things are looking very good.

I'd like to dig in a bit into `DatasetRefRemote` and `TransferDatasetRef` types and see if they can be simplified. Left some ideas.

src/adapter/http/src/data/mod.rs
rmn-boiko: Yeah, agree, added

src/adapter/http/src/data/account_handler.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub email: Option<String>,
+    pub display_name: AccountDisplayName,
+    pub account_type: AccountType,
+    pub avatar_url: Option<String>,
+    pub registered_at: DateTime<Utc>,
+    pub is_admin: bool,
+    pub provider: String,
+    pub provider_identity_key: String,
+}
+
```
zaychenko-sergei, line 50: Do we really need to explose this entire struct? For your task you would probably be fine with `id` and `account_name`, right?

src/adapter/http/src/data/account_handler.rs
```diff
@@ -0,0 +1,95 @@
...
+    match current_account_subject.as_ref() {
+        CurrentAccountSubject::Anonymous(_) => Err(ApiError::new_unauthorized()),
+        CurrentAccountSubject::Logged(account) => {
+            let auth_service = catalog.get_one::<dyn AuthenticationService>().unwrap();
+            let full_account_info_maybe = auth_service.account_by_id(&account.account_id).await?;
+            if let Some(full_account_info) = full_account_info_maybe {
+                return Ok(Json(full_account_info.into()));
+            }
+
+            Err(ApiError::not_found_without_body())
```
zaychenko-sergei, line 90: How is it possible we are logged in, but don't have a full account record?

src/adapter/http/src/data/dataset_info_handler.rs
```diff
@@ -0,0 +1,75 @@
...
+    let dataset_handle = dataset_repo
+        .resolve_dataset_ref(&dataset_id.clone().as_local_ref())
+        .await
+        .map_err(|err| match err {
+            GetDatasetError::NotFound(e) => ApiError::not_found(e),
+            GetDatasetError::Internal(e) => e.api_err(),
+        })?;
+
+    Ok(Json(dataset_handle.into()))
+}
```
zaychenko-sergei, line 73: I.e,, this would be a private dataset. I think we have to apply authorization logic before answering the dataset event exists.

src/adapter/http/src/data/workspace_info_handler.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub is_multi_tenant: bool,
```
zaychenko-sergei, line 30: Maybe version too

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,308 @@
...
+use internal_error::{InternalError, ResultIntoInternal};
+use kamu_core::*;
+use opendatafabric::{self as odf, DatasetRefRemote};
+use url::Url;
+
+use crate::UrlExt;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct RemoteAliasResolverImpl {
```
zaychenko-sergei, line 24: Note, there are 2 types of remote aliases: push & pull.
You've named everything every general, but only work with push aliases.
The naming of either the component, or particular methods should be improved.

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,308 @@
...
+                    .contains(remote_ref, remote_alias_kind)
+            {
+                return Ok(local_handle);
+            }
+
+            // No luck - now have to search through aliases
+            use tokio_stream::StreamExt;
+            let mut datasets = self.dataset_repo.get_all_datasets();
+            while let Some(dataset_handle) = datasets.next().await {
+                let dataset_handle = dataset_handle?;
```
zaychenko-sergei, line 205: Let's make sure we have a test for this state. I'm going to refactor this soon to list the handles from the database entries, but I'm not sure what exactly reproduces this state.

src/app/cli/src/cli_value_parser.rs
rmn-boiko: Yeah, you are right! I forgot that it will pass ID etc refs. Changed

src/adapter/http/src/data/account_handler.rs
rmn-boiko: Yeah. true, removed

src/infra/core/src/remote_alias_resolver_impl.rs
rmn-boiko: It is tasted in `test_push` scope

src/infra/core/src/remote_alias_resolver_impl.rs
rmn-boiko: I've already made some preparation where service accept Alias kind to support both push and pull. And once we will be ready to improve pull flow it will be reused

zaychenko-sergei (approved):

src/adapter/http/src/general/dataset_info_handler.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use crate::axum_utils::ensure_authenticated_account;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DatasetInfoResponse {
+    pub id: DatasetID,
+    pub account_name: Option<AccountName>,
```
sergiimk, line 35: Let's add `account_id` too. Name is good for visualization, but navigating from dataset to an account should be done by IDs that are resistant to renaming.

sergiimk (changes requested):

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,296 @@
...
+            if remote_repo_names.len() > 1 {
+                return Err(ResolveAliasError::AmbiguousRepository);
+            }
+            if let Some(repository_name) = remote_repo_names.first() {
+                repo_name = repository_name.clone();
+            } else {
+                return Err(ResolveAliasError::EmptyRepositoryList);
+            }
+        }
+        let remote_repo = self.remote_repo_reg.get_repository(&repo_name).int_err()?;
```
sergiimk, line 164: If `dataset_push_target_maybe == Some(Repository("blah"))` and "blah" repository does not exist will we get an internal error here instead of `RepositoryNotFound`?

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,296 @@
...
+impl RemoteAliasResolver for RemoteAliasResolverImpl {
+    async fn resolve_remote_alias(
+        &self,
+        local_dataset_handle: &odf::DatasetHandle,
+        dataset_push_target_maybe: Option<odf::DatasetPushTarget>,
+        remote_alias_kind: RemoteAliasKind,
+    ) -> Result<RemoteAliasRef, ResolveAliasError> {
+        let repo_name: odf::RepoName;
+        let mut account_name = None;
+        let mut dataset_name = None;
```
sergiimk, line 121: Might be better to rewrite this as:
```rust
let (repo_name, account_name, dataset_name) = if ... { .. } else { ... };
```
This way you get certainty that all branches of logic consider all three of these values and not simply forget to assign one of them.

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,296 @@
...
+            .int_err()?;
+        }
+        let transfer_dataset_name = dataset_name.clone().unwrap_or(
+            self.resolve_remote_dataset_name(
+                local_dataset_handle,
+                &remote_repo.url,
+                access_token_maybe.as_ref(),
+            )
+            .await?,
+        );
```
sergiimk, line 184: Isn't `resolve_remote_dataset_name` an expensive function? Written like this we will always be calling it even if `dataset_name` was not `None`.

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,296 @@
...
+        let client = reqwest::Client::new();
+        let header_map = Self::build_headers_map(access_token_maybe);
+
+        let workspace_info_response = client
+            .get(server_backend_url.join("info").unwrap())
+            .headers(header_map.clone())
+            .send()
+            .await
+            .int_err()?
+            .error_for_status()
```
sergiimk, line 231: Have you tested this with plain storage repos like `S3` or local FS? I don't see any logic that would differentiate those from ODF repos.

Even if a repo has HTTP(S) scheme it's not a fact that it's an ODF node. So we likely need some fallback logic for cases when `/info` is not present.

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,296 @@
...
+                http::HeaderValue::from_str(format!("Bearer {access_token}").as_str()).unwrap(),
+            );
+        };
+        header_map
+    }
+
+    // Return account name if remote workspace is in multi tenant mode
+    pub async fn resolve_remote_account_name(
+        server_backend_url: &Url,
+        access_token_maybe: Option<&String>,
```
sergiimk, line 220: Since you have extracted this helper as a type - you could pass token as constructor argument, not to have to pass it through every function's arguments.

sergiimk (changes requested):

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Recommendation: for ease of reading, use the following order:
 - Fixed
 -->
 
+## [Unreleased]
+### Changed
+- `kamu push <dataset>` command now can be called without `--to` reference and Alias or Remote dataset repository will be used as destination
+- `kamu login` command now will store repository to Repository registry. Name can be provided with `--repo-name` flag and to skip creating repo can be used `--skip-add-repo` flag
```
sergiimk, line 17: We also need to check if our instructions on pulling/pushing datasets in the Jupyter demo notebooks need to be updated or can be simplified.

src/domain/core/src/services/remote_aliases_registry.rs
```diff
@@ -46,3 +46,103 @@ impl From<GetDatasetError> for GetAliasesError {
...
+pub trait RemoteAliasResolver: Send + Sync {
+    // Resolve remote alias reference.
+    // Firstly try to resolve from AliasRegistry, if cannot do it
+    // try to resolve via repository registry
+    async fn resolve_remote_alias(
+        &self,
+        local_dataset_handle: &DatasetHandle,
+        dataset_push_target_maybe: Option<DatasetPushTarget>,
+        remote_alias_kind: RemoteAliasKind,
+    ) -> Result<RemoteAliasRef, ResolveAliasError>;
```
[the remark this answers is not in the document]
rmn-boiko: Yeah, agree

src/infra/core/src/remote_alias_resolver_impl.rs
rmn-boiko: Good catch, fixed

src/infra/core/src/remote_alias_resolver_impl.rs
rmn-boiko: Yeah, it is very expensive, but it was my idea, if the user does not provide dataset_name I am trying to fetch it from a remote server.

CHANGELOG.md
rmn-boiko: Updated

src/infra/core/src/remote_alias_resolver_impl.rs
```diff
@@ -0,0 +1,291 @@
...
+        let client = reqwest::Client::new();
+        let header_map = Self::build_headers_map(access_token_maybe);
+
+        let workspace_info_response = client
+            .get(server_backend_url.join("info").unwrap())
+            .headers(header_map.clone())
+            .send()
+            .await
+            .int_err()?;
+        if workspace_info_response.status().is_client_error() {
```
sergiimk, line 225: Do we have a test that does a push to localfs or S3 repos using the full "push" flow? (I know there are tests for low-level sync, but it doesn't engage this logic).

To me it looks that push will just crash here because `reqwest` doesn't know how to deal with `file://` or `s3://` schemes.

Issue #877, opened 2024-10-04
E2E: cover `kamu-cli` commands


s373r:
To facilitate fearless refactoring of the project, we need to cover all commands with E2E tests.

### Command list:
- [ ] `kamu add`
  - [x] `--recursive `
  - [ ] a specific dataset (regular case)
  - [ ] `--stdin`
  - [ ] `--replace`
  - [ ] `--name`
- [x] `kamu completions`
- [ ] `kamu config`
  - [ ] `list`
  - [ ] `get`
  - [ ] `set`
- [ ] `kamu delete`
  - [ ] a specific dataset (regular case)
  - [ ] `--recursive`
  - [ ] `--add`
- [ ] `kamu ingest`
  - [x] a specific dataset (regular case)
  - [ ] `--stdin`
  - [ ] `--source-name`
  - [ ] `--recursive`
- [x] `kamu init`
- [ ] `kamu inspect`
  - [ ] `kamu inspect lineage`
    - [ ] a specific dataset (regular case)
  - [ ] `kamu inspect query`
    - [ ] a specific dataset (regular case)
  - [ ] `kamu inspect schema`
    - [ ] a specific dataset (regular case)
- [x] `kamu list`
- [ ] `kamu log`
  - [ ] a specific dataset (regular case)
  - [ ] `--filter`
- [ ] `kamu login`
  - [ ] without subcommand
    - [ ] `--check`
    - [x] --access-token
  - [ ] `kamu login oauth`
  - [ ] `kamu login password`
- [ ] `kamu logout`
  - [ ] a specific server (regular case)
  - [ ] `--all`
- [ ] `kamu new`
  - [ ] `--root`
  - [ ] `--derivative`
- [x] (skip) `kamu notebook` 
- [ ] `kamu pull`
  - [ ] From:
    - [x] `odf+http`
    - [ ] S3
  - [ ] Arguments:
    - [ ] from a registered repository
    - [ ] `--as`
    - [ ] `--all`
    - [ ] `--recursive`
    - [ ] `--set-watermark`
    - [ ] `--force`
    - [ ] `--reset-derivatives-on-diverged-input`
- [ ] `kamu push`
  - [x] `--to`
  - [ ] `--all`
  - [ ] `--recursive`
  - [ ] `--no-alias`
  - [ ] `--force`
  - [ ] `--visibility`
- [ ] `kamu rename`
- [ ] `kamu reset`
- [ ] `kamu repo`
[...]

PR #883, opened 2024-10-08
E2E: add (or expand) `ingest`/`inspect`/`log`/`new`/`reset`/`search`/etc command tests

s373r:
- Related issue: https://github.com/kamu-data/kamu-cli/issues/877


src/e2e/app/cli/repo-tests/src/commands/test_ingest_command.rs
```diff
@@ -180,8 +184,289 @@ pub async fn test_push_ingest_from_file_snapshot_with_event_time(mut kamu: KamuC
...
+// Helpers
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 fn path(p: &Path) -> &str {
     p.as_os_str().to_str().unwrap()
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+async fn assert_ingest_data_to_player_scores_from_stdio<I, S, T>(
```
rmn-boiko, line 400: Optional for now: I think we can move it to a higher level. IT could be reused for different commands also

rmn-boiko (approved):

src/e2e/app/cli/repo-tests/src/commands/test_ingest_command.rs
s373r: Discussed in Slack: came to the conclusion that we don't need to

Issue #896, opened 2024-10-11
E2E: cover the flows


s373r:
### Types

- `Ingest`
- `ExecuteTransform`
- `HardCompaction`
- `Reset`

PR #914, opened 2024-10-21
E2E: cover flows

s373r:
## Description

Closes: #896

### Added
- E2E: cover all flow scenarios
- E2E: `KamuApiServerClientExt`: method grouping

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/adapter/http/src/e2e/system_time_handler.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct SetSystemTimeRequest {
+    pub new_system_time: DateTime<Utc>,
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[tracing::instrument(level = "info", skip_all)]
+pub async fn set_system_time_handler(
```
zaychenko-sergei, line 37: What is your opinion on whether we need to add these extensions into Swagger?

src/app/cli/src/explore/api_server.rs
```diff
@@ -175,6 +173,10 @@ impl APIServer {
             None
         };
 
+        router = router
+            .layer(Extension(gql_schema))
+            .layer(Extension(api_server_catalog));
+
```
zaychenko-sergei, line 179: Can you explain the change in layering order here?

src/e2e/app/cli/common/src/kamu_api_server_client.rs
```diff
@@ -76,6 +80,29 @@ impl KamuApiServerClient {
         Ok(())
     }
 
+    pub async fn set_system_time(&self, t: DateTime<Utc>) {
+        let endpoint = self.server_base_url.join("e2e/system_time").unwrap();
```
zaychenko-sergei, line 84: Since the number of REST endpoints is growing, we should be looking for a way to enumerate them or have a list of constants, rather than copypasting the suffixes in the code.

src/e2e/app/cli/repo-tests/src/test_flow.rs
```diff
@@ -7,13 +7,23 @@
...
+    DATASET_ROOT_PLAYER_SCORES_INGEST_DATA_NDJSON_CHUNK_3,
+};
+use opendatafabric::{DatasetAlias, DatasetName};
 use tokio_retry::strategy::FixedInterval;
 use tokio_retry::Retry;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-pub async fn test_get_dataset_list_flows(kamu_api_server_client: KamuApiServerClient) {
+pub async fn test_web_ui_get_dataset_list_flows(kamu_api_server_client: KamuApiServerClient) {
```
zaychenko-sergei, line 26: What is the motivation to rename it? There is no web ui behind it, it's just GraphQL API. Could be named as `test_gql_get_dataset_list_flows`, if you'd like to emphasize the transport mechanism.

src/e2e/app/cli/repo-tests/src/test_flow.rs
```diff
@@ -736,6 +746,903 @@ pub async fn test_dataset_trigger_flow(kamu_api_server_client: KamuApiServerClie
         .await;
 }
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub async fn test_trigger_flow_ingest(kamu_api_server_client: KamuApiServerClient) {
+    let temp_dir = tempfile::tempdir().unwrap();
```
zaychenko-sergei, line 752: I think these tests could benefit from another layer of abstraction over GQL API calls.
Currently it's really hard to follow what is happening in the test in general, because it's quite long, and uses GQL low-level templating. 
The GQL calls are not very specific to particular tests, these are just possible "steps" in the API, and the templates with parameterization could live in another utility layer.

zaychenko-sergei: Even if these GQL templates would only be used in any other test suites, and they don't qualify to become a more common test API, there will still be a benefit from a module-local helper. Currently there is signifficant duplication in the way flows are triggered and their results are matched against the expected outcome.

src/adapter/http/src/e2e/system_time_handler.rs
s373r: I think that `e2e`-related REST API routes should not appear in Swagger. 

In my opinion, these are internal test details that should not be visible to kamu-web-ui (main user Swagger)

src/app/cli/src/explore/api_server.rs
s373r: I needed to move the extensions' declaration below because I started using `catalog` in the freshly-added `set_system_time_handler()`.

Otherwise, `catalog` could not be found.

---

A fair question arises, why was `gql_schema` also moved? I thought it was better to have both extensions side by side.

src/e2e/app/cli/repo-tests/src/test_flow.rs
s373r: Since these are queries copied from `kamu-web-ui`. With the renaming, I wanted to separate the more complex tests from these.

src/e2e/app/cli/repo-tests/src/test_flow.rs
s373r: Fair point -- I'll make it easier to read

src/e2e/app/cli/repo-tests/src/test_flow.rs
s373r: makes sense -- will rename the tests

s373r: Done

src/e2e/app/cli/repo-tests/src/test_flow.rs
s373r: A little rethinking of some of the methods -- I think it's gotten better: what do you think?

src/e2e/app/cli/common/src/kamu_api_server_client.rs
s373r: Yes, I think about this periodically when I'm forced to use hardcode in tests. After the swagger integration that defines endpoint usage via the `#[utoipa::path()]` macro, I have no idea how to do it more or less nicely yet. 

Q: Maybe you have some ideas?

src/e2e/app/cli/common/src/kamu_api_server_client.rs
```diff
@@ -7,15 +7,20 @@
...
-use opendatafabric::DatasetAlias;
 use reqwest::{Method, Response, StatusCode, Url};
 use serde::Deserialize;
+use serde_json::json;
 use tokio_retry::strategy::FixedInterval;
 use tokio_retry::Retry;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub type AccessToken = String;
```
zaychenko-sergei, line 20: This is a bit confusing, as we have such type in the application.
Also, maybe we should rather use a tuple struct?

zaychenko-sergei (approved):

PR #917, opened 2024-10-21
Add `kamu-base:git` image

s373r:
## Description

We need the ability to build `kamu` from any branch for Windows. This PR adds a docker image that allows this.

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

images/kamu-base-git/podman/podman-priv.yaml
```diff
@@ -0,0 +1,20 @@
+version: v1
+kind: Pod
+metadata:
+  name: podman-priv
```
sergiimk, line 4: I'm confused. How all this kubernetes stuff is related to the docker image?

I don't see it being used anywhere.

images/kamu-base-git/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+    apt -y install curl wget gnupg unzip jq && \
+    . /etc/os-release && \
+    echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /" \
+    | tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && \
+    curl -L "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key" \
+    | apt-key add - && \
+    apt update && \
+    apt install ca-certificates && \
+    apt -y install podman fuse-overlayfs && \
+    rm -rf /var/lib/apt/lists/*
```
sergiimk, line 39: 90% of build time for `kamu-base` image is installing podman, aws cli, etc.

I wonder if for these dev builds we could just use `FROM kamu-data/kamu-base:latest` that comes with all this stuff in it, and we only replace the `kamu` binary.

This would significantly speed up the builds and save space too.

images/kamu-base-git/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```shell
+cd ./images/kamu-base-git/
+docker build -t kamu-base:git .
+```
+
+Or, you can also specify the desired branch:
+```shell
+docker build --build-arg KAMU_BRANCH=<REPLACE_ME> -t kamu-base:git .
+```
```
sergiimk, line 18: What we discussed with hauki team last week is this vision:
- every repo has a "dev build" GH action
  - runs on demand
  - parametrized by branch name / commit
  - generates deployment artifacts (docker images, helm charts)
- artifacts are stored in a dev image registry
  - separately from production images
  - a place we can easily clean up when needed
- a "Dev Deployment" GH action will exist e.g. in `kamu-deploy` repo
  - is parametrized with branch names per every component
  - it kicks off "dev build" actions in all repos
  - waits for completion
  - then proceeds to deploy the artifacts to dev environment

If we follow this plan - I think there still exists a need for "dev docker image", but there is no need to build kamu in docker.

I assume you are creating this image mainly for @dmitriy-borzenko to be able to test UI against latest branch changes, but I think it would be a lot more convenient for him to just trigger such image builds on GitHub instead of building locally:
- it's just easier to use
- there's less risk of human error - forgetting to pull, building wrong commit etc
- GH job can reuse cache and given how resource-hungry our rust builds are just not stress frontend dev's laptops

images/kamu-base-git/podman/podman-priv.yaml
s373r: This is the contents of the `podman` folder that was copied from `kamu-base` (`./images/kamu-base/podman`)
If it's not relevant here, it looks like we need to do a revision of what we supply in `kamu-base`.

What do you think about that?

images/kamu-base-git/Dockerfile
s373r: Totally makes sense

images/kamu-base-git/podman/podman-priv.yaml
s373r: Addendum: as part of this PR, these changes are no longer contained. But they also remain in `kamu-base`. This seems to be an echo from 2021. 

In a good way, we need to revisit the files that are packed into the `kamu-base` image.

images/kamu-base-git/Dockerfile
s373r: Done

images/kamu-base-git/README.md
s373r: > If we follow this plan - I think there still exists a need for "dev docker image", but there is no need to build kamu in docker.

I really like this plan, it's the way it's supposed to be.

> I assume you are creating this image mainly for @dmitriy-borzenko to be able to test UI against latest branch changes, but I think it would be a lot more convenient for him to just trigger such image builds on GitHub instead of building locally:

This PR is not offering an alternative, but a solution to the here and now -- since Dmitriy couldn't work with any branch of `kamu-cli` as we previously wanted. 

I'll be happy to remove this image as soon as the above GH actions are available, but for now I suggest leaving it as is, as it has already proven to be effective.

sergiimk: Ok let's solve the immediate problem, but later on I see this image likely reused (not replaced) in the "dev build" GH action.

@dmitriy-borzenko let us know if you find this build procedure too cumbersome - creating GH action to automate it should be a simple matter.

images/kamu-base-git/podman/podman-priv.yaml
sergiimk: Looks like originally I copied these files to `kamu-base` by accident. I will clean them up on master.

sergiimk (approved):

images/kamu-base-git/README.md
dmitriy-borzenko: At the moment, Dima's solution solves the problem. But if it can be simplified and automated, I will be only glad

Issue #828, opened 2024-09-14
Support default values in dataset env var templates


sergiimk:
Currently we allow ingest parametrization via env vars that are templated into value strings like so:
```yaml
fetch:
  kind: Container
  image: "ghcr.io/kamu-data/fetch-com.defillama:0.1.0"
  args:
    - --request-interval
    - '${{ env.request_interval }}'
    - protocols
    - chain-tvls
    - --top-n
    - '${{ env.top_n_protocols }}'
```

It would be very useful to allow dataset manifests to provide default values for parameters that are not specified (e.g. using a syntax similar to GitHub's templating engine):

```yaml
fetch:
  kind: Container
  image: "ghcr.io/kamu-data/fetch-com.defillama:0.1.0"
  args:
    - --request-interval
    - '${{ env.request_interval || 1 }}'
    - protocols
    - chain-tvls
    - --top-n
    - '${{ env.top_n_protocols || 10 }}'
```

PR #924, opened 2024-10-25
WIP: Support default values in dataset env var templates

demusdev:
## Description

Closes: #828 

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -46,6 +46,48 @@ pub struct FetchService {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub(crate) fn template_string<'a>(
+    s: &'a str,
+    lookup_fn: &dyn Fn(&str) -> Result<String, PollingIngestError>,
+) -> Result<Cow<'a, str>, PollingIngestError> {
```
demusdev, line 52: I'm not sure PollingIngestError is right error here. LF suggestions.

zaychenko-sergei: `PollingIngestError` is specific to polling ingest service, fetch service is a part of it,.
Since you are isolating the function, you can define your own error struct, specific for this algorithm. 
Then, it can be converted to whatever where it will be used.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -46,6 +46,48 @@ pub struct FetchService {
...
+    let mut s = Cow::from(s);
+    let re_tpl = regex::Regex::new(r"\$\{\{([^}]*)\}\}").unwrap();
+    let re_env =
+        regex::Regex::new(r"^env\.([a-zA-Z-_0-9]+)\s?(?:\|\|\s+([a-zA-Z-_0-9]+))?$").unwrap();
+
+    loop {
+        if let Some(ctpl) = re_tpl.captures(&s) {
+            let tpl_range = ctpl.get(0).unwrap().range();
```
zaychenko-sergei, line 60: Unless in tests, prefer `.expect("msg")` style to `.unwrap()`. It will be easier to locate the error by a more specific text.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -209,39 +251,15 @@ impl FetchService {
...
-            }
-        }
+        let lookup_fn = |env_name: &str| -> Result<String, PollingIngestError> {
+            let value = self
+                .dataset_key_value_svc
+                .find_dataset_env_var_value_by_key(env_name, dataset_env_vars)
+                .int_err()?;
+            let exposed_value = value.get_exposed_value();
+            Ok(exposed_value.to_string())
+        };
```
zaychenko-sergei, line 261: What happens, if variable is not defined? Making an internal error in this case might be too rough.
We would rather prefer printing a valuable error message, so think of what errors makes sense.
Its definitely not a `PollingIngestError`, it is a too wide kind for your situation.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +363,118 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+    use super::*;
+
+    fn gen_var() -> String {
+        let random_str: String = rand::thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(10)
+            .map(char::from)
+            .collect::<String>()
+            .to_uppercase();
+        format!("VAR_{random_str}")
```
zaychenko-sergei, line 391: We have an utility already, which might simplify, look for `get_random_name()`

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +363,118 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+        assert_eq!(result.unwrap(), "val1, val2");
+
+        // When at least one of vars is missing, returns error.
+        let result = template_string("{{ env.test1 }}, ${{ env.missing_var }}", &lookup_fn);
+        assert!(result.is_err());
+
+        // Default value is being used if present.
+        let result = template_string("foo=${{ env.test3 || default-val3 }}", &lookup_fn);
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), "foo=default-val3");
```
zaychenko-sergei, line 476: `assert_eq!(result, Ok("foo=default-val3"))`

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +363,118 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+        let result = fetch_svc.template_string(&template, &env_vars);
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), "222");
+
+        // When dataset_env_vars is empty and system env var is not set,
+        // we expect error.
+        std::env::remove_var(&var);
+        let result = fetch_svc.template_string(&template, &env_vars_empty);
+        assert!(result.is_err());
+        assert_matches!(result, Err(Internal(_)));
```
zaychenko-sergei, line 447: The first assertion is implicitly checked in the 2nd already.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +395,104 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+
+        // When dataset_env_vars is empty and system env var is not set,
+        // we expect error.
+        std::env::remove_var(&var);
+        let result = fetch_svc.template_string(&template, &env_vars_empty);
+        assert_matches!(result, Err(Internal(_)));
+
+        // The only supported prefix is "env."
+        let result = fetch_svc.template_string("${{ foo.bar }}", &env_vars_empty);
+        assert!(result.is_err());
```
sergiimk, line 470: Once you surface the errors in `PollingIngestError` enum I'd like to see `assert_matches!` here matching a specific type of the error. Otherwise there is a chance that we are getting a completely unrelated error here.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +395,104 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+            match env_name {
+                "test1" => Some(String::from("val1")),
+                "test2" => Some(String::from("val2")),
+                _ => None,
+            }
+        };
+
+        // Populates multiple vars.
+        let result = template_string("${{ env.test1 }}, ${{ env.test2 || default2 }}", &lookup_fn);
+        assert!(result.is_ok());
```
sergiimk, line 485: Don't assert for `is_ok()` - this will produce an unhelpful message `false != true`. Unwrap directly so you get a panic with a complete error message.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +395,104 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+        assert_eq!(result.unwrap(), "val1, val2");
+
+        // When at least one of vars is missing, returns error.
+        let result = template_string("{{ env.test1 }}, ${{ env.missing_var }}", &lookup_fn);
```
sergiimk, line 489: missing dollar sign in the first pattern?

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +395,104 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+        assert!(result.is_err());
+
+        // Default value is being used if present.
+        let result = template_string("foo=${{ env.test3 || default-val3 }}", &lookup_fn);
```
sergiimk, line 493: Sorry for not specifying this in the task requirements, but I think we should require quoting the string values. This is to in future allow expressions like: `${{ env.specific || env.general || 'default'}}` where `env.general` is treated as code rather than string value.

The world of templating engines is a complete mess with no standards. I based this super primitive regex implementation on [GitHub Actions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#using-the-vars-context-to-access-configuration-variable-values) so let's continue with their single-quoted string syntax for now. Later on we should pick a proper templating engine and standardize around it.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -46,6 +47,83 @@ pub struct FetchService {
...
+}
+
+pub fn template_string<'a>(
+    s: &'a str,
+    lookup_fn: &dyn Fn(&str) -> Option<String>,
+) -> Result<String, TemplateError> {
+    let mut s = Cow::from(s);
+    let re_tpl = regex::Regex::new(r"\$\{\{([^}]*)\}\}").unwrap();
+    let re_env =
+        regex::Regex::new(r"^env\.([a-zA-Z-_0-9]+)\s?(?:\|\|\s+([a-zA-Z0-9@:%_+-.~#?&/=]+))?$")
```
sergiimk, line 79: Assuming we support two types of default values:
- integral: `123`, `-123`
- single-quoted strings: `'asd!$#@#$@#'`, `''`
perhaps we can simplify the regex to `((-?[0-9]+)|('[^']*'))` i.e. instead of listing all characters that are allowed inside a string we match *everything* until we encounter a closing single quote.

sergiimk (changes requested):

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -208,40 +284,14 @@ impl FetchService {
...
-            }
-        }
+    ) -> Result<String, PollingIngestError> {
+        let lookup_fn = |env_name: &str| -> Option<String> {
+            self.dataset_key_value_svc
+                .find_dataset_env_var_value_by_key(env_name, dataset_env_vars)
+                .ok()
+                .map(|v| v.get_exposed_value().to_string())
+        };
+        template_string(s, &lookup_fn).map_err(|err| PollingIngestError::Internal(err.int_err()))
```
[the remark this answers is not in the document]
sergiimk: A very important rule for deciding whether something is `Internal` or explicit error:

We never return the details of `Internal` errors to the user from REST/GQL API. Internal errors may be related to infrastructure (e.g. or accounts, S3 buckets etc.) so we only write them to logs and not show to users not to reveal sensitive details.

In your case we'd definitely want to explain to the user why ingest failed, so it shouldn't be an internal error.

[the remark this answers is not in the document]
demusdev: The main reason I created separate errors is for being able easily move (and potentially reuse in different places) the templating code. Do you think we shouldn't bother until we'll really need this?

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ Recommendation: for ease of reading, use the following order:
 - Fixed
 -->
 
+## [Unreleased]
+### Added
+- Dataset definition: added possibility to set defaults in templates:
+```yaml
```
sergiimk, line 17: Nitpick: please indent the code block show it shows up under the list item

src/infra/core/src/ingest/fetch_service/core.rs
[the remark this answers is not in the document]
demusdev: > In your case we'd definitely want to explain to the user why ingest failed, so it shouldn't be an internal error.
Thanks for explanation! I was thinking it's kind of generic error. Now the purpose is pretty clear.

src/infra/core/src/ingest/fetch_service/core.rs
demusdev: Will do 👍 As far as I'm touching templates, WDYT about support them in image field es well?
```
image: "ghcr.io/kamu-data/fetch-com.defillama:${{ env.image_ver || '0.1.5' }}"
```

src/infra/core/src/ingest/fetch_service/core.rs
demusdev: I like this more. I was not sure so decided take just a set of valid url symbols for start.

src/infra/core/src/ingest/fetch_service/core.rs
sergiimk: Sure, that might be helpful in some cases.

src/infra/core/src/ingest/fetch_service/core.rs
```diff
@@ -345,3 +317,87 @@ impl std::convert::From<EventTimeSourceError> for PollingIngestError {
...
+
+        let (var, template) = var_and_template();
+
+        let env_vars_empty: HashMap<String, DatasetEnvVar> = HashMap::new();
+        let env_vars: HashMap<String, DatasetEnvVar> =
+            HashMap::from([(var.clone(), ds_env_var(&var, "222").unwrap())]);
+
+        // When dataset_env_vars is empty and system env var is set,
+        // its value is being used.
+        std::env::set_var(&var, "111");
```
sergiimk, line 373: It's not a good practice to mess with the global state (process environment) within the tests. I see that unfortunately one of our old tests does it too.

Since this test doesn't really care where the value comes from (from env vars or elsewhere) it might be much better to mock the `DatasetKeyValueService` to return a fixed set of values.

sergiimk: Actually it seems that you don't even need to check this logic here and can avoid sys env altogether.

sergiimk (commented):
Looks pretty good to me :+1: 
Once @zaychenko-sergei approves too - should be good to merge.

zaychenko-sergei (approved):

sergiimk (approved):
Good to merge (once tests are green)

Issue #857, opened 2024-09-26
SHOW TABLES query times out after a minute


zaychenko-sergei:
Depends on #986.

Note: after key blocks from #986 are implemented, this problem  will unlikely need anything else specifically, but still should be tested and re-measured.

---

Open a random dataset in `demo` environment, and queried `show tables`.
The query fails after 1 minutes with a server timeout:

![image](https://github.com/user-attachments/assets/8bbaf3a5-01cb-43fa-93e5-bbbbd47c9669)

Grafana tracing shows numerous schema queries for every dataset:

![image](https://github.com/user-attachments/assets/8e3b2ec2-adae-4616-a892-43ab5dbdcf51)

followed by endless repository read iterations:

![image](https://github.com/user-attachments/assets/9198e9ae-abfa-49a6-89e8-8a8160d64315)

Assuming this is caused with the following code fragment, that simply looks for `SetDataSchema` event in every dataset:

```
    #[tracing::instrument(level="info", skip_all, fields(dataset_handle = ?self.dataset_handle))]
    async fn init_table_schema(&self) -> Result<SchemaRef, InternalError> {
        let maybe_set_data_schema = self
            .dataset
            .as_metadata_chain()
            .accept_one(SearchSetDataSchemaVisitor::new())
            .await
            .int_err()?
            .into_event();

        if let Some(set_data_schema) = maybe_set_data_schema {
            set_data_schema.schema_as_arrow().int_err()
        } else {
            Ok(Arc::new(Schema::empty()))
        }
    }
```

PR #941, opened 2024-11-07
Introduced DatasetRegistry abstraction, encapsulating listing and resolution of datasets

zaychenko-sergei:
## Description

Related to #857.

Introduced `DatasetRegistry` abstraction, encapsulating listing and resolution of datasets.

Key changes:
- Registry is backed by database-stored dataset entries, which are automatically maintained
- Scope for `DatasetRepository` is now limited to support `DatasetRegistry` and in-memory dataset dependency graph
- New concept of `ResolvedDataset`: a wrapper arround `Arc<dyn Dataset>`, aware of dataset identity
- `DatasetRegistryRepoBridge` utility connects both abstractions in a simple way for testing needs
- Query and Dataset Search functions now consider only the datasets accessible for current user
- Core services now explicitly separate planning (transactional) and execution (non-transactional) processing phases
- Similar decomposition introduced in task system execution logic
- Revised implementation of core commands and services: `pull`, `push`, `reset`, `verify`, `compact`, setting watermark
- All commands, except `kamu ui` and `kamu system api-server` now require to be backed by a db transaction
- More parallelism from `pull` command, allowing to mix ingest/sync/transform operations of the same depth level
- Optimized `pull` flow, when a single non-recursive dataset is sent for processing
- Batched form for dataset authorization checks
- Ensuring correct transactionality for dataset lookup and authorization checks all over the code base
- Passing multi/single tenancy as an enum configuration instead of boolean
- Renamed outbox "durability" term to "delivery mechanism" to clarify the design intent
- Greatly reduced complexity and code duplication of many use case and service tests with `oop` macro for inheritance of harnesses

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/domain/core/src/entities/working_datasets_map.rs
```diff
@@ -0,0 +1,66 @@
...
+use std::sync::Arc;
+
+use opendatafabric::{DatasetHandle, DatasetID};
+
+use crate::Dataset;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Default)]
+pub struct WorkingDatasetsMap {
```
sergiimk, line 20: The name stumbled me a bit when I saw it in function parameters. Although it consists of maps on the inside for fast lookups, the overall data structure seems more similar to a `Set<ResolvedDataset>` to me. If we were to implement mut methods on it, it would certainly behave like a set having `fn insert(&mut self, ds: ResolvedDataset)`.

Should we name it `ResolvedDatasetSet`?

src/domain/core/src/services/dataset_registry.rs
```diff
@@ -0,0 +1,124 @@
...
+        &self,
+        dataset_ref: &DatasetRef,
+    ) -> Result<Option<DatasetHandle>, InternalError>;
+
+    async fn get_dataset_by_ref(
+        &self,
+        dataset_ref: &DatasetRef,
+    ) -> Result<Arc<dyn Dataset>, GetDatasetError>;
+
+    fn get_resolved_dataset_by_handle(&self, dataset_handle: &DatasetHandle) -> ResolvedDataset;
```
sergiimk, line 55: IMO it would simplify things a lot if we always returned `ResolvedDataset` everywhere - it would eliminate 2-3 methods from these interfaces.

Especially as per my other comment we make `ResolvedDataset` usable directly as a stand-in for `dyn Dataset` and allow simple conversion via `.into()`

src/domain/core/src/services/ingest/polling_ingest_service.rs
```diff
@@ -233,13 +219,6 @@ impl TemplateInvalidPatternError {
 // TODO: Revisit error granularity
 #[derive(Debug, Error)]
 pub enum PollingIngestError {
-    #[error(transparent)]
-    DatasetNotFound(
```
sergiimk, line 237: Oh this is so cool that some errors are not even possible at this layer any more!

src/domain/core/src/services/pull_request_planner.rs
```diff
@@ -0,0 +1,427 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait PullRequestPlanner: Send + Sync {
+    async fn build_pull_plan(
+        &self,
+        request: PullRequest,
+        options: &PullOptions,
+        tenancy_config: TenancyConfig,
+    ) -> Result<PullPlanIterationJob, PullResponse>;
```
sergiimk, line 31: Hm, how does this method act when `options.recursive` is `true`? Shouldn't it still return a Vec of iterations?

src/domain/core/src/services/sync_service.rs
```diff
@@ -24,31 +25,54 @@ use crate::*;
...
-    pub dst: DatasetRefAny,
+    pub src: SyncRequestSource,
+    pub dst: SyncRequestDestination,
 }
 
-#[derive(Debug, Clone)]
+pub struct SyncRequestSource {
+    pub dataset: Arc<dyn Dataset>,
+    pub src_ref: DatasetRefAny,
+    pub sync_ref: SyncRef,
```
sergiimk, line 47: This and `SyncRequestDestination` structs stand out to me compared to the simplicity of ingest/transform services.

I kinda expected something like:
```rust
pub struct SyncRequest {
    pub src: SyncRef,
    pub dst: SyncRef,
}

pub enum SyncRef {
  Local(ResolvedDataset),
  LocalNew(DatasetAlias),
  Remote(Arc<Url>),
}
```
and have a clear separation of layer that resolves `DatasetRefAny` into one of these `SyncRef`s (`SyncPlanner`?) so that `DatasetRef`s were not appearing in the low-level sync API just like they don't appear in pull plans.

Just looking at this struct I'm not sure what to expect from `src_ref` and `sync_ref` and how they overlap. Some comments would help.

src/domain/core/src/services/push_request_planner.rs
```diff
@@ -14,31 +14,53 @@ use opendatafabric::*;
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 pub struct PushResponse {
     /// Local dataset handle, if resolved
     pub local_handle: Option<DatasetHandle>,
     /// Destination reference, if resolved
     pub target: Option<DatasetPushTarget>,
     /// Result of the push operation
-    pub result: Result<SyncResult, PushError>,
+    pub result: Result<(SyncResult, Arc<dyn Dataset>), PushError>,
```
sergiimk, line 63: maybe extract dataset into a separate field?
```rust
pub target_dataset: Option<Arc<dyn Dataset>>
```

src/domain/core/src/services/remote_aliases_registry.rs
```diff
@@ -7,29 +7,28 @@
...
 
-#[async_trait]
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
 pub trait RemoteAliasesRegistry: Send + Sync {
     async fn get_remote_aliases(
         &self,
-        dataset_ref: &DatasetRef,
+        dataset: Arc<dyn Dataset>,
```
sergiimk, line 24: Aliases fundamentally don't belong in an ODF dataset, and likely will be one of the next few things we move into a DB. This will make this interface unusable. Perhaps it should work with `DatasetHandle`s, and for the time being the implementation will use repo to access the file `/<dataset>/info`, but later will switch to DB without breaking the interface.

src/domain/core/src/services/remote_aliases_registry.rs
```diff
@@ -38,31 +37,24 @@ pub enum GetAliasesError {
...
-#[async_trait]
+#[async_trait::async_trait]
 pub trait RemoteAliasResolver: Send + Sync {
     // Resolve remote push target.
     // Firstly try to resolve from AliasRegistry, if cannot do it
     // try to resolve via repository registry
     async fn resolve_push_target(
         &self,
-        local_dataset_handle: &DatasetHandle,
+        local_target: ResolvedDataset,
```
sergiimk, line 51: Same here, likely better to continue to pass only the handle, not to surface internal concern of where the aliases are currently stored in the interface.

sergiimk (commented):
Reviewed changes in `domain/` up until `transform_elaboration_service` - will continue tomorrow.

src/domain/core/src/entities/dataset.rs
```diff
@@ -64,6 +65,8 @@ pub trait Dataset: Send + Sync {
         checkpoint: Option<&CheckpointRef>,
     ) -> Result<ExecuteTransform, InternalError>;
 
+    fn try_get_ipfs_compatible_url(&self) -> Option<&Url>;
```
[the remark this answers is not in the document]
sergiimk: I initially thought that we will want the flexibility of separately storing different parts of the dataset (e.g. metadata blocks stored in fast durable storage, while raw data of derivative datasets is stored in some best-effort non-replicated storage).

But now I feel like we're arriving to a model where datasets in file storage are always stored in their complete form, but metadata blocks can optionally be in DB and lazily synchronized with the storage.

In this case we will always be able to get a top-level internal URL that refers to the entire dataset and not just its parts, thus `get_internal_url()` on Dataset level will make sense.

src/domain/core/src/services/transform/transform_elaboration_service.rs
```diff
@@ -0,0 +1,84 @@
...
+    ResolvedDataset,
+    TransformListener,
+    TransformOptions,
+    WorkingDatasetsMap,
+};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait TransformElaborationService: Send + Sync {
```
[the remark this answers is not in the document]
zaychenko-sergei: In theory it could be merged with the execution service, but definitely not with the planner.

src/domain/datasets/domain/src/services/dataset_entry_service.rs
```diff
@@ -0,0 +1,48 @@
...
+use internal_error::InternalError;
+use opendatafabric::AccountID;
+use thiserror::Error;
+
+use crate::DatasetEntry;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait DatasetEntryService: Sync + Send {
```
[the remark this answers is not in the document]
zaychenko-sergei: Not yet used, but likely will be in future as we add more high-level operations.

src/infra/core/src/transform/transform_helpers.rs
```diff
@@ -0,0 +1,309 @@
...
+        prev_offset: prev_query.as_ref().and_then(ExecuteTransform::last_offset),
+        vocab: set_vocab.unwrap_or_default().into(),
+        input_states,
+        prev_checkpoint: prev_query.and_then(|q| q.new_checkpoint.map(|c| c.physical_hash)),
+    })
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub(crate) async fn get_transform_input_from_query_input(
```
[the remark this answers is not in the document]
zaychenko-sergei: They are only called from elaboration service because of compacting feature, which forces the partial replanning activity over already resolved datasets.

I would rather rethink the need in that feature.

src/domain/core/src/services/push_request_planner.rs
zaychenko-sergei: Separate field - no, because we have nothing to assign when there is an error.
But, I've extracted a separate structure:

```
pub struct PushResponse {
    /// Local dataset handle, if resolved
    pub local_handle: Option<DatasetHandle>,
    /// Destination reference, if resolved
    pub target: Option<DatasetPushTarget>,
    /// Result of the push operation
    pub result: Result<PushResult, PushError>,
}

pub struct PushResult {
    pub sync_result: SyncResult,
    pub target_dataset: Arc<dyn Dataset>,
}
```

src/domain/core/src/services/pull_request_planner.rs
zaychenko-sergei: Not allowed. This entry is supposed to solely simplify life of the task executor.

```
#[async_trait::async_trait]
impl PullRequestPlanner for PullRequestPlannerImpl {
    #[tracing::instrument(level = "debug", skip_all, fields(?request, ?options))]
    async fn build_pull_plan(
        &self,
        request: PullRequest,
        options: &PullOptions,
        tenancy_config: TenancyConfig,
    ) -> Result<PullPlanIterationJob, PullResponse> {
        assert!(!options.recursive);
```

src/domain/core/src/entities/working_datasets_map.rs
zaychenko-sergei: I like the idea of storing `ResolvedDataset` there. A bit tricker with handles iteration, but I managed. So now we have `ResolvedDatasetsMap`.

src/domain/core/src/repos/dataset_repository.rs
```diff
@@ -47,22 +47,15 @@ pub struct CreateDatasetFromSnapshotResult {
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 #[async_trait]
-pub trait DatasetRepository: DatasetRegistry + Sync + Send {
-    fn is_multi_tenant(&self) -> bool;
-
-    async fn resolve_dataset_ref(
+pub trait DatasetRepository: Sync + Send {
```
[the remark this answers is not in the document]
zaychenko-sergei: Yes, it is storage de-facto. Wanna rename it now?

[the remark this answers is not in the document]
zaychenko-sergei: We decided not to rename in this PR

src/domain/core/src/services/dataset_registry.rs
zaychenko-sergei: Done

src/domain/core/src/entities/resolved_dataset.rs
```diff
@@ -0,0 +1,44 @@
...
+use opendatafabric::DatasetHandle;
+
+use super::Dataset;
+use crate::CreateDatasetResult;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Clone)]
+pub struct ResolvedDataset {
+    pub dataset: Arc<dyn Dataset>,
```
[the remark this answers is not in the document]
zaychenko-sergei: Had some success with `Deref<Target = Arc<dyn Dataset>>`

src/domain/core/src/services/sync_service.rs
zaychenko-sergei: Almost.

```
pub struct SyncRequest {
    pub src: SyncRef,
    pub dst: SyncRef,
}

pub enum SyncRef {
    Local(ResolvedDataset),
    LocalNew(DatasetAlias),
    Remote(SyncRefRemote),
}

pub struct SyncRefRemote {
    pub url: Arc<Url>,
    pub dataset: Arc<dyn Dataset>,
    pub original_remote_ref: DatasetRefRemote,
}
```

The original remote URL is important for progress reporting and some tests.

src/domain/core/src/services/remote_aliases_registry.rs
zaychenko-sergei: Solved

src/infra/core/src/sync_service_impl.rs
```diff
@@ -500,15 +343,20 @@ impl SyncServiceImpl {
...
+            src_dataset,
+        ))
     }
 
-    async fn add_to_ipfs(&self, src: &DatasetRef) -> Result<String, SyncError> {
-        let source_url = self.dataset_repo.get_dataset_url(src).await.int_err()?;
+    async fn add_to_ipfs(&self, src_dataset: &dyn Dataset) -> Result<String, InternalError> {
+        let source_url = src_dataset
+            .try_get_ipfs_compatible_url()
+            .expect("Dataset must provide IPFS-compatible URL");
```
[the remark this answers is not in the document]
zaychenko-sergei: Done

src/domain/core/src/entities/dataset.rs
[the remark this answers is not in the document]
zaychenko-sergei: Agree, done

Issue #939, opened 2024-11-05
Add version for outbox messages


rmn-boiko:
Currently, our outbox processor reads all messages from the database after the service starts. But if some message structure was changed it will cause panic.

The goal is to extend outbox messages with version and make sure service does not read messages with outdated version.
Possible solution:
current message:
`pub struct FlowConfigurationUpdatedMessage {
    pub event_time: DateTime<Utc>,
    pub flow_key: FlowKey,
    pub paused: bool, 
    pub rule: FlowConfigurationRule,
}
`

Expected message:
`pub struct OutboxMessage {
    message: Enum(existing messages),
    version: i32
}
`

PR #942, opened 2024-11-07
Add outbox version

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/939

This PR does not provide any changes for inmem implementation such as I assume it is not possible that we might have messages with different versions if we are using inmem repository

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/infra/messaging-outbox/sqlite/src/repos/sqlite_outbox_message_repository.rs
```diff
@@ -93,6 +94,7 @@ impl OutboxMessageRepository for SqliteOutboxMessageRepository {
                     occurred_on
                 FROM outbox_messages
                 WHERE {producer_filters}
+                and version = {OUTBOX_MESSAGE_VERSION}
```
s373r, line 97: Minor: `indoc::formatdoc!()` can be used in this (and similar cases) -- it is a combination of the indoc::indoc!() and format!() macros

src/infra/messaging-outbox/sqlite/src/repos/sqlite_outbox_message_repository.rs
```diff
@@ -108,12 +110,13 @@ impl OutboxMessageRepository for SqliteOutboxMessageRepository {
 
             use sqlx::Row;
             let mut query_stream = query.try_map(|event_row: SqliteRow| {
-                Ok(OutboxMessage {
-                    message_id: OutboxMessageID::new(event_row.get(0)),
-                    producer_name: event_row.get(1),
-                    content_json: event_row.get(2),
-                    occurred_on: event_row.get(3),
-                })
+                println!("event_row");
```
s373r, line 113: Oh, good old-fashioned debugging through prints :)

Btw, I advise you to use `dbg!()`, since they are caught via clippy -- I've added this rule, since I sometimes forgot to remove them (debug macros) myself.

src/utils/messaging-outbox/src/entities/outbox_message.rs
```diff
@@ -13,12 +13,36 @@ use crate::OutboxMessageID;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// Version of outbox message which will be bumped only if the structure contains
+// breaking changes in the message structure
+pub const OUTBOX_MESSAGE_VERSION: i32 = 1;
```
s373r, line 18: Q: If we don't allow the `OutboxMessage` user to select a version, how can we change the version for a particular message (`.content_json`) if it becomes different?

src/utils/messaging-outbox/src/entities/outbox_message.rs
```diff
@@ -13,12 +13,36 @@ use crate::OutboxMessageID;
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 #[derive(Debug, Clone)]
 pub struct OutboxMessage {
     pub message_id: OutboxMessageID,
     pub producer_name: String,
     pub content_json: serde_json::Value,
     pub occurred_on: DateTime<Utc>,
+    pub version: i32,
```
s373r, line 28: u32? I guess we don't need negative versions

src/infra/messaging-outbox/repo-tests/src/outbox_message_repository_test_suite.rs
```diff
@@ -341,6 +347,50 @@ pub async fn test_reading_messages_above_max_with_multiple_producers(catalog: &C
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub async fn test_outbox_messages_version(catalog: &Catalog) {
```
s373r, line 350: Q: Can also add not only an example with the past version, but also the addition of a future version?

s373r (commented):
Add minor comments / questions

src/infra/messaging-outbox/repo-tests/src/outbox_message_repository_test_suite.rs
rmn-boiko: We suppose we will work only with latest. I can imagine only 1 case when we need to ignore highest versions, it's when we deployed and reverted changes but to test it might be tricky, because once we rich a `higher mock version` we should always go back and change it.

src/utils/messaging-outbox/src/entities/outbox_message.rs
rmn-boiko: It will not touch users. 
We should cover next scenario.
During service startup we are reading all messages from db, but if we made breaking changes we should identify that service should not work with outdated messages. So we will bump it only with `OutboxMessage` breaking changes In RP

src/infra/messaging-outbox/sqlite/src/repos/sqlite_outbox_message_repository.rs
rmn-boiko: Looks good. Will use it

src/utils/messaging-outbox/src/entities/outbox_message.rs
rmn-boiko: It's kind tricky, because we need to perform additional cast to support postgres insert. They removed support of u32. I would keep as it is. WDYT?

s373r: Got it, thanks a lot for providing the context. Yes, you are right, I recall problems with Postgres :( It's a shame that one of the implementations imposes a type for us within the model.

Still, it doesn't look like (to me) a worthwhile reason to complicate the decision. For this reason, please add a short comment to the field why `i32` is chosen, otherwise the readers eye will really cling to it

src/utils/messaging-outbox/src/entities/outbox_message.rs
s373r: Interesting approach, I'll wait for comments from other teammates

src/utils/messaging-outbox/src/entities/outbox_message.rs
sergiimk: Personally I would any day prefer a bunch ugly type casts on repository implementation level to imprecise types at domain layer. Serialization should not dictate us the types. But I think we already made this trade-off in `OutboxMessageID` and perhaps other ID types, so won't insist in this simple case, but in general good domain types should make invalid state unrepresentable.

src/utils/messaging-outbox/src/entities/outbox_message.rs
sergiimk: I think it will eventually become a responsibility of the consumer to determine whether they can or cannot consume a specific version of event.

E.g. it's a common practice for a service to emit V1 and V2 of the same domain event during a transitional period and then stop producing V1 once all consumers have upgraded.

I'm not saying we should spend time solving the upgradeability problem now, but we should consider better where versions really belong so that at least we didn't have to migrate from one versioning scheme to another and create meta-versions :)

zaychenko-sergei: I agree that the version of events is a property of a particular producer.
We cannot have a single global version for all kinds of events, as this is a version of the payload, not the main `OutboxMessage` structure.

For now it should be easy, just declare separate constant near each message type definition, and use it in emitting code.

src/utils/messaging-outbox/src/entities/outbox_message.rs
zaychenko-sergei: Totally true. Besides. we have a lot of ugly casts in repository already, so a few more won't change the picture there, but the model would better  be clean.

src/utils/messaging-outbox/src/entities/outbox_message.rs
zaychenko-sergei: If we could somehow extract the version from message class metadata, that would also simplify the emission. We already have a `Message` trait:

```
pub trait Message: Clone + Serialize + for<'a> Deserialize<'a> + Send + Sync {}
```

Perhaps we could demand returning a version number here.

zaychenko-sergei (changes requested):
Before we go into supporting multiple versions of consumers, ..., I'd like to see at least a basic check that the version of the deserialized message is supported, which logs an error for now..

src/utils/messaging-outbox/src/entities/outbox_message.rs
sergiimk: Do individual messages currently have version number in them when serialized into JSON?

Do you think per-message versioning would make sense or you think per-producer is better?

Personally I think per-message versions work much better for gradual deprecation. If we version on producer (service) level, then to deprecate a single message type we'd need to emit ALL messages as both `v1` and `v2` for some time, i.e. increase `2x` the message flow of the service.

And if we're doing per-message versioning - should we still surface the version into a column or can we rely on JSON access capabilities of Postgres to read version from serialized data?

Issue #504, opened 2024-02-13
Add warning when trying to delete not syncronized remote dataset


rmn-boiko:
Currently, when deleting a remote dataset, we skip checking if it is up to date. Otherwise we do not allow delete dataset with existing derivatives without deleting them first.

The objective is to display a warning when the user attempts to delete a dataset that has not been fully synchronized with a remote.

PR #957, opened 2024-11-13
Add warning when trying to delete not synchronized dataset

demusdev:
## Description

Closes: #504 

## Console part
Print warning when dataset has remotes which are out of sync, or we failed to get remote status

Screenshot 2024-11-21 at 02 00 26

### Manual testing
1. Unpack [cli_504.zip](https://github.com/user-attachments/files/17755329/cli_504.zip)
2. Run `./reset.sh` to create kamu workspace and all the test entities
3. Try to replace/delete datasets: `kamu rm -a`, `kamu add test2.yaml --replace`

## GraphQL part
`datasets.metadata` response was extended to return push statuses

```bash
kamu system api-server --http-port 55667

DS_ID=$(kamu log test.test1 | rg -o 'did:odf:[0-9a-f]+') \
             GQL_QUERY="{datasets {byId (datasetId: \\\"$DS_ID\\\") {metadata {pushSyncStatuses {statuses {remote, result {... on CompareChainsResultStatus { message } ... on CompareChainsResultError { reason { message } } }}}}}}}" \
             DATA_RAW="{\"query\":\"$GQL_QUERY\"}" \
             curl -s 'http://127.0.0.1:55667/graphql' --data-raw $DATA_RAW \
             | jq .data.datasets.byId.metadata
```
<p align="center">
⬇️ ⬇️ ⬇️
</p>

```json
{
  "pushSyncStatuses": {
    "statuses": [
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/ahead",
        "result": {
          "message": "BEHIND"
        }
      },
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/in-sync",
        "result": {
          "message": "EQUAL"
        }
      },
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/unknown",
        "result": {
          "reason": {
            "message": "Remote dataset not found"
          }
        }
      },
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/behind",
        "result": {
          "message": "AHEAD"
        }
      }
    ]
  }
}
```

## Web UI part (TODO)
Fetch synchronisation status when user want to delete a dataset, show sync warnings inside confirmation dialog

Screenshot 2024-11-13 at 18 10 40

- [x] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [x] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/app/cli/src/commands/add_command.rs
```diff
@@ -14,16 +14,16 @@ use kamu::domain::*;
...
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 pub struct AddCommand {
-    interact: Arc<Interact>,
     resource_loader: Arc<dyn ResourceLoader>,
     dataset_repo: Arc<dyn DatasetRepository>,
     create_dataset_from_snapshot: Arc<dyn CreateDatasetFromSnapshotUseCase>,
     delete_dataset: Arc<dyn DeleteDatasetUseCase>,
+    helpers: Arc<dyn CommandHelpers>,
```
zaychenko-sergei, line 26: Perhaps, a component should not be named in plural form: `CommandHelper`

src/domain/core/src/entities/dataset.rs
```diff
@@ -126,6 +128,54 @@ pub struct CommitResult {
...
+pub enum SyncStatus {
+    InSync,
+    OutOfSync,
+}
+
+#[derive(Clone, Debug)]
+pub struct DatasetPushStatuses {
+    pub statuses: HashMap<DatasetRefRemote, Option<SyncStatus>>,
+}
+
```
zaychenko-sergei, line 143: Move these definitions into your new service

src/domain/core/src/entities/dataset.rs
```diff
@@ -126,6 +128,54 @@ pub struct CommitResult {
...
+    pub fn in_sync(&self) -> Vec<&DatasetRefRemote> {
+        self.statuses
+            .iter()
+            .filter_map(|(ds_ref, mb_status)| match mb_status {
+                Some(SyncStatus::InSync) => Some(ds_ref),
+                _ => None,
+            })
+            .collect()
+    }
+    pub fn out_of_sync(&self) -> Vec<&DatasetRefRemote> {
```
zaychenko-sergei, line 154: Have blank lines between methods

src/infra/core/src/push_status_service_impl.rs
```diff
@@ -0,0 +1,98 @@
...
+                    .as_metadata_chain()
+                    .resolve_ref(&BlockRef::Head)
+                    .await
+                    .ok()
+                    .map(|remote_head| {
+                        if remote_head == local_head {
+                            SyncStatus::InSync
+                        } else {
+                            SyncStatus::OutOfSync
+                        }
```
zaychenko-sergei, line 90: I wonder if it makes any sense to differentiate between the cases:
 - local dataset is ahead
 - remote dataset is ahead
 - local & remote have diverged
 
 We have something like this in `SyncService` where we compare chains.

src/app/cli/src/app.rs
```diff
@@ -467,6 +470,7 @@ pub fn configure_base_catalog(
     b.add::<CreateDatasetFromSnapshotUseCaseImpl>();
     b.add::<DeleteDatasetUseCaseImpl>();
     b.add::<RenameDatasetUseCaseImpl>();
+    b.add::<PushStatusServiceImpl>();
```
zaychenko-sergei, line 473: Place the registration near `PushService`. This is a list of use cases. No behavioral effect, but there is at least some order in this registraiton code.

src/infra/core/src/push_status_service_impl.rs
```diff
@@ -0,0 +1,98 @@
...
+    PushStatusService,
+    RemoteAliasKind,
+    RemoteAliasesRegistry,
+    SyncStatus,
+};
+use opendatafabric::{DatasetHandle, DatasetRefRemote};
+
+#[component(pub)]
+#[interface(dyn PushStatusService)]
+pub struct PushStatusServiceImpl {
```
zaychenko-sergei, line 29: This service could be covered with some tests.

src/infra/core/src/use_cases/mod.rs
```diff
@@ -20,3 +20,5 @@ pub use create_dataset_from_snapshot_use_case_impl::*;
 pub use create_dataset_use_case_impl::*;
 pub use delete_dataset_use_case_impl::*;
 pub use rename_dataset_use_case_impl::*;
+
+pub use crate::push_status_service_impl::*;
```
zaychenko-sergei, line 24: Should not be here

src/domain/core/src/services/push_status_service.rs
```diff
@@ -0,0 +1,30 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+use internal_error::InternalError;
+use opendatafabric::DatasetHandle;
+
+use crate::DatasetPushStatuses;
```
zaychenko-sergei, line 15: Normally we place `use` declarations above the separation boundary for better readability.

src/app/cli/src/commands/command_helpers.rs
```diff
@@ -0,0 +1,86 @@
...
+use opendatafabric::DatasetHandle;
+
+use crate::{CLIError, Interact};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait CommandHelper: Send + Sync {
+    async fn confirm_deletion(&self, dataset_handles: &[DatasetHandle]) -> Result<(), CLIError>;
+}
```
sergiimk, line 24: What is the point of separating this helper from the delete command?

To me types like "Utils" and "Helper" are always a sign of code smell. If you cannot give some component a clear name - it usually indicates that it's not logically decomposed correctly and you don't actually know what purpose it serves.

Also a good sanity test is to ask "what happens if all other commands need helpers?". Do we end up dumping code from multiple commands into this type? Or will we have `CommandHelper2` and `CommandHelper3`. This decomposition doesn't make sense to me.

And why do we need the complexity of hiding helper behind an interface and injecting via DI? Is it ever going to be replaced by another implementation?

src/domain/core/src/services/push_status_service.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use internal_error::InternalError;
+use opendatafabric::{DatasetHandle, DatasetRefRemote};
+
+use crate::utils::metadata_chain_comparator::CompareChainsResult;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait PushStatusService: Send + Sync {
```
sergiimk, line 20: Why do we need another service? It seems to me that this functionality would fit just fine in the existing `PushService`. They both operate on the same logical level of push aliases.

src/app/cli/src/commands/command_helpers.rs
```diff
@@ -0,0 +1,86 @@
...
+            if !statuses.all_synced() {
+                eprintln!(
+                    "{} dataset is out of sync with remote(s):",
+                    console::style(&hdl.alias).bold()
+                );
+                for (remote, status) in statuses.out_of_sync() {
+                    let status_desc = match status {
+                        CompareChainsResult::LhsAhead { .. } => "ahead of ",
+                        CompareChainsResult::LhsBehind { .. } => "behind ",
+                        CompareChainsResult::Divergence { .. } => "diverge from ",
```
sergiimk, line 56: `diverge` -> `diverged`?

src/domain/core/src/services/push_status_service.rs
```diff
@@ -0,0 +1,81 @@
...
+        &self,
+        dataset_handle: &DatasetHandle,
+    ) -> Result<DatasetPushStatuses, InternalError>;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Clone, Debug)]
+pub struct DatasetPushStatuses {
+    pub statuses: HashMap<DatasetRefRemote, Option<CompareChainsResult>>,
```
sergiimk, line 36: Weird to see `Option` here. When `None` status is returned?

I expected something more like `Result<CompareResult, CheckStatusError>`.

src/domain/core/src/services/push_status_service.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.statuses
+            .iter()
+            .filter_map(|(ds_ref, mb_status)| match mb_status {
+                Some(CompareChainsResult::Equal) => Some(ds_ref),
+                _ => None,
+            })
+            .collect()
+    }
+
+    pub fn out_of_sync(&self) -> Vec<(&DatasetRefRemote, &CompareChainsResult)> {
```
sergiimk, line 50: This kind of splitting logic seems very specific to how caller wants to process the results and thus probably belongs in the presentation layer rather than in domain.

I think domain services should have very simple return types that have minimal assumptions on how client will be processing them. Hash map in this case seems like an overkill, especially since all you do is iterate it, not access by key.

Personally I would simplify things down to:
```rust
async fn get_statuses(
     &self,
    dataset_handle: &DatasetHandle,
) -> Result<Vec<Result<CompareResult, StatusCheckError>>, InternalError>;
```

src/app/cli/src/commands/command_helpers.rs
demusdev: In both `add` and `delete` command we need to print sync warning and show delete confirmation. This code was moved to `confirm_deletion` function. As far as I didn't found better place, I decided to create `command_helpers` file in order to have such random helpers into that file. What would you propose here?

> And why do we need the complexity of hiding helper behind an interface and injecting via DI? Is it ever going to be replaced by another implementation?

interface is redundant, but I think it makes sense to keep it in DI, because it has 2 deps which we'll need to pass as function params otherwise.

sergiimk: Ah, I missed that `confirm_deletion` is also used by Add command.

Would composition work better here? Can Add command instantiate and call the Delete command for those datasets that already exist when `--replace` flag is present, let Delete do its thing, and then proceed if Delete was successful?

src/domain/core/src/services/push_status_service.rs
sergiimk: I don't quite understand what constitutes an `unknown` status. The dataset is either in sync, not in sync, or some error happened while we were checking. If there was an error - I think we should surface it as such so that presentation layer could decide how to display it.

src/domain/core/src/services/push_service.rs
```diff
@@ -122,3 +128,36 @@ impl From<GetDatasetError> for PushError {
...
+    ),
+    #[error(transparent)]
+    GetRefError(
+        #[from]
+        #[backtrace]
+        GetRefError,
+    ),
+    #[error("Remote URL not found")]
+    NoRemoteUrl,
+}
```
sergiimk, line 154: Currently your error is like a union of "everything that can go wrong under the hood" errors, but this is not very user friendly. For example, as a user I get `GetRefError` in response - how can I react to it, or what does it even tell me, really?

We consider errors a part of the API and design them NOT based on what can go wrong under the hood, but rather based on what logically can go wrong and how to communicate it best to the user ([this](https://github.com/kamu-data/kamu-cli/blob/e496fe036b58367a5119a1c25721d8c333ec65af/src/domain/core/src/entities/metadata_chain.rs#L614) is a good example).

I would start with including `Internal` error to represent cases where "some bad stuff happened under the hood". Then see what logical situations we should represent to make them distinguishable to the user. For example `GetRefError::NotFound` could be mapped into `RemoteDatasetNotFound` - so the user will know that remote dataset does not exist any more.

This alone might be enough for now - we can always come back and extract more logical errors later.

src/domain/core/src/services/push_status_service.rs
sergiimk: Ok, I see that in service API you are already representing errors explicitly and there is no "unknown" state. In GQL API I think we should do the same. [Here](https://github.com/kamu-data/kamu-cli/blob/e496fe036b58367a5119a1c25721d8c333ec65af/src/adapter/graphql/src/mutations/dataset_metadata_mut.rs#L153) is a good example of how we represent logical errors as enums.

src/domain/core/src/services/push_service.rs
```diff
@@ -29,6 +30,11 @@ pub trait PushService: Send + Sync {
         options: PushMultiOptions,
         sync_listener: Option<Arc<dyn SyncMultiListener>>,
     ) -> Vec<PushResponse>;
+    /// Returns sync status of all push remotes connected with a given dataset
+    async fn status_multi(
```
sergiimk, line 34: We use `_multi` prefix for methods that operate on multiple datasets at a time, but here you are working with just one. I would rename to `check_remote_status` to highlight that we're actually reaching out to remote repos here.

src/adapter/graphql/src/scalars/dataset_metadata.rs
```diff
@@ -0,0 +1,107 @@
...
+        Self { statuses }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// CompareChainsResult
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Enum, Serialize, Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CompareChainsResult {
```
zaychenko-sergei, line 56: We discussed using a derived macro:

`#[graphql(remote = ...)]`

This would allow to avoid customizing `From` implementation.

src/adapter/graphql/src/scalars/dataset_metadata.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct DatasetRefRemote(odf::DatasetRefRemote);
+
+impl From<odf::DatasetRefRemote> for DatasetRefRemote {
+    fn from(value: odf::DatasetRefRemote) -> Self {
+        DatasetRefRemote(value)
+    }
+}
+
```
zaychenko-sergei, line 86: This struct naturally belongs to `./src/adapter/graphql/src/scalars/dataset_id_name.rs`

src/app/cli/src/commands/delete_command.rs
```diff
@@ -20,57 +21,133 @@ use crate::Interact;
...
+                        CompareChainsResult::LhsBehind { .. } => {
+                            out_of_sync.push((remote, "behind"));
+                        }
+                        CompareChainsResult::Divergence { .. } => {
+                            out_of_sync.push((remote, "diverged from"));
+                        }
+                        _ => (),
+                    },
+                    Err(e) => unknown.push((remote, e)),
+                }
```
zaychenko-sergei, line 89: The beauty of Rust allows customizing only what's different:

```
match check_result {
    Ok(status) => {
        let maybe_reason = match status {
            CompareChainsResult::LhsAhead => Some("ahead of"),
            ...
            _ => None
        };
        if let Some(reason) = maybe_reason {
            out_of_sync.push((remote, reason));
        }
    }
    ...
}
```

src/app/cli/src/commands/delete_command.rs
```diff
@@ -20,57 +21,133 @@ use crate::Interact;
...
-            .try_collect()
-            .await?
+            match &self.target {
+                DeleteTarget::Handles(hs) => hs.clone(),
+                DeleteTarget::Patterns(ps) => {
+                    filter_datasets_by_local_pattern(self.dataset_repo.as_ref(), ps.clone())
+                        .try_collect()
+                        .await?
+                }
+            }
```
zaychenko-sergei, line 150: Can we avoid clone here? I think you can move instead in both occurrences of `.clone()`

src/app/cli/src/commands/delete_command.rs
```diff
@@ -20,57 +21,133 @@ use crate::Interact;
...
     async fn run(&mut self) -> Result<(), CLIError> {
-        if self.dataset_ref_patterns.is_empty() && !self.all {
+        let target_is_empty = match &self.target {
+            DeleteTarget::Handles(hs) => hs.is_empty(),
+            DeleteTarget::Patterns(ps) => ps.is_empty(),
+        };
+
+        if target_is_empty && !self.all {
             return Err(CLIError::usage_error("Specify a dataset or use --all flag"));
         }
```
zaychenko-sergei, line 138: The best is to move it to `validate_args` override from `Command` trait

src/app/cli/src/commands/command_helpers.rs
zaychenko-sergei: A command calling another command is an example of even worse decomposition.

zaychenko-sergei: I clearly see this as a CLI-level service. Perhaps the name `CommandHelper` isn't brilliant, we could have `ConfirmDeleteService` at CLI level. It really does not need to have a trait and can be injected in commands directly, similarly to `WorkspaceService`.

zaychenko-sergei: Or it could be a `CliInteractionService`, `CliPromptService`, or whatever similar.

src/domain/core/src/services/push_service.rs
demusdev: done

src/domain/core/src/services/push_service.rs
demusdev: done

src/adapter/graphql/src/scalars/dataset_metadata.rs
demusdev: done

src/app/cli/src/commands/delete_command.rs
demusdev: done

src/domain/core/src/services/push_status_service.rs
demusdev: done

src/app/cli/src/commands/delete_command.rs
demusdev: done

src/app/cli/src/commands/command_helpers.rs
demusdev: Switched back to service approach (with no trait)

src/app/cli/src/commands/add_command.rs
```diff
@@ -283,18 +291,17 @@ impl Command for AddCommand {
...
+                    self.dataset_repo.clone(),
+                    self.delete_dataset.clone(),
+                    DeleteTarget::Handles(already_exist),
+                    self.dependency_graph_service.clone(),
+                    self.push_service.clone(),
+                    self.interact.clone(),
+                    false,
+                    false,
+                );
+                delete_cmd.run().await?;
```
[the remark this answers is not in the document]
demusdev: Discussion moved to this thread https://github.com/kamu-data/kamu-cli/pull/957#discussion_r1845235065

src/e2e/app/cli/repo-tests/src/commands/test_delete_command.rs
```diff
@@ -9,10 +9,16 @@
 
 use kamu_cli_e2e_common::{
     DATASET_DERIVATIVE_LEADERBOARD_SNAPSHOT_STR,
+    DATASET_ROOT_PLAYER_SCORES_INGEST_DATA_NDJSON_CHUNK_1,
```
sergiimk, line 12: I'm not a big fan of reuse across test suites. In my experience it leads to dumping grounds of setup code that grow huge, hard to maintain, and fragile. Would caution against `kamu_api_server_client_ext.rs` becoming that (cc @s373r).

I think inlining test data or reusing at most on the test suite (module) level is better for readability and maintainability of tests.

Just sharing an opinion, not calling for action as part of this PR.

sergiimk (commented):
Good job with the tests! On a high level all looks good to me.

resources/schema.gql
```diff
@@ -264,6 +264,23 @@ type CompactionMetadataOnly {
 	recursive: Boolean!
 }
 
+union CompareChainsResult = CompareChainsResultStatus | CompareChainsResultError
+
+type CompareChainsResultError {
+	error: String!
```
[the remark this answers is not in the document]
demusdev: done:
```
❯ DS_ID=$(kamu log test.test1 | rg -o 'did:odf:[0-9a-f]+') \
             GQL_QUERY="{datasets {byId (datasetId: \\\"$DS_ID\\\") {metadata {pushSyncStatuses {statuses {remote, result {... on CompareChainsResultStatus { message } ... on CompareChainsResultError { reason { message } } }}}}}}}" \
             DATA_RAW="{\"query\":\"$GQL_QUERY\"}" \
             curl -s 'http://127.0.0.1:55667/graphql' --data-raw $DATA_RAW \
             | jq .data.datasets.byId.metadata
{
  "pushSyncStatuses": {
    "statuses": [
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/ahead",
        "result": {
          "message": "BEHIND"
        }
      },
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/in-sync",
        "result": {
          "message": "EQUAL"
        }
      },
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/unknown",
        "result": {
          "reason": {
            "message": "Remote dataset not found"
          }
        }
      },
      {
        "remote": "file:///Users/andrii/kamu/test/cli_504/repo1/behind",
        "result": {
          "message": "AHEAD"
        }
      }
    ]
  }
}
```

src/e2e/app/cli/repo-tests/src/commands/test_delete_command.rs
demusdev: I totally agree in general. But I don't think it will be much harm if we'll have "just a dataset" with "some" data chunks, for the cases where we don't check anything specific like transformations, template strings, etc.

zaychenko-sergei (changes requested):

src/e2e/app/cli/repo-tests/src/commands/test_delete_command.rs
s373r: I'll duplicate a comment that is relevant to the thread, but for some reason fell out of it:

> > In my experience it leads to dumping grounds of setup code that grow huge, hard to maintain, and fragile
>
> I totally disagree. The reuse of test elements is great, hierarchy of test steps is great, not repeating yourself is gret. Tests are a 1st class citizen and have to be designed equally well as the main product, and in this case they don't become mess.

(c) @zaychenko-sergei, https://github.com/kamu-data/kamu-cli/pull/957#discussion_r1856225241

src/app/cli/src/app.rs
```diff
@@ -159,6 +160,7 @@ pub async fn run(workspace_layout: WorkspaceLayout, args: cli::Cli) -> Result<()
         let output_config = configure_output_format(&args, &workspace_svc);
         base_catalog_builder.add_value(output_config.clone());
         base_catalog_builder.add_value(Interact::new(args.yes, output_config.is_tty));
+        base_catalog_builder.add::<ConfirmDeleteService>(); // Should be added after Interact
```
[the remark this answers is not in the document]
demusdev: done

src/adapter/graphql/tests/tests/test_gql_metadata.rs
```diff
@@ -7,21 +7,192 @@
...
+use url::Url;
 
 use crate::utils::authentication_catalogs;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub struct MockRemoteStatusService {}
+
+#[async_trait::async_trait]
+impl RemoteStatusService for MockRemoteStatusService {
```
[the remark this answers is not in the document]
demusdev: done

src/infra/core/tests/tests/test_remote_status_service.rs
```diff
@@ -0,0 +1,393 @@
...
+use opendatafabric::{
+    AccountName,
+    DatasetHandle,
+    DatasetID,
+    DatasetKind,
+    DatasetRef,
+    DatasetRefRemote,
+    MetadataBlock,
+    Multihash,
+};
```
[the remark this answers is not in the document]
demusdev: done

src/infra/core/tests/tests/test_remote_status_service.rs
```diff
@@ -0,0 +1,393 @@
...
+        todo!()
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+struct FakeDatasetFactory {
+    datasets: HashMap<String, Arc<dyn Dataset>>,
+}
+
```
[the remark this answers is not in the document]
demusdev: done

src/infra/core/tests/tests/test_remote_status_service.rs
```diff
@@ -0,0 +1,393 @@
...
+        .prev(root_hash, 0)
+        .build()
+}
+
+fn chain() -> MetaChain {
+    MetadataChainImpl::new(
+        MetadataBlockRepositoryImpl::new(ObjectRepositoryInMemory::new()),
+        ReferenceRepositoryImpl::new(NamedObjectRepositoryInMemory::new()),
+    )
+}
```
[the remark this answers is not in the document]
demusdev: done

src/adapter/graphql/tests/tests/test_gql_metadata.rs
```diff
@@ -7,21 +7,192 @@
...
+            .add::<SystemTimeSourceDefault>()
+            .add::<DatasetRegistryRepoBridge>()
+            .add_value(TenancyConfig::SingleTenant)
+            .add_value(MockRemoteStatusService {})
+            .bind::<dyn RemoteStatusService, MockRemoteStatusService>();
+
+        NoOpDatabasePlugin::init_database_components(&mut b);
+
+        b.build()
+    };
```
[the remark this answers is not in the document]
demusdev: done

zaychenko-sergei (approved):

PR #960, opened 2024-11-14
Do not show usage error for --all flag

demusdev:
## Description

When --all flag is set for the `repo delete` command, and there are no repositories to delete, do not shoow usage error.

### Before
Screenshot 2024-11-14 at 19 54 03

### After
Screenshot 2024-11-27 at 22 02 31

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/app/cli/src/commands/repository_delete_command.rs
```diff
@@ -46,6 +46,12 @@ impl RepositoryDeleteCommand {
 #[async_trait::async_trait(?Send)]
 impl Command for RepositoryDeleteCommand {
     async fn run(&mut self) -> Result<(), CLIError> {
+        if self.names.is_empty() && !self.all {
+            return Err(CLIError::usage_error(
+                "Specify a repository or use --all flag",
+            ));
+        }
+
```
s373r, line 54: Please use (override) `Command::validate_args()` for this check

s373r: Also, I think we should work through all the variations
1) no names & no `--all`
2) no names & `--all`
3) some names & no `--all`
4) some names & `--all`

---

I think some of these options are redundant and an error message should be provided for them

src/app/cli/src/commands/repository_delete_command.rs
s373r, line 1: Optional: add E2E tests -- Here it would be great to check our errors (output messages)

PS. I'd like to help make these, if you decide to make them, let me know and I'll advise on the best way to approach it

s373r (changes requested):
Added minor comments

src/app/cli/src/commands/repository_delete_command.rs
demusdev: done

src/app/cli/src/commands/repository_delete_command.rs
demusdev: done

src/app/cli/src/commands/delete_command.rs
```diff
@@ -58,10 +58,16 @@ impl DeleteCommand {
...
+        match (self.dataset_ref_patterns.len(), self.all) {
+            (0, true) => Ok(()),
+            (len, false) if len > 0 => Ok(()),
+            (0, false) => Err(CLIError::usage_error("Specify dataset(s) or pass --all")),
+            (len, true) if len > 0 => Err(CLIError::usage_error(
+                "You can either specify dataset(s) or pass --all",
+            )),
+            _ => Err(CLIError::usage_error(
+                "Invalid combination of arguments".to_owned(),
+            )),
```
s373r, line 70: Legend:
```rust
/*1*/  (0, true) => Ok(()),
/*2*/  (len, false) if len > 0 => Ok(()),
/*3*/  (0, false) => Err(CLIError::usage_error("Specify dataset(s) or pass --all")),
/*4*/  (len, true) if len > 0 => Err(CLIError::usage_error(
                "You can either specify dataset(s) or pass --all",
            )),
/*5*/  _ => Err(CLIError::usage_error(
                "Invalid combination of arguments".to_owned(),
            )),
```

---

a) 
```rust
/*1*/  (0, true) => Ok(()),
/*4*/  (len, true) if len > 0 => Err(CLIError::usage_error(
                "You can either specify dataset(s) or pass --all",
            )),
```
We don't need an additional predicate (`if len > 0`) for `/*4*/` since the zero length is handled by branch `/*1*/`

b) Same for pair `/*2*/` and `/*3*/`
c) It looks like branch `/*5*/` is not needed since it can't be called

s373r: (Optional) d) Please sort the branches, e.g. by pairs, or in ascending order. We have four options here, roughly:
```
0 0
0 1
1 0
1 1
```

In my opinion, it might simplify the readability a little bit

src/app/cli/src/commands/repository_delete_command.rs
s373r: Hmm, I see that you did it for `DeleteCommand`, but I don't see it for `RepositoryDeleteCommand` -- do we need `match { ... }` here too?

src/e2e/app/cli/repo-tests/src/commands/test_delete_command.rs
```diff
@@ -333,3 +333,28 @@ pub async fn test_delete_warning(mut kamu_node_api_client: KamuApiServerClient)
...
+        Some(["You can either specify dataset(s) or pass --all"]),
+    )
+    .await;
+
+    kamu.assert_failure_command_execution(
+        ["delete"],
+        None,
+        Some(["Specify dataset(s) or pass --all"]),
+    )
+    .await;
```
s373r, line 357: can we also add the case of `delete player-scores` for completeness?

s373r (commented):
Minor comments added

src/app/cli/src/commands/repository_delete_command.rs
demusdev: My bad. `delete` command got new validation for free :) I think it makes sense to keep it for consistency. Now added to `repo delete` as well

src/app/cli/src/commands/delete_command.rs
demusdev: Not agree to remove guard, as matches order change may affect logic which I would like to avoid. But agree that it looks like it needs improvement. I changed to matching against slice.
Order was changed as requested.

src/e2e/app/cli/repo-tests/src/commands/test_delete_command.rs
demusdev: done

zaychenko-sergei (approved):

Issue #854, opened 2024-09-26
Persistent storage of dataset dependencies graph


zaychenko-sergei:
Currently we are using an in-memory representation of dataset dependency graph in `DependencyGraphService`.
The model is based on explicit graph representation implemented with `petgraph` crate.

The dependencies data for this graph is lazily loaded just once via repository trait `DependencyGraphRepository`, implemented by `DependencyGraphRepositoryInMemory` component, The invocation happens on first use of the graph queries, which in API-server case happens when the first user after server startup opens the search page, as it provides data on count of downstream datasets:

![image](https://github.com/user-attachments/assets/bc5d46c9-3219-4349-b796-6c88ec79edfc)

Later, the consistency of the in-memory dependencies is maintained via dataset events (created, deleted, dependencies updated).

The [measurements](https://grafana.sha.stg.internal.kamu.dev/explore?schemaVersion=1&panes=%7B%225oi%22:%7B%22datasource%22:%22365dd1cc-3781-43c1-86a1-a3f78c6ab420%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bjob%3D%5C%22demo%2Fkamu-api-server%5C%22%7D%20%7C%3D%20%60%60%20%7C%20json%20%7C%20level%20%3D%20%60INFO%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22365dd1cc-3781-43c1-86a1-a3f78c6ab420%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221727283349335%22,%22to%22:%221727283355571%22%7D%7D,%22sS-%22:%7B%22datasource%22:%22b4feed3c-7158-4213-be67-14a716270591%22,%22queries%22:%5B%7B%22query%22:%223392a5b7a74a143b1cd8a8bcfabfb996%22,%22queryType%22:%22traceql%22,%22refId%22:%22A%22,%22limit%22:20,%22tableType%22:%22traces%22%7D%5D,%22range%22:%7B%22from%22:%221727283349335%22,%22to%22:%221727283355571%22%7D%7D%7D&orgId=1) show that for ~120 datasets hosted in S3, this initialization takes over 3s.

This can be explained with full scan of dataset repository in `DependencyGraphRepositoryInMemory`, doing:
 - full listing of dataset keys in S3 bucket (`ListObjectsV2`)
[...]

PR #973, opened 2024-11-27
854 persistent storage of dataset dependencies graph

zaychenko-sergei:
## Description

Closes: #854 

- [x] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/infra/datasets/postgres/src/repos/postgres_dataset_dependency_repository.rs
```diff
@@ -0,0 +1,208 @@
...
+#[async_trait::async_trait]
+impl DatasetDependencyRepository for PostgresDatasetDependencyRepository {
+    async fn stores_any_dependencies(&self) -> Result<bool, InternalError> {
+        let mut tr = self.transaction.lock().await;
+
+        let connection_mut = tr.connection_mut().await?;
+
+        let has_data = sqlx::query_scalar!(
+            r#"
+            SELECT EXISTS (SELECT * FROM dataset_dependencies LIMIT 1) as has_data
```
s373r, line 44: Q: out of curiosity, why not just `COUNT(*)`?

src/infra/datasets/repo-tests/src/dataset_dependencies_repository_test_suite.rs
```diff
@@ -0,0 +1,603 @@
...
+                downstream_dataset_id: entry_baz.id,
+                upstream_dataset_ids: vec![entry_bar.id],
+            },
+        ]),
+    );
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub async fn test_dependency_fanins(catalog: &Catalog) {
```
s373r, line 171:
```
test_dependency_fanins
                ^^^^^^
```
Q: out of curiosity, it's about https://en.wikipedia.org/wiki/Fan-in ?

src/domain/core/src/messages/core_message_types.rs
```diff
@@ -49,6 +50,10 @@ impl DatasetLifecycleMessage {
         })
     }
 
+    pub fn about_2_delete(dataset_id: DatasetID) -> Self {
```
s373r, line 53: Minor: `about_to_delete`?

s373r (commented):
5 cents from my side

src/infra/datasets/postgres/src/repos/postgres_dataset_dependency_repository.rs
zaychenko-sergei: I assume this query will at most look at 1 row, while `COUNT` is likely doing more work. Even if it's not loading all rows, it's still sort of a computation over table pages.

While we only need to know if table is empty...

src/domain/core/src/messages/core_message_types.rs
zaychenko-sergei: SMS slang from 90s.
to, two => 2
for => 4

src/infra/datasets/postgres/src/repos/postgres_dataset_dependency_repository.rs
s373r: Yeah, makes a lot of sense -- thank you

src/domain/core/src/messages/core_message_types.rs
s373r: Yes, I understand and read it easily (btw like leetspeak: 1337, etc. just look at my nickname :) )

I mean,
Q: do we need to use it in the naming? I don't see that kind of slang anywhere else in the code base

src/domain/core/src/messages/core_message_types.rs
```diff
@@ -23,8 +23,9 @@ const DATASET_LIFECYCLE_OUTBOX_VERSION: u32 = 1;
 pub enum DatasetLifecycleMessage {
     Created(DatasetLifecycleMessageCreated),
     DependenciesUpdated(DatasetLifecycleMessageDependenciesUpdated),
-    Deleted(DatasetLifecycleMessageDeleted),
     Renamed(DatasetLifecycleMessageRenamed),
+    About2Delete(DatasetLifecycleMessageAbout2Delete),
```
sergiimk, line 27: Why do we need to separate "about to delete" and "deleted" messages?

I thought that one benefit of combining `immediate` and `durable` delivery mechanisms in one messaging system is that things that need to happen inside of the same transaction can subscribe to the same event as those that need to happen outside of the transaction.

Here it almost feels like `About2Delete` event is intended ONLY to be delivered via `immediate` mode, which seems strange as delivery is currently a property of a consumer.

src/domain/datasets/domain/src/repos/dataset_dependency_repository.rs
```diff
@@ -0,0 +1,90 @@
...
+use internal_error::InternalError;
+use opendatafabric::DatasetID;
+use thiserror::Error;
+use tokio_stream::Stream;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait DatasetDependencyRepository: Send + Sync {
+    async fn stores_any_dependencies(&self) -> Result<bool, InternalError>;
```
sergiimk, line 21: How about replacing this with `edges_count()` as a more versatile method.

src/domain/datasets/domain/src/repos/dataset_dependency_repository.rs
```diff
@@ -0,0 +1,90 @@
...
+    async fn remove_all_dependencies_of(&self, dataset_id: &DatasetID)
+        -> Result<(), InternalError>;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub struct DatasetDependencies {
+    pub downstream_dataset_id: DatasetID,
+    pub upstream_dataset_ids: Vec<DatasetID>,
```
sergiimk, line 46: Why don't we stream just the ID pairs (as they are stored in the DB) and let the caller aggregate however they want? Seems like an unnecessary complication of a (what should be low-level) interface.

sergiimk (approved):
Left some suggestions and questions, but looks great overall :+1:

src/domain/core/src/messages/core_message_types.rs
zaychenko-sergei: Зануды!

src/domain/datasets/domain/src/repos/dataset_dependency_repository.rs
zaychenko-sergei: We only need to know if the repository is empty.
Is empty != Count edges from performance point of view.
If we will need to know the number of edges, we will add a separate API.

Somehow, standard containers have both `is_empty` and `len`.
Especially with more advanced structures like linked lists and trees this  this makes a lot more sense, right?
So, it's the same, I'm hoping to use cheaper queries for quick tests.

src/domain/datasets/domain/src/repos/dataset_dependency_repository.rs
zaychenko-sergei: We already had this modeling before. I don't see a motivation to reduce the level of abstraction.

src/domain/core/src/messages/core_message_types.rs
zaychenko-sergei: I've tried cascade deletions on foreign keys in new tables, and this worked better, so `AboutToDelete` is no longer needed.

src/domain/datasets/domain/src/repos/dataset_dependency_repository.rs
zaychenko-sergei: Inserts are looking quite nicely currently.

PR #988, opened 2024-12-11
Fix non-sequential offsets on ingest

sergiimk:
## Description

DataFusion was optimizing-out the `repartition(1)` call which led to parallel serialization of parquet files where offsets could be reordered across column batches.

The fix is to sort data before writing it instead of repartititoning - this make DF preserve the ordering.

I also added checks for physical plans to detect the next time when some optimization rules may affect our order guarantees.


zaychenko-sergei (approved):

src/e2e/app/cli/repo-tests/src/test_smart_transfer_protocol.rs
```diff
@@ -897,8 +897,8 @@ async fn test_smart_push_all_smart_pull_all(
             +--------+----+----------------------+----------------------+----------+-----------+-------+
             | offset | op | system_time          | match_time           | match_id | player_id | score |
             +--------+----+----------------------+----------------------+----------+-----------+-------+
-            | 2      | 0  | 2050-01-02T03:04:05Z | 2000-01-02T00:00:00Z | 2        | Charlie   | 90    |
-            | 3      | 0  | 2050-01-02T03:04:05Z | 2000-01-02T00:00:00Z | 2        | Alice     | 70    |
+            | 2      | 0  | 2050-01-02T03:04:05Z | 2000-01-02T00:00:00Z | 2        | Alice     | 70    |
+            | 3      | 0  | 2050-01-02T03:04:05Z | 2000-01-02T00:00:00Z | 2        | Charlie   | 90    |
```
sergiimk, line 901: This is interesting. Looks like this test was relying on wrong ordering bug, as this root dataset's PK is `(match_id, player_id)`

demusdev (approved):
Works for me 👍

Issue #353, opened 2023-12-13
Support bulk data export


sergiimk:
## Requirements
For cases when plain data is needed we should add ability to dump data in a plain format.

- Support multiple output formats:
  - all json variations (soa, aos, aoa, ndjson)
  - csv
  - parquet
- Support basic partitioning (n records per file)
- Allow export while querying via `kamu sql -c`

### Changes to `sql` command
New options:
- `--output-path (-o)` - path to the destination file or directory
- `--partition-size`
  - when set, data stored to several destination files, limiting number of records per file by specified amount
  - files should have ending with partition number
  - cause error when set without `--output-path` option

**Examples:**
Creates single file `/path/to/file.csv`:
```sh
kamu sql --command "select * from x" --output-format csv --output-path "/path/to/file.csv"
```

Creates multiple files `/path/to/data/..._01.csv, /path/to/data/..._02.csv, ...`:
```sh
kamu sql --command "select * from x" --output-format csv --output-path "/path/to/data" --partition-size 100
```

### New `kamu export` command 
Allows to export an entire dataset by name.
Works analogously to `kamu sql --command "select * from <dataset> order by offset`.
When `--output-path` is not set - it defaults to `$CWD/<dataset-name>`.

**Examples:**
Creates a single file: `/path/to/file.csv`
```sh
kamu export some.dataset --output-format csv --output-path "/path/to/file.csv"
```

Creates directory `some.dataset` in current directory and dumps data using some default (TBD) partitioning size:
```sh
kamu export some.dataset --output-format parquet
```

## Implementation notes
- We should maximally reuse the DataFusion's behavior of `DataFrame::write_*` function here - before implementing partitioning let's consider what partitioning values it provides and how it names files. It would be best if when saving parquet files all partitioning work was done by DataFusion and we didn't have to maintain custom code.
- DataFusion does not support all JSON variations that we support, so we may end up needing to manually partition files anyway. If DataFusion can support at least one JSON format - we can leave the rest as unimplemented and revisit them in future - le'ts keep the scope of this ticket small.

PR #989, opened 2024-12-12
Support bulk data export

demusdev:
## Description

Closes: #353 

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅ - NO EFFECT
    
  - [ ] Workspace layout and metadata: ✅  - NO EFFECT
    
  - [ ] Configuration: ✅  - NO EFFECT
    
  - [ ] Container images: ✅  - NO EFFECT
- [ ] Observability:
    
  - [ ] Tracing / [metrics](https://github.com/kamu-data/kamu-standards/blob/master/metrics_design.md): ✅  - NO NEED
    
  - [ ] Alerts: ✅  - NO NEED
- [ ] Documentation:
    
- [ ] Downstream effects:
    
  - [ ] [kamu-node](https://github.com/kamu-data/kamu-node): ✅  - NO EFFECT
    
  - [ ] [kamu-web-ui](https://github.com/kamu-data/kamu-web-ui): ✅  - NO EFFECT
    

src/app/cli/src/cli.rs
```diff
@@ -367,6 +368,27 @@ pub struct Delete {
...
+
+/// Export a dataset
+#[derive(Debug, clap::Args)]
+#[command(after_help = r#"
+This command exports a dataset to a file(s) of a given format.
+TODO: add extend doc
+"#)]
+pub struct Export {
+    #[arg(index = 1)]
+    pub dataset: String,
```
sergiimk, line 381: Better to use a more specific type

src/app/cli/src/cli.rs
```diff
@@ -367,6 +368,27 @@ pub struct Delete {
...
+
+    #[arg(long)]
+    pub output_path: String, //todo pathbuf
+
+    #[arg(long, value_parser = ["parquet", "json", "csv"])]
+    pub output_format: String, //todo value_parser = ["csv", ...]
+
+    #[arg(long)]
+    pub partition_size: Option<usize>,
```
sergiimk, line 390: Unclear if it's in bytes or records

src/app/cli/src/commands/export_command.rs
```diff
@@ -0,0 +1,70 @@
...
+    // async fn validate_args(&self) -> Result<(), CLIError> {
+    //     self.parse_format().map(|_| ())
+    // }
+
+    async fn run(&mut self) -> Result<(), CLIError> {
+        let format = self.parse_format()?;
+        let query = format!("select * from '{}'", self.dataset);
+        let rows_exported = self
+            .export_service
+            .export_to_fs(&query, &self.output_path, format, self.partition_size)
```
sergiimk, line 65: Should we support single-file export option too?

src/domain/core/src/services/export_service.rs
```diff
@@ -0,0 +1,36 @@
...
+use internal_error::InternalError;
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Service
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait ExportService: Send + Sync {
+    async fn export_to_fs(
+        &self,
+        sql_query: &String,
```
sergiimk, line 21: Would it be better to decouple ExportService and QueryService by simply accepting a `DataFrame` here instead of SQL?

src/infra/core/src/query_service_impl.rs
```diff
@@ -73,6 +75,18 @@ impl QueryServiceImpl {
...
 
+        if let Some(overrides) = session_config_overrides {
+            if let Some(val) = overrides.target_partitions {
+                cfg.options_mut().execution.target_partitions = val;
+            }
+            if let Some(val) = overrides.minimum_parallel_output_files {
+                cfg.options_mut().execution.minimum_parallel_output_files = val;
+            }
+            if let Some(val) = overrides.soft_max_rows_per_output_file {
+                cfg.options_mut().execution.soft_max_rows_per_output_file = val;
```
sergiimk, line 86: Is there no way to override this in `write_parquet`? 

I would like to avoid tying `QueryService` interface so closely to DataFusion internals by surfacing such in-depth config parameters.

demusdev: I didn't find options.
This is where it's being used: https://github.com/apache/datafusion/blob/2464703c84c400a09cc59277018813f0e797bb4e/datafusion/core/src/datasource/file_format/write/demux.rs#L150-L150
And TaskContext (SessionContext.state) comes from `write_parquet`, where new dataframe is being created: https://github.com/apache/datafusion/blob/2464703c84c400a09cc59277018813f0e797bb4e/datafusion/core/src/dataframe/parquet.rs#L85-L85
And latter comes from context we create.

If you want to have these services completely independent, we can just have session creation code (mostly)dup in new service, or move `create_session` and `create_session_with_config_overrides` to a third service and have it as a dependency for QueryService and ExportService.

src/domain/core/src/services/export_service.rs
demusdev: we need to create specific session context with single target partition and specific rows count per target file.

src/app/cli/src/commands/export_command.rs
demusdev: yes. it's very early draft PR which I created just for visibility. I plan to add description and showcase tests

src/app/cli/src/cli.rs
demusdev: I plan to add params documentation later in this PR

src/domain/core/src/services/export_service.rs
demusdev: Makes sense! I will change it.

src/infra/core/src/query_service_impl.rs
demusdev: My prev statement is not quite true. I figured out how to reassemble DataFrame with hacked session config.

src/app/cli/src/cli.rs
zaychenko-sergei: It's not about the docs, it's about the name

demusdev: but name was take from ticket description: https://github.com/kamu-data/kamu-cli/issues/353
should we change it to something like `--rows-per-file`?

zaychenko-sergei: Yes, the switch is unclear similarly. Maybe like this or maybe `--records-per-file`.

src/app/cli/src/commands/export_command.rs
```diff
@@ -0,0 +1,99 @@
...
+use crate::{CLIError, Command};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct ExportCommand {
+    export_service: Arc<dyn ExportService>,
+    query_service: Arc<dyn QueryService>,
+    dataset_ref: DatasetRef,
+    output_path: Option<PathBuf>,
+    output_format: String,
```
zaychenko-sergei, line 25: Don't we have enums like the following in other CLI commands:

```
#[derive(Debug, Clone, Copy, clap::ValueEnum)]
pub enum OutputFormat {
    /// Comma-separated values
    #[clap(name = "csv")]
    Csv,

    /// Array of Structures format
    #[clap(name = "json")]
    Json,

    /// One Json object per line - easily splittable format
    #[clap(name = "ndjson")]
    NdJson,

    /// Structure of arrays - more compact and efficient format for encoding
    /// entire dataframe
    #[clap(name = "json-soa")]
    JsonSoA,

    /// Array of arrays - compact and efficient and preserves column order
    #[clap(name = "json-aoa")]
    JsonAoA,

    /// A pretty human-readable table
    #[clap(name = "table")]
    Table,
}
```

zaychenko-sergei: I don't mean using this particular one, but perhaps your own `ExportFormat` listing valid options.

src/app/cli/src/commands/export_command.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        let df = self
+            .query_service
+            .get_data(&self.dataset_ref)
+            .await
+            .map_err(CLIError::failure)?;
+
+        let mut default_path: PathBuf = PathBuf::new();
+        default_path.push(self.dataset_ref.to_string());
+        default_path.push(""); // ensure trailing slash to have it as a dir
```
zaychenko-sergei, line 77: Maybe something like this:

```   
   let path = env::current_dir()?;
   path.join(self.dataset_ref.to_string())
```   
also, should we add some file extension depending on the format?

src/app/cli/src/commands/sql_shell_command.rs
```diff
@@ -42,9 +44,13 @@ pub struct SqlShellCommand {
     engine_prov_config: Arc<EngineProvisionerLocalConfig>,
     output_config: Arc<OutputConfig>,
     container_runtime: Arc<ContainerRuntime>,
+    export_service: Arc<dyn ExportService>,
     command: Option<String>,
     url: Option<String>,
     engine: Option<SqlShellEngine>,
+    output_path: Option<PathBuf>,
+    partition_size: Option<usize>,
+    supported_export_formats: Vec<OutputFormat>,
```
zaychenko-sergei, line 53: I guess you can declare it as `const` in the module. Does not feel like it's a state.

src/app/cli/src/output/output_config.rs
```diff
@@ -98,6 +102,25 @@ pub enum OutputFormat {
...
+impl Display for OutputFormat {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let str_val = match *self {
+            OutputFormat::Csv => "csv",
+            OutputFormat::Json => "json",
+            OutputFormat::NdJson => "ndjson",
+            OutputFormat::JsonSoA => "json-soa",
+            OutputFormat::JsonAoA => "json-aoa",
+            OutputFormat::Table => "table",
+            OutputFormat::Parquet => "parquet",
```
zaychenko-sergei, line 120: And why custom display was necessary? Can't we live with the automatic version?

src/domain/core/src/services/errors.rs
```diff
@@ -0,0 +1,24 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/// Wraps [`datafusion::error::DataFusionError`] error to attach a backtrace at
+/// the earliest point
+#[derive(Error, Debug)]
+#[error("DataFusion error")]
+pub struct DataFusionError {
+    #[from]
+    pub source: datafusion::error::DataFusionError,
```
zaychenko-sergei, line 22: It feels like a boundary between domains.. 
Do we get any benefit from depending on a particular error type here?

See how it's done for `InternalError`:

```
pub type BoxedError = Box<dyn std::error::Error + Send + Sync>;

#[derive(Error, Debug)]
#[error("Internal error")]
pub struct InternalError {
    #[source]
    source: BoxedError,
    context: Option<String>,
    #[backtrace]
    backtrace: Option<Backtrace>,
}
```

src/domain/core/src/services/export_service.rs
```diff
@@ -0,0 +1,98 @@
...
+impl Display for ExportFormat {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let str_val = match *self {
+            ExportFormat::Parquet => "parquet",
+            ExportFormat::Csv => "csv",
+            ExportFormat::NdJson => "ndjson",
+        };
+        f.write_str(str_val)
+    }
+}
```
zaychenko-sergei, line 51: There is `strum` crate we use in many places, maybe it can help to avoid this boilerplate?

src/domain/core/src/services/export_service.rs
```diff
@@ -0,0 +1,98 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Errors
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Error)]
+#[error("{context}")]
+pub struct NotImplementedError {
+    pub context: String,
+}
```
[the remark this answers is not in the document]
zaychenko-sergei: There is `unimplemented!` macro that is a specific kind of panic for this situation.

src/app/cli/src/commands/export_command.rs
zaychenko-sergei: As I understand, the extensions are automatically attached by the `DataFusion` writers. Did you test what happens if I specify particular extensions in the path? Wrong extensions for the selected format?

src/app/cli/src/output/output_config.rs
zaychenko-sergei: `strum`?

src/domain/core/src/services/errors.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub backtrace: Backtrace,
+}
```
zaychenko-sergei, line 24: Why not to keep it where it was? It's accessible in the module namespace.

src/domain/core/src/services/export_service.rs
```diff
@@ -0,0 +1,98 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Service
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait ExportService: Send + Sync {
+    async fn export_to_fs(
+        &self,
+        df: DataFrame,
+        path: &Path,
```
zaychenko-sergei, line 28: What confuses a bit is that is not simpliy a path, rather a prefix of the path.
Maybe some comments?

zaychenko-sergei, line 28: If we compare it to just extending `QueryService` with the `export_to_fs` method, is there any benefit in separate interface? It seems they travel together.

src/app/cli/src/commands/export_command.rs
demusdev: it accepts any extension you specify. you simply need to have any. otherwise it will assume path as a dir

demusdev: here's related datafusion writer https://github.com/apache/datafusion/blob/1fc776981496c0f0648ece8069992c65a7021c83/datafusion/core/src/datasource/file_format/write/demux.rs#L67

src/domain/core/src/services/export_service.rs
demusdev: it's a path. either path to file or to directory. if it's a file path (something with extension and without trailing path delimiter), DataFusion will store data into it. Otherwise, it treats path as a dir and will create files (with generated names, and extension related to export format) inside.

src/domain/core/src/services/export_service.rs
zaychenko-sergei: @sergiimk Can you explain how `DataFrame` object will help. Isn't this `DataFusion`'s specific trait?
I understand Spark and Flink have a similar concept, but I guess it's not the same interface?
Or do we plan to wrap muiltiple types with our own abstraction?

src/app/cli/src/output/output_config.rs
demusdev: We use `fmt` when printing error message, where we need to display supported formats. And we need custom formatter, to support all custom mappings like "NdJson" -> "ndjson", "JsonSoA" -> "json-soa"

src/domain/core/src/services/export_service.rs
demusdev: I think logically it's a different things, `QueryService` is for preparing/reading data, and `ExportService` is for writing. When I started working on code, I was expecting something much more complex than just a few lines of code so created a separate service.
But with the current code state, having it separated or not is more matter of taste. I just left as it was.

zaychenko-sergei (approved):

zaychenko-sergei:
@demusdev wait for @sergiimk 's approval.

src/domain/core/src/services/export_service.rs
sergiimk: `DataFrame` can wrap Arrow buffers and our long-term goal is to make Arrow the only data interchange format in Kamu. 

There is a high chance Spark and other engines will soon natively support Arrow format. If they don't we'll imlement adapters. But the idea is that `DataFrame` will be able to represent output data from any engine in ODF, not only DataFusion.

src/domain/core/src/services/export_service.rs
```diff
@@ -0,0 +1,80 @@
...
+        format: &ExportFormat,
+        records_per_file: Option<usize>,
```
sergiimk, line 29: Let's add `ExportOpts` struct and move this parameter there. This way users will be able to pass `ExportOpts::default()` where they don't care about the details, and we will be able to add more options without breaking every place where this method is called.

This is a pattern a lot of our services follow.

src/domain/core/src/services/export_service.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Debug, Error)]
+pub enum ExportError {
+    #[error(transparent)]
+    Internal(
+        #[from]
+        #[backtrace]
+        InternalError,
+    ),
+    #[error(transparent)]
+    DataFusionError(DataFusionError),
```
sergiimk, line 68: Aren't all DF errors still internal errors? Why do we need to differentiate them?

sergiimk (approved):
Left a few small suggestions but looks good overall :ship:

src/app/cli/src/cli.rs
```diff
@@ -367,6 +369,40 @@ pub struct Delete {
...
+
+    /// Export destination. Dafault is `<current workdir>/<dataset name>`
+    #[arg(long)]
+    pub output_path: Option<PathBuf>,
+
+    /// Output format
+    #[arg(long, value_parser = parsers::export_format)]
+    pub output_format: ExportFormat,
+
+    /// Number of records per file, if stored into a directory. Default is 5m.
```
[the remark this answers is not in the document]
demusdev: I think it makes sense to let the engine to choose default, so I voting for the first option

resources/cli-reference.md
```diff
@@ -254,6 +255,35 @@ Delete local datasets matching pattern:
...
+
+**Arguments:**
+
+* `<DATASET>` — Local dataset reference
+
+**Options:**
+
+* `--output-path <OUTPUT_PATH>` — Export destination. Dafault is `<current workdir>/<dataset name>`
+* `--output-format <OUTPUT_FORMAT>` — Output format
+* `--records-per-file <RECORDS_PER_FILE>` — Number of records per file, if stored into a directory. Default is 5m. It's a soft limit. For a sake of export performance real number of records may be slightly different
```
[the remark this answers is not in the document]
demusdev: done

src/domain/core/src/services/export_service.rs
demusdev: `ExportError::Internal` is thrown when we failed to calculate exported rows.
For all other cases we throw `DataFusionError`. I took QueryService as an example, so new command and changes to `sql` one conforms existent interface. 
With this approach we sometimes may show user not useful details. But if we want to improve this, we need to be careful with filtering DataFusion errors in order to provide users with details when their sql is invalid, or they cannot store data to a specific location. I think we need to make changes to both Query and External service at the same time. And if we'll go with this direction, I guess it makes sense to implement it in scope of separate ticket.

demusdev: For the context, here're some of DataFusion errors exposed to users.
Triggered from QueryService:
```bash
❯ kamu sql --command "select random_field_name from 'test.data2'" --output-format parquet --output-path ./out/parquet/1
Error:
  0: DataFusion error
  1: Schema error
  2: No field named random_field_name. Valid fields are "test.data2".offset, "test.data2".op, "test.data2".system_time, "test.data2".event_time, "test.data2".uuid1.
  
❯ kamu sql --command "select * from 'test.data2' qwe 212" --output-format parquet --output-path /nix/1.parquet
Error:
  0: DataFusion error
  1: SQL error: ParserError("Expected end of statement, found: 212")
  2: sql parser error: Expected end of statement, found: 212

❯ kamu sql --command "select * from 'does.not.exists'" --output-format parquet --output-path ./out/parquet/1
Error:
  0: DataFusion error
  1: Error during planning: table 'kamu.kamu.does.not.exists' not found
```
Triggered from ExportService:
```bash❯ kamu sql --command "select 1/0" --output-format parquet --output-path ./out/parquet/1
Error:
  0: DataFusion error
  1: Arrow error
  2: Divide by zero error
  
❯ kamu sql --command "select * from 'test.data2'" --output-format parquet --output-path /out/parquet/1
Error:
  0: DataFusion error
  1: IO error
  2: Generic LocalFileSystem error
  3: Unable to create dir /out/parquet/1
  4: Read-only file system (os error 30)
  
❯ kamu sql --command "select * from 'test.data2'" --output-format parquet --output-path /nix/1.parquet
Error:
  0: DataFusion error
  1: IO error
  2: Generic LocalFileSystem error
  3: Unable to open file /nix/1.parquet#1
  4: Permission denied (os error 13)  
  
❯ kamu sql --command "select * from 'test.data2' limit -1" --output-format parquet --output-path /nix/1.parquet
Error:
  0: DataFusion error
  1: Optimizer rule 'eliminate_limit' failed
caused by
Error during planning: LIMIT must be >= 0, '-1' was provided
  2: Error during planning: LIMIT must be >= 0, '-1' was provided
```

sergiimk: All errors you provided as examples above look like planning-phase errors to me, so If I'm not mistaken they should all be raised from the `QueryServiceImpl` while we are setting up the logical plan, not from `ExportServiceImpl`.

Here it makes sense for `QueryService` to return `DataFusionError` instead of `InternalError` because very likely all errors will be caused by the invalid user input.

This logical plan in form of a `DataFrame` is then passed into the `ExportService` where we add a few extra steps:
- repartition + file sync - a planning error on at phase will indicate we badly messed up the code
- finally we execute the plan - an error at this phase will likely be IO-related (out of space, permissions etc).

In both cases there isn't much what user can do about them. And the latter execution-phase errors when used in `kamu-node` context can reveal details of internal infrastructure (e.g. some S3 bucket names or object keys).

demusdev: Some errors like division by zero and invalid limit value are spotted on execution phase (in ExportService). I believe there are a few more useful ones I didn't spot. And also IO errors in example above looks useful as well. I'd preserve them either as DataFusion errors, or create specific ones, and would map corresponding DataFusion errors to that new ones.

Issue #984, opened 2024-12-10
Refactoring: Separate planning and execution phases in key dataset manipulation services


zaychenko-sergei:
During previous refactoring when introducing `DatasetRegistry` (#941) we've introduced explicit planning and execution phases in `ingest`, `sync` and `transform` operations, so that the registry access is concentrated in planning phase only that is being run under a database transaction. This allowed to run the remaining executions without a transaction at all, which is critical for Task Execution context.

Since now we are moving towards storing more data, such as HEAD refs, key blocks, summaries - in the database, there is a need to push more processing into similar plan/execute pattern.

In particular:
- compacting dataset (plan how the blocks will be merged an collect data file URLs, but merge and commit separately)
- resetting dataset (very simple plan, confirm the arguments or apply defaults)
- operations, which use `DataWriterDataFusion`, in particular:
    - polling ingest
    - push ingest
    - set watermark 

A few extra notes:
- `DataWriterDataFusion` already has a natural data structure representing a plan - `DataWriterMetadataState`, it can be returned as a result of planning phase, and consumed at the execution phase
- Polling ingest is already represented in planning hierarchy, we create `PullIngestItem`. It could be extended with `DataWriterMetadataState`.
- Setting a watermark or Push ingest are not yet represented anyhow in Flows/Tasks systems, however it's highly likely that at least the Push ingest will, so we better be prepared and have a similar architecture as polling ingest, compacting, and reset.

In case of CLI, use cases would manage invoking both phases.
At the level of tasks, the executor is already split on planning and execution phases, and they simply need to be extended to support these changes.

PR #994, opened 2024-12-17
984 refactoring separate planning and execution phases in key dataset manipulation services

zaychenko-sergei:
## Description

Closes: #984 

Extracted "planner" and "executor" for compacting, reset, set watermark, push ingest, partially polling ingest.
Renamed long-running "executors" to "agents".
Introduced `MetadataQueryService` to absorb simple queries that do not have to be defined at the level of metadata chian from the interface point of view.

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
  - [ ] Alerts: ✅
- [ ] Documentation:
  - [ ] [Changelog](./CHANGELOG.md): ✅ - TODO
    
  - [ ] [Public documentation](https://github.com/kamu-data/kamu-docs/): ✅ - NO EFFECT
- [ ] Downstream effects:
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): ✅ - TODO
  - [ ] [kamu-web-ui](https://github.com/kamu-data/kamu-web-ui): ✅ - NO EFFECT
  - [x] [release-queue](https://github.com/orgs/kamu-data/projects/4/views/1) - TODO


src/infra/core/src/services/metadata_query_service_impl.rs
```diff
@@ -0,0 +1,98 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[component(pub)]
+#[interface(dyn MetadataQueryService)]
+pub struct MetadataQueryServiceImpl {}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+impl MetadataQueryService for MetadataQueryServiceImpl {
```
sergiimk, line 25: Would these methods be better suited in `MetadataChainExt` extension trait?

To me a service encapsulates complex interactions between several components, but here we have nothing but a stateless wrapper on top of `MetadataChain`'s visitor interface.

`MetadataChain` is a place where we also validate complex domain rules about what sequences of events are valid an which aren't (e.g. two active push sources with same name) so I don't think an argument that this is some "higher-level logic" works in this case either.

zaychenko-sergei: I'm not sure. I have a feeling that for high level queries to dataset's current state, we shouldn't be considering that there is some metadata chain behind it. We just want to know key characteristics of the dataset's current state. The fact that this maps to `MetadataChain` algorithm is an implementation detail. Besides, these questions could be answered with database in future.

zaychenko-sergei: `MetadataChainExt` deals with a much lower level of abstraction. This is mostly about visiting the blocks one way or another,  emphasizing explicit knowledge of chain structure.

src/infra/ingest-datafusion/src/writer.rs
```diff
@@ -857,16 +841,16 @@ impl DataWriter for DataWriterDataFusion {
...
-    dataset: Arc<dyn Dataset>,
+    target: ResolvedDataset,
     ctx: SessionContext,
     block_ref: BlockRef,
     metadata_state: Option<DataWriterMetadataState>,
 }
 
 impl DataWriterDataFusionBuilder {
-    pub fn new(dataset: Arc<dyn Dataset>, ctx: SessionContext) -> Self {
+    pub fn new(target: ResolvedDataset, ctx: SessionContext) -> Self {
```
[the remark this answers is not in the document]
zaychenko-sergei: Yes, simplified!

Issue #891, opened 2024-10-10
GQL, `DatasetsMut::create_empty()` & `DatasetsMut::create_from_snapshot()`: return correct alias in multi-tenant environments 


s373r:
### Describe the bug

When calling methods, we can't get the actual alias

### Steps To Reproduce

1. Init a workspace
```shell
kamu init --multi-tenant
```

2. Start API server
```shell
KAMU_AUTH_GITHUB_CLIENT_ID=1 \
  KAMU_AUTH_GITHUB_CLIENT_SECRET=2 \
  kamu system api-server \
  --address 0.0.0.0 \
  --http-port 8080 \
  --get-token
```

3. Open http://0.0.0.0:8080
4. Using the received token from step 2. create an empty dataset
```graphql
mutation {
  datasets {
    createEmpty(datasetKind: ROOT, datasetAlias: "empty-root-dataset") {
      message
      ... on CreateDatasetResultSuccess {
        dataset {
          alias
          owner {
            accountName
          }
        }
      }
    }
  }
}

```

### Current Behavior

```diff
 {
   "data": {
     "datasets": {
       "createEmpty": {
         "message": "Success",
         "dataset": {
-          "alias": "empty-root-dataset",
           "owner": {
             "accountName": "s373r"
           }
         }
       }
     }
   },
   "extensions": {}
 }
```

### Expected Behavior

```diff
 {
   "data": {
     "datasets": {
       "createEmpty": {
         "message": "Success",
         "dataset": {
+          "alias": "s373r/empty-root-dataset",
           "owner": {
             "accountName": "s373r"
           }
         }
       }
     }
   },
   "extensions": {}
 }
```

### kamu system info

```sh
build:
  appVersion: 0.204.5
...
```

### Anything else?

We need to read the current user at creation, and use the correct alias based on that.

Also, the `KamuApiServerClientExt` used in E2E tests needs to be updated.

PR #1008, opened 2024-12-23
Prefix dataset alias with account name

demusdev:
## Description

Closes: #891

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [ ] Workspace layout and metadata: ✅ - NO EFFECT
    
  - [ ] Configuration: ✅ - NO EFFECT
    
  - [ ] Container images: ✅ - NO EFFECT
- [x] Observability:
    
     - NO NEED
  - [ ] Alerts: ✅ - NO NEED
- [x] Documentation:
    
  - [ ] [Public documentation](https://github.com/kamu-data/kamu-docs/): ✅ - NO NEED
  
- [ ] Downstream effects:
    

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Recommendation: for ease of reading, use the following order:
 - Fixed
 -->
 
+## [Unreleased]
+### Fixed
+- GQL: datasets/createEmpty returns alias prefixed with account name
```
s373r, line 16: Comment: the changes do not only apply to this method

s373r, line 16: Based on the description, it seems that there will be an account alias (`kamu/ds` not `ds`) for ST as well, is that right?

If so, it's unexpected behavior

src/e2e/app/cli/repo-tests/src/rest_api/test_gql_query.rs
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+        "#,
+         ))
+        .await;
+    let token = login_response["auth"]["login"]["accessToken"]
+        .as_str()
+        .map(ToOwned::to_owned)
+        .unwrap();
+    kamu_api_server_client.set_token(Some(token));
```
s373r, line 87: Can be replaced by `kamu_api_server_client.auth().login_as_kamu().await;`

src/e2e/app/cli/repo-tests/src/rest_api/test_gql_query.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use kamu_cli_e2e_common::{KamuApiServerClient, KamuApiServerClientExt};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub async fn test_gql_query_mut_create_empty_returns_correct_alias_mt(
+    mut kamu_api_server_client: KamuApiServerClient,
+) {
+    let token = kamu_api_server_client.auth().login_as_e2e_user().await;
+    kamu_api_server_client.set_token(Some(token));
```
s373r, line 18: `login_as_e2e_user()` method escapes the token installation logic, so the `token` does not need to be set explicitly

src/e2e/app/cli/repo-tests/src/rest_api/test_gql_query.rs
```diff
@@ -0,0 +1,131 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use kamu_cli_e2e_common::{KamuApiServerClient, KamuApiServerClientExt};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub async fn test_gql_query_mut_create_empty_returns_correct_alias_mt(
```
s373r, line 14: We also need to add a test for another related GQL method as well

src/infra/core/src/repos/dataset_repository_local_fs.rs
```diff
@@ -904,12 +904,12 @@ impl DatasetStorageStrategy for DatasetMultiTenantStorageStrategy {
...
         raw_alias: &DatasetAlias,
     ) -> Result<DatasetAlias, ResolveDatasetError> {
-        Ok(if let Some(account_name) = &raw_alias.account_name {
-            let (_, canonical_account_name) = self.resolve_account_dir(account_name).int_err()?;
-            DatasetAlias::new(Some(canonical_account_name), raw_alias.dataset_name.clone())
-        } else {
-            raw_alias.clone()
-        })
+        let account_name = self.effective_account_name(raw_alias);
+        let (_, canonical_account_name) = self.resolve_account_dir(account_name).int_err()?;
```
s373r, line 908: Q: if we get the `account_name` right after `effective_account_name()` call, then why do we need to call `resolve_account_dir()`?

s373r (changes requested):
Added comments and questions

CHANGELOG.md
demusdev: in single tenant workspace (also before this PR changes), `datasets.byOwnerAndName` returns alias with `kamu` prefix, but `datasets.byId` does not. I guess we should return it only for multi-tenant workspace, otherwise, client couldn't find the type of a workspace.
1. Should we remove prefix from `datasets.byOwnerAndName` response?
2. Should this endpoint be allowed for single tenant workspace at all?
3. Should we remove `owner` part from responses in case of singe-tenant workspace?

You may check calls here: https://gist.github.com/demusdev/768e3acb2184d5fa031ed9d24e5593c0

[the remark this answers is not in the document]
demusdev: Didn't know there were anchors with issue link. Thanks for the hint!
I found the only place in tests: https://github.com/kamu-data/kamu-cli/blob/62ac6732c094ddbc8c2945ece3bfc01be3453d8d/src/e2e/app/cli/common/src/kamu_api_server_client_ext.rs#L496
Is it correct assumption that alias format is `account/dataset-name`, where `account/` prefix is optional, and present only in case of multi-tenant workspace? Otherwise, I don't see options to construct correct `DatasetAlias`.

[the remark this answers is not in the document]
demusdev: There's also issue mentioning in main code: https://github.com/kamu-data/kamu-cli/blob/62ac6732c094ddbc8c2945ece3bfc01be3453d8d/src/adapter/graphql/src/mutations/datasets_mut.rs#L112

Should we do anything more, or both TODO's can be removed from that place?

src/infra/core/src/repos/dataset_repository_local_fs.rs
demusdev: I guessed it covers some corner cases with checking path/updating alias prefix, that's why I left the call.
I believe @zaychenko-sergei can say more about that function.

src/e2e/app/cli/repo-tests/src/rest_api/test_gql_query.rs
demusdev: done

src/e2e/app/cli/repo-tests/src/rest_api/test_gql_query.rs
demusdev: done

CHANGELOG.md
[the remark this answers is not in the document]
demusdev: tests were updated

src/infra/core/src/repos/dataset_repository_local_fs.rs
zaychenko-sergei: I don't really understand your change here, plz explain what you are trying to solve.

src/e2e/app/cli/common/src/kamu_api_server_client_ext.rs
```diff
@@ -376,6 +377,17 @@ pub struct DatasetApi<'a> {
...
+    fn get_alias(alias_str: &str) -> DatasetAlias {
+        if let Some((account, dataset)) = alias_str.split_once('/') {
+            DatasetAlias::new(
+                Some(AccountName::new_unchecked(account)),
+                DatasetName::new_unchecked(dataset),
+            )
+        } else {
+            DatasetAlias::new(None, DatasetName::new_unchecked(alias_str))
+        }
+    }
```
sergiimk, line 389: `DatasetAlias` implements `std::str::FromStr` so you should be able to do:
```rust
let alias: DatasetAlias = "foo/bar".parse()?;
```

src/infra/core/src/repos/dataset_repository_local_fs.rs
demusdev: the idea of the change was that there account should be always present in dataset alias in the multi-tenant workspace.

zaychenko-sergei (approved):

s373r (approved):

Issue #850, opened 2024-09-25
Implement batch loading of event sourcing aggregates (Flows, Tasks)


zaychenko-sergei:
Currently we do not support batch loading of flows & tasks, and load them in sequential loops like below:

```
            // TODO: implement batch loading
            for flow_id in relevant_flow_ids {
                let flow = Flow::load(flow_id, self.flow_event_store.as_ref()).await.int_err()?;
                yield flow.into();
            }
```

Knowing a set of queries (id's in our case), it should be possible to load all related events in 1 SQL query in chronological order.
Then `event_sourcing` crate could post-process the loaded collection of events, and re-construct multiple aggregates in 1 call.

This should optimize database roundtrips, esp. for listing flows:

![image](https://github.com/user-attachments/assets/9eac9938-4aba-4008-9084-80093527ac9a)

s373r:
Cool, tracing is the best visual way to know that optimization is needed. 

Yes, we knew this before, but when you can see this need on the graph, superb!

zaychenko-sergei:
@demusdev Consult with @rmn-boiko for detailed supervision

PR #1024, opened 2025-01-04
Implement batch loading of event sourcing aggregates (Flows, Tasks)

demusdev:
## Description

Closes: #850

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [ ] Workspace layout and metadata: ✅ - NO EFFECT
    
  - [ ] Configuration: ✅ - NO EFFECT
    
  - [ ] Container images: ✅ - NO EFFECT
- [ ] Observability:
    
  - [ ] Tracing / [metrics](https://github.com/kamu-data/kamu-standards/blob/master/metrics_design.md): ✅ - NO NEED
    
  - [ ] Alerts: ✅ - NO NEED
- [ ] Documentation:
    
  - [ ] [Public documentation](https://github.com/kamu-data/kamu-docs/): ✅ - NO NEED
  
- [ ] Downstream effects:
    
### DIFF
--- a/src/domain/flow-system/services/src/flow/flow_query_service_impl.rs
+++ b/src/domain/flow-system/services/src/flow/flow_query_service_impl.rs
@@ -13,7 +13,7 @@ use std::sync::Arc;
 use chrono::{DateTime, Utc};
 use database_common::PaginationOpts;
 use dill::{component, interface, Catalog};
-use futures::{StreamExt, TryStreamExt};
+use futures::TryStreamExt;
 use internal_error::ResultIntoInternal;
 use kamu_core::DatasetOwnershipService;
 use kamu_flow_system::*;
@@ -47,16 +47,17 @@ impl FlowQueryServiceImpl {
         }
     }
 
-    fn flow_state_stream<'a>(&'a self, flow_ids: FlowIDStream<'a>) -> FlowStateStream<'a> {
+    fn flow_state_stream(&self, flow_ids: Vec<FlowID>) -> FlowStateStream {
         Box::pin(async_stream::try_stream! {
             // 32-items batching will give a performance boost,
             // but queries for long-lived datasets should not bee too heavy.
             // This number was chosen without any performance measurements. Subject of change.
-            let batch_size = 32;
-            let mut chunks = flow_ids.try_chunks(batch_size);
-            while let Some(item) = chunks.next().await {
-                let ids = item.int_err()?;
-                let flows = Flow::load_multi(ids, self.flow_event_store.as_ref()).await.int_err()?;
+            let chunk_size = 32;
+            for chunk in flow_ids.chunks(chunk_size) {
+                let flows = Flow::load_multi(
+                    chunk.to_vec(),
+                    self.flow_event_store.as_ref()
+                ).await.int_err()?;
                 for flow in flows {
                     yield flow.int_err()?.into();
                 }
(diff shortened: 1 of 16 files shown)


src/domain/flow-system/services/src/flow/flow_query_service_impl.rs
```diff
@@ -46,6 +46,18 @@ impl FlowQueryServiceImpl {
             agent_config,
         }
     }
+    fn flow_state_stream<'a>(&'a self, flow_ids: FlowIDStream<'a>) -> FlowStateStream<'a> {
```
zaychenko-sergei, line 49: Formatting: missing blank line

src/utils/event-sourcing/src/event_store.rs
```diff
@@ -20,6 +20,26 @@ pub trait EventStore<Proj: Projection>: Send + Sync {
...
+        Box::pin(async_stream::try_stream! {
+          for query in queries {
+            let mut stream = self.get_events(&query, GetEventsOpts::default());
+            while let Some(event) = stream.next().await {
+              let (event_id, event) = event?;
+              yield (query.clone(), event_id, event)
+            }
+          }
+        })
+    }
```
zaychenko-sergei, line 41: Why is this method not using multi-load from event store?

demusdev: the idea was to have custom implementation for postges-based event stores, but use default one for inmem and sqlite (where performance is not critical)
would you like to have multi-load based implementation for non-prod cases as well?

zaychenko-sergei (approved):

src/utils/event-sourcing/src/aggregate.rs
```diff
@@ -116,6 +117,50 @@ where
...
+
+        let mut event_stream = event_store.get_events_multi(queries.clone());
+        let mut agg_results: HashMap<Proj::Query, Result<Self, LoadError<Proj>>> = HashMap::new();
+
+        while let Some(res) = event_stream.next().await {
+            // When failed to read at least one event from source stream,
+            // function returns error result immediately
+            let (query, event_id, event) = res?;
+            let agg_result: Result<Aggregate<Proj, Store>, LoadError<Proj>> = match agg_results
+                .remove(&query)
```
sergiimk, line 141: Hm, why do we need to remove and re-insert the aggregate on every event? Can't we get a mutable reference and update it in place?

src/domain/flow-system/services/src/flow/flow_state_helper.rs
```diff
@@ -0,0 +1,45 @@
...
+use kamu_flow_system::{Flow, FlowEventStore, FlowID, FlowStateStream};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct FlowStateHelper {
+    flow_event_store: Arc<dyn FlowEventStore>,
+}
+
+#[component(pub)]
+impl FlowStateHelper {
```
sergiimk, line 23: What's the reason to have this in a separate component? This looks like it could've easily been a part of `FlowEventStore`. I don't see us ever needing to stub or create a different implementation of this component, so seems like it complicates DI configuration for no apparent benefit?

src/utils/event-sourcing/src/aggregate.rs
demusdev: Makes sense. I just missed this. I think it should be like the following. Do you want me to create chore PR with the change?
```Rust
if let Some(agg_result) = agg_results.get_mut(&query) {
    if let Ok(agg) = agg_result {
        if let Err(err) = agg.apply_stored(event_id, event) {
            *agg_result = Err(err.into());
        }
    }
} else {
    let agg_result =
        Self::from_stored_event(query.clone(), event_id, event).map_err(Into::into);
    agg_results.insert(query, agg_result);
}
```

src/domain/flow-system/services/src/flow/flow_state_helper.rs
demusdev: Agree. Should I move it in a chore PR together with another change?

src/utils/event-sourcing/src/aggregate.rs
demusdev: changing in scope of #1031

src/domain/flow-system/services/src/flow/flow_state_helper.rs
demusdev: changing in scope of #1031

Issue #1040, opened 2025-01-20
Prototype email gateway


zaychenko-sergei:
Main topic:

https://github.com/kamu-data/kamu-node/issues/55

Scope of this ticket:
- Emails are fire-and-forget / best effort
- Select an email gateway service
- Create email gateway service with implementation for the chosen tech
- Create a set of event handlers that listens to event bus events and sends emais

PR #1048, opened 2025-01-24
1040 prototype email gateway

zaychenko-sergei:
## Description

Closes: #1040 

Integration of email gateway (Postmark):
  - Defined `EmailSender` crate and prototyped `Postmark`-based implementation
  - CLI config support for gateway settings (API key, sender address & name)
  - Applied `askoma` templating engine and defined base HTML-rich template for all emails
  - Emails are fire-and-forget / best effort
  - First email implemented: account registration

Added GQL suport to query and update email on the currently logged account.

Emails are mandatory for Kamu accounts now:
 - predefined users need to specify an email in config
 - predefined users are auto-synced at startup in case they existed before
 - GitHub users are queried for primary verified email, even if it is not public
 - migration code for the database existing users

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [ ] Alerts: ✅
- [x] Documentation:
    
  - [x] [Public documentation](https://github.com/kamu-data/kamu-docs/): - NO EFFECT ✅
  
- [ ] Downstream effects:
    
  - [x] [kamu-node](https://github.com/kamu-data/kamu-node): - UPDATES REQUIRED ✅
    
  - [x] [kamu-web-ui](https://github.com/kamu-data/kamu-web-ui): - NON-BLOCKING, BUT CAN USE NEW GQL API ✅
    
  - [x] [release-queue](https://github.com/orgs/kamu-data/projects/4/views/1) - DEPLOY UPDATES REQUIRED


src/domain/accounts/services/Cargo.toml
```diff
@@ -23,12 +23,15 @@ doctest = false
...
 database-common = { workspace = true }
+email-gateway = { workspace = true }
 init-on-startup = { workspace = true }
 internal-error = { workspace = true }
 kamu-accounts = { workspace = true }
+messaging-outbox = { workspace = true }
 odf = { workspace = true }
 time-source = { workspace = true }
 
+askama = { version = "0.12" }
```
sergiimk, line 34: Hm, I expected to see this dependency in an infra layer, not in a domain crate

src/domain/accounts/services/Cargo.toml
```diff
@@ -23,12 +23,15 @@ doctest = false
 
 [dependencies]
 database-common = { workspace = true }
+email-gateway = { workspace = true }
```
sergiimk, line 26: This seems like an odd coupling. I expected that in account domain layer we only emit messages about account life cycle and react to them elsewhere.

The fact that these messages trigger an email notification is IMO neither a account domain concern, nor even an infra-level account service concern. Sending emails notifs is more like Kamu Node-level business logic concern.

I would suggest moving the templating and email sending into a crate in `kamu-node`. Seems like one of those things that we'll never need on CLI level.

src/domain/accounts/services/src/notifiers/account_lifecycle_notifier.rs
```diff
@@ -0,0 +1,107 @@
...
+        &self,
+        created: &AccountLifecycleMessageCreated,
+    ) -> Result<(), InternalError> {
+        let registration_email = RegistrationEmail {
+            username: &created.display_name,
+        };
+        let rendered_registration_body = registration_email.render().unwrap();
+
+        self.email_sender
+            .send_email(
```
sergiimk, line 64: Yea, as per my other comment this looks completely out of place in account domain crate.

src/domain/accounts/services/Cargo.toml
zaychenko-sergei: I don't agree, Askama is not something of infrastructure layer. It's basically a string conversion by temlate.
No matter what is the technology to send an email is, it is still an HTML page.

src/adapter/graphql/src/mutations/account_mut.rs
```diff
@@ -24,8 +25,97 @@ impl AccountMut {
...
+    ) -> Result<UpdateEmailResult> {
+        let Ok(new_email) = Email::parse(&new_email) else {
+            return Ok(UpdateEmailResult::InvalidEmail(UpdateEmailInvalid {
+                dummy: false,
+            }));
+        };
+
+        let account_repo = from_catalog_n!(ctx, dyn AccountRepository);
+        match account_repo
+            .update_account_email(&self.account.id, new_email.clone())
```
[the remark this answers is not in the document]
zaychenko-sergei: Any account mutation entry point is gated in general - a user must be logged in and must match to the account being modified:

```
#[Object]
impl AccountsMut {
    async fn by_id(&self, ctx: &Context<'_>, account_id: AccountID) -> Result<Option<AccountMut>> {
        check_logged_account_id_match(ctx, &account_id)?;
        ...
     }

    async fn by_name(
        &self,
        ctx: &Context<'_>,
        account_name: AccountName,
    ) -> Result<Option<AccountMut>> {
        check_logged_account_name_match(ctx, &account_name)?;
        ...
    }
}
```

Regarding extracting a whole use case, feels like an overkill.
It might not be a perfect design (since authorization is in the handler), but adding a use case layer just to update a little attribute feels unmotivated for now.

src/domain/accounts/services/Cargo.toml
zaychenko-sergei: It's a good point. It's there, as is there was no other place to hold it. I guess I will start a Kamu-node branch pointing on this branch.

src/domain/accounts/services/Cargo.toml
zaychenko-sergei: But if we move the email generation to Node, then everything about `Askama` and emails, basically 80% of this branch, moves to Node.

Issue #1036, opened 2025-01-15
Poor first pull user experience


rmn-boiko:
Implement similar solution for pull command, we did in scope of this epic:

https://github.com/kamu-data/kamu-cli/issues/450

PR #1052, opened 2025-01-29
Pull command ux improve

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/1036

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/domain/task-system/services/src/task_agent_impl.rs
```diff
@@ -149,15 +149,18 @@ impl TaskAgentImpl {
...
                 |task_definition_planner: Arc<dyn TaskDefinitionPlanner>| async move {
                     task_definition_planner
                         .prepare_task_definition(&task.logical_plan)
                         .await
                 },
             )
-            .await?;
+            .await
+        else {
+            return Ok(TaskOutcome::Failed(TaskError::Empty));
```
zaychenko-sergei, line 162: Add a trace for the error, it is swallowed completely right now.

rmn-boiko: We already have on the inside level, but yeah, will add here too

src/e2e/app/cli/repo-tests/src/test_flow.rs
```diff
@@ -1177,6 +1178,224 @@ pub async fn test_trigger_flow_reset(mut kamu_api_server_client: KamuApiServerCl
...
+            .await,
+        FlowTriggerResponse::Success(_)
+    );
+
+    kamu_api_server_client.flow().wait(&dataset_id).await;
+
+    kamu_api_server_client
+        .graphql_api_call_assert(
+            &get_dataset_list_flows_query(&dataset_id),
+            Ok(&indoc::indoc!(
```
sergiimk, line 1251: Consider using `json!()` macro when comparing results. With it you can compare only the nodes you need (avoiding the very deep nestedness), you don't need to use `replace` to insert dynamic values into the string, and the test will be more resilient to changes. 

[Example](https://github.com/kamu-data/kamu-cli/blob/master/src/adapter/http/tests/tests/test_data_query.rs#L444-L447).

src/infra/core/src/services/remote/remote_alias_resolver_impl.rs
```diff
@@ -186,6 +186,61 @@ impl RemoteAliasResolver for RemoteAliasResolverImpl {
...
+                let remote_repo = self.remote_repo_reg.get_repository(&alias.repo_name)?;
+                if remote_repo.url.is_s3_protocol() {
+                    return Ok(remote_repo
+                        .url
+                        .join(&format!("{}/", alias.local_alias()))
+                        .unwrap());
+                }
+
+                let transfer_url = if remote_repo.url.is_odf_protocol() {
+                    remote_repo.url.clone().odf_to_transport_protocol()?
```
sergiimk, line 214: Surprised to see `odf_to_transport_protocol` on this level. Wouldn't this cause what would be a SmTP push to become a SiTP push? Shouldn't we preserve the `odf+` prefix for the sync service to pick the correct transport protocol?

src/infra/core/src/services/remote/remote_alias_resolver_impl.rs
```diff
@@ -186,6 +186,61 @@ impl RemoteAliasResolver for RemoteAliasResolverImpl {
...
+            )),
+            odf::DatasetRefRemote::Url(url) => {
+                let mut dataset_url = url.as_ref().clone();
+                dataset_url.ensure_trailing_slash();
+                Ok(dataset_url)
+            }
+            odf::DatasetRefRemote::Alias(alias)
+            | odf::DatasetRefRemote::Handle(odf::metadata::DatasetHandleRemote { alias, .. }) => {
+                let remote_repo = self.remote_repo_reg.get_repository(&alias.repo_name)?;
+                if remote_repo.url.is_s3_protocol() {
```
sergiimk, line 206: Is this only for S3, or should this apply to all non-`odf` protocols? e.g. if we support push via IPFS or WebDAV - I don't think we should be mutating URLs of those protocols. Only in ODF protocol we can reason about multi-tenancy.

sergiimk (commented):
Let a few question and suggestions.

src/e2e/app/cli/repo-tests/src/test_flow.rs
rmn-boiko: Yeah, good suggestion, changed

src/infra/core/src/services/remote/remote_alias_resolver_impl.rs
rmn-boiko: We are using transport variant to fetch token and work with recent API endpoints

src/infra/core/src/services/remote/remote_alias_resolver_impl.rs
rmn-boiko: Good catch, I've changed to check not a specific type but of only.

src/infra/core/src/services/remote/remote_alias_resolver_impl.rs
sergiimk: I still don't quite understand.

I see that the URL returned from the call to `remote_alias_resolver.resolve_pull_url()` is:
- used to construct a `Dataset` instance
- then wrapped into the `SyncRefRemote`

So I don't see how after we stripped out `odf+https` into plain `https` can the `SyncService` understand that SmartTP should be used instead of SimpleTP.

I understand the need to convert to HTTP(s) at some level, but I expected it to be much lower-level logic.

Am I misunderstanding the logical flow here?

rmn-boiko: At the end we return `combine_remote_url` which will add of back, plaint `https` is needed to call API endpoints and search in token store. Not sure about correct level, but it we use it right after we perform  `odf_to_transport_protocol`

Issue #1098, opened 2025-02-23
Jupyter demo: Unable to launch spark engine


sergiimk:
### Describe the bug

Looks like something broke in jupyter.demo.kamu.dev environment making it unable to run local spark engine.

Spark is only used by some advanced examples, so this issue is not a show stopper.

### Steps To Reproduce

1. Login to https://jupyter.demo.kamu.dev env
2. Open terminal
3. 
    ```sh
    cd 01\ -\ Kamu\ Basics\ \(COVID-19\ example\)/
    ./init-chapter-2.sh
    kamu sql server --engine spark --livy
    ```

### Current Behavior

Observed error:
```
Error:
  0: Internal error
  1: Timed out after 30s
Help us by reporting this problem at https://github.com/kamu-data/kamu-cli/issues
```
In `cat .kamu/run/livy.err.txt`:
```
Error: invalid config provided: cannot set hostname when running in the host UTS namespace: invalid configuration
```

### Expected Behavior

Spark launches fine.

### kamu system info

```sh
appVersion: 0.224.0
```

PR #1115, opened 2025-02-28
Fix spark container spawn inside jupyter

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/1098

## Testing plan

- [x] Tested locally on ubuntu with docker/podman
- [x] Tested locally on macos with docker/podman
- [x] Tested locally on window using virtual machine with docker/podman
- [x] Tested jupyter(docker)+ podman as a container runtime inside

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/images/demo/Dockerfile.jupyter
+++ b/images/demo/Dockerfile.jupyter
@@ -13,7 +13,7 @@ USER root
 # Podman & tools
 # Source: https://github.com/containers/podman/blob/056f492f59c333d521ebbbe186abde0278e815db/contrib/podmanimage/stable/Dockerfile
 RUN apt update && \
-    apt -y install ca-certificates curl wget gnupg unzip jq podman fuse-overlayfs && \
+    apt -y install ca-certificates curl wget gnupg unzip jq podman fuse-overlayfs catatonit && \
     apt-get clean && rm -rf /var/lib/apt/lists /var/cache/apt/archives
 
 COPY podman/containers.conf /etc/containers/containers.conf
--- a/images/demo/Dockerfile.jupyter
+++ b/images/demo/Dockerfile.jupyter
@@ -13,7 +13,7 @@ USER root
 # Podman & tools
 # Source: https://github.com/containers/podman/blob/056f492f59c333d521ebbbe186abde0278e815db/contrib/podmanimage/stable/Dockerfile
 RUN apt update && \
-    apt -y install ca-certificates curl wget gnupg unzip jq podman fuse-overlayfs catatonit && \
+    apt -y install ca-certificates curl wget gnupg unzip jq podman fuse-overlayfs && \
     apt-get clean && rm -rf /var/lib/apt/lists /var/cache/apt/archives
 
 COPY podman/containers.conf /etc/containers/containers.conf
--- a/images/demo/jupyter/.kamuconfig
+++ b/images/demo/jupyter/.kamuconfig
@@ -4,7 +4,6 @@ kind: CLIConfig
 content:
   engine:
     runtime: podman
-    networkNs: host
     maxConcurrency: 1
     startTimeout: 60s
   protocol:
--- a/images/demo/podman/containers.conf
+++ b/images/demo/podman/containers.conf
@@ -1,8 +1,6 @@
 [containers]
-netns="host"
 userns="host"
 ipcns="host"
-utsns="host"
 cgroupns="host"
 cgroups="disabled"
 log_driver = "k8s-file"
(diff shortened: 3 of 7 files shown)


images/demo/Dockerfile.jupyter
```diff
@@ -13,7 +13,7 @@ USER root
 # Podman & tools
 # Source: https://github.com/containers/podman/blob/056f492f59c333d521ebbbe186abde0278e815db/contrib/podmanimage/stable/Dockerfile
 RUN apt update && \
-    apt -y install ca-certificates curl wget gnupg unzip jq podman fuse-overlayfs && \
+    apt -y install ca-certificates curl wget gnupg unzip jq podman fuse-overlayfs catatonit && \
```
s373r, line 16: if the image has been modified, you need to update the image version

s373r (changes requested):

images/demo/Dockerfile.jupyter
rmn-boiko: Yeah, once I'll get approves will do it in release PR

images/demo/podman/containers.conf
```diff
@@ -1,8 +1,6 @@
 [containers]
-netns="host"
 userns="host"
 ipcns="host"
-utsns="host"
```
s373r, line 5: Suggestion: write a comment on why some networks allow host mode and some do not

src/app/cli/src/explore/sql_shell_impl.rs
```diff
@@ -81,7 +81,7 @@ impl SqlShellImpl {
                 .stderr(std::fs::File::create(&spark_stderr_path).int_err()?);
 
             container_builder = if let Some(p) = port {
-                container_builder.network("host").map_port_with_address(
+                container_builder.map_port_with_address(
```
s373r, line 84: Q: in the case of Docker, does this change also work?

rmn-boiko: I've added testing steps I;ve done. 
Answering your question, yes, it was in my testing plan.
The only possible issue might be, if we will change our jupyter to next setup(docker inside docker). I didn't cover such case, but think it is not in scope

images/demo/podman/containers.conf
rmn-boiko: Thanks, I've added comments, for covered changes

rmn-boiko, line 5: Removing next configuration allow us to resolve problem described in the next issue with podman: https://github.com/containers/podman/issues/11969

images/demo/Dockerfile.jupyter
rmn-boiko, line 16: `catatonit` tool needs to be installed to make our `kamu sql server --engine spark` command work fine.
We are trying to start the container with init flag, and podman by default, is trying to use this tool

images/demo/podman/containers.conf
s373r: Sorry I didn't clarify right away -- I was referring to the comments in `images/demo/podman/containers.conf`

images/demo/jupyter/.kamuconfig
```diff
@@ -4,7 +4,6 @@ kind: CLIConfig
 content:
   engine:
     runtime: podman
-    networkNs: host
```
sergiimk, line 7: I think the necessity for using host networking came from kubernetes. As Jupyter runs inside k8s pod it seemed not to be able to create another network NS inside it.

I don't see testing under k8s in your test plan, so I'd recommend:
- Extracting the full podman run command and testing it in `jupyter.demo.kamu.dev` environment
- You can run spark container manually and then see if our python client can [connect to it](https://github.com/kamu-data/kamu-client-python/tree/master?tab=readme-ov-file#using-with-spark)

src/app/cli/src/explore/sql_shell_impl.rs
sergiimk: The number of different configurations and considerations is getting very complex.

While your memory is fresh, I suggest creating a test suite similar to [the one I made for notebooks](https://github.com/kamu-data/kamu-testing/blob/master/test-cases/cli/notebooks.md) and describe different configurations that we want to be checked, e.g.:
- SQL shell
  - datafusion / spark engine
  - docker / podman / podman-in-k8s runtime
- SQL server
  - flight-sql / spark+thrift / spark+livy protocol
  - docker / podman / podman-in-k8s runtime

sergiimk (approved):

Issue #1116, opened 2025-02-28
Private Datasets: Milestone 2. Access Sharing


s373r:
### Context

- RFC: [link](https://docs.google.com/document/d/1qbay9-tiGfBQRO4SKDY8nvoTIl3iGq2vRazZRG6NGp0)

The 1st milestone (https://github.com/kamu-data/kamu-cli/issues/676) implemented the Private Datasets framework, which focused on the Dataset Owner. 

The goal of this milestone is to implement the ability to grant (and revoke) access to work with a dataset to another user

### Expectations

- Expand `DatasetAction::{ Read, Write }` with new `Maintain` & `Own`
  - In other words, `Maintainer` role is added to `Reader` and `Editor` roles
- New GQL API:
  - ```
    Dataset { 
        collaboration: DatasetCollaboration {
            accountRoles() -> { DatasetAccessRole, Account }
        }
    }
    DatasetMut { 
        collaboration: DatasetCollaborationMut {
            setRole(accountId, role)
            unsetRoles([accountId])
        }
    }
    ```
- Update GQL `Dataset::permissions()` to reflect the new permissions
  - TBA
- To implement user input auto-completion, we introduce a new endpoint:
  - ```
    Search::name_lookup(
        query: String,
        filters: LookupFilters,
        page: Option<usize>,
        per_page: Option<usize>,
    ) -> Result<NameLookupResultConnection> { ... }
    
    pub enum NameLookupResult {
        Account(Account),
        // Reserved for future use:
        // Dataset
        // Organization,
        // Issue,
    }
    
    LookupFilters {
        by_account: Option<AccountLookupFilter>,
    }
    
    AccountLookupFilter {
        // Empty for now
    }
    ```
- Use cases are covered by tests:
  - **UC1**: As a dataset maintainer, I can see a list of users and their roles that have access to a dataset.
  - **UC2**: As a dataset maintainer, I can grant access (via role) with another user to the dataset as a Reader or Editor.
    - **NOTE1**: A user can have only one role in relation to a dataset.
    - **NOTE2**: We can add access to the public dataset as well. Let's imagine such a case:
      - Intention: we're going to make the dataset private, with a guarantee that required users won't lose access.
      - We give access to the users in advance.
      - Change the visibility on the dataset to private.
      - Result: the users did not lose access.
    - **NOTE3**: Role is granted instantly and does not require a confirmation from another user.
[...]

PR #1121, opened 2025-03-04
Private Datasets: Milestone 2. Access Sharing

s373r:
## Description

Closes: #1116

### Added
- Private Datasets, access sharing:
  - ReBAC: there can be only one role between a subject (account) and an object (dataset)
  - ReBAC: Added `RebacDatasetRegistryFacade` to simplify work with authorization validation
  - GQL: CRUD Collaboration API
  - GQL: Implemented API for searching by account name
  - OSO: Added `maintain` & `own` permissions in addition to `read` & `write`
### Fixed
- Private Datasets:
  - GQL: Correct processing of access permissions when viewing flows of another account

- [ ] Compatibility:
    
  - [x] Network APIs: :warning: Updated
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/e2e/app/cli/common/src/kamu_api_server_client_ext.rs
```diff
@@ -1020,36 +1259,130 @@ pub struct DatasetDependencies {
...
-    NotFound,
+    #[error("Dataset not found")]
+    DatasetNotFound,
+    #[error("Access")]
+    Access,
     #[error(transparent)]
     Internal(#[from] InternalError),
 }
 
+impl PartialEq for SetDatasetVisibilityError {
```
zaychenko-sergei, line 1312: 1) Normally the errors are not compared, rather tested with `assert_matches!`

2) If you customize `PartialEq`, is there a reason not to define `Eq`?

zaychenko-sergei (approved):

src/e2e/app/cli/common/src/kamu_api_server_client_ext.rs
s373r: Done

src/infra/auth-rebac/postgres/src/repos/postgres_rebac_repository.rs
```diff
@@ -255,15 +255,16 @@ impl RebacRepository for PostgresRebacRepository {
...
-                )
+        .map_err(|e| {
+            match e {
+                sqlx::Error::Database(e) if e.is_unique_violation() => {
+                    InsertEntitiesRelationError::some_role_is_already_present(
+                        subject_entity,
+                        object_entity,
+                    )
+                }
+                _ => InsertEntitiesRelationError::Internal(e.int_err()),
```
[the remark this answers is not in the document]
s373r: > Recent experience tells that we should better not rely on this, esp. for tests, as this marks the transaction as failed - very hard to debug.

I understand, I've encountered this behavior once before -- at the same time, I don't want the tests to dictate the rules.

> In a similar situation in dataset entries, I've added an additional SELECT query, without removing this part. It would softly detect most of the typical cases without breaking transactions. For rare racing case, the current solution still works as a fallback.

Yes, I pointed that out in https://github.com/kamu-data/kamu-cli/pull/1147

In this case, I leave it as is, since we're doing pre-cleaning of roles (in `RebacServiceImpl`)

PR #1136, opened 2025-03-12
Semantic search

sergiimk:
## Description

Addresses: kamu-data/kamu-node#131

Introduces semantic search using OpenAI API for embeddings and Qdrant as a vector store.

## Example config
```yaml
  search:
    # Optional: CLI defaults to OpenAI API and will look for OPENAI_API_KEY env var if this is not configured
    embeddingsEncoder:
      kind: openAi
      apiKey: sk-proj-...
   # Optional: CLI defaults to `qdrantContainer`, spawning a background container during command execution
    vectorRepo:
      kind: qdrant
      url: http://localhost:6334
```

## Example output
image
image

## Issues & Limitations:
- Currently supports re-indexing on startup or lazily upon the first query. Will extend to support reacting to dataset creation / updates / deletes in separate PR
- Due to #1137 the search command currently leaks the Qdrant container

- [ ] Unit and integration tests added ❌
  - The tests will be non trivial to add so skipping for now
  - Will research adding an in-memory impl of vector store for CLI use and testing
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
    - https://github.com/kamu-data/kamu-node/pull/204
    

src/domain/search/services/src/search_service_local_impl.rs
```diff
@@ -0,0 +1,158 @@
...
+            self.vector_repo.as_deref(),
+        ) else {
+            return Err(NatLangSearchNotEnabled.into());
+        };
+
+        // Encode prompt
+        let prompt_vec = embeddings_encoder
+            .encode(vec![prompt.to_string()])
+            .await
+            .int_err()?
```
rmn-boiko, line 74: Would be nice to return here some InvalidPrompt error instead of internal

src/domain/search/services/src/search_service_local_impl.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        // Filter by accessibility
+        let dataset_handles = self
+            .dataset_action_authorizer
+            .filter_datasets_allowing(dataset_handles, kamu_core::auth::DatasetAction::Read)
+            .await?;
+
+        // Merge found points with accessible handles
+        let dataset_handles_by_id: std::collections::BTreeMap<odf::DatasetID, odf::DatasetHandle> =
```
rmn-boiko, line 135: Simple HashMap should be enough?

src/infra/search/qdrant/src/vector_repo_qdrant.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    async fn client(&self) -> Result<&qdrant_client::Qdrant, InternalError> {
+        self.client
+            .get_or_try_init(async || self.init_client().await)
+            .await
+    }
+
+    async fn init_client(&self) -> Result<qdrant_client::Qdrant, InternalError> {
+        use ::qdrant_client::qdrant::*;
```
rmn-boiko, line 50: move this on top of file, such as 3 methods need this

src/adapter/graphql/src/queries/search.rs
```diff
@@ -92,6 +95,58 @@ impl Search {
...
     }
+
+    /// Searches for datasets and other objects managed by the
+    /// current node using a prompt in natural language
+    #[tracing::instrument(level = "info", name = Search_query, skip_all, fields(?per_page))]
+    async fn query_natural_language(
+        &self,
+        ctx: &Context<'_>,
+        prompt: String,
+        // page: Option<usize>,
```
s373r, line 106: Q: maybe add `page` to the schema already, but just don't use it yet?

src/app/cli/src/services/config/models.rs
```diff
@@ -836,6 +843,216 @@ impl IdentityConfig {
...
+        Self::OpenAi(EmbeddingsEncoderConfigOpenAi::default())
+    }
+}
+
+#[skip_serializing_none]
+#[derive(Debug, Clone, Merge, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
+pub struct EmbeddingsEncoderConfigOpenAi {
+    pub url: Option<String>,
```
s373r, line 978: `Option<Url>`

src/app/cli/src/services/config/models.rs
```diff
@@ -836,6 +843,216 @@ impl IdentityConfig {
...
+    }
+}
+
+#[skip_serializing_none]
+#[derive(Debug, Clone, Merge, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
+pub struct VectorRepoConfigQdrant {
+    #[merge(skip)]
+    pub url: String,
```
s373r, line 1019: `Url`

src/infra/search/openai/src/embeddings_encoder_openai.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use internal_error::*;
+use kamu_search::EmbeddingsEncoder;
+use secrecy::ExposeSecret as _;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, Clone)]
+pub struct EmbeddingsEncoderConfigOpenAI {
+    pub url: Option<String>,
```
s373r, line 20: `Option<Url>`?

src/infra/search/qdrant/src/vector_repo_qdrant.rs
```diff
@@ -0,0 +1,182 @@
...
+impl VectorRepository for VectorRepositoryQdrant {
+    async fn num_points(&self) -> Result<usize, InternalError> {
+        let res = self
+            .client()
+            .await?
+            .collection_info(&self.config.collection_name)
+            .await
+            .int_err()?;
+
+        Ok(usize::try_from(res.result.unwrap().points_count.unwrap()).unwrap())
```
s373r, line 91: 3 `unwrap()` in a row looks scary -- if this is normal in this situation, please add a `// Safety:` comment

src/infra/search/qdrant/src/vector_repo_qdrant.rs
```diff
@@ -0,0 +1,182 @@
...
+        };
+
+        self.client()
+            .await?
+            .upsert_points_chunked(
+                qdrant_client::qdrant::UpsertPointsBuilder::new(
+                    &self.config.collection_name,
+                    points,
+                ),
+                1000,
```
s373r, line 121: Please add a comment re `1000` const

s373r (commented):
Add minor comments

src/domain/search/services/src/search_service_local_impl.rs
sergiimk: On kamu's side we don't have a clear concept of what invalid prompt is - it's up to embeddings model to reject it. We can introduce a better error type for this later if we notice common failure cases (e.g. unsupported languages, or prompt being too long).

src/domain/search/services/src/search_service_local_impl.rs
sergiimk: My intuition was that we will usually encounter very small number of elements here (page size being let's say ~10-50 items), and BTreeMap should perform better on tiny sets.

Chatgpt seems to agree:
image

src/app/cli/src/services/config/models.rs
sergiimk: I was thinking about this, but then saw Qdrant for example supporting "special URLs" like `:memory:` for in-memory impelmentation. Such value would not parse into `Url` and will be rejected, so in case of external components I thought it's better to leave URL validation to them.

src/infra/search/qdrant/src/vector_repo_qdrant.rs
sergiimk: replaced with `expect()` with nice messages

src/infra/search/qdrant/src/vector_repo_qdrant.rs
sergiimk: moved into a constant

src/infra/search/qdrant/src/vector_repo_qdrant.rs
sergiimk: I kinda like this approach with local imports:
- unlike wildcard import at the top of the file, it allows me not to pollute the global namespace with a bunch of Qdrant-specific names
- and it avoids very long import statement for all these builders if I were to import them one-by one

Happy to discuss if we want to allow such things in our code convention or not.

PR #1173, opened 2025-04-01
Query performance pt.2

sergiimk:
In this PR:
- I externalize Datafusion configuration
  - [These config options](https://datafusion.apache.org/user-guide/configs.html) can now be included in our config filles
  - The options can be separated into base (applies to all), ingest, batch query, and compaction-specific options - so all of these can be tuned differently
  - I introduce new wrapper types that then carry these configs via DI (so a lot of tests were affected)
- In `QueryOptions` I differentiate the "unspecified" (`None`) from empty (`Some([])`) list of input datasets
  - Thanks to this a query like `select 1` will resolve of an empty list of datasets, allowing us to skip listing schemas of all datasets like before to create the Datafusion schema
  - This makes `select 1` query take milliseconds
- Queries like `show tables` still result in "unspecified" list of datasets - i.e. potentially can touch all datasets available to the user
  - But thanks to query analysis I detect `show tables` and pass a new `does_not_need_exact_schema` hint
  - So although we still end up listing all datasets, we are able to avoid loading the schemas of each dataset (traversing metadata) which is expensive


src/infra/core/src/testing/mock_dataset_action_authorizer.rs
```diff
@@ -78,6 +116,22 @@ impl MockDatasetActionAuthorizer {
...
             .returning(|_, _| Ok(()));
         mock_dataset_action_authorizer
+            .expect_classify_dataset_ids_by_allowance()
+            .returning(|ids, _| {
+                Ok(ClassifyByAllowanceIdsResponse {
+                    authorized_ids: ids,
+                    unauthorized_ids_with_errors: Vec::new(),
+                })
+            });
+        mock_dataset_action_authorizer
```
sergiimk, line 126: Notice: `mockall` mocks panic if _none_ of the `expect_*`ed methods was called. After I changed `QueryService` to use batch `classify_dataset_ids_by_allowance` methods the tests started crashing.

To solve this, in `MockDatasetActionAuthorizer::allowing()` and `denying()` I add `exect`s for both singular and batch methods, and if at least one of them is called - the mock will be happy.

sergiimk: @zaychenko-sergei I will go ahead an merge this to propagate changes to the node faster, but wanted to point out this specific change to you, as I saw a lot of complexity forming in this mock class.

Issue #1175, opened 2025-04-03
Restrict force push/pull with rewriting seed block


rmn-boiko:
### Describe the bug

Currently, we allow to force push/pull datasets even if they rewrite the seed block. But since we are storing the dataset in a folder named by its id it causes issues with handling such datasets.

### Steps To Reproduce

Create two different datasets
push first dataset to remote server
push second dataset by using first dataset URL with force flag

### Current Behavior

We rewrite all blocks in dataset by it is findable by primary dataset id

### Expected Behavior

Push/pull flow fails with human human-readable error

### kamu system info

```sh
latest
```

PR #1176, opened 2025-04-03
Restrict force push/pull with seed rewrite

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/1175

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
+++ b/src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
@@ -27,6 +27,7 @@ use kamu_datasets::{
     MESSAGE_PRODUCER_KAMU_HTTP_ADAPTER,
 };
 use messaging_outbox::{Outbox, OutboxExt};
+use odf::dataset::MetadataChainExt;
 use odf::Multihash;
 use tracing::Instrument;
 use url::Url;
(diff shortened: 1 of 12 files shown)


src/adapter/http/src/smart_protocol/axum_server_push_protocol.rs
```diff
@@ -373,9 +363,20 @@ impl AxumServerPushProtocolInstance {
...
 
-        axum_write_payload::<DatasetPushMetadataAccepted>(
+        if let Some(dataset_handle) = self.maybe_dataset_handle.as_ref()
+            && push_request.force_update_if_diverged
+            && let Some((_, first_incoming_block)) = new_blocks.front()
+            && let odf::MetadataEvent::Seed(seed_event) = &first_incoming_block.event
+            && seed_event.dataset_id != dataset_handle.id
+        {
+            return Err(PushServerError::ValidationError(
+                PushValidationError::SeedBlockRewriteRestricted,
```
sergiimk, line 373: Seems like the error doesn't reflect what the actual check is.

What if new seed block has same dataset ID but different dataset kind (switching root to derivative) - does our code tolerate that?

What if ID and kind are the same, but only timestamps and overall block hash has changed - is there a reason for us to reject such force push?

I'd definitely agree that you should not be able to change ID during the force-push. I'd also restrict changing kind for simplicity, until we find a good reason to allow it. But a seed block that only differs in timestamp and hash seems harmless.

src/adapter/http/src/smart_protocol/ws_tungstenite_client.rs
```diff
@@ -616,7 +626,15 @@ impl SmartTransferProtocolClient for WsSmartTransferProtocolClient {
...
             let dst = if let Some(dst) = dst {
-                dst
+                // Check is incoming seed is different from existing
+                if transfer_options.force_update_if_diverged
+                    && let Some((_, first_incoming_block)) = new_blocks.front()
+                    && let odf::MetadataEvent::Seed(seed_event) = &first_incoming_block.event
+                    && &seed_event.dataset_id != dst.get_id()
+                {
+                    InternalError::bail("Rewriting seed block is restricted")?;
+                }
```
zaychenko-sergei, line 636: Can these conditions be isolated in a common handler? It seems we have largely copied them from
`axum_server_push_protocol.rs`

zaychenko-sergei (approved):

Issue #1019, opened 2024-12-27
GQL: datasets.byOwnerAndName should not return account prefixed alias in a single-tenant workspace


andriidemus:
### Describe the bug

We parse prefixed alias, as an alias from multi-tenant workspace.
For the consistency, it makes sense to make `datasets.byOwnerAndName` return unprefixed alias in case of singe-tenant workspace.

[Here are](https://gist.github.com/demusdev/768e3acb2184d5fa031ed9d24e5593c0) several API calls examples in both single- and multi-tenant workspaces

### Steps To Reproduce

1. Create single tenant workspace
```bash
kamu init
```
3. Start server
```bash
kamu system api-server --http-port 55667 --get-token
export KAMU_AUTH_TOKEN="<token>"
```
5. Create empty dataset
```bash
curl -s 'http://127.0.0.1:55667/graphql' \
  --header "Authorization: Bearer $KAMU_AUTH_TOKEN" \
  --data-raw '{"query": "mutation { datasets { createEmpty(datasetKind: ROOT, datasetAlias: \"empty-root-dataset\") { message ... on CreateDatasetResultSuccess { dataset { alias } } } } }"}'

```
7. Get dataset alias
```bash
curl -s 'http://127.0.0.1:55667/graphql' \
  --header "Authorization: Bearer $KAMU_AUTH_TOKEN" \
  --data-raw '{"query": "{ datasets { byOwnerAndName (accountName: \"kamu\", datasetName: \"empty-root-dataset\") {alias owner {accountName}} } }"}' \
  | jq .data.datasets.byOwnerAndName.alias
```

### Current Behavior

Dataset alias prefixed with account name: "<account_name>/<dataset>"
```
{
  "data": {
    "datasets": {
      "byOwnerAndName": {
        "alias": "kamu/empty-root-dataset",
      }
    }
  },
  "extensions": {}
}
```

### Expected Behavior

Dataset alias format without account name prefix
```
{
  "data": {
    "datasets": {
      "byOwnerAndName": {
        "alias": "empty-root-dataset",
      }
    }
  },
  "extensions": {}
}
```

### kamu system info

```sh
build:
  appVersion: 0.215.0
```

s373r:
Offtop: @demusdev , you can run a GQL query using the following command:
```shell
kamu system api-server gql-query '{ apiVersion }'
```

PR #1179, opened 2025-04-04
GQL: `DatasetsMut::create_empty()` & `DatasetsMut::create_from_snapshot()`: alias name validation in multi-tenant mode

s373r:
## Description

Closes: #1019 

- The task being closed required that the account name not be returned in the alias (single-tenant)
- After double-checking, it turns out this is already happening
- But the possibility of a minor collision was found in the case of the multi-tenant mode

### Added
- GQL: `DatasetsMut::create_empty()` & `DatasetsMut::create_from_snapshot()`: alias name validation in multi-tenant mode

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/domain/datasets/services/src/utils/create_dataset_use_case_helper.rs
+++ b/src/domain/datasets/services/src/utils/create_dataset_use_case_helper.rs
@@ -11,23 +11,25 @@ use std::sync::Arc;
 
 use dill::component;
 use internal_error::{ErrorIntoInternal, InternalError};
-use kamu_accounts::CurrentAccountSubject;
 use kamu_core::{ResolvedDataset, TenancyConfig};
 use kamu_datasets::{
     CreateDatasetError,
+    CreateDatasetFromSnapshotError,
     DatasetLifecycleMessage,
     DatasetReferenceCASError,
+    IncorrectAliasAccountNameError,
     NameCollisionError,
     MESSAGE_PRODUCER_KAMU_DATASET_SERVICE,
 };
 use messaging_outbox::{Outbox, OutboxExt};
+use thiserror::Error;
 
 use crate::{CreateDatasetEntryError, DatasetEntryWriter};
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[component(pub)]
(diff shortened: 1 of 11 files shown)


zaychenko-sergei (approved):

src/domain/datasets/services/src/utils/create_dataset_use_case_helper.rs
```diff
@@ -185,3 +181,27 @@ impl CreateDatasetUseCaseHelper {
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Error, Debug)]
+pub enum CanonicalDatasetAliasError {
+    #[error(transparent)]
+    IncorrectAliasAccountName(#[from] IncorrectAliasAccountNameError),
```
sergiimk, line 188: I disagree a bit with this error type. I don't think that user `alice` trying to create a dataset under `bob/foo` is an "incorrect alias error" - if that was the case - our API should not have accepted a multi-tenant `DatasetAlias` to begin with and only accepted `DatasetName`.

I definitely see in future cases where `alice` creates a dataset under another account, e.g. an organization that she is a part of. So in this case a more natural errors to me are:
- **unauthorized** - you don't have permissions to create a dataset under this different users
- **not found** - the account you're referring to does not exist (or you don't have permissions to see it)

sergiimk (changes requested):

src/domain/datasets/services/src/utils/create_dataset_use_case_helper.rs
sergiimk: While we don't have organizations - I would just return unauthorized for all cases

s373r: > I definitely see in future cases where alice creates a dataset under another account, e.g. an organization that she is a part of

A fair comment

zaychenko-sergei (approved):

Issue #1186, opened 2025-04-09
Device Flow Authentication


s373r:
### Preamble

RFC: [link](https://docs.google.com/document/d/1WjMDKBOzO1pf0-NdhBKm0A0gZIyHAFWqKcOTKhuBAjk)

### Sequence Diagram

![Image](https://github.com/user-attachments/assets/fc20de44-cc62-43c4-aaab-f6bf515f4c23)

### Requirements

1. New `POST /platform/token/device` endpoint:
  - Two modes of operation: with and without device_code;
  - If device_code is not passed, generate; otherwise, check if the required token is available.
  - The data model is described in the RFC ([Sequence Diagram section](https://docs.google.com/document/d/1WjMDKBOzO1pf0-NdhBKm0A0gZIyHAFWqKcOTKhuBAjk/edit?tab=t.0#heading=h.7x5kmh7dhoe3))
2. Implement `DeviceCodeRepository`:
  - Store the correspondence between `device_code` and the received user token.
  - Implement an `InitOnStartup` job that will clear the table of expired data.
3. Update `LoginService` to use Device Flow instead of Loopback Redirect Flow.
4. Extend the GQL `AuthMut::login()` method by adding an optional device_code parameter

PR #1188, opened 2025-04-09
Device Flow Authentication

s373r:
## Description

Closes: #1186

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

migrations/postgres/20250415100644_device_codes.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE device_codes
```
zaychenko-sergei, line 1: let's clearly indicate this is related to oauth in our names

zaychenko-sergei: who knows what kind of IoT device modeling we will have tomorrow

migrations/postgres/20250415100644_device_codes.sql
```diff
@@ -0,0 +1,13 @@
...
+(
+    device_code            uuid        NOT NULL PRIMARY KEY,
+    device_code_created_at timestamptz NOT NULL,
+    device_code_expires_at timestamptz NOT NULL,
+    token_iat              BIGINT,
+    token_exp              BIGINT,
```
zaychenko-sergei, line 7: Aren't iat/exp the same data as two fields above?

migrations/sqlite/20250415100644_device_codes.sql
```diff
@@ -0,0 +1,13 @@
...
+    device_code_created_at TEXT        NOT NULL,
+    device_code_expires_at TEXT        NOT NULL,
+    token_iat              INTEGER,
+    token_exp              INTEGER,
+    token_last_used_at     TEXT,
+    account_id             VARCHAR(100) REFERENCES accounts (id)
+);
+
+CREATE INDEX idx_device_codes_device_code_expires_at
+    ON device_codes (DATETIME(device_code_expires_at) DESC);
```
zaychenko-sergei, line 13: It's interesting. Please double check if we apply this trick in other tables where sorting by time might be expected.

src/adapter/http/src/axum_utils.rs
```diff
@@ -13,6 +13,21 @@ use kamu_accounts::{AnonymousAccountReason, CurrentAccountSubject};
...
+macro_rules! from_catalog_n {
+    ($catalog:ident, $T:ty) => {{
+        $catalog.get_one::<$T>().unwrap()
+    }};
+    ($catalog:ident, $T:ty, $($Ts:ty),+) => {{
+        ( $catalog.get_one::<$T>().unwrap(), $( $catalog.get_one::<$Ts>().unwrap() ),+ )
+    }};
+}
+
+pub(crate) use from_catalog_n;
```
zaychenko-sergei, line 29: I think we had a similar thing in GQL adapter. Does this mean it should be a part of `dill`?

src/adapter/http/src/http_server_dataset_router.rs
```diff
@@ -150,14 +158,294 @@ pub async fn platform_login_handler(
...
-#[allow(clippy::unused_async)]
+#[expect(clippy::unused_async)]
 pub async fn platform_token_validate_handler(catalog: Extension<Catalog>) -> Result<(), ApiError> {
     ensure_authenticated_account(&catalog).api_err()?;
     Ok(())
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+/// <https://datatracker.ietf.org/doc/html/rfc8628#section-3.1>
```
zaychenko-sergei, line 169: This file became very big. Consider splitting on multiple Rust files.

src/app/cli/src/services/odf_server/login_service.rs
```diff
@@ -263,9 +241,103 @@ impl LoginService {
...
+        &self,
+        client: &reqwest::Client,
+        login_polling_url: &Url,
+        device_code: &str,
+    ) -> Result<DeviceAccessTokenResult, InternalError> {
+        let response = client
+            .post(login_polling_url.clone())
+            .form(&[
+                // We do not use this value, but we are required to pass it.
+                ("client_id", "kamu"),
```
zaychenko-sergei, line 323: It's used twice at least, so make a constant

src/domain/accounts/domain/src/services/device_code_service.rs
```diff
@@ -0,0 +1,97 @@
...
+            Ok(device_token) => Ok(Some(device_token)),
+            Err(E::NotFound(_)) => Ok(None),
+            Err(E::Internal(e)) => Err(e),
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[nutype(sanitize(trim), validate(not_empty), derive(AsRef))]
```
zaychenko-sergei, line 84: Love this stuff. We should identify other places in our project, where such value objects with validation could be beneficial.

src/domain/accounts/services/src/services/device_code_service_impl.rs
```diff
@@ -0,0 +1,116 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[dill::component]
+#[dill::interface(dyn DeviceCodeService)]
+#[dill::interface(dyn InitOnStartup)]
+#[dill::meta(InitOnStartupMeta {
+    job_name: JOB_KAMU_ACCOUNTS_DEVICE_CODE_SERVICE,
+    depends_on: &[
+        JOB_KAMU_ACCOUNTS_PREDEFINED_ACCOUNTS_REGISTRATOR,
+        JOB_KAMU_DATASETS_DATASET_ENTRY_INDEXER
```
zaychenko-sergei, line 47: Why does it need any dataset entries at all?
I would rather avoid any dependencies of accounts domain from datasets domain, as the other direction of dependency exists already.

src/adapter/http/tests/harness/client_side_harness.rs
```diff
@@ -195,6 +196,10 @@ impl ClientSideHarness {
         b.add_value(kamu::utils::ipfs_wrapper::IpfsClient::default());
         b.add_value(IpfsGateway::default());
 
+        b.add::<DeviceCodeServiceImpl>();
+        b.add::<InMemoryDeviceCodeRepository>();
+        b.add::<DummyDatasetEntryIndexer>();
```
zaychenko-sergei, line 201: Not sure I understood a need in dataset entry indexer here and in a few more test groups.
The authentication tasks should be totally independent on datasets domain.

src/domain/accounts/services/src/dependencies.rs
```diff
@@ -0,0 +1,28 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use dill::CatalogBuilder;
+
+use crate::*;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub fn register_dependencies(catalog_builder: &mut CatalogBuilder, needs_indexing: bool) {
```
zaychenko-sergei, line 16: Do we have any more domains level, where it's still not done like that?

src/infra/accounts/postgres/src/repos/postgres_device_code_repository.rs
```diff
@@ -0,0 +1,156 @@
...
+    async fn cleanup_expired_device_codes(&self) -> Result<(), CleanupExpiredDeviceCodesError> {
+        let mut tr = self.transaction.lock().await;
+
+        let connection_mut = tr.connection_mut().await?;
+
+        sqlx::query!(
+            r#"
+            DELETE
+            FROM device_codes
+            WHERE device_code_expires_at < NOW()
```
zaychenko-sergei, line 145: Is this taking timezone differences into account?

migrations/sqlite/20250415100644_device_codes.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE device_codes
+(
+    device_code            VARCHAR(36) NOT NULL PRIMARY KEY,
+    device_code_created_at TEXT        NOT NULL,
+    device_code_expires_at TEXT        NOT NULL,
+    token_iat              INTEGER,
+    token_exp              INTEGER,
+    token_last_used_at     TEXT,
+    account_id             VARCHAR(100) REFERENCES accounts (id)
```
zaychenko-sergei, line 9: Cascade delete

migrations/postgres/20250415100644_device_codes.sql
s373r: Yes, you're right -- the previous name was too generic

Fixed

migrations/postgres/20250415100644_device_codes.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE device_codes
+(
+    device_code            uuid        NOT NULL PRIMARY KEY,
+    device_code_created_at timestamptz NOT NULL,
```
rmn-boiko, line 4: WDYT to revisit such entity duplications in our tables e.g
table `device_codes` -> field `device_code_created_at`. Would be more readable to use `created_at`.

src/adapter/graphql/src/mutations/auth_mut.rs
```diff
@@ -26,11 +26,19 @@ impl AuthMut {
         ctx: &Context<'_>,
         login_method: String,
         login_credentials_json: String,
+        device_code: Option<String>,
     ) -> Result<LoginResponse> {
+        let device_code = if let Some(raw_device_code) = device_code {
+            let device_code = DeviceCode::try_new(&raw_device_code).int_err()?;
```
rmn-boiko, line 32: Maybe some InvalidDecodeError instead Internal here?

migrations/sqlite/20250415100644_device_codes.sql
s373r: `accounts` & `accounts_passwords` tables were updated (SQLite, PostgreSQL). In the case of MySQL, on the contrary, `LOWER()` calls have been removed, since columns are already case-insensitive by default.

Done

src/domain/accounts/domain/src/repos/device_code_repository.rs
```diff
@@ -0,0 +1,78 @@
...
+use internal_error::InternalError;
+use thiserror::Error;
+
+use crate::{DeviceCode, DeviceToken, DeviceTokenCreated, DeviceTokenParamsPart};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait DeviceCodeRepository: Send + Sync {
+    async fn create_device_code(
```
rmn-boiko, line 19: Dont like to see `create_device_code` which accept `device_code_created`.
Suggest to rename to `save_device_code`

src/domain/accounts/domain/src/repos/device_code_repository.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[error(transparent)]
+    NotFound(#[from] DeviceTokenFoundError),
+
+    #[error(transparent)]
+    Internal(#[from] InternalError),
+}
+
+#[derive(Error, Debug)]
+#[error("Dataset token for device_code '{device_code}' not found")]
+pub struct DeviceTokenFoundError {
```
rmn-boiko, line 68: NotFound*

src/app/cli/src/services/odf_server/login_service.rs
```diff
@@ -29,134 +32,107 @@ use crate::odf_server;
...
-        Ok(WebServer {
-            server_future,
-            local_addr,
-        })
-    }
+        let polling_start_time = self.time_source.now();
+        let polling_interval = Duration::seconds(
+            device_authorization_response
+                .interval
+                .unwrap_or(5)
```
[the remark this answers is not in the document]
s373r: The value 5 is taken from the RFC as a mandatory value:
>    /// OPTIONAL.  The minimum amount of time in seconds that the client
>    /// SHOULD wait between polling requests to the token endpoint.  If no
>    /// value is provided, clients MUST use 5 as the default.

Hid the magic number in `DeviceAuthorizationResponse::interval_or_default()`.

Fixed

src/app/cli/src/services/odf_server/login_service.rs
s373r: Done

src/domain/accounts/domain/src/services/device_code_service.rs
s373r: Yeah, I like it a lot too -- we'll replace types as we go along

src/adapter/http/tests/harness/client_side_harness.rs
s373r: Deleted it -- it was being added since I mistakenly listed it as a dependency

Fixed

src/adapter/http/src/http_server_dataset_router.rs
s373r: Extracted all platform related to `./platform` directory

Done

src/domain/accounts/services/src/dependencies.rs
s373r: Added some similar stuff, but not for domains, but for `kamu-cli` entities.

The kamu namespace remains, which needs to be split up, as it has a huge number of dependencies.

Partly done.

migrations/postgres/20250415100644_device_codes.sql
s373r: Yeah, I was tempted to remove the prefixes too, but in this case, we're actually storing two entities: the `device_code` and the optional associated `token`.

Thus, to avoid confusion between the two, I decided to keep the explicit prefixes.

src/adapter/graphql/src/mutations/auth_mut.rs
s373r: At the moment, we don't have a mechanism to pass the message to GQL error message:
```rs
pub enum GqlError {
    Internal(InternalError),
    Access(odf::AccessError),
    Gql(async_graphql::Error),
}
```

That's on the one hand. On the other hand, I agree with you that it should be an understandable error for the calling party. 

Thus, to achieve this, I will prepare a separate GQL scalar for it

s373r: After adding the scalar, the output in case of an error is very clear:
```json
{
  "data": null,
  "extensions": {},
  "errors": [
    {
      "message": "Failed to parse \"DeviceCode\": failed to parse a UUID",
      "locations": [
        {
          "line": 3,
          "column": 118
        }
      ],
      "path": [
        "auth",
        "login"
      ]
    }
  ]
}
```

Done

src/domain/accounts/domain/src/repos/device_code_repository.rs
s373r: Yes indeed, in the repository we only save, not create 

Renamed

src/domain/accounts/domain/src/repos/device_code_repository.rs
s373r: Good find! Thank you!

Fixed

src/infra/accounts/postgres/src/repos/postgres_device_code_repository.rs
s373r: For record: after writing the tests, no additional changes were required

rmn-boiko (approved):

zaychenko-sergei (approved):

PR #1189, opened 2025-04-10
`opendatafabric-openapi-mt-client`: autogenerated client for ODF REST API

s373r:
## Description

- This PR adds an automatically generated client (based on the OpenAPI (mt) schema) to interact with the REST API.
- Generation using [openapi-generator-cli](https://openapi-generator.tech/) with [rust (client) generator](https://openapi-generator.tech/docs/generators/rust).
    - The generator is used pretty much as is, except for changing the `Cargo.toml` template.

### Added
- `opendatafabric-openapi-mt-client`: autogenerated client for ODF REST API

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

Makefile
```diff
@@ -331,3 +334,39 @@ codegen-graphql-schema:
 .PHONY: codegen-cli-reference
 codegen-cli-reference:
 	cargo nextest run -p kamu-cli generate_reference_markdown
+
+
+.PHONY: codegen-openapi-mt-client
+codegen-openapi-mt-client:
```
sergiimk, line 340: Oh my, I don't know how I feel about including such codegenned client crate...

Having it tied to CI, and using typed interfaces would bring us same benefits we are reaping from GQL codegen in Web UI already.

But if we were to adopt this there's a few things I'd like to see:

1) Make sure it doesn't bring in new heavy depenencies, ie.:
- It depents on same versions of e.g. `reqwest` and `tokio` that we are using and not pulls in a whole separate stack
  - I already see an issue that it pulls in `openssl` while the rest of the code is using `rustls` - this is a big no-no for me

2) Let's get rid of auto-generated markdown docs - they just duplicate existing OpenAPI spec and don't bring any value.

3) I don't think it belongs in `odf` layer as it uses endpoints from Kamu

s373r: > Having it tied to CI, and using typed interfaces would bring us same benefits we are reaping from GQL codegen in Web UI already, so it probably makes sense.

To be honest, I've been eyeing autogeneration since the days of E2E test integration. In those, I manually make HTTP calls on the spot (cheap and fast).

Once again, when I needed to make an HTTP request to `kamu-node` (in https://github.com/kamu-data/kamu-cli/pull/1188), I thought it's time to take a couple of hours to make a typed REST API client.

> Make sure it doesn't bring in new heavy depenencies, ie.:

Revised the dependencies, the unnecessary ones are now unused.

Fixed

> Let's get rid of auto-generated markdown docs - they just duplicate existing OpenAPI spec and don't bring any value.

Done

> I don't think it belongs in odf layer as it uses endpoints from Kamu

Fair point -- moved it to `src/utils/kamu-api-client-mt` 

Fixed

zaychenko-sergei:
I guess we need it for API server more than for CLI.
In CLI it would likely benefit to e2e flows at most, while for API server there is a potential for public use.
For API server we might want to provide such auto-generated clients for several more languages, like Python, Java, Node.JS.

Issue #318, opened 2023-12-13
Improve CLI scriptability


sergiimk:
We often need to script some automation around the tool (e.g. workspace setup). We also often run the CLI under CI pipelines. In both cases there are a lot of issues with ergonomics in output and error handling.

Details TBD:
- Add `kamu inspect summary <dataset>` command
- Consider filtering of log messages `kamu log --filter seed`
- Commands don’t output any status under non-tty - do we need some middleground between spinners and logging?
    - See https://github.com/clap-rs/clap/pull/4765

PR #1207, opened 2025-04-19
SKU: Molecule

sergiimk:
This delivery includes:
- VersionedFile dataset "archetype" GQL API
- Collection dataset "archetype" GQL API

Start review with API overview and examples here:
https://docs.google.com/document/d/1Gdh1ml93wrQhD-3z0Qkcd9HjU1h7QOTEe20EMw_O2rg/edit?tab=t.0


src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/collection_mut.rs
```diff
@@ -0,0 +1,427 @@
...
+            {
+                CollectionUpdateResult::Success(r) => {
+                    expected_head = r.new_head.clone().into();
+                    res = Some(CollectionUpdateResult::Success(r));
+                }
+                r => {
+                    return Ok(r);
+                }
+            }
+        }
```
sergiimk, line 212: We rely on sorting by `(pk, op)` in merge strategies, but this fails for changelog streams that can contain multiple retractions/corrections for the same key within a single data batch. I will ticket up and reference this problem here, but this workaround of splitting changes into multiple blocks should last us for a while.

src/adapter/graphql/tests/tests/test_gql_dataset_adapter_viersioned_file.rs
```diff
@@ -0,0 +1,1009 @@
...
+    //                                     url
+    //                                     headers { key value }
+    //                                     expiresAt
+    //                                 }
+    //                             }
+    //                         }
+    //                     }
+    //                 }
+    //             }
+    //             "#
```
sergiimk, line 897: The harness is using GQL API in embedded mode so with its current configuration I cannot test the full path of direct upload/download. Will likely remove this and eventually address this via e2e test.

src/adapter/graphql/src/queries/datasets/dataset.rs
```diff
@@ -74,6 +70,42 @@ impl Dataset {
...
+
+        let Some(odf::MetadataEvent::AddPushSource(source)) = source else {
+            return Ok(false);
+        };
+
+        let push_source_columns: std::collections::BTreeSet<String> = source
+            .read
+            .schema()
+            .cloned()
+            .unwrap_or_default()
```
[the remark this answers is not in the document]
sergiimk: thanks, fixed

src/adapter/graphql/src/queries/datasets/adapters/collection_entry.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[common_macros::method_names_consts(const_value_prefix = "GQL: ")]
+#[ComplexObject]
+impl CollectionEntry {
+    /// Resolves the reference to linked dataset
+    #[tracing::instrument(level = "info", name = CollectionEntry_as_dataset, skip_all)]
+    pub async fn as_dataset(&self, ctx: &Context<'_>) -> Result<Option<Dataset>> {
+        match Dataset::try_from_ref(ctx, &self.reference.as_local_ref()).await? {
+            TransformInputDataset::Accessible(v) => Ok(Some(v.dataset)),
+            TransformInputDataset::NotAccessible(_) => Ok(None),
```
s373r, line 125: Minor: In a good way, we need to rename `TransformInputDataset` in the generated code, since we started using `Dataset::try_from_ref()` more often -- e.g., `MaybeAccessibleDataset` or something like that

src/adapter/graphql/src/queries/datasets/adapters/versioned_file.rs
```diff
@@ -0,0 +1,180 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug)]
+pub struct VersionedFile {
+    state: DatasetRequestStateWithOwner,
+}
+
+impl VersionedFile {
+    pub fn new(state: DatasetRequestStateWithOwner) -> Self {
```
s373r, line 24: Minor: pass as a ref

src/adapter/graphql/src/queries/datasets/adapters/versioned_file.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        let total_count = df.clone().count().await.int_err()?;
+
+        let df = df
+            .sort(vec![col("version").sort(false, false)])
+            .int_err()?
+            .limit(page * per_page, Some(per_page))
+            .int_err()?;
+
+        let records = df.collect_json_aos().await.int_err()?;
```
s373r, line 146: Q: do we need to take the `total_count` based on `records`?

src/adapter/graphql/src/queries/datasets/adapters/versioned_file.rs
```diff
@@ -0,0 +1,180 @@
...
+impl VersionedFile {
+    pub fn new(state: DatasetRequestStateWithOwner) -> Self {
+        Self { state }
+    }
+
+    pub async fn get_entry(
+        &self,
+        ctx: &Context<'_>,
+        as_of_version: Option<FileVersion>,
+        as_of_block_hash: Option<Multihash<'static>>,
```
s373r, line 32: Q: How about crossing these two parameters into one enum?

Thus, the user will not be able to pass them both at the same time

src/adapter/graphql/src/queries/datasets/dataset_request_state.rs
```diff
@@ -19,7 +19,7 @@ use crate::queries::*;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
s373r, line 22: > DatasetRequestState & DatasetRequestStateWithOwner

Note: that is heavy cloning

src/adapter/graphql/src/scalars/data_query.rs
```diff
@@ -71,15 +67,6 @@ impl DataQueryResult {
...
     }
 
-    pub fn no_schema_yet(format: DataBatchFormat, limit: u64) -> DataQueryResult {
-        DataQueryResult::Success(DataQueryResultSuccess {
-            schema: None,
-            data: DataBatch::empty(format),
-            datasets: None,
-            limit,
-        })
-    }
```
s373r, line 81: It would be great to add a note about this in CHANGELOG

src/adapter/graphql/tests/tests/test_gql_dataset_adapter_collection.rs
```diff
@@ -0,0 +1,954 @@
...
+
+        let (_catalog_anonymous, catalog_authorized) = authentication_catalogs(&base_catalog).await;
+
+        // TODO: Yuck
+        let catalog_authorized = catalog_authorized
+            .builder_chained()
+            .add_value(kamu_adapter_http::AccessToken {
+                token: "<secret>".into(),
+            })
+            .build();
```
s373r, line 687: Q: do we need a token here since `Tracing` has been added to `schema_quiet`?

src/adapter/graphql/tests/tests/test_gql_dataset_adapter_viersioned_file.rs
```diff
@@ -0,0 +1,941 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+                "#
+            ))
+            .variables(async_graphql::Variables::from_json(json!({
+                "datasetAlias": "x",
+            }))),
```
s373r, line 58: Not a comment:

like this approach with `.variables'!

src/app/cli/src/commands/system_api_server_run_command.rs
```diff
@@ -155,13 +154,17 @@ impl Command for APIServerRunCommand {
                 s(format!("http://{}", api_server.local_addr())).bold(),
             );
             eprintln!("{}", s("Use Ctrl+C to stop the server").yellow());
+        }
 
-            if self.get_token {
+        if self.get_token {
+            if self.output_config.is_tty && self.output_config.verbosity_level == 0 {
```
s373r, line 160: > self.output_config.is_tty && self.output_config.verbosity_level == 0

Minor: let's make this a separate method

PS. I don't fully understand why we don't use `self.output_config.verbosity_level > 0` instead.

src/app/cli/src/commands/system_api_server_run_command.rs
```diff
@@ -155,13 +154,17 @@ impl Command for APIServerRunCommand {
...
                 eprintln!(
                     "{} {}",
-                    s("JWT token:").green().bold(),
+                    s("Access token:").green().bold(),
                     s(access_token).dim()
                 );
+            } else {
+                tracing::warn!("Access token: {access_token}");
```
s373r, line 167: info?

src/domain/core/src/services/ingest/push_ingest_executor.rs
```diff
@@ -154,13 +154,16 @@ pub enum PushIngestError {
 
 pub enum DataSource {
     Url(url::Url),
+    Buffer(bytes::Bytes),
     Stream(Box<dyn AsyncRead + Send + Unpin>),
 }
 
+// TODO: Should this be Debug trait?
```
s373r, line 161: It looks like yes. Only in the case of `url` we should take into account the fact that `Debug` outputs the whole (huge) structure there, not the value itself

src/infra/core/src/services/ingest/push_ingest_executor_impl.rs
```diff
@@ -206,6 +206,13 @@ impl PushIngestExecutorImpl {
                     _ => Err(format!("Unsupported source: {url}").int_err().into()),
                 }
             }
+            DataSource::Buffer(buf) => {
+                // Save buffer to a file so datafusion can read it
+                // TODO: We likely can avoid this and build a DataFrame directly
+                tracing::info!(path = ?temp_path, "Copying buffer into a temp file");
+                tokio::fs::write(&temp_path, buf).await.int_err()?;
```
s373r, line 213: Q: Please remind me whether `args.operation_dir` will be cleared after a successful `do_ingest()` iteration?

Since `temp_path` is technically not a temporary folder that will delete itself

s373r (commented):
Added minor comments/questions

src/adapter/graphql/src/queries/datasets/adapters/versioned_file.rs
sergiimk: we calculate `total_count` before sorting and limiting the `df` to fetch current page, so I think it's correct - `total_count` should reflect how many versions there are in total, not just on the current page

src/adapter/graphql/src/scalars/data_query.rs
sergiimk: Done

src/adapter/graphql/tests/tests/test_gql_dataset_adapter_collection.rs
sergiimk: I left a comment to remove this token and the ugly dependency from GQL to HTTP adapter once we move direct upload/download test into e2e layer.

sergiimk: Never mind - I went ahead and moved direct uploads to e2e tests

src/app/cli/src/commands/system_api_server_run_command.rs
sergiimk: We have #318 to improve this mess - for now I stay on the side of consistency with other code.

src/app/cli/src/commands/system_api_server_run_command.rs
sergiimk: I want token message to be clearly visible

src/domain/core/src/services/ingest/push_ingest_executor.rs
sergiimk: Changed to Debug to produce something like:
```
Url("https://url-as-string")
Bytes(...)
Stream(...)
```

src/infra/core/src/services/ingest/push_ingest_executor_impl.rs
sergiimk: Currently we clear it on every run of CLI tool. We don't clear it in node yet and will need to implement a GC for that.

Issue #766, opened 2024-08-13
ReBAC: absorb the `is_admin` attribute from the Accounts domain


s373r:
### Context

Currently, the existing `Account::is_admin` flag, which can only be raised in the case of preconfigured accounts. The `accounts` table stores the flag.

ReBAC allows us to apply flexible, dynamic rules, including to the account.

### Expectations

- [ ] Move the `is_admin` flag to the ReBAC domain
- [ ] Update the existing code, according to the new storage location
- [ ] Write migrations for existing environments

PR #1223, opened 2025-04-25
Move is_admin field to rebac

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/766

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

migrations/sqlite/20250425151059_remove_is_admin_from_account.sql
```diff
@@ -0,0 +1,56 @@
...
+        account_type,
+        avatar_url,
+        registered_at,
+        provider,
+        provider_identity_key
+    FROM accounts;
+
+DROP TABLE accounts;
+ALTER TABLE accounts_new RENAME TO accounts;
+
```
zaychenko-sergei, line 51: Note: this trick will not work, there are several foreign keys pointing to this account.

rmn-boiko: The whole migration is under `PRAGMA foreign_keys = OFF;`

src/adapter/auth-oso-rebac/src/oso_resource_service_impl.rs
```diff
@@ -53,79 +53,21 @@ impl OsoResourceServiceImplStateHolder {
...
 
-pub struct OsoResourceServiceImpl {
-    state_holder: Arc<OsoResourceServiceImplStateHolder>,
+pub struct OsoDatasetResourceServiceImpl {
     dataset_entry_svc: Arc<dyn DatasetEntryService>,
     rebac_service: Arc<dyn RebacService>,
-    account_service: Arc<dyn AccountService>,
 }
 
 #[component(pub)]
```
s373r, line 61: Minor: 

Tip: if you use `#[dill::component]` over the structure declaration, you won't have to write `new()`

migrations/postgres/20250425151059_remove_is_admin_from_account.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO auth_rebac_properties (entity_type, entity_id, property_name, property_value)
+SELECT 
+  'account' AS entity_type,
+  id::VARCHAR AS entity_id,
+  'is_admin' AS property_name,
+  is_admin::TEXT AS property_value
+FROM accounts;
+
+ALTER TABLE accounts
+DROP COLUMN is_admin;
```
s373r, line 10: I think since we are introducing the `can_provision_accounts` property as part of this PR, we should also add its creation to the migration.

PS. Add the end new line in the migration files :)

s373r (changes requested):

src/adapter/graphql/Cargo.toml
```diff
@@ -42,6 +42,7 @@ kamu-core = { workspace = true }
 kamu-datasets = { workspace = true }
 kamu-flow-system = { workspace = true }
 kamu-flow-system-services = { workspace = true }
+kamu-adapter-auth-oso-rebac = { workspace = true }
```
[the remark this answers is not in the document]
rmn-boiko: Totally agree.
I moved logic and removed `OsoAccountResourceServiceImpl` entity

migrations/postgres/20250425151059_remove_is_admin_from_account.sql
rmn-boiko: I removed `can_provision_accounts` property adding to not mix logic in 1 PR

migrations/sqlite/20250425151059_remove_is_admin_from_account.sql
zaychenko-sergei: Since we are talking about Kamu CLI use case mostly now, it might be easier to force workspace migration to drop the database. It will be re-indexed automatically. For CLI use cases, everything is there, no external data is strictly necessary.

But at least, make sure that existing CLI workspaces don't crash with this migration.

src/adapter/graphql/src/guards.rs
```diff
@@ -61,10 +62,21 @@ impl AdminGuard {
...
-            CurrentAccountSubject::Logged(a) if a.is_admin => Ok(()),
+            CurrentAccountSubject::Logged(a) => {
+                if rebac_service
+                    .get_account_properties(&a.account_id)
+                    .await
+                    .int_err()?
+                    .is_admin
+                {
+                    return Ok(());
+                }
```
zaychenko-sergei, line 77: For me as the developer-user, this is too complex API. I want something simpler, it will be a frequent need.

zaychenko-sergei: You are copying this code over and over in your own PR.

src/adapter/graphql/tests/tests/test_gql_data.rs
```diff
@@ -66,6 +73,14 @@ async fn create_catalog_with_local_workspace(
...
             .add::<ObjectStoreBuilderLocalFs>()
             .add::<LoginPasswordAuthProvider>()
+            .add::<RebacServiceImpl>()
+            .add::<RebacServiceImplCacheState>()
+            .add::<InMemoryRebacRepository>()
+            .add_value(DefaultAccountProperties { is_admin: false })
+            .add_value(DefaultDatasetProperties {
+                allows_anonymous_read: false,
+                allows_public_read: false,
+            })
```
zaychenko-sergei, line 83: Can we have these defined as shared constants somewhere? I think they repeated as well in a few suites, including the old one about datasets.

src/domain/accounts/services/Cargo.toml
```diff
@@ -47,7 +48,9 @@ uuid = { version = "1", default-features = false }
 
 [dev-dependencies]
 kamu-accounts = { workspace = true, default-features = false, features = ["testing"] }
+kamu-auth-rebac-services = { workspace = true }
```
zaychenko-sergei, line 51: Even though these are dev-dependencies, please double check we didn't create a loop here.

src/domain/auth-rebac/domain/src/entities/property.rs
```diff
@@ -152,8 +162,6 @@ impl From<DatasetPropertyName> for PropertyName {
 )]
 #[strum(serialize_all = "snake_case")]
 pub enum AccountPropertyName {
-    // TODO: Private Datasets: absorb the `is_admin` attribute from the Accounts domain
-    //       https://github.com/kamu-data/kamu-cli/issues/766
     IsAnAdmin,
```
zaychenko-sergei, line 165: IsAnAdmin => IsAdmin

src/adapter/graphql/src/guards.rs
rmn-boiko: Created Ext trait for rebac_service and encapsulated logic here

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -48,7 +22,29 @@ pub type DefaultDatasetProperties = DatasetProperties;
...
+        Self {
+            state: RwLock::new(State::default()),
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 pub struct RebacServiceImpl {
+    cache_state: Arc<RebacServiceImplCacheState>,
```
s373r, line 47:
```rust
// from
cache_state: Arc<RebacServiceImplCacheState>
// to
state: Arc<RwLock<State>>,
```

Having an entire `dill` component to store state that will not be used outside of `RebacServiceImpl` is redundant -- please store the state inside the service

PS. re `pub struct State`: in our case, we don't need a `pub` modifier for `State`, since we don't plan to use this structure outside of this module. If we were planning to do so, we would need to name it not such common, e.g., `RebacServiceImplState` (but that's not our current case).

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -10,36 +10,10 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use dill::{component, interface};
+use dill::{component, interface, *};
```
s373r, line 13: Minor/optional: we recently agreed to use `dill` without imports

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -87,7 +85,21 @@ impl RebacService for RebacServiceImpl {
...
+        if let Some(account_properties_cache) = writable_state
+            .account_properties_cache_map
+            .get_mut(account_id.as_str())
+        {
+            match property_name {
+                AccountPropertyName::IsAdmin => {
+                    account_properties_cache.is_admin = property_value.is_true();
+                }
+            }
+        };
```
s373r, line 100: Unfortunately, we can never retrieve anything from the cache since we don't write anything to it.

In this case, we would be well served by the entry API from `HashMap<K, V>. Like here, for example:

```rust
    // InMemoryRebacRepository
    async fn set_entity_property(
        &self,
        entity: &Entity,
        property_name: PropertyName,
        property_value: &PropertyValue,
    ) -> Result<(), SetEntityPropertyError> {
        let mut writable_state = self.state.write().await;

        let entity_properties = writable_state
            .entities_properties_map
            .entry(entity.clone().into_owned())
            .or_insert_with(EntityProperties::new);

        entity_properties.insert(property_name, property_value.clone().into_owned().into());

        Ok(())
    }
```

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -103,14 +115,42 @@ impl RebacService for RebacServiceImpl {
...
+        if let Some(account_properties_cache) = writable_state
+            .account_properties_cache_map
+            .get_mut(account_id.as_str())
+        {
+            match property_name {
+                AccountPropertyName::IsAdmin => {
+                    account_properties_cache.is_admin = false;
+                }
+            }
+        };
```
s373r, line 130: Since, we currently have only one property, we need to remove `account_properties_cache` pair from the state (read from the map) completely

src/adapter/auth-oso-rebac/src/oso_resource_service_impl.rs
s373r: No changes

src/adapter/auth-oso-rebac/src/schema.polar
```diff
@@ -3,13 +3,13 @@
...
-    # │ anonymous (reserved) │ bool   │
-    # │ is_admin             │ bool   │
-    # └──────────────────────┴────────┘
+    # ┌────────────────────────┬────────┐
+    # │        Field           │  Type  │
+    # ├────────────────────────┼────────┤
+    # │ account_id             │ String │
+    # │ anonymous (reserved)   │ bool   │
+    # │ is_admin               │ bool   │
+    # └────────────────────────┴────────┘
```
s373r, line 12: Please revert unnecessary changes

s373r (changes requested):

src/domain/auth-rebac/services/src/rebac_service_impl.rs
rmn-boiko: Agree, changed

src/adapter/auth-oso-rebac/src/oso_resource_service_impl.rs
```diff
@@ -84,48 +42,18 @@ impl OsoResourceServiceImpl {
...
-            // let account_properties = self
-            //     .rebac_service
-            //     .get_account_properties(&account.id)
-            //     .await
-            //     .int_err()?;
-
-            UserActor::logged(&account.id, account.is_admin)
+        let account = match self.account_service.get_account_by_id(account_id).await {
+            Ok(found_account) => found_account,
+            Err(e) => return Err(e.into()),
```
s373r, line 47: Previously, we requested `account` to read `account.is_admin` from it. 

At this point, it looks like we can remove the `account_service.get_account_by_id()` call at all.

In that case, `GetUserActorError::NotFound` will no longer be needed either.

src/domain/auth-rebac/services/src/rebac_service_impl.rs
```diff
@@ -87,7 +69,18 @@ impl RebacService for RebacServiceImpl {
...
             .set_entity_property(&account_entity, property_name.into(), property_value)
-            .await
+            .await?;
+
+        // Update the cache state
+        let mut writable_state = self.cache_state.write().await;
+        let account_properties = writable_state
+            .account_properties_cache_map
+            .entry(account_id.to_string())
+            .or_insert_with(AccountProperties::default);
```
s373r, line 79: Use `self.default_account_properties`

s373r (approved):

zaychenko-sergei (approved):

PR #1226, opened 2025-05-01
Account creation api

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-node/issues/224

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/app/cli/src/app.rs
```diff
@@ -822,6 +823,24 @@ pub fn register_config_in_catalog(
     }
     //
 
+    // Did secret key encryption configuration
+    // TODO: decide should we panic if the key is not set
+    let did_encryption_config = config
```
rmn-boiko, line 828: Extremely important!!!
Accrding to requirements we are storing encrypted did secret keys in DB for both `dataset and account` entities. Logic requires Encryption key. But now We are pushing users to always set this key for kamu-cli and kamu-node. usage. 
I want your thoughts on whether we should generate some encryption key automatically and store it for the user? Or we will force users to set it always.

Another my thought was to move storing did keys to the GQL level which will make possible to use this value optional for cli usage

resources/schema.gql
```diff
@@ -134,6 +139,8 @@ type AccountMut {
 	Update account email
 	"""
 	updateEmail(newEmail: String!): UpdateEmailResult!
+	createAccount(accountName: String!, email: String): CreateAccountResult!
+	modifyPassword(accountName: AccountName!, password: String!): ModifyPasswordResult!
```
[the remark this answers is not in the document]
sergiimk: Is `modifyPassword` an admin-only endpoint?
Please add docstrings to all new endpoints to explain their purpose.

src/adapter/graphql/src/mutations/account_mut.rs
```diff
@@ -57,6 +70,81 @@ impl AccountMut {
...
+        let Ok(account_name) = odf::AccountName::from_str(&account_name) else {
+            return Ok(CreateAccountResult::InvalidAccountName(
+                AccountNameInvalid::default(),
+            ));
+        };
+
+        let new_email = email.unwrap_or(
+            if let Some(parent_account_parts) = self.account.email.to_string().split_once('@') {
+                format!(
+                    "{}_{}@{}",
```
sergiimk, line 91: Why use underscore, which most likely will produce an invalid email, when we could use `+` to direct all child account notifications to the parent?

src/app/cli/src/services/config/models.rs
```diff
@@ -75,6 +76,9 @@ pub struct CLIConfig {
     /// Uploads configuration
     #[merge(strategy = merge_recursive)]
     pub uploads: Option<UploadsConfig>,
+
+    /// Did secret key encryption configuration
+    pub did_encryption: Option<DidSecretEncryptionConfig>,
```
sergiimk, line 81: I'd like us to have a more hierarchical configuration, grouping options into a few big sections.

In the node config we have `auth` section, where I think DID secrets encryption would fit. Perhaps let's add such a section into CLI and nest this configuration under it?

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
```diff
@@ -98,6 +99,8 @@ pub struct AccountConfig {
     #[serde(default)]
     pub is_admin: bool,
     #[serde(default)]
+    pub can_provision_accounts: bool,
```
sergiimk, line 102: Not necessarily now, but I think it would make more sense for us to model these not as `is_x` flags but rather a list of `properties: [admin, can_provision_accounts]` and later relations. Otherwise we'll:
- get tired extending this type with all possible properties
- make system rigid - we want to allow for all kinds of custom ReBAC properties here

src/domain/datasets/domain/src/repos/dataset_did_secret_key_repository.rs
```diff
@@ -0,0 +1,68 @@
...
+// by the Apache License, Version 2.0.
+
+use crypto_utils::DidSecretKey;
+use internal_error::InternalError;
+use thiserror::Error;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait DatasetDidSecretKeyRepository: Send + Sync {
```
sergiimk, line 17: I'm second-guessing whether it makes sense to separate DID secret storage between accounts and datasets. Who knows what other entities we will be generating DIDs for - maybe it's better to store them all in one table. Ideally we should have a mechanism to tell by DID whether it belongs to a dataset, account or some other type. We somewhat have this with `did:odf` vs `did:key` but we could start embedding entity types into ODF DIDs too.

What do you think @rmn-boiko @zaychenko-sergei ?

src/utils/crypto-utils/src/entities/did_secret_encryption_config.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[skip_serializing_none]
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct DidSecretEncryptionConfig {
+    /// The encryption key must be a 32-character alphanumeric string, which
+    /// includes both uppercase and lowercase Latin letters (A-Z, a-z) and
+    /// digits (0-9).
+    ///
+    /// # Example
```
sergiimk, line 27: Please add a shell command that generates such a string - this is very helpful when deploying.

resources/schema.gql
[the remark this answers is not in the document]
rmn-boiko: Yeah, it's under `AdminGuard`
Added docstring

src/adapter/graphql/src/mutations/account_mut.rs
rmn-boiko: Both `_` and `+` are allowed symbols and will be valid emails. But I used an underscore by mistake. Changed to `+`

src/domain/datasets/domain/src/repos/dataset_did_secret_key_repository.rs
rmn-boiko: Yeah, I would love to have single place for all DID's. 
But want to hear your thoughts on how you can see entities in DID? You think about `did:account:<id>` and `did:dataset:<id>`? Or something more general?

src/utils/crypto-utils/src/entities/did_secret_encryption_config.rs
rmn-boiko: Added

src/domain/datasets/domain/src/repos/dataset_did_secret_key_repository.rs
sergiimk: - We used the most simple form `did:key:<public key>` for accounts.
- In DimaP's wallet auth work it looks like we will adopt `did:ethr:<public key hash>` or `did:pkh` for identifying EVM accounts.
- We still only use `did:odf:<public key>` for datasets.

So we currently can non-ambiguously understand what key is for what entity.

If we need more entity types we could extend to `did:odf:<something>:<key>` or use first few bytes in key part to encode the entity type and other extra information (similarly how [IPFS CID](https://docs.ipfs.tech/concepts/content-addressing/#how-cids-are-created) is structured).

src/adapter/graphql/src/mutations/account_mut.rs
sergiimk: Yeah sorry, by "invalid" I actually meant "email that doesn't exist"

resources/schema.gql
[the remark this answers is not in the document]
s373r: Optional: hide admin methods from GQL schema

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
```diff
@@ -131,6 +131,17 @@ impl AccountConfig {
...
     }
 
+    pub fn set_properties(mut self, properties: Vec<AccountPropertyName>) -> Self {
+        for property in properties {
+            let existing_properties = self.properties.get_or_insert_with(Vec::new);
+
+            if !existing_properties.contains(&property) {
+                existing_properties.push(property);
+            }
+        }
```
s373r, line 141: Q: can we just do an assignment based on the account name?

```rust
self.properties = properties;
```

src/domain/accounts/services/src/services/account_service_impl.rs
```diff
@@ -129,6 +161,63 @@ impl AccountService for AccountServiceImpl {
         self.account_repo
             .search_accounts_by_name_pattern(name_pattern, filters, pagination)
     }
+
+    async fn create_account(
+        &self,
+        account_name: &odf::AccountName,
+        email: email_utils::Email,
+        password: String,
```
s373r, line 169: Q: maybe allocate a new type (`nutype`) that will not allow empty password and length less than 8?

src/domain/accounts/services/src/services/predefined_accounts_registrator.rs
```diff
@@ -56,9 +56,13 @@ impl PredefinedAccountsRegistrator {
...
+            return Ok(());
+        };
+
+        for name in account_properties {
             self.rebac_service
-                .set_account_property(account_id, name, &value)
+                .set_account_property(account_id, *name, &boolean_property_value(true))
                 .await
                 .int_err()?;
         }
```
s373r, line 68: I think this solution may be fragile, and it is better to explicitly set the values. We have default values in case something is not explicitly specified

src/infra/accounts/repo-tests/src/accounts_repository_test_utils.rs
```diff
@@ -21,9 +21,10 @@ pub(crate) fn make_test_account(
     email: &str,
     provider: &str,
     provider_identity_key: &str,
+    account_id_maybe: Option<odf::AccountID>,
```
s373r, line 24: Q: why do we need this parameter if generating from the name always gives a stable result?

src/infra/accounts/repo-tests/src/password_hash_repository_test_suite.rs
```diff
@@ -76,3 +96,72 @@ pub async fn test_store_couple_account_passwords(catalog: &Catalog) {
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub async fn test_modify_password(catalog: &Catalog) {
```
s373r, line 100: Please add a test with an attempt to set an empty password

s373r: PS. The same applies to the password hash

src/utils/crypto-utils/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[lib]
+doctest = false
+
+
+[dependencies]
+aes-gcm = { version = "0.10.3" }
+argon2 = { version = "0.5" }
+internal-error = { workspace = true }
+odf = { workspace = true }
```
s373r, line 28: Please extract to own group

src/utils/crypto-utils/src/aes_gcm.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Debug)]
+struct AesGcmError(aes_gcm::Error);
+
+impl std::fmt::Display for AesGcmError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "AES-GCM error")
+    }
+}
+
+impl std::error::Error for AesGcmError {}
```
s373r, line 69: Q: Why not just `thiserror::Error`?

src/utils/crypto-utils/src/argon2_hash.rs
```diff
@@ -0,0 +1,71 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use aes_gcm::aead::OsRng;
+use argon2::PasswordHash;
+use password_hash::{PasswordHasher, PasswordVerifier, SaltString};
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub fn get_argon2_hash(password: &str, hashing_mode: PasswordHashingMode) -> String {
```
s373r, line 16: Optional: group by struct (e.g., `Argon2Hasher`) with static methods (`hash()`, `verify()`)

src/utils/crypto-utils/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl From<ParseEncryptionKey> for EncryptionError {
+    fn from(value: ParseEncryptionKey) -> Self {
+        match value {
+            ParseEncryptionKey::InvalidEncryptionKeyLength => Self::InvalidEncryptionKey,
+            ParseEncryptionKey::InternalError(err) => Self::InternalError(err),
+        }
+    }
+}
```
s373r, line 56: Please extract to own file

src/utils/random-strings/src/common.rs
```diff
@@ -0,0 +1,34 @@
...
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use rand::distributions::{Alphanumeric, Uniform};
+use rand::prelude::Distribution;
+
+pub fn get_random_string(
```
s373r, line 13: I would like to suggest splitting this method into two:
- `get_random_name()` -- keep the old 
- `get_random_string()` -- add a new

`get_random_name()`, based on where it is used, was intended to return readable names (as `Docker` does, for example)

s373r (commented):
Added comments/questions

src/infra/accounts/repo-tests/src/accounts_repository_test_utils.rs
rmn-boiko: In test we always were using `new_seeded_ed25519` method for AccountID which always return static value. But to test new logic with pregenerated AccountID with randomly generated Private part of did.

src/utils/random-strings/src/common.rs
rmn-boiko: Do we really want to duplicate 2 pretty similar methods. 
New method `get_random_string()` with `Alphanumeric` param works similar to old one

src/utils/crypto-utils/src/aes_gcm.rs
rmn-boiko: Yeah, you're right. Changed

migrations/postgres/20250426151069_did_secret_keys_table.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE dataset_did_secret_keys(
```
zaychenko-sergei, line 1: All our migrations have separating comments. Please follow the format of other migrations.

migrations/sqlite/20240701092841_dataset_env_vars_table.sql
```diff
@@ -4,7 +4,7 @@ CREATE TABLE dataset_env_vars(
     id VARCHAR(36) NOT NULL PRIMARY KEY,
     key VARCHAR(200) NOT NULL,
     value BLOB NOT NULL,
-    secret_nonce BLOKB,
+    secret_nonce BLOB,
```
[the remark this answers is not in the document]
zaychenko-sergei: Yes, you can't modify existing migrations. You need to create a migration with new/old  column mapping, as changing the type will unlikely be possible directly in SQlite.

migrations/postgres/20250426151069_did_secret_keys_table.sql
```diff
@@ -0,0 +1,13 @@
...
+   secret_key BYTEA NOT NULL,
+   owner_id VARCHAR(100) NOT NULL REFERENCES accounts(id) ON DELETE CASCADE
+);
+
+CREATE TABLE account_did_secret_keys(
+   account_id VARCHAR(100) NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
+   secret_nonce BYTEA NOT NULL,
+   secret_key BYTEA NOT NULL,
+   owner_id VARCHAR(100) NOT NULL REFERENCES accounts(id) ON DELETE CASCADE
+);
```
zaychenko-sergei, line 13: What is the difference between `owner_id` and `account_id` in this context?

migrations/postgres/20250426151069_did_secret_keys_table.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE dataset_did_secret_keys(
+   dataset_id VARCHAR(100) NOT NULL REFERENCES dataset_entries(dataset_id) ON DELETE CASCADE,
+   secret_nonce BYTEA NOT NULL,
+   secret_key BYTEA NOT NULL,
+   owner_id VARCHAR(100) NOT NULL REFERENCES accounts(id) ON DELETE CASCADE
```
zaychenko-sergei, line 5: Why do we need `owner_id` here? We have `owner_id` in `DatasetEntry` already. Do you see any reason for denormalization from the beginning?

src/adapter/graphql/src/mutations/account_mut.rs
```diff
@@ -57,6 +70,81 @@ impl AccountMut {
...
+                    field: err.account_field.to_string(),
+                }),
+            ),
+            Err(CreateAccountError::Internal(e)) => Err(e.into()),
+        }
+    }
+
+    /// Reset password for selected account. Allowed only for admin users
+    #[graphql(guard = "AdminGuard::new()")]
+    async fn modify_password(
```
zaychenko-sergei, line 127: Why the permission check is different from `create_account`? I don't think we are going to give Molecule an admin account for the whole system.

migrations/postgres/20250426151069_did_secret_keys_table.sql
rmn-boiko: Very soon we will add possibility to create datasets in one account from another. And to have here entity who creates it will be very useful.

src/adapter/graphql/src/utils.rs
```diff
@@ -223,6 +224,27 @@ pub(crate) fn logged_account(ctx: &Context<'_>) -> bool {
...
+    let rebac_service = from_catalog_n!(ctx, dyn RebacService);
+
+    let account_properties = rebac_service
+        .get_account_properties(account_id)
+        .await
+        .int_err()?;
+
+    if account_properties.is_admin || account_properties.can_provision_accounts {
+        return Ok(());
+    }
```
zaychenko-sergei, line 240: Anti-pattern "feature envy" here.
This business logic should be encapsulated somewhere else in a domain-specific service rather than in GQL adapter.

migrations/postgres/20250426151069_did_secret_keys_table.sql
rmn-boiko: `owner_id` is Id of account who created this account, and `account_id` is Id of created account

src/adapter/graphql/src/mutations/account_mut.rs
```diff
@@ -57,6 +70,81 @@ impl AccountMut {
...
+            Ok(created_account) => Ok(CreateAccountResult::Success(CreateAccountSuccess {
+                account: AccountView::from_account(created_account),
+            })),
+            Err(CreateAccountError::Duplicate(err)) => Ok(
+                CreateAccountResult::NonUniqueAccountField(AccountFieldNonUnique {
+                    field: err.account_field.to_string(),
+                }),
+            ),
+            Err(CreateAccountError::Internal(e)) => Err(e.into()),
+        }
```
zaychenko-sergei, line 122: Think of extracting a use case in `accounts` domain. This has nothing to do with GQL in particular.

src/adapter/graphql/src/mutations/account_mut.rs
```diff
@@ -57,6 +70,81 @@ impl AccountMut {
...
+            .await
+        {
+            Ok(_) => Ok(ModifyPasswordResult::Success(
+                ModifyPasswordSuccess::default(),
+            )),
+            Err(ModifyPasswordError::AccountNotFound(_)) => Ok(
+                ModifyPasswordResult::AccountNotFound(AccountNotFound::default()),
+            ),
+            Err(ModifyPasswordError::Internal(e)) => Err(e.into()),
+        }
```
zaychenko-sergei, line 145: A bit smaller, but still smells like use case

src/domain/datasets/services/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
```diff
@@ -86,6 +115,16 @@ impl CreateDatasetFromSnapshotUseCase for CreateDatasetFromSnapshotUseCaseImpl {
...
 
+        if let Some(did_secret_encryption_key) = &self.did_secret_encryption_key {
+            let dataset_did_secret_key =
+                DidSecretKey::try_new(&dataset_did.1, did_secret_encryption_key.expose_secret())
+                    .int_err()?;
+            self.dataset_did_secret_key_repo
+                .save_did_secret_key(&dataset_did.0, logged_account_id, &dataset_did_secret_key)
+                .await
+                .int_err()?;
+        }
```
zaychenko-sergei, line 126: Don't we need the same functionality in "create dataset" w/o snapshot use case? I think it should be placed in the shared create helper that we have, and invoked from both use cases.

src/infra/accounts/repo-tests/src/password_hash_repository_test_suite.rs
zaychenko-sergei: The value type you suggested (nutype-based) would handle those better.

src/adapter/graphql/src/utils.rs
rmn-boiko: I would disagree with that. This is purely responsibility of Oso rebac for gql methods. I don't see this logic on domain level. 
Probably confusing thing that I call directly `rebac_service` instead Oso. The reason I did it that recently we migrate caching level to the `rebac_service` level. I can agree to change the calling to Oso service but can't see any other place for this logic

src/domain/datasets/services/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
rmn-boiko: It is a good and tricky question. Goal of epic to store all Did keys created via node. But `create_dataset_use_case` is used only in transfer protocol. And seems we don't need to store dids for pushed/pulled dataset

src/adapter/graphql/src/mutations/account_mut.rs
rmn-boiko: Agree, moved.

migrations/postgres/20250426151069_did_secret_keys_table.sql
zaychenko-sergei: Then `owner_id` => `creator_id`.

src/adapter/graphql/src/mutations/account_mut.rs
zaychenko-sergei: Who wil be modifying the passwords, us for all customers?

src/domain/datasets/services/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
zaychenko-sergei: That is weird. We allow creating datasets in UI without specifying any snapshot.

src/adapter/graphql/src/utils.rs
zaychenko-sergei: I expect something like "account facade", so that I can simply ask:

```
let account_facade = from_catalog_n!(ctx, dyn RebacAccountFacade);
account_facade.can_provision_accounts(account_id).await
```

While checking for admins or knowing this data exists in account properties - is a ReBAC domain implementation details.

src/domain/datasets/services/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
rmn-boiko: YEah, but under the hood we still calling `create_from_snapshot_impl` with empty snapshot

examples/archive/commercial-fishing/.kamuconfig
```diff
@@ -4,7 +4,7 @@ content:
   users:
     predefined:
       - accountName: kamu
-        isAdmin: true
+        properties: [isAdmin]
```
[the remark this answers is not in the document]
rmn-boiko: Yeah, I wanted to do it but a bit scared of additional migration. But agree it would fit better

src/domain/accounts/domain/src/entities/password.rs
```diff
@@ -0,0 +1,55 @@
...
+// by the Apache License, Version 2.0.
+
+use nutype::nutype;
+use thiserror::Error;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[nutype(
+    sanitize(trim),
+    validate(with = validate_password, error = PasswordValidationError),
```
zaychenko-sergei, line 17: There is an existing length validation in `nutype`: `len_char_min`. The non-ascii part could be expressed via Regex, but custom is fine there too.

src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
```diff
@@ -0,0 +1,62 @@
...
+                format!(
+                    "{}+{}@{}",
+                    parent_account_parts.0, account_name, parent_account_parts.1
+                )
+            } else {
+                format!("{account_name}@example.com")
+            };
+
+            Email::parse(&email_str).int_err()?
+        });
```
zaychenko-sergei, line 46: Let's encapsulate the validation inside `Email` type.

src/domain/datasets/services/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
```diff
@@ -19,20 +19,55 @@ use kamu_datasets::{
...
+    pub fn new(
+        current_account_subject: Arc<CurrentAccountSubject>,
+        system_time_source: Arc<dyn SystemTimeSource>,
+        did_generator: Arc<dyn DidGenerator>,
+        dataset_registry: Arc<dyn DatasetRegistry>,
+        create_helper: Arc<CreateDatasetUseCaseHelper>,
+        did_secret_encryption_config: Arc<DidSecretEncryptionConfig>,
+        did_secret_key_repo: Arc<dyn DidSecretKeyRepository>,
+    ) -> Self {
+        Self {
```
zaychenko-sergei, line 58: I still don't understand why we are extending "create from snapshot" and not extending "create" similarly.

migrations/postgres/20250426151069_did_secret_keys_table.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TYPE did_entity_type AS ENUM ('account', 'dataset');
+
+/* ------------------------------ */
+
+CREATE TABLE did_secret_keys(
+   entity_type  did_entity_type NOT NULL,
+   entity_id    VARCHAR(100) NOT NULL,
+   secret_nonce BYTEA NOT NULL,
+   secret_key   BYTEA NOT NULL,
+   creator_id   VARCHAR(100) NOT NULL REFERENCES accounts(id) ON DELETE CASCADE
```
zaychenko-sergei, line 12: Doubts here:
- the concepts of "Account" and "DidSecret" feel closely related (it feels like account exclusively owns 0..N secrets), so the cascade delete looks adequate
- it feels mismatching to put them in separate DDD domains, as separate domains should not use something like cascade delete and should be losely connected via outbox messages.

My gut feeling is that there is no motivation to place "secrets" in separate domain, it naturally belongs to "accounts" domain. If we are keeping these separate for some motivating reason, then we should not have this database-level coupling.

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
s373r: Yes, I confused you by mentioning the account name -- I apologize

I wanted to emphasize that if the method is called `set_properties()`, the expected behavior is setter behavior, and in this case we have an update.

And it's not quite clear to me how to delete properties, for example -- hence the suggestion to make a more explicit (and simple) API

examples/archive/commercial-fishing/.kamuconfig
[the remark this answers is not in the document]
s373r: It's a little unclear to me -- there's a reaction, but no change

src/domain/datasets/domain/src/repos/dataset_did_secret_key_repository.rs
s373r: > In DimaP's wallet auth work it looks like we will adopt did:ethr:<public key hash> or did:pkh for identifying EVM accounts.

Cross-posting: `did:pkh` has been selected

src/utils/crypto-utils/src/aes_gcm.rs
s373r: `Display` implementation can also be hidden via `#[error("AES-GCM error")`

src/utils/random-strings/src/common.rs
s373r: These methods, different semantically. For example, here are the random names that Docker generates:
image

That is, they are readable and can be easily reproduced.

In my opinion, we can keep `get_random_name()`, which will just do the `get_random_string()` call internally, so as not to change all the usage locations -- and then replace the `get_random_name()` implementation with something more appropriate whenever we want it

src/utils/crypto-utils/src/aes_gcm.rs
rmn-boiko: Changed, thanks

src/domain/accounts/services/src/services/predefined_accounts_registrator.rs
```diff
@@ -55,8 +55,26 @@ impl PredefinedAccountsRegistrator {
         account_id: &AccountID,
         account_config: &AccountConfig,
     ) -> Result<(), InternalError> {
+        let default_account_properties = vec![
+            AccountPropertyName::is_admin(false),
+            AccountPropertyName::can_provision_accounts(false),
+        ];
```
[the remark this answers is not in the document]
rmn-boiko: Used this value

migrations/postgres/20250426151069_did_secret_keys_table.sql
rmn-boiko: Changed

src/domain/accounts/domain/src/entities/password.rs
rmn-boiko: Yeah, just don't want to use here regex, and use single custom validation

src/utils/random-strings/src/common.rs
rmn-boiko: Agree with that, Changed

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
rmn-boiko: Such as it is mainly used in testing I changed to rewrite properties

src/domain/datasets/services/src/use_cases/create_dataset_from_snapshot_use_case_impl.rs
rmn-boiko: `CreateDatasetUseCase` is used only in Smtp protocol. And during this flow we already expect generated seed block with dataset Did without its private part. So we can't implement this logic here and moreover in current situation we don't have to

src/adapter/graphql/src/mutations/account_mut.rs
rmn-boiko: I'm not sure 1 line deserve to be a separate usecase

zaychenko-sergei (approved):

PR #1251, opened 2025-05-26
Add panic tracing message

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-node/issues/241

Log example: 
Screenshot from 2025-05-27 14-15-54

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

s373r (commented):
Two questions have been added

src/utils/observability/src/panic_hook.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::panic;
+use std::sync::Arc;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub fn extend_panic_hook() {
+    let default_hook = Arc::new(panic::take_hook());
+
+    panic::set_hook(Box::new(move |info| {
+        default_hook(info);
```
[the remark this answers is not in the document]
rmn-boiko: Yeah, we still print default panic in stdout and in addition we add tracing print for better observability

src/utils/observability/src/panic_hook.rs
```diff
@@ -0,0 +1,45 @@
...
+            .map(|loc| format!("{}:{}:{}", loc.file(), loc.line(), loc.column()))
+            .unwrap_or_else(|| "unknown location".into());
+
+        tracing::error!(
+            target: "panic_hook",
+            message = %msg,
+            location = %location,
+            backtrace = %backtrace,
+            "panic occurred"
+        );
```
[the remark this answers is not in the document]
rmn-boiko: I've added how it looks now

src/utils/observability/src/panic_handler.rs
```diff
@@ -0,0 +1,60 @@
...
+        "Unknown panic payload"
+    };
+
+    let backtrace = BACKTRACE
+        .lock()
+        .unwrap()
+        .take()
+        .unwrap_or("Backtrace not found".to_string());
+
+    tracing::error!(err = "Unhandled panic caught", panic_msg, %backtrace);
```
sergiimk, line 52: The way this is logged is different from how we log other things so (given how sensitive and error prone `tracing` macros are) this makes me question how it looks in the log.

Recommend changing this to the style of all our other tracing events:
```rust
tracing::errror!(panic_msg, %backtrace, "Unhandled panic caught");
```

src/utils/observability/src/panic_handler.rs
```diff
@@ -0,0 +1,60 @@
...
+
+// Such as we can't get real backtrace in catch_unwind block
+// we need place to store trace if such occurs and then print it in logs
+lazy_static! {
+    static ref BACKTRACE: Arc<Mutex<Option<String>>> = Arc::new(Mutex::new(None));
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub fn set_panic_hook() {
```
sergiimk, line 28: Let's give this a better name like `set_hook_capture_panic_backtraces_no_propagate` and add documentation that is meant to be always used in combination with `CatchPanicLayer` and the other handler.

rmn-boiko: Yeah, agree, changed.
And tested for all panic types(GQL, Rest, Smtp, outbox, flightsql)

sergiimk: I'm not clear on how this works for outbox for example?

`CatchPanicLayer::custom(panic_handler)` is the only thing that logs the backtrace, but hook is set for all panics.

e.g. if you panic right after your `set_hook` call I'd expect that program just exists without printing anything.

rmn-boiko: I kept default panic hook which print simple panic message into stdout. 
We can't rid from it to keep printing panics for another cli commands. 
Wdyt about next solution:
1. Move tracing logic into `set_hook_capture_panic_backtraces_no_propagate` method
2. Call this method on the command level(I think about adding some property `needs_catching_panics` for api_sever, sql, ui etc commands)
3. We can keep `catch_panic` layer to return Internal error but remove recent logic where we store backtrace in separate place.

src/utils/observability/src/panic_handler.rs
```diff
@@ -0,0 +1,52 @@
...
+        };
+
+        tracing::error!(error_msg, error_backtrace = %backtrace, "Unhandled panic caught");
+    }));
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[allow(clippy::needless_pass_by_value)]
+pub fn panic_handler(_err: Box<dyn Any + Send + 'static>) -> Response<Body> {
```
sergiimk, line 45: would it be more appropriate to move this into `observability::axum` submodule?

src/utils/observability/src/axum.rs
```diff
@@ -164,3 +170,43 @@ pub async fn unknown_fallback_handler(
     );
     (axum::http::StatusCode::NOT_FOUND, "Not Found")
 }
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub fn set_hook_capture_panic_backtraces_no_propagate(propagate: bool) {
```
sergiimk: suggest renaming to `set_hook_trace_panics` to reflect what it's actually doing

sergiimk: btw I was suggesting to move only `panic_handler` which is axum-specific to this module - panic tracing could've stayed in its own module

PR #1266, opened 2025-06-06
`CreateAccountUseCaseImpl`: added sending `AccountLifecycleMessage`

s373r:
### Fixed
- `CreateAccountUseCaseImpl`: added sending `AccountLifecycleMessage`

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
+++ b/src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
@@ -14,9 +14,11 @@ use email_utils::Email;
 use internal_error::{InternalError, ResultIntoInternal};
 use kamu_accounts::{
     Account,
+    AccountLifecycleMessage,
     AccountService,
     CreateAccountError,
     CreateAccountUseCase,
+    MESSAGE_PRODUCER_KAMU_ACCOUNTS_SERVICE,
     PasswordHashRepository,
 };
 use random_strings::AllowedSymbols;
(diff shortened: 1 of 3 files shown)


src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
```diff
@@ -58,6 +63,21 @@ impl CreateAccountUseCaseImpl {
...
+
+    async fn notify_account_created(&self, new_account: &Account) -> Result<(), InternalError> {
+        use messaging_outbox::OutboxExt;
+
+        self.outbox
+            .post_message(
+                MESSAGE_PRODUCER_KAMU_ACCOUNTS_SERVICE,
+                AccountLifecycleMessage::created(
+                    new_account.id.clone(),
+                    new_account.email.clone(),
```
rmn-boiko, line 75: Q: should we send notification to created account email? Such as sometimes it will not be an existing email maybe better to send it to creator account email?

s373r: @rmn-boiko 

> Q: should we send notification to created account email?

If there is no mail, it will be generated

> maybe better to send it to creator account email

Good point -- but I think what we care about at this point is the fact that the message was sent

zaychenko-sergei (approved):

PR #1267, opened 2025-06-06
Refactor: Add `update_versioned_file_use_case`

rmn-boiko:

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/adapter/graphql/src/mutations/dataset_mut/dataset_mut.rs
```diff
@@ -206,10 +206,9 @@ impl DatasetMut {
     }
 
     /// Downcast a dataset to a versioned file interface
-    async fn as_versioned_file(&self, ctx: &Context<'_>) -> Result<Option<VersionedFileMut>> {
-        Ok(Some(VersionedFileMut::new(
-            self.dataset_request_state.resolved_dataset(ctx).await?,
-        )))
+    #[expect(clippy::unused_async)]
+    async fn as_versioned_file(&self) -> Result<VersionedFileMut> {
+        Ok(VersionedFileMut::new(&self.dataset_request_state))
```
zaychenko-sergei: Why? The dataset may not represent a version filed, it would be an error.

zaychenko-sergei (commented):
1) Use case tests?
2) Consider placing these files in datasets domain instead. It might be another new domain as well, but for now that is not obvious. We are only going to leave general orchestration logic in the core domain, step by step.

src/adapter/graphql/src/mutations/dataset_mut/dataset_mut.rs
rmn-boiko: Yeah, right. Changed

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
```diff
@@ -7,141 +7,91 @@
...
+                let mut digest = sha3::Sha3_256::new();
+                let mut buf = [0u8; 2048];
+
+                loop {
+                    let read = stream.read(&mut buf).await.int_err()?;
+                    if read == 0 {
+                        break;
+                    }
+                    digest.update(&buf[..read]);
+                }
```
zaychenko-sergei, line 73: I think there are many snippets like this over our code base, and you did not invent it. This is pretty-much low-level for an API handler, so I suggest that we make a reusable utility and look where it could replace the buffering hack elsewhere

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
```diff
@@ -7,141 +7,91 @@
...
+                    odf::Multihash::new(odf::metadata::Multicodec::Sha3_256, &digest).unwrap();
+
+                // Get the stream again and copy data from uploads to storage using computed
+                // hash
+                // TODO: PERF: Should we create file in the final storage directly to avoid
+                // copying?
+                let content_stream = upload_service
+                    .upload_token_into_stream(&upload_token.0)
+                    .await
+                    .int_err()?;
```
zaychenko-sergei, line 86: Not sure on this logic, we are de-facto scanning the file twice, if we keep both digest and the stream. Can't we do the digest computation at the same time we are uploading the file? Imagine we sent 200mb of data, it could be a noticeable performance problem.

src/domain/datasets/domain/src/entities/versioned_file.rs
```diff
@@ -0,0 +1,83 @@
...
+use file_utils::{MediaType, MediaTypeRef};
+use odf::Multihash;
+
+pub type FileVersion = u32;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[nutype::nutype(derive(AsRef, Clone, Debug, Into))]
+pub struct ExtraDataFields(serde_json::Value);
+
```
zaychenko-sergei: What did you achieve with this nutype? I don't see any validation or transformation, and the semantic meaning "extra data" is not telling much compared to `some_stuff: serde_json::Value`

src/domain/datasets/domain/src/entities/versioned_file.rs
```diff
@@ -0,0 +1,83 @@
...
+            extra_data,
+        }
+    }
+
+    pub fn to_record_data(self) -> serde_json::Value {
+        let mut record: serde_json::Value = self.extra_data.into();
+        record["version"] = self.version.into();
+        record["content_hash"] = self.content_hash.to_string().into();
+        record["content_length"] = self.content_length.into();
+        record["content_type"] = self.content_type.clone().into();
```
zaychenko-sergei: So, if there is a schema to this "extra data", why don't we have a struct for this instead?

src/domain/datasets/domain/src/use_cases/update_version_file_use_case.rs
```diff
@@ -0,0 +1,100 @@
...
+        expected_head: Option<odf::Multihash>,
+        extra_data: Option<ExtraDataFields>,
+    ) -> Result<UpdateVersionFileResult, UpdateVersionFileUseCaseError>;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct ContentArgs {
+    pub content_length: usize,
+    pub content_hash: odf::Multihash,
```
zaychenko-sergei, line 35: Not an expert in the topic, but maybe we could avoid double scanning of the content, if this hash was optional?

src/domain/datasets/services/src/use_cases/mod.rs
```diff
@@ -15,6 +15,7 @@ mod delete_dataset_use_case_impl;
 mod get_dataset_downstream_dependencies_use_case_impl;
 mod get_dataset_upstream_dependencies_use_case_impl;
 mod rename_dataset_use_case_impl;
+mod update_version_file_use_case_impl;
```
zaychenko-sergei: Let's maybe split these use cases by submodules, so that it will be easier to cut in future. One part for pure dataset use cases, another part for version files, maybe we will have some collection use cases as well soon.

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+    async fn get_latest_version(
+        &self,
+        dataset_ref: &odf::DatasetRef,
+    ) -> Result<(FileVersion, odf::Multihash), InternalError> {
+        // TODO: Consider retractions / corrections
+        let query_res = self
+            .query_svc
+            .tail(dataset_ref, 0, 1, GetDataOptions::default())
+            .await
+            .int_err()?;
```
zaychenko-sergei: I think that needs to be rather an SQL query, if we have to consider corrections and retractions

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        let Some(df) = query_res.df else {
+            return Ok((0, query_res.block_hash));
+        };
+
+        let last_version = df
+            .select_columns(&["version"])
```
zaychenko-sergei: Where did this came from? If we have a schema column names for versioned files, they should be constants.

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+                .remove("content_length")
+                .unwrap_or_default()
+                .as_u64()
+                .unwrap_or_default(),
+        )
+        .unwrap();
+        let content_type =
+            MediaType::from(record.remove("content_type").unwrap().as_str().unwrap());
+        let content_hash = odf::Multihash::from_multibase(
+            record.remove("content_hash").unwrap().as_str().unwrap(),
```
zaychenko-sergei: Same here, many references to implicit schema. What if schema columns are changed .. should the poor maintainer be looking for spread of strings all over the project?

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+#[async_trait::async_trait]
+impl UpdateVersionFileUseCase for UpdateVersionFileUseCaseImpl {
+    #[tracing::instrument(level = "info", name = UpdateVersionFileUseCaseImpl_execute, skip_all, fields(%dataset_handle.id))]
+    async fn execute(
+        &self,
+        dataset_handle: &odf::DatasetHandle,
+        content_args_maybe: Option<ContentArgs>,
+        expected_head: Option<odf::Multihash>,
+        extra_data: Option<ExtraDataFields>,
+    ) -> Result<UpdateVersionFileResult, UpdateVersionFileUseCaseError> {
```
zaychenko-sergei: Where is an authorization check?
If it is still in the API layer, consider moving it here (except for basic authentication).

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+            new_version,
+            content_args.content_hash.clone(),
+            content_args.content_length,
+            content_args.content_type,
+            extra_data,
+        );
+
+        let ingest_result = self
+            .push_ingest_data_use_case
+            .execute(
```
zaychenko-sergei: What is the advantage of calling use case from use case here?

rmn-boiko: Such as it is a part of updating versioned file we want to keep calling push ingest use case inside a new one, to be able correctly reuse it in the future

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
rmn-boiko: I've tried to implement a quick solution. But got failed. Will cut a separate ticket to handle this

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
rmn-boiko: There are 1-2 places with similar logic. But that logic is mixed with writing to file.

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
sergiimk: In-band uploads in GQL are specifically meant for testing with small files, e.g. <10 MiB.

For out-of-band - the file is first uploaded to a temporary storage and then needs to be copied into the dataset under the the name corresponding to the hash. Since there is no "move" operation in S3 - we have to compute the hash before we copy data, so not sure how you could avoid reading data twice.

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
```diff
@@ -393,6 +332,7 @@ pub enum UpdateVersionResult {
     Success(UpdateVersionSuccess),
     CasFailed(UpdateVersionErrorCasFailed),
     InvalidExtraData(UpdateVersionErrorInvalidExtraData),
+    TooLarge(UploadVersionErrorTooLarge),
```
sergiimk: Don't think this error belongs in `UpdateVersionResult`. It is returned from `finish_upload_new_version` - and it would be weird to make users expect `TooLarge` errors at this stage as it should've been already rejected at `start_upload` stage.

What I think we need is a separate error for `upload_new_version` - an in-band case that combines `start_upload` and `finish_upload` into one. This one is not meant for production, so it should be OK to change its error type.

We should warn customers though before deploying.

src/domain/datasets/services/src/use_cases/versioned_files/update_version_file_use_case_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+            last_entity.version += 1;
+
+            last_entity
+        };
+
+        if entity.content_length > self.upload_config.max_file_size_in_bytes() {
+            return Err(UpdateVersionFileUseCaseError::TooLarge(
+                UploadTooLargeError::new(
+                    entity.content_length,
+                    self.upload_config.max_file_size_in_bytes(),
```
sergiimk: I only see one size limit but I think we need two:
- Overall file content limit
- Limit for in-band upload (see my other comments for ideas on this)

src/adapter/graphql/src/mutations/dataset_mut/adapters_mut/versioned_file_mut.rs
sergiimk: As per my other comment, I think we need two limits:
- overall limit for file content size
- limit for in-band upload - to prevent abuse of testing-only API

The latter one is not something we can easily enforce. Yes we can check the length of data in the handler, but by that time **the abuse is already done** - our API got tied up processing absurdly large request.

Perhaps exceeding in-band limits should be handled on the protocol level - by imposing overall limits on the size of the request payload. i.e. so that GQL library will read at most 1MiB from the TCP stream before rejecting the request outright.

Large requests is a general anti-DoS measure which we could skip for now and track in our security review ticket:
https://github.com/kamu-data/planning-internal/issues/62

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
sergiimk: Better yet - I recently re-wrote parts of the code to use `serde` (see `VersionedFileRecord` in `versioned_file_entry.rs`). Instead of constants we could use a helper struct to deserialize the data.

src/domain/datasets/services/src/use_cases/update_version_file_use_case_impl.rs
rmn-boiko: Currently we do not support retractions for versioned files.

PR #1288, opened 2025-07-04
Login required platform mode

rmn-boiko:

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/app/cli/src/explore/web_ui_server.rs
+++ b/src/app/cli/src/explore/web_ui_server.rs
@@ -133,10 +133,6 @@ impl WebUIServer {
             },
         };
 
-        let access_token = Self::acquire_access_token(server_catalog.clone(), &login_instructions)
-            .await
-            .expect("Token not retrieved");
-
         let web_ui_url = Url::parse(&web_ui_url).expect("URL failed to parse");
 
         let default_protocols = Protocols::default();
@@ -150,14 +146,13 @@ impl WebUIServer {
             }))
             .build();
 
+        let access_token = Self::acquire_access_token(web_ui_catalog.clone(), &login_instructions)
+            .await
+            .expect("Token not retrieved");
+
         let mut open_api_router = OpenApiRouter::with_openapi(
             kamu_adapter_http::openapi::spec_builder(crate::app::VERSION, "").build(),
         )
-        .route(
-            "/assets/runtime-config.json",
-            axum::routing::get(runtime_configuration_handler),
-        )
-        .route("/ui-config", axum::routing::get(ui_configuration_handler))
         .route("/graphql", axum::routing::post(graphql_handler))
         .nest(
             "/odata",
@@ -185,6 +180,11 @@ impl WebUIServer {
         }
 
         let (router, api) = open_api_router
+            .route(
+                "/assets/runtime-config.json",
+                axum::routing::get(runtime_configuration_handler),
+            )
+            .route("/ui-config", axum::routing::get(ui_configuration_handler))
             .nest(
                 "/platform",
                 kamu_adapter_http::platform::root_router(allow_anonymous),


src/adapter/flight-sql/Cargo.toml
```diff
@@ -23,6 +23,7 @@ doctest = false
 
 [dependencies]
 kamu-accounts = { workspace = true }
+kamu-accounts-services = { workspace = true, features = ["testing"] }
```
zaychenko-sergei: Bad sign: implementation-level dependency is unexpected here

zaychenko-sergei: If it is only for testing, place it in `[dev-dependencies]`

src/adapter/graphql/src/root.rs
```diff
@@ -39,6 +39,7 @@ impl Query {
     ///
     /// Datasets are historical streams of events recorded under a certain
     /// schema.
+    #[graphql(guard = "AuthPolicyGuard")]
```
zaychenko-sergei: A guard can be set on the struct level. Maybe not a good idea here, but try this in other places to avoid copying guard for each method.

image

src/adapter/http/Cargo.toml
```diff
@@ -36,6 +36,7 @@ http-common = { workspace = true }
 # TODO: Adapters should depend only on kamu-domain crate and be implementation-agnostic
 kamu = { workspace = true }
 kamu-accounts = { workspace = true }
+kamu-accounts-services = { workspace = true }
```
zaychenko-sergei: Again, a bad dependency, unless it's only used in [dev-dependencies] section.

src/adapter/http/src/middleware/auth_policy_layer.rs
```diff
@@ -0,0 +1,93 @@
...
+impl AuthPolicyLayer {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl<Svc> Layer<Svc> for AuthPolicyLayer {
+    type Service = AuthPolicyMiddleware<Svc>;
+
+    fn layer(&self, inner: Svc) -> Self::Service {
```
zaychenko-sergei, line 35: How is this one different from FlightSQL? Can't we just have a common single Axum middleware?

src/domain/accounts/domain/src/entities/auth_config.rs
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+impl Default for AuthConfig {
+    fn default() -> Self {
+        Self {
+            allow_anonymous: Some(false),
```
zaychenko-sergei: I think for CLI we will rather want it to be allowed by default.

src/domain/accounts/services/src/services/auth_policy_service.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use kamu_accounts::CurrentAccountSubject;
+use thiserror::Error;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[async_trait::async_trait]
+pub trait AuthPolicyService: Sync + Send {
+    fn check(&self) -> Result<(), AuthPolicyCheckError>;
+}
```
zaychenko-sergei: The name `check` tells me nothing about the context of what this thing is doing.

src/app/cli/src/app.rs
```diff
@@ -911,6 +911,17 @@ pub fn register_config_in_catalog(
     }
     //
 
+    // Authentication configuration
+    catalog_builder.add_value(config.auth.clone().unwrap());
+    if let Some(auth_config) = config.auth.as_ref()
+        && auth_config.allow_anonymous.unwrap()
+    {
+        catalog_builder.add::<kamu_accounts_services::AllowAnonymousAuthPolicyServiceImpl>();
+    } else {
```
zaychenko-sergei: Can't we just skip dressing the endpoints with the related middleware, when login required mode is off? I thought we are going to have middleware only when the access should be restricted, othewise - simply skip checking anything. Right now you are injecting a middleware and a Null-object style service, which just heats the servers.

src/domain/accounts/services/src/services/auth_policy_service.rs
zaychenko-sergei: The abstract trait should leave in "domain" part of the bounded context, not in the implementation, which are "services".

resources/schema.gql
```diff
@@ -169,6 +170,8 @@ type AccountMut {
 	Access to the mutable flow configurations of this account
 	"""
 	flows: AccountFlowsMut!
+	createAccessToken(tokenName: String!): CreateTokenResult!
+	revokeAccessToken(tokenId: AccessTokenID!): RevokeResult!
```
zaychenko-sergei: Maybe create another API sub-group here and in queries, which is dedicated to tokens management?

src/adapter/http/src/platform/router.rs
```diff
@@ -10,19 +10,31 @@
...
+    router = router
+        .routes(routes!(handlers::file_upload_prepare_post_handler))
+        .routes(routes!(handlers::file_upload_post_handler))
+        .routes(routes!(handlers::file_upload_get_handler));
+
+    if !allow_anonymous {
+        router = router.layer(AuthPolicyLayer::new()); // Mutate router without moving
+    }
+
+    router = router
```
zaychenko-sergei, line 31: Maybe make some comments about the endpoints above requiring auth, and the endpoints below not-requiring it

PR #1300, opened 2025-07-16
observability: make dill dependency optional

sergiimk:
This allows using `observability` crate in other projects that may not want `dill` as a dependency, or may need to compile in stable rust that `dill` does not yet support.


s373r (approved):
LGTM

PR #1304, opened 2025-07-21
Open gql playground in anonymous mode

rmn-boiko:


src/adapter/http/src/middleware/auth_policy_layer.rs
```diff
@@ -62,6 +63,10 @@ impl<Svc> AuthPolicyMiddleware<Svc> {
     pub async fn is_auth_graphql_operation_request(
         request: Request<Body>,
     ) -> (Request<Body>, bool) {
+        if request.method() == http::Method::GET {
```
zaychenko-sergei, line 67: Add a comment this is important for playground.

Issue #1307, opened 2025-07-21
GraphQL object access improvements


sergiimk:
We encountered several issues with GQL that impact the productivity. This ticket is to address these issues by extending the GQL API to have more convenient top-level accessors, better graph connectivity, and better batch request support:

## Subject account accessor
It will be very useful to be able to inspect information about the current account based on the identity in the auth token.

Suggestion is to add `me` endpoint for queries and mutations:
```gql
{
  accounts {
    me: Account
  }
}
```

## Connect accounts and datasets
It would be useful to have a graph connection between an account and datasets it owns:
```gql
{
  accounts { me {
    ownedDatasets(<pagination>): DatasetConnection!
  } }
}
```

## Accessing datasets by reference
Sometimes you have a dataset reference that can be ID or an alias (`<account>/<name>`) and want to access the dataset uniformly without having to parse it:
```gql
{
  datasets {
    byRef(datasetRef: ...): Dataset
  }
}
```

## Batch access
Sometime we want to query same information about multiple objects at once, so we could use batch accessors:
```gql
{
  accounts {
    byIds(datasetIds: [DatasetID!]!, allowMissing: bool!): [Dataset!]
    byNames(accountNames: [AccountName!]!, allowMissing: bool!): [Account!]
  }
  datasets {
    byIds(datasetIds: [DatasetID!]!, allowMissing: bool!): [Dataset!]
    byRefs(datasetRefs: [DatasetRef!]!, allowMissing: bool!): [Dataset!]
  }
}
```
NOTE: Consider `allowMissing` parameter that controls whether returned list should just skip the references it could not resolve or return `null` for the entire request altogether

NOTE: ID-based batch accessors could be useful for mutations too.

## Encoding selection
When working with data/schema/metadata encoding our current scheme looks like this:
```gql
{
  getData(..., format: ND_JSON) {
    data {
      format
      content
    }
  }
  getOtherData(..., format: ND_JSON) {
    data {
      format
      content
    }
  }
}
```
Above encoding spreads into every single operation involving data/schema/metadata, which is cumbersome.

An alternative approach could be
```gql
{
  getData(...) {
    data(format: ND_JSON) {
      format
      content
    }
  }
  getOtherData(...) {
    data(format: ND_JSON) {
      format
      content
    }
  }
}
```
which:
- keeps encoding selection within the block object itself
- simplifies operation signatures
- seems a lot more "graph-friendly"

PR #1316, opened 2025-07-30
Add extended events block gql api

rmn-boiko:
## Description

Closes: #issue

- [ ] Compatibility:
    
  - [ ] Network APIs: ✅
    
  - [ ] Configuration: ✅
    
  - [ ] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
+++ b/src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
@@ -25,6 +25,73 @@ pub struct DatasetMetadata<'a> {
     dataset_request_state: &'a DatasetRequestState,
 }
 
+impl<'a> DatasetMetadata<'a> {
+    #[tracing::instrument(level = "info", skip_all)]
+    async fn get_current_polling_source(
+        &'a self,
+        ctx: &Context<'_>,
+    ) -> Result<
+        Option<(
+            odf::Multihash,
+            odf::metadata::MetadataBlockTyped<odf::metadata::SetPollingSource>,
+        )>,
+    > {
+        let metadata_query_service = from_catalog_n!(ctx, dyn kamu_core::MetadataQueryService);
+
+        let resolved_dataset = self.dataset_request_state.resolved_dataset(ctx).await?;
+        let result = metadata_query_service
+            .get_active_polling_source(resolved_dataset)
+            .await
+            .int_err()?;
+
+        Ok(result)
+    }
+
+    #[tracing::instrument(level = "info", skip_all)]
+    async fn get_current_push_sources(
+        &'a self,
+        ctx: &Context<'_>,
+    ) -> Result<
+        Vec<(
+            odf::Multihash,
+            odf::metadata::MetadataBlockTyped<odf::metadata::AddPushSource>,
+        )>,
+    > {
+        let metadata_query_service = from_catalog_n!(ctx, dyn kamu_core::MetadataQueryService);
+
+        let resolved_dataset = self.dataset_request_state.resolved_dataset(ctx).await?;
+
+        let result = metadata_query_service
+            .get_active_push_sources(resolved_dataset)
+            .await
+            .int_err()?;
+
+        Ok(result)
+    }
+
+    #[tracing::instrument(level = "info", skip_all)]
+    async fn get_current_transform(
+        &'a self,
+        ctx: &Context<'_>,
+    ) -> Result<
+        Option<(
+            odf::Multihash,
+            odf::metadata::MetadataBlockTyped<odf::metadata::SetTransform>,
+        )>,
+    > {
+        let metadata_query_service = from_catalog_n!(ctx, dyn kamu_core::MetadataQueryService);
+
+        let resolved_dataset = self.dataset_request_state.resolved_dataset(ctx).await?;
+        let result = metadata_query_service
+            .get_active_transform(resolved_dataset)
+            .await?;
+
+        Ok(result)
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 #[common_macros::method_names_consts(const_value_prefix = "Gql::")]
 #[Object]
 impl<'a> DatasetMetadata<'a> {
@@ -140,27 +207,46 @@ impl<'a> DatasetMetadata<'a> {
     /// Current polling source used by the root dataset
     #[tracing::instrument(level = "info", name = DatasetMetadata_current_polling_source, skip_all)]
     async fn current_polling_source(&self, ctx: &Context<'_>) -> Result<Option<SetPollingSource>> {
-        let metadata_query_service = from_catalog_n!(ctx, dyn kamu_core::MetadataQueryService);
+        let source_maybe = self.get_current_polling_source(ctx).await?;
 
-        let resolved_dataset = self.dataset_request_state.resolved_dataset(ctx).await?;
-        let source = metadata_query_service
-            .get_active_polling_source(resolved_dataset)
-            .await
-            .int_err()?;
+        Ok(source_maybe.map(|(_hash, block)| block.event.into()))
+    }
 
-        Ok(source.map(|(_hash, block)| block.event.into()))
+    /// Current polling source block used by the root dataset
+    #[tracing::instrument(level = "info", name = DatasetMetadata_current_polling_source_block, skip_all)]
+    async fn current_polling_source_block(
+        &self,
+        ctx: &Context<'_>,
+    ) -> Result<Option<MetadataBlockExtended>> {
+        let source_maybe = self.get_current_polling_source(ctx).await?;
+
+        let account =
+            Account::from_dataset_alias(ctx, &self.dataset_request_state.dataset_handle().alias)
+                .await?
+                .expect("Account must exist");
+
+        Ok(if let Some((hash, block)) = source_maybe {
+            Some(
+                MetadataBlockExtended::new(
+                    ctx,
+                    hash,
+                    block.into(),
+                    account,
+                    Some(MetadataManifestFormat::Yaml),
+                )
+                .await?,
+            )
+        } else {
+            None
+        })
     }
 
     /// Current push sources used by the root dataset
     #[tracing::instrument(level = "info", name = DatasetMetadata_current_push_sources, skip_all)]
     async fn current_push_sources(&self, ctx: &Context<'_>) -> Result<Vec<AddPushSource>> {
-        let metadata_query_service = from_catalog_n!(ctx, dyn kamu_core::MetadataQueryService);
-
-        let resolved_dataset = self.dataset_request_state.resolved_dataset(ctx).await?;
-        let mut push_sources: Vec<AddPushSource> = metadata_query_service
-            .get_active_push_sources(resolved_dataset)
-            .await
-            .int_err()?
+        let mut push_sources: Vec<AddPushSource> = self
+            .get_current_push_sources(ctx)
+            .await?
             .into_iter()
             .map(|(_hash, block)| block.event.into())
             .collect();
@@ -168,6 +254,43 @@ impl<'a> DatasetMetadata<'a> {
         push_sources.sort_by(|a, b| a.source_name.cmp(&b.source_name));
 
         Ok(push_sources)
+    }
+
+    /// Current push source blocks used by the root dataset
+    #[tracing::instrument(level = "info", name = DatasetMetadata_current_push_source_blocks, skip_all)]
+    async fn current_push_source_blocks(
+        &self,
+        ctx: &Context<'_>,
+    ) -> Result<Vec<MetadataBlockExtended>> {
+        let push_sources = self.get_current_push_sources(ctx).await?;
+
+        let mut result = Vec::new();
+        let account =
+            Account::from_dataset_alias(ctx, &self.dataset_request_state.dataset_handle().alias)
+                .await?
+                .expect("Account must exist");
+
+        for (hash, block) in push_sources {
+            result.push(
+                MetadataBlockExtended::new(
+                    ctx,
+                    hash,
+                    block.into(),
+                    account.clone(),
+                    Some(MetadataManifestFormat::Yaml),
+                )
+                .await?,
+            );
+        }
+
+        result.sort_by(|a, b| match (&a.event, &b.event) {
+            (MetadataEvent::AddPushSource(a_event), MetadataEvent::AddPushSource(b_event)) => {
+                a_event.source_name.cmp(&b_event.source_name)
+            }
+            _ => unreachable!("Expected only AddPushSource type"),
+        });
+
+        Ok(result)
     }
 
     /// Sync statuses of push remotes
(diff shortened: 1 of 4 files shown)


resources/schema.gql
```diff
@@ -1245,6 +1245,7 @@ type DatasetMetadata {
 	Current vocabulary associated with the dataset
 	"""
 	currentVocab: SetVocab
+	extendedBlocksByEventType(eventType: MetadataEventType!): [MetadataBlockExtended!]!
```
sergiimk: `MetadataBlockExtended` type is called like that because it includes a few convenience things in addition to regular `MetadataBlock` like:
- `blockHash` (hash is not stored in the `odf::MetadataBlock` because it is only known after serialization)
- `author` as a link to GQL `Account` type

So it's just a more "GQL friendly` version of the `MetadataBlock`.

Having "extended" in the names of operations doesn't add any information and only will confuse users, so I'd prefer this to be named just `blocksByEventType`.

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
```diff
@@ -273,6 +325,62 @@ impl<'a> DatasetMetadata<'a> {
...
+        let block_infos: Vec<(odf::Multihash, odf::metadata::MetadataBlock)> = match event_type {
+            MetadataEventType::SetPollingSource => {
+                let mut blocks = vec![];
+                if let Some((hash, block)) = self.get_current_polling_source(ctx).await? {
+                    blocks.push((hash, block.into()));
+                }
+                blocks
+            }
+            MetadataEventType::AddPushSource => self
+                .get_current_push_sources(ctx)
```
sergiimk: Hm, in the operation name `extended_blocks_by_event_type` nothing tells me that it returns only latest / current / projection of the events. From its name and signature I expected it to return ALL blocks of the specified type.

I think we should either have:
- A low level API for iterating blocks of a certain type
  - We already have `MetadataChain::blocks(page: Int, perPage: Int): MetadataBlockConnection!` and could extend it to support type filters
- Or a high level API for **current** state of transforms / push / poll sources

We already have:
- `currentTransform: SetTransform`
- `currentPollingSource: SetPollingSource`
- `currentPushSources: [AddPushSource!]!`

so if the goal of this PR is to expose them as blocks instead of events, perhaps we could change the return type of the above and do a breaking change?

src/adapter/graphql/src/scalars/metadata.rs
```diff
@@ -27,25 +27,52 @@ pub struct MetadataBlockExtended {
     pub author: Account,
     pub event: MetadataEvent,
     pub sequence_number: u64,
+    pub encoded: Option<String>,
```
sergiimk: We have a similar encoding thing for data and schema and I was thinking about the ergonomics of it lately.

Instead of having something like this:
```gql
{
  giveSomeBlocks(..., encoding: YAML) {
    blockHash
    encoded
  }
  giveSomeOtherBlocks(..., encoding: YAML) {
    blockHash
    encoded
  }
}
```
where encoding spreads into every operation involving blocks, we could do instead:
```gql
{
  giveSomeBlocks(...) {
    blockHash
    encoded(encoding: YAML)
  }
  giveOtherBlocks(...) {
    blockHash
    encoded(encoding: YAML)
  }
}
```
which keeps encoding selection within the block object itself.

It's a lot more "graph-friendly" IMO. What do you think?

Also noted this suggestion on: https://github.com/kamu-data/kamu-cli/issues/1307

sergiimk (changes requested):

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
rmn-boiko: modify these methods this was my very first idea. But during the discussion with Dmytriy and Sergiy, introducing breaking changes for these methods will cause I lot of work on ui side, which we would like to avoid

src/adapter/graphql/src/scalars/metadata.rs
rmn-boiko: Yeah, looks much better. Added

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
```diff
@@ -273,6 +325,60 @@ impl<'a> DatasetMetadata<'a> {
...
+        let resolved_dataset = self.dataset_request_state.resolved_dataset(ctx).await?;
+
+        let mut visitors_list: Vec<MetadataVisitor> =
+            event_types.iter().map(|event| (*event).into()).collect();
+
+        let mut visitors: Vec<
+            &mut dyn odf::dataset::MetadataChainVisitor<Error = odf::dataset::Infallible>,
+        > = visitors_list
+            .iter_mut()
+            .map(MetadataVisitor::as_visitor)
```
sergiimk: `MetadataVisitor` wrapper type feels a bit awkward to me. It's less of a type and more of a collection of mapping functions...

Please take a look at how we solved a similar problem in `dataset_metadata_handler` (`src/adapter/http/src/data/metadata_handler.rs`) and let me know what do you think about that approach.

Instead of a vector where some visitors may be added and some aren't based on filtering condition - it relies on a neat trick that `Option<T: Visitor>` also implements `Visitor` - allowing to use a static array.

Because of this it also does not require boxing into `ExtractBlock` - you can operate with individual visitors knowing their exact types.

src/odf/dataset/src/services/metadata_chain_visitors.rs
```diff
@@ -284,6 +302,54 @@ impl MetadataChainVisitor for SearchSingleDataBlockVisitor {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub struct SearchAddPushSourcesVisitor {
```
sergiimk: My understanding of `metadataProjection` API is that we want to get **active** push sources, not just all push sources that were ever declared.

Active means there is an `AddPushSource` event that has not been undone by `DisablePushSource` event.

Currently we don't support `DisablePushSource` event properly, so all `AddPushSource`s == all active sources, so I'm not asking you to implement `DisablePushSource` handling logic here.

But I think we should name this visitor `SearchActivePushSourcesVisitor` instead to reflect its intended function.

src/odf/dataset/src/services/metadata_chain_visitor.rs
```diff
@@ -94,6 +94,14 @@ pub trait MetadataChainVisitor: Send {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+pub trait ExtractBlock {
```
sergiimk: With the approach used in `dataset_metadata_handler` we probably will not need this trait

sergiimk (changes requested):

src/adapter/graphql/src/scalars/metadata.rs
sergiimk: You said "Added" but I still see the old approach being used.

rmn-boiko: Yeah, sorry. I've added to mutation params but forgot to add it in response

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
rmn-boiko: Changed to this solution

src/odf/dataset/src/services/metadata_chain_visitors.rs
rmn-boiko: Agree, renamed and left comments.

src/odf/dataset/src/services/metadata_chain_visitor.rs
rmn-boiko: Removed

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
```diff
@@ -273,6 +325,156 @@ impl<'a> DatasetMetadata<'a> {
...
+        let mut result = vec![];
+
+        // Add all blocks from visitors
+        if let Some(inner) = attachments_visitor
+            && let Some((hash, block)) = inner.into_hashed_block()
+        {
+            result.push(
+                MetadataBlockExtended::new(ctx, hash, block.into(), account.clone(), encoding)
+                    .await?,
+            );
```
sergiimk: how about extracting this whole body (creation of `MetadataBlockExtended` and push into vector) into a lambda, so that for every listener you could do something as short as:
```rust
attachmetns_visitor.and_then(|v| v.into_hashed_block()).map(add_block);
```
this way when we change how `MetadataBlockExtended` is constructed - we won't have to change it in a dozen places.

Collecting `(hash, block)` tuples into a temporary vec and then mapping over it would work well too.

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
```diff
@@ -273,6 +325,156 @@ impl<'a> DatasetMetadata<'a> {
             .into_event()
             .map(Into::into))
     }
+
+    #[tracing::instrument(level = "info", name = DatasetMetadata_metadata_projection, skip_all)]
+    async fn metadata_projection(
+        &self,
+        ctx: &Context<'_>,
+        event_types: Vec<MetadataEventType>,
+        encoding: Option<MetadataManifestFormat>,
```
sergiimk: Let's move this parameter down to the point where user requests encoded block content.

i.e. instead of having a field `pub encoded: Option<EncodedBlock>` - make `encoded` a method that accepts encoding as a parameter

rmn-boiko: Changed

resources/schema.gql
```diff
@@ -2556,6 +2563,18 @@ See: https://github.com/kamu-data/open-data-fabric/blob/master/open-data-fabric.
...
+	SET_POLLING_SOURCE
+	SET_VOCAB
+	SET_ATTACHMENTS
+	SET_INFO
+	SET_LICENSE
+	SET_DATA_SCHEMA
+	SET_TRANSFORM
+	ADD_PUSH_SOURCE
+}
+
```
s373r, line 2577: We have the following GQL type:
```gql
union MetadataEvent = AddData
    | ExecuteTransform
    | Seed
    | SetPollingSource
    | SetTransform
    | SetVocab
    | SetAttachments
    | SetInfo
    | SetLicense
    | SetDataSchema
    | AddPushSource
    | DisablePushSource
    | DisablePollingSource
```

Q: Can we use it?

src/odf/dataset/src/services/metadata_chain_visitors.rs
```diff
@@ -284,6 +285,43 @@ impl MetadataChainVisitor for SearchSingleDataBlockVisitor {
...
+        Decision::NextOfType(Flag::ADD_PUSH_SOURCE)
+    }
+
+    fn visit(&mut self, (hash, block): HashedMetadataBlockRef) -> Result<Decision, Self::Error> {
+        let flag = Flag::from(&block.event);
+
+        if flag == Flag::ADD_PUSH_SOURCE {
+            self.hashed_blocks
+                .push((hash.clone(), block.clone().into_typed().unwrap()));
+        }
```
s373r, line 318: Please add a TODO re `DisablePushSource`

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
s373r: It's a tricky spot. 

Since all visitors want their own block type + there's async (`MetadataBlockExtended::new(...).await`), which isn't easy to pack into standard mapping methods.

s373r: As an option, we could consider creating optional futures and subsequently launching them:
```rust
       // Variant 1.
       // Add all blocks from visitors
        let mut futures: Vec<
            Pin<Box<dyn Future<Output = StdResult<MetadataBlockExtended, InternalError>> + Send>>,
        > = Vec::with_capacity(9);

        if let Some(fut) =
            attachments_visitor
                .and_then(|v| v.into_hashed_block())
                .map(|(hash, block)| async {
                    MetadataBlockExtended::new(ctx, hash, block.into(), account.clone(), encoding)
                        .await
                })
        {
            futures.push(Box::pin(fut));
        }
        if let Some(fut) =
            info_visitor
                .and_then(|v| v.into_hashed_block())
                .map(|(hash, block)| async {
                    MetadataBlockExtended::new(ctx, hash, block.into(), account.clone(), encoding)
                        .await
                })
        {
            futures.push(Box::pin(fut));
        }
        // ... other visitors

        let result = futures::future::try_join_all(futures).await?;

        Ok(result)
```

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
s373r: > Option<T: Visitor> also implements Visitor

Offtop: elegantly

src/adapter/graphql/src/queries/datasets/dataset_metadata.rs
s373r: I'll check a couple more options

UPD: updated the proposed option

s373r: How much more expressive it has become is a big question...

I think we should get rid of this part ...:
```rust
if let Some(fut) = ... {
    futures.push(Box::pin(fut));
}
```
... , first, collect it in `Vec<Option<Fut>>`, then filter it from None (`.flatten()`), but I couldn't achieve this quickly

s373r: Considering that we don't need a typed block, it might be cleaner:
```rust
       // Variant 2.
        // Add all blocks from visitors
        let mut maybe_hashed_blocks: Vec<Option<(odf::Multihash, odf::metadata::MetadataBlock)>> = vec![
            attachments_visitor
                .and_then(|v| v.into_hashed_block())
                .map(|(h, b)| (h, b.into())),
            info_visitor
                .and_then(|v| v.into_hashed_block())
                .map(|(h, b)| (h, b.into())),
            // ... other SINGLE block visitors
        ];
        maybe_hashed_blocks.extend(
            push_sources_visitor
                .map(|v| v.into_hashed_blocks())
                .into_iter()
                .flatten()
                .map(|(h, b)| Some((h, b.into()))),
        );
        let futures_iter = maybe_hashed_blocks
            .into_iter()
            .flatten() // Skip nones
            .map(|(hash, block)| async {
                MetadataBlockExtended::new(ctx, hash, block.into(), account.clone(), encoding).await
            });

        let result = futures::future::try_join_all(futures_iter).await?;

        Ok(result)
```

resources/schema.gql
rmn-boiko: I've tried to somehow reuse this type at the beginning.  But unfortunatly it cannot be used as an Input object

s373r (approved):
LGTM

PR #1321, opened 2025-08-03
Introduce ODF schema

sergiimk:
## Description

Implements ODF schema migrations and improves archetype support per [ODF RFC-016, RFC-017](https://github.com/open-data-fabric/open-data-fabric/pull/116)

New looks of `inspect schema` command:
image

Metadata of a versioned file containing a summary of linked objects:
image

- [x] Compatibility:
    
  - [x] Network APIs: ❌
    
    A manual migration is needed to define schemas for existing versioned file and collections dataset. A script was developed and tested: https://github.com/kamu-data/kamu-deploy/pull/336
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
    https://github.com/kamu-data/kamu-node/pull/273
    
    This work will generate a bunch of new tasks in UI, but those don't need to be lock-step


src/e2e/app/cli/repo-tests/src/commands/test_log_command.rs
```diff
@@ -120,29 +121,19 @@ pub async fn test_log_command(kamu: KamuCliPuppet) {
...
+                        DataField::i64("offset"),
+                        DataField::i32("op"),
+                        DataField::timestamp_millis_utc("system_time"),
+                        DataField::timestamp_millis_utc("match_time").optional(),
+                        DataField::i64("match_id").optional(),
+                        DataField::string("player_id").optional(),
+                        DataField::i64("score").optional(),
+                    ],
+                    extra: None,
+                }),
```
s373r, line 136: That looks much better! :)

src/infra/core/src/services/query/mod.rs
```diff
@@ -308,6 +327,9 @@ impl KamuTable {
             .await;
 
         let options = ParquetReadOptions {
+            // TODO: PERF: Schema here comes from the metadata chain and will contain logical types
```
s373r: Q: Can we send some hints over here?

src/odf/data-utils/src/testing/mod.rs
```diff
@@ -15,6 +15,16 @@ use crate::data::DataFrameExt;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[allow(clippy::needless_pass_by_value)]
```
s373r: Unnecessary allow

src/odf/metadata/src/schema/arrow_conversions.rs
```diff
@@ -0,0 +1,1139 @@
...
+            ArrowDataType::Union(_, _) => {
+                return Err(UnsupportedSchema::new(
+                    "Union types are not yet supported in ODF schema",
+                ));
+            }
+            ArrowDataType::Dictionary(_, _) => {
+                return Err(UnsupportedSchema::new(
+                    "Dictionary types are not yet supported in ODF schema",
+                ));
+            }
```
s373r, line 360: Q: Could it be `Map`?

src/odf/metadata/src/serde/flatbuffers/proxies_generated.rs.orig
s373r, line 1: It seems like this is an unnecessary file

src/adapter/graphql/src/queries/datasets/adapters/collection.rs
```diff
@@ -20,23 +20,47 @@ pub struct Collection {
...
+
+        let extra_columns_schema =
+            odf::utils::schema::parse::parse_ddl_to_odf_schema(&extra_columns_ddl.join(", "))?;
+
+        let schema = odf::schema::DataSchema::new(
+            [
+                odf::schema::DataField::i64("offset"),
+                odf::schema::DataField::i32("op"),
+                odf::schema::DataField::timestamp_millis_utc("system_time"),
+                odf::schema::DataField::timestamp_millis_utc("event_time"),
```
zaychenko-sergei: Perhaps default system column names should have constants defined somewhere in ODF layer

src/adapter/graphql/src/queries/datasets/adapters/versioned_file.rs
```diff
@@ -28,7 +28,41 @@ impl VersionedFile {
...
+
+        let extra_columns_schema =
+            odf::utils::schema::parse::parse_ddl_to_odf_schema(&extra_columns_ddl.join(", "))?;
+
+        let schema = odf::schema::DataSchema::new(
+            [
+                odf::schema::DataField::i64("offset"),
+                odf::schema::DataField::i32("op"),
+                odf::schema::DataField::timestamp_millis_utc("system_time"),
+                odf::schema::DataField::timestamp_millis_utc("event_time"),
```
zaychenko-sergei: It repeats = can be generalized for reuse

src/adapter/graphql/src/queries/datasets/dataset.rs
```diff
@@ -71,46 +70,24 @@ impl Dataset {
...
-
-    pub(crate) async fn is_collection(dataset: &ResolvedDataset) -> Result<bool> {
-        Self::push_source_has_columns(dataset, &["path", "ref"]).await
-    }
-
-    async fn push_source_has_columns(dataset: &ResolvedDataset, columns: &[&str]) -> Result<bool> {
-        let (source, _merge) = match dataset
+    pub(crate) async fn get_archetype(
+        dataset: &ResolvedDataset,
+    ) -> Result<Option<odf::schema::ext::DatasetArchetype>> {
```
zaychenko-sergei: Consider caching similarly to other fields in `DatasetRequestState`.

src/adapter/graphql/src/scalars/data_schema.rs
```diff
@@ -17,89 +19,93 @@ use crate::prelude::*;
...
+    OdfYaml,
     Parquet,
     ParquetJson,
-    ArrowJson,
 }
 
-#[derive(SimpleObject, Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DataSchema {
+    pub schema: Arc<odf::schema::DataSchema>,
```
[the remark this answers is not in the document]
zaychenko-sergei: Also, I don't understand, where does it convert to JSON/YAML?

src/adapter/graphql/src/scalars/data_schema.rs
```diff
@@ -17,89 +19,93 @@ use crate::prelude::*;
 
 #[derive(Enum, Debug, Clone, Copy, PartialEq, Eq)]
 pub enum DataSchemaFormat {
+    ArrowJson,
+    OdfJson,
+    OdfYaml,
```
zaychenko-sergei, line 24: Should we deprecate access to older schema formats?

src/adapter/graphql/src/scalars/dataset_archetype.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl From<DatasetArchetype> for odf::schema::ext::DatasetArchetype {
+    fn from(value: DatasetArchetype) -> Self {
+        match value {
+            DatasetArchetype::Collection => Self::Collection,
+            DatasetArchetype::VersionedFile => Self::VersionedFile,
+        }
+    }
+}
```
zaychenko-sergei: `#[graphql(remote = "odf::schema::ext::DatasetArchetype")] `

resources/schema.gql
```diff
@@ -298,6 +298,10 @@ type AddData {
 	when resuming.
 	"""
 	newSourceState: SourceState
+	"""
+	ODF extensions.
+	"""
+	extra: ExtraAttributes
```
zaychenko-sergei, line 356: UI might need a higher-level view model for these attributes. Of course, there could be really custom key-value system shows as is, but perhaps for ODF-defined and Kamu-defined attributes, UI needs a representation of higher-level intent.

src/adapter/graphql/src/scalars/odf_generated.rs
```diff
@@ -50,6 +50,8 @@ pub struct AddData {
     /// should be carried, i.e. only the last state per source is considered
     /// when resuming.
     pub new_source_state: Option<SourceState>,
+    /// ODF extensions.
+    pub extra: Option<ExtraAttributes>,
```
zaychenko-sergei, line 54: What about "new_linked_objects" ?

src/infra/core/tests/tests/ingest/test_push_ingest.rs
```diff
@@ -410,10 +411,142 @@ async fn test_ingest_push_schema_stability() {
...
+                "metadata": {},
+                "nullable": false,
+            }, {
+                "name": "event_time",
+                "data_type": {
+                    "Timestamp": [
+                        "Millisecond",
+                        "UTC",
+                    ],
+                },
```
zaychenko-sergei, line 526: Repetetive code for asserting standard columns

src/infra/core/src/services/query/mod.rs
sergiimk: This comment is slightly misleading, as above in this file I use `default_buffer_encoding` setting in `odf_schema.to_arrow()` call to make datafusion use `*View` types for efficiency.

This is not a perfect solution, but should get the job done.

I'll update the comment.

src/odf/metadata/src/schema/arrow_conversions.rs
sergiimk: No, dictionary here is not like python dict - here it actually means a compact encoding for a `List` with low-cardinality values.

Example:
```
Logical column: ["red", "red", "blue", "green", "red"]
Physical representation:
  values: ["red", "blue", "green"]
  indexes: [0, 0, 1, 2, 0]
```

We can easily support this in ODF schema (via encoding annotation on a `List` type), but we don't use this representation anywhere yet, so I decided to save time and not implement tests for it yet.

src/odf/metadata/src/serde/flatbuffers/proxies_generated.rs.orig
sergiimk: thanks for catching, will remove and figure out what produces it

src/adapter/graphql/src/queries/datasets/dataset.rs
sergiimk: done!

src/adapter/graphql/src/scalars/data_schema.rs
sergiimk: I think it's still useful to see how data looks on disk and in memory. But perhaps if Parquet schema is requested we should be reading it from file and not from metadata chain. Will need to think more about it before changing the API.

src/adapter/graphql/src/scalars/dataset_archetype.rs
sergiimk: nice!

resources/schema.gql
sergiimk: I started with exposing `DatasetArchetype` in GQL as a separate endpoint so UI didn't have to extract it from schema annotations.

ODF Data schema itself only can be requested as encoded in JSON/YAML, so there are no GQL objects which we could extend with things like field `description`. UI will need to learn how to work with ODF schema and its extensions, which I think is OK.

The extra attrs you pointed out are attached to a `AddData` metadata event which currently can only have `opendatafabric.org/inkedObjects` attribute. We could surface it into an endpoint ... but I'm actually considering a path similar to `DataSchema` - perhaps we should not create GQL objects for every metadata block/event, and only expose those as encoded in JSON/YAML/Flatbuffers in our APIs.

src/adapter/graphql/src/scalars/odf_generated.rs
sergiimk: This is a code-generated object, so I can't rename it without changing ODF schema.

We could add a helper endpoint, but as per my other comment I'm not yet sure this is the right path.

src/adapter/graphql/src/queries/datasets/adapters/collection.rs
sergiimk: Introduced `DataSchema::builder().with_changelog_system_fileds()` for reuse

src/adapter/graphql/src/queries/datasets/adapters/versioned_file.rs
sergiimk: Introduced `DataSchema::builder().with_changelog_system_fileds()` for reuse

src/infra/core/tests/tests/ingest/test_push_ingest.rs
sergiimk: In this test I'd like to be very explicit and write out schema in full, to ensure we catch any potential drift.

Issue #1318, opened 2025-07-30
Wrong authorization header behavior


dmitriy-borzenko:
Reproduce:

<img width="1863" height="791" alt="Image" src="https://github.com/user-attachments/assets/b7152d61-3b7d-4288-bd6b-8b6bcfdd70bd" />

Result:

<img width="1720" height="961" alt="Image" src="https://github.com/user-attachments/assets/54d89c0f-1da2-4aa7-aa88-6bd289a0f2de" />

PR #1324, opened 2025-08-08
Correctly handle empty authentication token error

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/1318

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Recommendation: for ease of reading, use the following order:
 ### Added
 - GQL: `dataset_metadata`: Added new method `MetadataProjection` to get `[MetadataExtendedBlock]` filtered by types and range
 ### Fixed
-- Correctly handle empty authentication token error
+- Return bad request response for requests with empty bearer authentication token
 
 ## [0.245.3] - 2025-07-29
 ### Fixed
--- a/src/adapter/http/src/middleware/authentication_layer.rs
+++ b/src/adapter/http/src/middleware/authentication_layer.rs
@@ -140,7 +140,7 @@ where
                 if let Ok(access_token) = request.extract_parts::<Option<BearerHeader>>().await {
                     access_token.map(|th| AccessToken::new(th.token()))
                 } else {
-                    return Ok(unauthorized_access_response());
+                    return Ok(bad_request_response(Some("Invalid Bearer token header")));
                 };
 
             let base_catalog = request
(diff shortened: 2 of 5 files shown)


src/adapter/http/src/middleware/authentication_layer.rs
```diff
@@ -136,11 +136,12 @@ where
...
-            let maybe_access_token = request
-                .extract_parts::<Option<BearerHeader>>()
-                .await
-                .unwrap()
-                .map(|th| AccessToken::new(th.token()));
+            let maybe_access_token =
+                if let Ok(access_token) = request.extract_parts::<Option<BearerHeader>>().await {
+                    access_token.map(|th| AccessToken::new(th.token()))
+                } else {
+                    return Ok(unauthorized_access_response());
```
sergiimk: Looks like at this stage we only care about `Authorization` header being well-formed, not about it containing a valid token.

Using [OAuth 2 RFC](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) as inspiration - it recommends returning 400 when the request (including headers) is malformed.

401 is used when token itself is expired or malformed (but the `Authrization` header as a whole makes sense).

While I would prefer 400 in this situation, as I think `Bearer ` is invalid request more than auth failure, 400 has a drawback as it's more tricky for user to understand **which part** of the request was wrong. Ideally we would also return the JSON payload that contains an error message, but the exact format needs to be designed first.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Recommendation: for ease of reading, use the following order:
 ## [Unreleased]
 ### Added
 - GQL: `dataset_metadata`: Added new method `MetadataProjection` to get `[MetadataExtendedBlock]` filtered by types and range
+### Fixed
+- Correctly handle empty authentication token error
```
sergiimk: "Correctly" is subjective - let's explain how exactly it will react to what.

src/adapter/http/src/middleware/authentication_layer.rs
rmn-boiko: Changed

src/adapter/http/Cargo.toml
```diff
@@ -24,7 +24,7 @@ doctest = false
 [features]
 default = []
 
-e2e = ["dep:http-body-util"]
+e2e = []
```
zaychenko-sergei, line 27: Safe to remove, if it does not switch anything else

rmn-boiko: It still refers to e2e part of crate. I would keep it

Issue #1270, opened 2025-06-09
`PredefinedAccountsRegistrator`: using use cases from the Accounts domain


s373r:
### Context
- Depended on https://github.com/kamu-data/kamu-cli/issues/1269
- `PredefinedAccountsRegistrator` currently uses `AccountRepository` directly to create and update accounts

### Requirements

- Refactoring: Use `CreateAccountUseCase` to create accounts
- Refactoring: Introduce a use case for account updating
- E2E: Cover `PredefinedAccountsRegistrator`

PR #1345, opened 2025-08-26
PredefinedAccountsRegistrator: using use cases from the Accounts domain

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/1270

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
```diff
@@ -35,9 +35,13 @@ pub trait CreateAccountUseCase: Send + Sync {
...
 
 #[derive(bon::Builder, Default)]
-pub struct CreateAccountUseCaseOptions {
+pub struct PredefinedAccountFields {
     pub email: Option<Email>,
     pub password: Option<Password>,
+    pub display_name: Option<String>,
+    pub provider: Option<String>,
+    pub id: Option<odf::AccountID>,
+    pub avatar_url: Option<String>,
```
s373r: Q: How about making `PredefinedAccountFields` a `CreateAccountUseCaseOptions` field?

This way, we will retain the ability to extend the options.

src/domain/accounts/services/src/services/predefined_accounts_registrator.rs
```diff
@@ -91,19 +94,24 @@ impl PredefinedAccountsRegistrator {
...
+                None,
+                &account.account_name,
+                PredefinedAccountFields::builder()
+                    .email(account.email.clone())
+                    .maybe_password(Some(account_config.password.clone()))
+                    .display_name(account.display_name.clone())
+                    .provider(account.provider)
+                    .maybe_avatar_url(account.avatar_url)
+                    .id(account.id)
+                    .build(),
```
s373r: Opinion: A fluid interface is suitable for options, but in the case of structure initialization, the approach may become less readable.

[Optional] Q: How about leaving the fluid interface for the options, but initializing the structure itself (which can become the field for these options) as usual?

src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
```diff
@@ -82,30 +119,79 @@ impl CreateAccountUseCaseImpl {
...
+                )
+                .await
+                .int_err()?;
+        }
+
+        // Notify about account creation only if account was created by another account
+        // otherwise created account via predefined registration will skip notification
+        if creator_account.is_some() {
+            self.notify_account_created(&account).await?;
+        }
```
s373r: Q: What do you think about adding a flag in the options, something like "notify_account_created: bool"?

src/domain/accounts/services/src/use_cases/update_account_use_case_impl.rs
```diff
@@ -0,0 +1,125 @@
...
+    UpdateAccountUseCaseError,
+};
+use messaging_outbox::Outbox;
+
+use crate::utils;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[dill::component(pub)]
+pub struct UpdateInnerAccountUseCaseImpl {
```
s373r: I would like to propose a solution with one -- but I am still thinking it over.

Just in case:
Q: Did you split the use case into two for testing purposes?

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ Recommendation: for ease of reading, use the following order:
     - response timeout
     - bad status code in response
 ## Changed
+- Refactoring:
+  - Added new `UpdateAccountUseCase` and replace usage of similar methods
+  - PredefinedAccountRegistrator now uses `use_case` methods instead of service methods
```
s373r:
```suggestion
  - `PredefinedAccountRegistrator` now uses `UpdateAccountUseCase` methods instead of `AccountService` methods
```

src/domain/accounts/services/src/use_cases/update_account_use_case_impl.rs
rmn-boiko: Not really. 
Inside `UpdateUseCase` I've added access check(which checks CurrentAccountSubject matching).
But `PredefinedAccountRegistrator` does not need this check. So for higher levels such as API I'm using general `UpdateUseCase` and inside `PredefinedAccountRegistrator` I'm using this method

src/domain/accounts/services/src/use_cases/update_account_use_case_impl.rs
```diff
@@ -0,0 +1,130 @@
...
```
zaychenko-sergei: You probably mean "internal". "Inner" would mean an account is wrapped into another bigger account, like матрешка

src/domain/accounts/services/src/use_cases/update_account_use_case_impl.rs
zaychenko-sergei: Instead of 2 use cases, we could have 2 execute methods: `execute` and `execute_internal`. Or, in case the authorization check is the only difference, there could simply be a "preauthorized: bool" parameter.

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
```diff
@@ -35,9 +35,13 @@ pub trait CreateAccountUseCase: Send + Sync {
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 #[derive(bon::Builder, Default)]
-pub struct CreateAccountUseCaseOptions {
+pub struct PredefinedAccountFields {
```
zaychenko-sergei: I don't really see benefit in this rename. We will likely be implementing a Signup function sooner or later, and that would mean "create" is invoked not just for predefined accounts.

How about `AccountFields` simply?

src/domain/accounts/domain/src/use_cases/update_account_use_case.rs
```diff
@@ -10,37 +10,44 @@
...
-////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-// Errors
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 #[derive(Debug, Error)]
-pub enum RenameAccountError {
-    #[error(transparent)]
-    Duplicate(AccountErrorDuplicate),
-
+pub enum UpdateAccountUseCaseError {
```
zaychenko-sergei: Minor: normally "use case" is not a part of error types.

src/domain/accounts/domain/src/use_cases/update_account_use_case.rs
```diff
@@ -10,37 +10,44 @@
...
+
+impl From<UpdateAccountError> for UpdateAccountUseCaseError {
+    fn from(err: UpdateAccountError) -> Self {
+        match err {
+            UpdateAccountError::NotFound(e) => UpdateAccountUseCaseError::NotFound(e),
+            UpdateAccountError::Duplicate(e) => UpdateAccountUseCaseError::Duplicate(e),
+            UpdateAccountError::Internal(e) => UpdateAccountUseCaseError::Internal(e),
+        }
+    }
+}
```
zaychenko-sergei: If that's because of "Access" only, it's easier to have just 1 error type both for use case and service.

src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
```diff
@@ -82,30 +119,79 @@ impl CreateAccountUseCaseImpl {
...
         account_name: &odf::AccountName,
-        options: CreateAccountUseCaseOptions,
+        predefined_fields: PredefinedAccountFields,
     ) -> Result<Account, CreateAccountError> {
-        let email = if let Some(email) = options.email {
+        let email = if let Some(email) = predefined_fields.email {
             email
         } else {
             Self::generate_email(creator_account, account_name)?
         };
```
zaychenko-sergei: Prefer a separate entry point for "derived email" instead of `Option<&Account>`. You could have normal `execute` and a `execute_derived` taking a creator account, which internally may share most of the algorithm via private methods.

zaychenko-sergei: Same story for notification - notify in "execute_derived" only without `if` statement.

src/domain/accounts/services/src/use_cases/update_account_use_case_impl.rs
rmn-boiko: It's not just to skip auth check. But to resolve issue, the `PredefinedRegistrator` startup job is executed before we are injecting `CurrentAccountSubject`.
First my solution was to inject `Catalog` and make two methods. Where for auth check I'm getting `CurrentAccountSubject` from catalog. Will change it back

src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
rmn-boiko: Changed, by moving it to separate method

src/domain/accounts/services/src/services/predefined_accounts_registrator.rs
rmn-boiko: Removed it, by moved this logic to separate method

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
rmn-boiko: Put back options structure

PR #1350, opened 2025-08-29
ringbuf: 0.4

s373r:
## Description

Closes: #1349

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/infra/core/src/services/ingest/prep_service.rs
+++ b/src/infra/core/src/services/ingest/prep_service.rs
@@ -200,12 +200,14 @@ impl Read for ReaderHelper<'_> {
 
 struct DecompressZipStream {
     ingress: std::thread::JoinHandle<Result<(), PollingIngestError>>,
-    consumer: ringbuf::Consumer<u8, Arc<ringbuf::HeapRb<u8>>>,
+    consumer: ringbuf::CachingCons<Arc<ringbuf::HeapRb<u8>>>,
     done_recvr: std::sync::mpsc::Receiver<usize>,
 }
 
 impl DecompressZipStream {
     fn new(mut input: Box<dyn Stream>, sub_path: Option<String>) -> Self {
+        use ringbuf::traits::*;
+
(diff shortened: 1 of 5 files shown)


src/infra/core/src/services/ingest/prep_service.rs
```diff
@@ -229,7 +231,14 @@ impl DecompressZipStream {
                         while producer.is_full() {
                             std::thread::sleep(std::time::Duration::ZERO);
                         }
-                        let read = producer.read_from(&mut file, None).unwrap();
+
+                        let Some(read) = producer.read_from(&mut file, None).map(Result::unwrap)
```
s373r, line 235:
```
// 0.3 (before)
Reads at most `count` bytes from `Read` instance and appends them to the ring buffer.
If `count` is `None` then as much as possible bytes will be read.

Returns `Ok(n)` if `read` succeeded. `n` is number of bytes been read.
`n == 0` means that either `read` returned zero or ring buffer is full.

If `read` is failed then original error is returned. In this case it is guaranteed that no items was read from the reader.
To achieve this we read only one contiguous slice at once. So this call may read less than `remaining` items in the buffer even if the reader is ready to provide more.
```                      

```
// 0.4 (after)
Reads at most `count` bytes from `Read` instance and appends them to the ring buffer.
If `count` is `None` then as much as possible bytes will be read.

Returns:
+ `None`: ring buffer is empty or `count` is `0`. In this case `read` isn't called at all.
+ `Some(Ok(n))`: `read` succeeded. `n` is number of bytes been read. `n == 0` means that `read` also returned `0`.
+ `Some(Err(e))` `read` is failed and `e` is original error. In this case it is guaranteed that no items was read from the reader.
  To achieve this we read only one contiguous slice at once. So this call may read less than `vacant_len` items in the buffer even if the reader is ready to provide more.
```

sergiimk (approved):
lgtm :+1:

Issue #1285, opened 2025-06-24
`dev`: Investigate search results behavior


s373r:
### Context

![Image](https://github.com/user-attachments/assets/af8e500f-84ff-47d9-8c2e-194e783de8a8)
(c) https://platform.dev.kamu.dev/molecule.vitarna/~select/datasets

We have access to datasets from another account, but don't see them in search results

### Current Behavior

On the screenshot

### Expected Behavior

In the search results, all datasets of the related account

PR #1357, opened 2025-09-02
`GQL: Search::query()`: case insensitive search

s373r:
:warning: Post-actions:
- [ ] Backport to the main branch 

---

## Description

Related: #1285

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/adapter/graphql/src/mutations/molecule_mut.rs
+++ b/src/adapter/graphql/src/mutations/molecule_mut.rs
@@ -29,12 +29,15 @@ impl MoleculeMut {
     async fn create_project(
         &self,
         ctx: &Context<'_>,
-        ipnft_symbol: String,
+        mut ipnft_symbol: String,
         ipnft_uid: String,
         ipnft_address: String,
         ipnft_token_id: U256,
     ) -> Result<CreateProjectResult> {
-        use datafusion::logical_expr::{col, lit};
+        ipnft_symbol.make_ascii_lowercase();
+        let lowercase_ipnft_symbol = ipnft_symbol;
+
+        use datafusion::prelude::*;
 
         let (
             subject,
@@ -85,7 +88,7 @@ impl MoleculeMut {
                 .filter(
                     col("ipnft_uid")
                         .eq(lit(&ipnft_uid))
-                        .or(col("ipnft_symbol").eq(lit(&ipnft_symbol))),
+                        .or(lower(col("ipnft_symbol")).eq(lit(&lowercase_ipnft_symbol))),
                 )
                 .int_err()?;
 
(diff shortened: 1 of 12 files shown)


src/adapter/graphql/src/mutations/molecule_mut.rs
```diff
@@ -112,7 +115,7 @@ impl MoleculeMut {
             .unwrap();
 
         let project_account_name: odf::AccountName =
-            format!("{}.{ipnft_symbol}", molecule_account.account_name)
+            format!("{}.{lowercase_ipnft_symbol}", molecule_account.account_name)
```
s373r, line 118: I have a feeling that it is also necessary to apply migrations to accounts that contain upper case letters (and, as a result, for `dataset_entries` as well).

```sql
SELECT DISTINCT owner_name
FROM dataset_entries;
-- +---------------------+
-- |owner_name           |
-- +---------------------+
-- |molecule.dev.vitarna |
-- |molecule.dev.MKO     |
-- |molecule.dev.CATTY   |
-- |molecule.dev.PSYMARK |
-- |molecule.dev.vitadent|
-- |molecule.VITALONG    |
-- +        ...          +
-- +---------------------+

SELECT DISTINCT account_name, display_name
FROM accounts;
-- +----------------------+----------------------+
-- |account_name          |display_name          |
-- +----------------------+----------------------+
-- |molecule.vita-fast    |molecule.VITA-FAST    |
-- |molecule.dev.nicotinic|molecule.dev.NICOTINIC|
-- |        ...           |        ...           |
```

The presence of upper case in `dataset_entries.owner_name` led to incorrect results during the search (before these changes).

Q: Do you think we need migrations, or can we just update the search code (GQL: `Search::query()`)?

zaychenko-sergei: Why do we need to change the case of how we store it, if we can make case-insensitive comparisons just during search?

s373r: @zaychenko-sergei , we convert account names to lowercase when saving (unless it is a web3 provider).
At the same time, we do not do that (lowercasing) when saving `dataset_entries`.

Thus, we will exclude the possibility of uppercase in `dataset_entries.owner_name`.

--

I agree that it's two-way solution, but it seems that this would be more correct.

zaychenko-sergei: Then you need a db migration

s373r: Migrations added -- Done

zaychenko-sergei (approved):

Issue #1263, opened 2025-06-05
`latest-with-data-mt` image: some datasets do not have dependencies


s373r:
### Describe the bug

Within `ghcr.io/kamu-data/kamu-base:latest-with-data-mt` image, it was discovered that some datasets, such as `kamu/account.tokens.portfolio.market-value` have no dependencies:

> The current dataset is derivative. Not enough data to build the graph. 

(c) http://localhost:4200/kamu/account.tokens.portfolio.market-value/lineage

### Steps To Reproduce

1. Start the API server from the image:
```shell
docker run -it --rm -p 8080:8080 -e KAMU_AUTH_GITHUB_CLIENT_ID=1 -e KAMU_AUTH_GITHUB_CLIENT_SECRET=2 ghcr.io/kamu-data/kamu-base:latest-with-data-mt kamu system api-server --address 0.0.0.0 --http-port 8080 --external-address 127.0.0.1
```
2. Run `kamu-web-ui` locally
3. Open http://localhost:4200/kamu/account.tokens.portfolio.market-value/lineage

### Current Behavior

![Image](https://github.com/user-attachments/assets/5b6403ef-b40e-44da-84a7-1274f3da7e50)

(c) http://localhost:4200/kamu/account.tokens.portfolio.market-value/lineage

### Expected Behavior

![Image](https://github.com/user-attachments/assets/be06723c-828a-4f76-94a0-c178fa4f091d)

(c) like at https://platform.demo.kamu.dev/kamu/account.tokens.portfolio.market-value/lineage

### kamu system info

```sh
0.240.1
```

### Anything else?

To dramatically speed up workspace generation, apply the following tricks:

- Apply patch:
```patch
Subject: [PATCH] patches
---
Index: images/kamu-base-with-data-mt/init-workspace.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/images/kamu-base-with-data-mt/init-workspace.py b/images/kamu-base-with-data-mt/init-workspace.py
--- a/images/kamu-base-with-data-mt/init-workspace.py	(revision 2cc7b7813d72f0b6861b2fa482f197a106462e23)
+++ b/images/kamu-base-with-data-mt/init-workspace.py	(date 1749136307412)
@@ -7,7 +7,9 @@
 ###############################################################################
 
 CURRENT_PATH = Path(__file__).resolve().parent
[...]

PR #1365, opened 2025-09-08
`DependencyGraphImmediateListener:` fix a message processing race condition

s373r:
## Description

Notes:
- `DatasetKeyBlockUpdateHandler` is guaranteed to be called before `DependencyGraphImmediateListener`.
  - `DependencyGraphImmediateListener` depends on the results of `DatasetKeyBlockUpdateHandler`, as it relies on key blocks cached in the database.
- `DatasetKeyBlockUpdateHandler` started working earlier, but due to its asynchronous nature, it finished its work later than `DependencyGraphImmediateListener`.

Closes: #1263

- [ ] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,10 +38,14 @@ Recommendation: for ease of reading, use the following order:
 ### Fixed
 - Crash when multiple unlabeled webhook subscriptions are defined within the same dataset
 - Restrict dataset creation with duplicate transform inputs.
-- (#1263) DependencyGraphImmediateListener: fix a message processing race condition.
+- (#1263) `DependencyGraphImmediateListener`: fixed a message processing race condition.
   - Parallel processing message of one type, in the case of handlers that were dependent on each other, 
     could lead to incorrect formation of the upstream/downstream dependency list 
     when pulling/pushing existing remote datasets. 
+- (#1263) `OsoDatasetAuthorizer::classify_dataset_ids_by_allowance()`.
+  - Fixed a bug that returned an empty `unresolved_resources` list 
+    when there were no dataset entries. This could happen when pulling a dataset 
+    that had upstream dependencies but no dependencies itself.
 
 ## [0.247.0] - 2025-08-28
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ Recommendation: for ease of reading, use the following order:
 ### Fixed
 - Crash when multiple unlabeled webhook subscriptions are defined within the same dataset
 - Restrict dataset creation with duplicate transform inputs.
+- (#1263) DependencyGraphImmediateListener: fix a message processing race condition.
+  - Parallel processing message of one type, in the case of handlers that were dependent on each other, 
+    could lead to incorrect formation of the upstream/downstream dependency list 
+    when pulling/pushing existing remote datasets. 
 
 ## [0.247.0] - 2025-08-28
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Recommendation: for ease of reading, use the following order:
 - Flows: each input contribution to batching rule is reflected as an update of start condition,
    so that flow history may show how many accumulated records were present at the moment of each update
 - GQL: `Search::query()`: case insensitive search.
+- (#1263): `images/kamu-base-with-data-mt`: make datasets public by default.
 ### Fixed
 - Crash when multiple unlabeled webhook subscriptions are defined within the same dataset
 - Restrict dataset creation with duplicate transform inputs.
--- a/src/domain/core/src/auth/dataset_action_authorizer.rs
+++ b/src/domain/core/src/auth/dataset_action_authorizer.rs
@@ -269,6 +269,16 @@ impl DatasetActionUnauthorizedError {
             .into(),
         ))
     }
+
+    pub fn unresolved(dataset_ref: odf::DatasetRef, action: DatasetAction) -> Self {
+        Self::Access(odf::AccessError::Unauthorized(
+            DatasetActionUnresolvedError {
+                action,
+                dataset_ref,
+            }
+            .into(),
+        ))
+    }
 }
 
 #[derive(Debug, Error)]
@@ -296,6 +306,13 @@ impl std::fmt::Display for DatasetsActionNotEnoughPermissionsError {
         write!(f, "'")?;
         Ok(())
     }
+}
+
+#[derive(Debug, Error)]
+#[error("User attempts to get '{action}' permission for an unresolved dataset '{dataset_ref}'")]
+pub struct DatasetActionUnresolvedError {
+    pub action: DatasetAction,
+    pub dataset_ref: odf::DatasetRef,
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/domain/core/src/auth/dataset_action_authorizer.rs
+++ b/src/domain/core/src/auth/dataset_action_authorizer.rs
@@ -282,26 +282,6 @@ pub struct DatasetActionNotEnoughPermissionsError {
     pub dataset_ref: odf::DatasetRef,
 }
 
-#[derive(Debug, Error)]
-pub struct DatasetsActionNotEnoughPermissionsError {
-    pub action: DatasetAction,
-    pub dataset_refs: Vec<odf::DatasetRef>,
-}
-
-impl std::fmt::Display for DatasetsActionNotEnoughPermissionsError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "User has no '{}' permission in datasets: '", self.action)?;
-        for (i, item) in self.dataset_refs.iter().enumerate() {
-            if i > 0 {
-                write!(f, ", ")?;
-            }
-            write!(f, "{item}")?;
-        }
-        write!(f, "'")?;
-        Ok(())
-    }
-}
-
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 #[derive(Debug, Error)]
--- a/src/domain/core/src/auth/dataset_action_authorizer.rs
+++ b/src/domain/core/src/auth/dataset_action_authorizer.rs
@@ -253,7 +253,11 @@ impl oso::ToPolar for DatasetAction {
 #[derive(Debug, Error)]
 pub enum DatasetActionUnauthorizedError {
     #[error(transparent)]
-    Access(odf::AccessError),
+    Access(
+        #[from]
+        #[backtrace]
+        odf::AccessError,
+    ),
 
     #[error(transparent)]
     Internal(#[from] InternalError),
@@ -263,16 +267,6 @@ impl DatasetActionUnauthorizedError {
     pub fn not_enough_permissions(dataset_ref: odf::DatasetRef, action: DatasetAction) -> Self {
         Self::Access(odf::AccessError::Unauthorized(
             DatasetActionNotEnoughPermissionsError {
-                action,
-                dataset_ref,
-            }
-            .into(),
-        ))
-    }
-
-    pub fn unresolved(dataset_ref: odf::DatasetRef, action: DatasetAction) -> Self {
-        Self::Access(odf::AccessError::Unauthorized(
-            DatasetActionUnresolvedError {
                 action,
                 dataset_ref,
             }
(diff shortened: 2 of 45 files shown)


src/domain/core/src/auth/dataset_action_authorizer.rs
```diff
@@ -298,6 +308,13 @@ impl std::fmt::Display for DatasetsActionNotEnoughPermissionsError {
     }
 }
 
+#[derive(Debug, Error)]
+#[error("User attempts to get '{action}' permission for an unresolved dataset '{dataset_ref}'")]
```
sergiimk: I don't quite understand this error. If dataset is unresolved (either does not exist or not visible to the subject due to permissions) shouldn't it be some kind of `NotFound` error?

src/domain/datasets/services/src/services/refs/dataset_reference_service_impl.rs
```diff
@@ -79,12 +108,25 @@ impl DatasetReferenceService for DatasetReferenceServiceImpl {
             "Setting dataset reference in persistent storage"
         );
 
-        // Try repository operation
+        // Signal the beginning of the reference update process
+        self.outbox
+            .post_message(
+                MESSAGE_PRODUCER_KAMU_DATASET_REFERENCE_SERVICE,
```
sergiimk: Do I understand this fix correctly?
- `DatasetReferenceMessageUpdated` has two consumers `A` and `B`
- Messages to `A` and `B` are delivered concurrently
- But `B` depends on `A` and needs `A` to apply its state changes **before** `B` starts processing the message
- Fix: Split original message into `Updating` listened by A and `Updated` listened by B to guarantee sequential processing

If so, this probably indicates a problem in where we drew transactional boundaries and dependencies between subsystems.

sergiimk: If both handlers are **immediate** (handle event inside the same transaction as what generated it) - this is less of a concern, but the event splitting solution still doesn't scale to 3+ subscribers.

zaychenko-sergei (changes requested):
As I understood, the problem comes from this block in dependency recompute. Basically, it needs to know fresh `SetTransform`, which might not be indexed yet.

```
pub(crate) async fn extract_modified_dependencies_in_interval(
    metadata_chain: &dyn odf::MetadataChain,
    head: &odf::Multihash,
    maybe_tail: Option<&odf::Multihash>,
) -> Result<DependencyChange, InternalError> {
    let mut new_upstream_ids: HashSet<odf::DatasetID> = HashSet::new();

    let mut set_transform_visitor = odf::dataset::SearchSetTransformVisitor::new();
    let mut seed_visitor = odf::dataset::SearchSeedVisitor::new();

    use odf::dataset::MetadataChainExt;
    metadata_chain
        .accept_by_interval(
            &mut [&mut set_transform_visitor, &mut seed_visitor],
            Some(head),
            maybe_tail,
        )
        .await
        .int_err()?;

    if let Some(event) = set_transform_visitor.into_event() {
        for new_input in &event.inputs {
            if let Some(id) = new_input.dataset_ref.id() {
                new_upstream_ids.insert(id.clone());
            }
        }
    }

    // 3 cases:
    //  - we see `SetTransform` event that is relevant now (changed)
    //  - we don't see it and stop where asked (unchanged)
    //  - we don't see it and reach seed (dropped)
    if !new_upstream_ids.is_empty() {
        Ok(DependencyChange::Changed(new_upstream_ids))
    } else if seed_visitor.into_event().is_some() {
        Ok(DependencyChange::Dropped)
    } else {
        Ok(DependencyChange::Unchanged)
    }
}

```

So, we have a dependency like this:

 reference updated  =>   key blocks recomputed  =>   dependencies updated 
 
However, on today it looks like this:

    reference updated => key blocks recomputed
                                      => dependencies updated
                                      
which causes the race condition in case key block indexes a new item that might affect the dependency.

The solution should be to revise the structure of outbox events and, probably, to introduce a separate event indicating "key blocks introduced". It should only be fired if new content really has new key blocks, skipping data updates.

The dependency listener thus should bind to "key blocks introduced", rather than "reference updated".

I also suggest that you review other listeners of "reference updated" and evaluate the possibility of similar error.

src/domain/datasets/services/src/services/refs/dataset_reference_service_impl.rs
s373r: > Do I understand this fix correctly?

Yes, you have correctly described the root problem and the solution.

> If both handlers are immediate 

Yes, that's exactly our case

> but the event splitting solution still doesn't scale to 3+ subscribers.

At the time of reading, I don't understand why — I thought that adding a new event that would clearly be processed earlier might be the solution to the problem. 

In a [parallel comment](https://github.com/kamu-data/kamu-cli/pull/1365#pullrequestreview-3199854347), an alternative solution is proposed -- I will examine it in light of concerns about the subscribers scaling.

src/domain/core/src/auth/dataset_action_authorizer.rs
s373r: After thinking about it, I agree with you—`NotFound` is suitable here. Additional clarification of the reasons for the error is not particularly important here.

src/domain/datasets/services/src/services/key_blocks/dataset_key_block_indexing_job.rs
```diff
@@ -166,6 +166,7 @@ pub(crate) async fn collect_dataset_key_blocks_in_range(
     // Resulting blocks
     let mut key_blocks = Vec::new();
 
+    // TODO: PERF: iterate only over key blocks from (!) storage
```
s373r: I started implementation, even created `SearchKeyBlocksVisitor`, but got stuck debugging situations where `MetadataChainDatabaseBackedImpl` needs to call iteration, but without cache.

In order not to delay the resolution of the current situation, I suggest we return to this later.

src/domain/datasets/services/src/services/refs/dataset_reference_service_impl.rs
s373r: > we drew transactional boundaries and dependencies between subsystems.

The issue has been solved by introducing a new message `DatasetKeyBlocksMessage::Introduced(_)`

Done

src/domain/core/src/auth/dataset_action_authorizer.rs
s373r: To verify access for a single dataset, the previous error (`DatasetActionUnauthorizedError`) has been retained. 

For multiple checks, `MultipleDatasetActionUnauthorizedError` has been added, which contains `NotFound(odf::DatasetNotFoundError)`.

Done

src/domain/datasets/services/src/services/refs/dataset_reference_service_impl.rs
sergiimk: TBH I'm not a fan of `DatasetKeyBlocksMessage` solution - to me this message looks like leaking a low-level caching concern into top-level interface which is our event bus. It's almost as bad as if we would call `update()` on the key block caching subsystem directly instead of publishiing a message.

I think a better solution could be to make key blocks cache "race-safe" - i.e. it could detect that the cache it has is valid for head H1 and dataset is already on H2, therefore it would need to fall back on full scan between `[H1, H2]`.

Since it's immediate mode and those `DatasetKeyBlocksMessage` events won't appear on cross-component event bus, I'm OK with not delaying the working fix for architectural perfection, but I think it's important we find better ways to deal with such problems in future without introducing multiple "PreUpdate" and "PostUpdate" message types that we will get lost in.

cc @zaychenko-sergei

CHANGELOG.md
```diff
@@ -35,9 +35,18 @@ Recommendation: for ease of reading, use the following order:
...
    so that flow history may show how many accumulated records were present at the moment of each update
 - GQL: `Search::query()`: case insensitive search.
+- (#1263): `images/kamu-base-with-data-mt`: make datasets public by default.
 ### Fixed
 - Crash when multiple unlabeled webhook subscriptions are defined within the same dataset
 - Restrict dataset creation with duplicate transform inputs.
+- (#1263) `DependencyGraphImmediateListener`: fixed a message processing race condition.
+  - Parallel processing message of one type, in the case of handlers that were dependent on each other, 
+    could lead to incorrect formation of the upstream/downstream dependency list 
+    when pulling/pushing existing remote datasets.
```
zaychenko-sergei, line 45: Probably obsolete?

src/domain/core/src/auth/dataset_action_authorizer.rs
```diff
@@ -300,10 +304,41 @@ impl std::fmt::Display for DatasetsActionNotEnoughPermissionsError {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Debug, Error)]
+pub enum MultipleDatasetActionUnauthorizedError {
```
zaychenko-sergei: This looks pretty much individual to me rather then multi.
It's a vector here making the multi-effect, so re-consider naming:
```
    pub unauthorized_handles_with_errors:
        Vec<(odf::DatasetHandle, MultipleDatasetActionUnauthorizedError)>
```

src/domain/datasets/domain/src/messages/dataset_key_blocks_message.rs
```diff
@@ -0,0 +1,69 @@
...
+
+const DATASET_REFERENCE_OUTBOX_VERSION: u32 = 1;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/// Represents messages related to dataset key blocks.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum DatasetKeyBlocksMessage {
+    /// Message indicating that dataset key blocks have been introduced.
+    Introduced(DatasetKeyBlocksMessageIntroduced),
```
zaychenko-sergei: This name makes an impression of "first introduction".
I guess we are rather "appending" than "introducing".

src/domain/datasets/domain/src/messages/dataset_key_blocks_message.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl DatasetKeyBlocksMessage {
+    pub fn introduced(
+        dataset_id: &odf::DatasetID,
+        block_ref: &odf::BlockRef,
+        tail_key_block_hash: &odf::Multihash,
+        head_key_block_hash: &odf::Multihash,
```
zaychenko-sergei: Maybe also a union of flags indicating type of blocks? That would make an easy exit possible without iterating, if we don't have `SetTransform` i.e.

src/domain/datasets/services/src/services/key_blocks/dataset_key_block_update_handler.rs
```diff
@@ -39,21 +42,67 @@ use crate::collect_dataset_key_blocks_in_range;
...
+                );
+                return Ok(());
+            }
+            Err(odf::DatasetRefUnresolvedError::Internal(e)) => Err(e),
+        }?;
+
+        let CollectKeyBlockResponse {
+            key_blocks,
+            divergence_detected: _, // We don't care about divergence here
+        } = self.update_dataset_key_blocks(target, message).await?;
```
zaychenko-sergei: I'm not sure I understand how divergence is handled. But perhaps a valid behavior is to indicate it as a flag and allow to re-compute derived data from scratch. We can later optimize if divergence point is not that far from the end of the chain.

CHANGELOG.md
s373r: No, this is a relevant information -- here I am describing situations (more for the user) in which a fixable problem could be encountered.

src/domain/datasets/domain/src/messages/dataset_key_blocks_message.rs
s373r: A very good idea

src/domain/core/src/auth/dataset_action_authorizer.rs
s373r: Renamed to `ClassifyByAllowanceDatasetActionUnauthorizedError` (yes, Java class name length-like)

Done

src/domain/datasets/services/src/services/key_blocks/dataset_key_block_update_handler.rs
s373r: > I'm not sure I understand how divergence is handled. 

The logic of the methods has not changed since you wrote them. I just added `CollectKeyBlockResponse` from `update_dataset_key_blocks()`.

> But perhaps a valid behavior is to indicate it as a flag and allow to re-compute derived data from scratch

I agree, it would indeed be better if a receiving handler could distinguish between `reset` and `force push`.

Done

src/domain/datasets/domain/src/messages/dataset_key_blocks_message.rs
s373r: Fair enough, I hadn't thought about it that way.

Renamed; Done

src/domain/datasets/domain/src/messages/dataset_key_blocks_message.rs
s373r: Done

PS. `extract_modified_dependencies_in_interval()`: hints optimizations added.

zaychenko-sergei (approved):

src/domain/core/src/auth/dataset_action_authorizer.rs
s373r: UPD: `MultipleDatasetActionUnauthorizedError` -> `ClassifyByAllowanceDatasetActionUnauthorizedError`

src/domain/datasets/services/src/services/refs/dataset_reference_service_impl.rs
s373r: > but I think it's important we find better ways to deal with such problems in future without introducing multiple "PreUpdate" and "PostUpdate" message types that we will get lost in.

I will try to find the best solutions in the future

PR #1366, opened 2025-09-08
Experiment: `BlockRef`: add `CheapClone` trait

s373r:


src/odf/dataset/src/entities/block_ref.rs
```diff
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /// References are named pointers to metadata blocks
-#[derive(Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Debug)]
```
sergiimk: In future I see this enum expanding to support `BlockRef::Named(Arc<str>)`, so don't think it can be `Copy`.

sergiimk: I would really like us to have a way to differentiate `Clone` and `CheapClone` though. I saw multiple threads on Rust Reddit about this, so perhaps we can find a library that provides a trait for cheap cloning and standardise around it?

zaychenko-sergei: Yeah, ref would contain names of branches at a magic future moment.

sergiimk: I suggest we adopt this crate:
https://crates.io/crates/cheap-clone

It's as simple as it gets, and we could easily mark many of our reference / ID types as cheaply clonable.

s373r: Integrated `cheap-clone` crates for `BlockRef` as an experiment.

Pros:
- visual difference on the calling side

Cons:
- It is not possible to highlight all previous existing `clone()` 
  - (`Clone` trait is a super trait for `CheapClone`)

Issue #1363, opened 2025-09-05
Improve vectorization of new GQL endpoints


sergiimk:
This is a continuation of #1307.

In #1359 we made some shortcuts to add new APIs that didn't utilize vectorization when requesting multiple accounts / datasets.

This is a follow up ticket to address the comments made by @zaychenko-sergei on that PR.

The proposal is to unify the vecorized lookup APIs on repository level towards the following shape:

```rust
async fn get_accounts_by_ids(
    &self,
    account_ids: &[odf::AccountID],
) -> Result<BatchLookup<AccountId, Account, AccountNotFound>, InternalError>;

pub struct BatchLookup<By, T, Err> {
  pub found: Vec<T>,
  pub not_found: Vec<(By, Err)>,
}
```

PR #1372, opened 2025-09-10
Generalization of vector operation processing code in repositories

s373r:
## Description

Closes: #1363

- [ ] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/utils/database-common/src/entities.rs
```diff
@@ -168,3 +171,94 @@ pub struct EventModel {
...
+        let mut not_found = Vec::with_capacity(ids.len() - found.len());
+
+        for id in ids {
+            if !found_ids.contains(*id) {
+                let cloned_id = (*id).clone();
+                let not_found_err = (options.not_found_err_fn)(*id);
+
+                not_found.push((cloned_id, not_found_err));
+            }
+        }
```
zaychenko-sergei: O(N^2)

Consider storing `found_ids` in a `HashSet`.

src/adapter/graphql/src/mutations/datasets_mut.rs
```diff
@@ -113,58 +113,55 @@ impl DatasetsMut {
...
-                    if skip_missing {
-                        continue;
-                    }
-                    return Err(GqlError::gql(format!(
-                        "Unresolved dataset: {}",
-                        dataset_request_state.dataset_handle().id
-                    )));
-                }
-            };
+                write!(
```
rmn-boiko: Shouldn't we use here `tracing::error`?

src/domain/datasets/services/src/services/entry/dataset_entry_service_impl.rs
```diff
@@ -159,6 +165,135 @@ impl DatasetEntryServiceImpl {
...
+                    account_name
+                }
+                None => {
+                    single_tenant_count += 1;
+                    self.current_account_subject.account_name_or_default()
+                }
+            })
+            .collect::<Vec<_>>();
+
+        if single_tenant_count > 0 && multi_tenant_count > 0 {
```
rmn-boiko: Maybe we also would like to check if we just got single_tenant_count > 0.
Now it seems we will lookup accounts with such query `[None, None, None]`

s373r: Interesting — I agree, there is room for optimization here

src/utils/database-common/src/entities.rs
s373r: yep, but  `found_ids` is already `HashSet<id>`

s373r: Renamed to `found_ids_set` for additional clarity for the reader

Done

src/domain/datasets/services/src/services/entry/dataset_entry_service_impl.rs
s373r: Done

src/adapter/graphql/src/mutations/datasets_mut.rs
s373r: I don't think so, because in the case of `skip_missing`, we skip those datasets that are not accessible -- for the user, they do not exist (not an app internal error, just a regular use case).

However, there was an error in the case of `!skip_missing && !resolution.forbidden.is_empty()`, but with`resolution.insufficient.is_empty()` -- we did not enter the coded path -- will fix

s373r: `ClassifyDatasetRefsByAccessResponse::try_get_error_message()` has been extracted into a method to remove this logic from the GQL area.

Done

zaychenko-sergei (approved):

PR #1381, opened 2025-09-17
Chore: Transactions management improvements

sergiimk:
This PR enables us to register both `TransactionRef` and `TransactionRefT<DB>` in the catalog, removing a bunch of boilerplate dynamic casting from every repo implementation.


src/utils/database-common/src/transactions/db_transaction.rs
```diff
@@ -0,0 +1,177 @@
...
+    fn register(self: Arc<Self>, catalog_builder: &mut dill::CatalogBuilder) {
+        // Add typed wrapper
+        catalog_builder.add_value(TransactionRefT {
+            inner: Arc::clone(&self),
+        });
+
+        // Add erased wrapper
+        catalog_builder.add_value(TransactionRef { inner: self });
+    }
+}
```
sergiimk, line 144: This is the essential part. We get here via dynamic dispatch from `TransactionRef::register` and knowing real type can register both typed and type-erased wrappers that point to the same state insteance.

Issue #1191, opened 2025-04-13
Regression: Resuming containerized ingest after cancelling


sergiimk:
### Describe the bug

It seems we had a significant regression during migration of CLI to SQLite.

It's related to running entire command in one transaction:
- I have a dataset that fetches several years of history via API (io.codex.tokens.olhcv)
- I have to request data in batches not to hit API limits
- Every X requests ingest container exists and returns `has_more=true` in order to preserve the current progress and go for another loop

Previously I could interrupt the ingest process via Ctrl+X and resume it from the last checkpoint, but now if I interrupt - no progress is saved because transaction is rolled back!

This problem likely also exists for `pull --all`

### Steps To Reproduce

Use some of the iterative ingest datasets e.g. `io.codex.ohlcv`

### Current Behavior

Upon hitting `Ctrl+C` the entire progress is lost.

### Expected Behavior

It should be possible to interrupt ingest loops and recursive pulls without loosing all the progress.

### kamu system info

```sh
latest master
```

### Anything else?

The correct thing to do is likely to extract all looping and recursive operations (ingest + has_more=true, pull --all, etc.) to use-case level and open a transaction per iteration.

PR #1388, opened 2025-09-23
Handle iterative ingest process in separate transactions

rmn-boiko:
## Description

Closes: https://github.com/kamu-data/kamu-cli/issues/1191

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Recommendation: for ease of reading, use the following order:
 ### Fixed
 - `kamu pull`: perform pulling each dataset in separate transaction and allow to interrupt run and some pulls will be saved. 
   It also helps to keep ingest iterations progress saved even if some iteration failed
+- Investigation: potential unstable ordering of dataset entry listings (#1405).
 
 ## [0.249.1] - 2025-09-25
 ### Fixed
--- a/src/infra/datasets/postgres/.sqlx/query-4106a49bfb09634154ebde02f2815f3e8b4720e559b1e71091a4d39d391f8416.json
+++ b/src/infra/datasets/postgres/.sqlx/query-4106a49bfb09634154ebde02f2815f3e8b4720e559b1e71091a4d39d391f8416.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "\n                SELECT dataset_id   as \"id: _\",\n                    owner_id     as \"owner_id: _\",\n                    owner_name,\n                    dataset_name as name,\n                    created_at   as \"created_at: _\",\n                    kind         as \"kind: _\"\n                FROM dataset_entries\n                WHERE owner_id = $1\n                LIMIT $2 OFFSET $3\n                ",
+  "query": "\n            SELECT dataset_id   as \"id: _\",\n                   owner_id     as \"owner_id: _\",\n                   owner_name,\n                   dataset_name as name,\n                   created_at   as \"created_at: _\",\n                   kind         as \"kind: _\"\n            FROM dataset_entries\n            WHERE dataset_id = ANY($1)\n            ORDER BY owner_name, dataset_name\n            ",
   "describe": {
     "columns": [
       {
@@ -46,9 +46,7 @@
     ],
     "parameters": {
       "Left": [
-        "Text",
-        "Int8",
-        "Int8"
+        "TextArray"
       ]
     },
     "nullable": [
@@ -60,5 +58,5 @@
       false
     ]
   },
-  "hash": "07bc67ff4b2e223da7236c8cc0b2b1b38f45b6d1569d63b4dbd16317e48d1bfb"
+  "hash": "4106a49bfb09634154ebde02f2815f3e8b4720e559b1e71091a4d39d391f8416"
 }
--- a/src/infra/datasets/postgres/.sqlx/query-88c65d5e30d6a07aca8d7f00b505f49ba599d2240e45d3d4279b28477b927d74.json
+++ b/src/infra/datasets/postgres/.sqlx/query-88c65d5e30d6a07aca8d7f00b505f49ba599d2240e45d3d4279b28477b927d74.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "\n                SELECT\n                    dataset_id   as \"id: _\",\n                    owner_id     as \"owner_id: _\",\n                    owner_name   as owner_name,\n                    dataset_name as name,\n                    created_at   as \"created_at: _\",\n                    kind         as \"kind: _\"\n                FROM dataset_entries\n                ORDER BY dataset_name ASC\n                LIMIT $1 OFFSET $2\n                ",
+  "query": "\n                SELECT dataset_id   as \"id: _\",\n                    owner_id     as \"owner_id: _\",\n                    owner_name,\n                    dataset_name as name,\n                    created_at   as \"created_at: _\",\n                    kind         as \"kind: _\"\n                FROM dataset_entries\n                WHERE owner_id = $1\n                ORDER BY dataset_name\n                LIMIT $2 OFFSET $3\n                ",
   "describe": {
     "columns": [
       {
@@ -46,6 +46,7 @@
     ],
     "parameters": {
       "Left": [
+        "Text",
         "Int8",
         "Int8"
       ]
@@ -59,5 +60,5 @@
       false
     ]
   },
-  "hash": "3d11399dd129f45de4c9e1e74fad20445276ada75b4fc000d8d5c86b34dc4950"
+  "hash": "88c65d5e30d6a07aca8d7f00b505f49ba599d2240e45d3d4279b28477b927d74"
 }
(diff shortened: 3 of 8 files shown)


src/app/cli/src/commands/pull_command.rs
```diff
@@ -115,19 +115,26 @@ impl PullCommand {
...
+                            self.refs.clone(),
+                            current_account_name,
+                            self.tenancy_config,
+                        )
+                        .try_collect()
+                        .await
+                    } else {
+                        Ok(vec![])
+                    }
+                },
```
zaychenko-sergei: If you move this block to a separate method, you could utilize #[transactional_method2] macro to make it more compact.

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -64,8 +60,25 @@ impl PullDatasetUseCaseImpl {
...
+                            .await?;
+                            let (iteration, read_errors) = Self::make_authorization_read_checks(
+                                iteration,
+                                dataset_action_authorizer,
+                            )
+                            .await?;
+                            Ok((iteration, write_errors, read_errors))
+                        },
+                    )
+                    .await?;
```
zaychenko-sergei: Similar situation, prefer little high-level method calls, where method uses #[transactional_method] decorator

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -106,15 +125,25 @@ impl PullDatasetUseCaseImpl {
...
+                                            psi,
+                                            sync_options,
+                                            sync_svc.clone(),
+                                            dataset_registry.clone(),
+                                            remote_alias_registry.clone(),
+                                            maybe_listener,
+                                        )
+                                        .await
+                                    },
+                                )
```
zaychenko-sergei: Just attach #[transactional_method3] to `sync`. Yes, there is even a 3-arg version of the macro!

src/domain/core/src/entities/resolved_datasets_map.rs
```diff
@@ -63,6 +63,21 @@ impl ResolvedDatasetsMap {
             resolved_dataset.detach_from_transaction();
         }
     }
+
+    fn get_all_values(&self) -> Vec<ResolvedDataset> {
+        self.resolved_datasets_by_id.values().cloned().collect()
+    }
+
+    pub async fn refresh_dataset_from_registry(&mut self, dataset_registry: &dyn DatasetRegistry) {
```
zaychenko-sergei: dataset**s**

sergiimk (commented):
Overall this may be passable for a quick fix, but definitely adds more technical debt than I'd like. Left more thoughts in Slack.

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -374,90 +524,42 @@ impl PullDatasetUseCaseImpl {
...
-                // Elaborate error
-                Err(e) => (
-                    pti.target,
-                    Err(PullError::TransformError(TransformError::Elaborate(e))),
-                ),
-            }
-        }
-
-        let transform_result = run_transform(
+        let transform_result = Self::run_transform_inner(
```
sergiimk: Looks like `transform` method is not doing anything but calling `run_transform_inner` - can those be combined?

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -361,11 +456,66 @@ impl PullDatasetUseCaseImpl {
         })
     }
 
+    #[transactional_static_method1(dataset_registry: Arc<dyn DatasetRegistry>)]
```
sergiimk, line 426: > duplicating question from the design doc not to forget

Let's say I run `pull --all` on a workspace with complex dataset dependency graph.

`PullDatasetUseCaseImpl` builds a high-level plan of jobs, split into batches (unfortunately called "iterations") where every job in a batch is independent of another.

`PullDatasetUseCaseImpl::pull_by_plan` then executes those batches sequentially and runs jobs in each batch in parallel (using `JoinSet`).

This PR seems to open DB transaction per every job right when it starts executing it.

But given the concurrency limitations of SQLite, how does this currently work? Will the parallel jobs be blocked when trying to acquire a DB transaction? Are we actually exploiting parallelism or end up running all jobs effectively sequentially?

zaychenko-sergei: From what I see, it's only locking a transaction for a short periods of time to:
 - obtain initial list of datasets to process
 - prepare plans / elaborate transform plans
 - write resulting reference

The only exception is `sync`. Here we could probably think of more decoupling. The lower levels are not very well prepared to this, since syncs were never really a part of Task system.

zaychenko-sergei (approved):

src/adapter/http/src/smart_protocol/ws_tungstenite_client.rs
```diff
@@ -519,6 +521,59 @@ impl WsSmartTransferProtocolClient {
...
+                    )),
+                    ..Default::default()
+                },
+            )
+            .await
+            .int_err()?;
+
+        let new_dst_head = transactional_target
+            .as_metadata_chain()
+            .resolve_ref(&odf::BlockRef::Head)
```
sergiimk: Doesn't use case return the updated head reference? Kinda odd that we need to read it back from the chain right after writing it.

src/domain/core/src/entities/resolved_datasets_map.rs
```diff
@@ -63,6 +63,21 @@ impl ResolvedDatasetsMap {
             resolved_dataset.detach_from_transaction();
         }
     }
+
+    fn get_all_values(&self) -> Vec<ResolvedDataset> {
+        self.resolved_datasets_by_id.values().cloned().collect()
+    }
+
+    pub async fn refresh_datasets_from_registry(&mut self, dataset_registry: &dyn DatasetRegistry) {
+        for resolved_dataset in self.get_all_values() {
```
sergiimk: Since you are not changing the size of the hash map, nor changing the keys - I think this refresh can be done in place via [`iter_mut`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.iter_mut)

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -361,11 +426,78 @@ impl PullDatasetUseCaseImpl {
...
+            .set_ref(
+                &odf::BlockRef::Head,
+                new_head,
+                odf::dataset::SetRefOpts {
+                    validate_block_present: true,
+                    check_ref_is: Some(Some(old_head)),
+                },
+            )
+            .await
+            .int_err()?;
```
sergiimk: Note: `int_err()` is not great here - we should handle at least the possibility of concurrent `head` update and report this error properly.

Since this issue was not introduced in this PR, mb let's leave a TODO?

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -317,42 +339,85 @@ impl PullDatasetUseCaseImpl {
...
-                    },
+    ) -> Result<PollingIngestResult, PollingIngestError> {
+        let mut combined_result = None;
+        let mut metadata_state = pii.metadata_state.clone();
+
+        loop {
+            match polling_ingest_svc
+                .ingest(
+                    pii.target.clone(),
+                    metadata_state.clone(),
```
sergiimk: Cloning state here and then re-building it from metadata in `update_ref_transactionally` seems quite inefficient.

Previously we were reusing `DataWriter` and its state across all iterations.

Perhaps we should update `PollingIngestResult` to return new state info, so it could be reused for subsequent write without recomputing?

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
```diff
@@ -361,11 +426,78 @@ impl PullDatasetUseCaseImpl {
...
+                    validate_block_present: true,
+                    check_ref_is: Some(Some(old_head)),
+                },
+            )
+            .await
+            .int_err()?;
+
+        if returns_metadata {
+            return Ok(Some(
+                DataWriterMetadataState::build(
```
sergiimk: this has no relation to updating the reference and I assume is only added here to reuse the transaction

ideally it should be removed - I proposed in another comment how we can avoid recomputing writer state

src/adapter/http/src/smart_protocol/ws_tungstenite_client.rs
rmn-boiko: it doesn't but yeah, I agree it makes sense to return it. Changed

src/domain/core/src/entities/resolved_datasets_map.rs
rmn-boiko: good catch. Used `values_mut` as a better solution

src/infra/core/src/use_cases/pull_dataset_use_case_impl.rs
rmn-boiko: removed

zaychenko-sergei (approved):

Issue #1377, opened 2025-09-15
Ensure case-insensitive storage of account names


s373r:
### Context

At the moment, we convert the account name to lowercase in the repositories -- for a more correct implementation of the ODF protocol, we need to keep account names as they are.

Based on a [Slack discussion](https://kamu-data.slack.com/archives/C02FVH5C04E/p1757608592234219?thread_ts=1757549398.416429&cid=C02FVH5C04E)

PR #1436, opened 2025-11-04
Dataset indexes stores  in lowercase representation

rmn-boiko:


src/domain/datasets/services/src/services/entry/dataset_entry_indexer.rs
```diff
@@ -151,12 +151,14 @@ impl DatasetEntryIndexer {
...
-                    .unwrap_or(&DEFAULT_ACCOUNT_NAME)
-                    .clone(),
+                odf::AccountName::new_unchecked(
+                    &dataset_handle
+                        .alias
+                        .account_name
+                        .as_ref()
+                        .unwrap_or(&DEFAULT_ACCOUNT_NAME)
+                        .to_lowercase(),
+                ),
```
s373r, line 161: On the contrary, we do not want to keep them in lower case. There is a ticket on this topic: https://github.com/kamu-data/kamu-cli/issues/1377

PR #1428, opened 2025-10-29
DataFusion `to_table()` UDTF

s373r:
<details>
  <summary>Quick demo</summary>

Negative:
Screenshot From 2025-10-30 01-35-17

Positive:
Screenshot From 2025-10-30 01-36-07

</details>

## Description

Closes: #1408 

### Added
- SQL: Added DataFusion `to_table()` UDTF that projects data from ledger datasets (#1428).

## Checklist before requesting a review

- [x] Unit and integration tests added
  
- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Workspace layout and metadata: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Tracing / [metrics](https://github.com/kamu-data/kamu-standards/blob/master/metrics_design.md): ✅
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
  - [x] [Changelog](./CHANGELOG.md): ✅
    
  - [ ] [Public documentation](https://github.com/kamu-data/kamu-docs/): ✅
  
- [ ] Downstream effects:
    
  - [ ] [kamu-node](https://github.com/kamu-data/kamu-node): ✅
    
  - [ ] [kamu-web-ui](https://github.com/kamu-data/kamu-web-ui): ✅
    
  - [ ] [release-queue](https://github.com/orgs/kamu-data/projects/4/views/1)

PR #1437, opened 2025-11-04
DataFusion `to_table()` UDTF: follow-ups

s373r:
Fixes @sergiimk's comments from #1428

### DIFF
--- a/src/infra/datafusion-udf/tests/tests/test_to_table.rs
+++ b/src/infra/datafusion-udf/tests/tests/test_to_table.rs
@@ -7,13 +7,22 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use datafusion::error::DataFusionError as DFError;
 use datafusion::prelude::*;
 use indoc::{formatdoc, indoc};
 use kamu_core::{GetDataResponse, MockQueryService, QueryService};
 use pretty_assertions::assert_matches;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+static REGULAR_TABLE_DATASET_ID: LazyLock<odf::DatasetID> =
+    LazyLock::new(|| odf::DatasetID::new_seeded_ed25519(b"REGULAR_TABLE_DATASET_ID"));
+static NO_SCHEMA_TABLE_DATASET_ID: LazyLock<odf::DatasetID> =
+    LazyLock::new(|| odf::DatasetID::new_seeded_ed25519(b"NO_SCHEMA_TABLE_DATASET_ID"));
+static UNAUTHORIZED_TABLE_DATASET_ID: LazyLock<odf::DatasetID> =
+    LazyLock::new(|| odf::DatasetID::new_seeded_ed25519(b"UNAUTHORIZED_TABLE_DATASET_ID"));
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
@@ -65,6 +74,7 @@ async fn test_wrong_arguments() {
 async fn test_dataset_not_found() {
     let ctx = create_session_context();
 
+    // Handle
     assert_matches!(
         ctx.sql(indoc!(
             "
@@ -76,6 +86,19 @@ async fn test_dataset_not_found() {
         Err(DFError::External(e))
             if e.to_string() == "Dataset not found: kamu/unknown-table"
     );
+    // ID
+    let not_found_table_dataset_id = odf::DatasetID::new_generated_ed25519().1;
+    assert_matches!(
+        ctx.sql(&formatdoc!(
+            "
+            SELECT *
+            FROM to_table(`{not_found_table_dataset_id}`)
+            ",
+        ))
+        .await,
+        Err(DFError::External(e))
+            if e.to_string() == format!("Dataset not found: {not_found_table_dataset_id}")
+    );
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/domain/core/src/services/query_service.rs
```diff
@@ -93,14 +93,6 @@ pub trait QueryService: Send + Sync {
         options: GetDataOptions,
     ) -> Result<GetDataResponse, QueryError>;
 
-    /// Extended version of [get_data()](Self::get_data()) that allows
-    /// retrieving requested metadata as well.
-    async fn get_data_with_metadata(
-        &self,
-        dataset_ref: &odf::DatasetRef,
-        options: GetDataWithMetadataOptions,
-    ) -> Result<GetDataWithMetadataResponse, QueryError>;
```
s373r, line 102: > `get_data_with_metadata` looks like such a narrow specific method that gathers a bunch of loosely-related information ...
>
> I wonder if it would be better to have `QueryService::get_changelog_projection()` method that encapsulates everything that is currently going on in this UDTF.
>
> Seems like a much simpler API, less prone to misuse.
>
> This would also allow `QueryService` in future to speed up projections e.g. by caching them in a database somewhere - all this logic would stay hidden inside of the service.
(c) https://github.com/kamu-data/kamu-cli/pull/1428#discussion_r2479647940

src/infra/datafusion-udf/src/to_table.rs
s373r, line 123: > There is a ["remote catalog" pattern](https://github.com/apache/datafusion/blob/main/datafusion-examples/examples/remote_catalog.rs#L78) that DF authors suggest to avoid needing `async` calls during query planning.
> 
> I have a very ugly implementation of it in `QueryServiceImpl::resolve_query_state` that needs a re-work.
> 
> It basically relies on:
> - pre-parsing and analyzing SQL query
> - resolving everything that needs `async` before passing query to datafusion
> - having the query planner pull already pre-resolved tables from the session context.
> 
> I don't know if this pattern can be applied to this case.
> 
> Perhaps in `resolve_query_state` we could add handling of table functions to pre-resolve dataset and store it in the session context?
> 
> Can `TableFunctionImpl` even access `SessionContext`?
(c) https://github.com/kamu-data/kamu-cli/pull/1428#discussion_r2479668469

src/infra/datafusion-udf/tests/tests/test_to_table.rs
s373r, line 82: > can you also test that we can query by DID? `to_table('did:odf:...')`
> (c) https://github.com/kamu-data/kamu-cli/pull/1428#discussion_r2479671671

s373r: Added several tests -- Done

src/domain/core/src/services/query_service.rs
s373r: Makes sense. `QueryService::get_changelog_projection()` implemented -- Done

src/infra/datafusion-udf/src/to_table.rs
s373r: > `QueryServiceImpl::resolve_query_state` .

I studied your work when thinking about how to solve my case. You did great work, and your code helped me get up to speed much faster.

> There is a ["remote catalog" pattern](https://github.com/apache/datafusion/blob/main/datafusion-examples/examples/remote_catalog.rs#L78) that DF authors suggest to avoid needing async calls during query planning

This approach will definitely work for `QueryServiceImpl::resolve_query_state()`, but unfortunately not for our case here:
```rust
        let mut ctx = SessionContext::new_with_config_rt(config, runtime);
        let state = ctx.state();

        let statement = state
            .sql_to_statement(
                "SELECT * FROM to_table(`kamu/unauthorized-table`)",
                &state.config().options().sql_parser.dialect,
            )
            .unwrap();
        let references = state.resolve_table_references(&statement).unwrap();

        dbg!(references);
        // [src/infra/core/src/services/query_service_impl.rs:89:9] references = [
        //     Bare {
        //         table: "to_table", // ⚠️ no arguments :(
        //     },
        // ]
```

> Perhaps in `resolve_query_state` we could add handling of table functions to pre-resolve dataset and store it in the session context?

In this case, with `TableFunctionImpl` (trait), we don't have access to full SQL initially, only its function arguments.
Think this doesn't fit the current case. Since the execution chain is `ToTableUdtf` -> `QueryServiceImpl`, not reverse. 

We could apply the proposed approach if we analyzed any SQL (same-named method), tried to find UDTF there, and then manually called it -- but I think this will be excessive in our situation.

> Can `TableFunctionImpl` even access `SessionContext`?

Not directly (rather, `SessionContext` has access to the function). But! If we make a separate `TableProvider` implementation for `ToTableProvider`, we'll get `Session` access in the `scan()` method:

```rust
impl TableProvider for ToTableProvider {
    fn as_any(&self) -> &dyn std::any::Any {
        self
    }

    fn table_type(&self) -> TableType {
        todo!()
    }

    fn schema(&self) -> SchemaRef {
        todo!()
    }

    async fn scan(
        &self,
        state: &dyn Session,
        projection: Option<&Vec<usize>>,
        filters: &[Expr],
        limit: Option<usize>,
    ) -> Result<Arc<dyn ExecutionPlan>> {
        todo!()
    }
}
```

One of my first implementations did exactly this. But had to abandon this since we need to know the schema immediately BEFORE calling `scan()`, which we couldn't afford due to the synchronous method (`TableFunctionImpl::call(&self, args: &[Expr]) -> DfResult<Arc<dyn TableProvider>>`).

sergiimk (approved):
lgtm 👍

src/infra/datafusion-udf/src/to_table.rs
sergiimk: Yeah it's a tricky issue. We survived for a long time with blocking->async call hacks before, so probably will be fine with this one too. Let's see if Datafusion improvement will be implemented soon enough before thinking of workarounds.

PR #1455, opened 2025-11-18
SKU/molecule, Phase 2

s373r:


resources/schema.gql
```diff
@@ -3065,28 +3107,205 @@ type ModifyPasswordWrongOldPassword implements ModifyPasswordResult {
...
+	dataset: Dataset!
+	byId(id: String!): MoleculeAnnouncementEntryV2!
+}
+
+type MoleculeDataRoomDatasetV2 {
+	"""
+	Access the underlying core Dataset
+	"""
+	dataset: Dataset!
+	entries(pathPrefix: CollectionPath, maxDepth: Int, page: Int, perPage: Int): MoleculeDataRoomEntryV2Connection!
```
sergiimk: Data room entries listing should also have `filters` parameter to enable filtering by `tags`, `categories`.

Additionally I think we should include `accessLevel` into the filters too. Currently Molecule fetches all entries and filters them on their end based on what access level user has, but this doesn't work well with pagination. I think instead they should be telling us what access levels are allowable and we will account for those in filtering on our side.

resources/schema.gql
```diff
@@ -3065,28 +3107,205 @@ type ModifyPasswordWrongOldPassword implements ModifyPasswordResult {
...
+		attachments: [String!],		moleculeAccessLevel: String!,		moleculeChangeBy: String!,		categories: [String!]!,		tags: [String!]!
+	): CreateAnnouncementResult!
+}
+
+type MoleculeAnnouncementsDatasetV2 {
+	"""
+	Access the underlying core Dataset
+	"""
+	dataset: Dataset!
+	byId(id: String!): MoleculeAnnouncementEntryV2!
```
sergiimk: This type is missing `tail()` endpoint to allow listing the latest announcements with filters

s373r: Done

PR #1467, opened 2025-11-28
Molecule APIv2: activity

s373r:

PR #1470, opened 2025-12-01
Molecule APIv2: global data room activity finalization

s373r:
Follow-ups for:
- https://github.com/kamu-data/kamu-cli/pull/1467

### DIFF
--- a/src/adapter/graphql/src/queries/datasets/adapters/collection_entry.rs
+++ b/src/adapter/graphql/src/queries/datasets/adapters/collection_entry.rs
@@ -14,7 +14,7 @@ use crate::queries::Dataset;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Clone)]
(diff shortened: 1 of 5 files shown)


src/adapter/graphql/src/mutations/molecule_mut/v2/molecule_data_room_mut_v2.rs
```diff
@@ -546,6 +546,81 @@ impl MoleculeDataRoomMutV2 {
...
     }
+
+    async fn append_global_data_room_activity(
+        &self,
+        ctx: &Context<'_>,
+        inserted_records: Vec<(
+            odf::metadata::OperationType,
+            kamu_datasets::CollectionEntryRecord,
+        )>,
+        activity_type: MoleculeDataRoomFileActivityType,
```
zaychenko-sergei, line 557: What a nice boundary for the use case. Why not to push most of it there

PR #1473, opened 2025-12-04
Feature/schema migrations

sergiimk:
## Description

Adresses: kamu-data/planning-internal#108

- `odf::DataSchema` now supports a `diff()` method that allows detailed comparisons between two schemas
- `SetDataSchema` event adds a limited supports for schema migrations, allowing to extend the schema with new optional columns

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [x] Observability:
    
  - [x] Alerts: ✅
- [x] Documentation:
    
- [x] Downstream effects:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Recommendation: for ease of reading, use the following order:
 ## [Unreleased]
 ### Added
 - GQL: `set_trigger` allows to schedule `HARD_COMPACTION` flows
+- `odf::DataSchema` now supports a `diff()` method that allows detailed comparisons between two schemas
+- `SetDataSchema` event adds a limited supports for schema migrations, allowing to extend the schema with new optional columns
 ### Changed
 - Refactoring: `ResolvedDataset` and `DatasetRegistry` moved to `kamu-datasets` domain
 - Experiment: Introduce `WriteCheckedDataset` and `ReadCheckedDataset` types, 


src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
```diff
@@ -856,6 +856,47 @@ impl<'a> ValidateSetDataSchemaVisitor<'a> {
...
+            DataSchemaDiffItem::FieldAdded { field_rhs } => {
+                if field_rhs.is_optional() {
+                    // It's OK to introduce a new optional field as this maintains compatibility
+                    // from publisher and consumer perspective
+                    true
+                } else {
+                    false
+                }
+            }
+            DataSchemaDiffItem::FieldRemoved => false,
```
zaychenko-sergei, line 880: Is field removal really a breaking change? Maybe we could just ignore the old redundand field in the old records?

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
```diff
@@ -856,6 +856,47 @@ impl<'a> ValidateSetDataSchemaVisitor<'a> {
...
+                index_rhs: _,
+                field_lhs: _,
+                field_rhs: _,
+            } => true,
+            DataSchemaDiffItem::FieldTypeChanged => false,
+            // TODO: Field becoming optional is relaxing the schema and a compatibile change from
+            // producer perspective, but it's a breaking change from consumer perspective and needs
+            // to be cautioned about and communicated downstream
+            DataSchemaDiffItem::FieldBecameOptional => false,
+            DataSchemaDiffItem::FieldBecameRequired => false,
```
zaychenko-sergei, line 892: Field becoming required does not sound like a breaking change from consumers point of view.

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
```diff
@@ -856,6 +856,47 @@ impl<'a> ValidateSetDataSchemaVisitor<'a> {
...
+                }
+            }
+            DataSchemaDiffItem::FieldRemoved => false,
+            DataSchemaDiffItem::FieldReordered {
+                index_lhs: _,
+                index_rhs: _,
+                field_lhs: _,
+                field_rhs: _,
+            } => true,
+            DataSchemaDiffItem::FieldTypeChanged => false,
```
zaychenko-sergei, line 887: Maybe there are exceptions to this, such as numbers casting?

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
```diff
@@ -856,6 +856,47 @@ impl<'a> ValidateSetDataSchemaVisitor<'a> {
...
+    fn check_allowed_schema_migration_rec(diff: &DataSchemaDiffItem<'_>) -> bool {
+        #[expect(clippy::match_same_arms)]
+        match diff {
+            DataSchemaDiffItem::SchemaAttributesChanged => true,
+            DataSchemaDiffItem::FieldAttributesChanged => true,
+            DataSchemaDiffItem::FieldAdded { field_rhs } => {
+                if field_rhs.is_optional() {
+                    // It's OK to introduce a new optional field as this maintains compatibility
+                    // from publisher and consumer perspective
+                    true
```
zaychenko-sergei, line 875: I understand consumer's view, but not publisher's.
I.e., what do we do, if we attempt to use v2 API to write to non-migrated data room?
I could imagine that publisher pushes new optional fields to a dataset with an old schema, and they will be just ignored, but the publisher still needs to find out of incompatibility somehow

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
sergiimk: I want to remove migration restrictions one small step at a time after carefully thinking through all ramifications.

At present we only need to be able to add optional columns, so I'd rather forbid field removal until it is needed and then add sufficient test coverage.

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
sergiimk: But it is from a writer (ingest) perspective.

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
sergiimk: Yes, expanding a type e.g. from `i32` to `i64` is something we could allow in future, but no point spend time on it now.

src/odf/dataset-impl/src/entities/metadata_chain_validators.rs
sergiimk: > I.e., what do we do, if we attempt to use v2 API to write to non-migrated data room?

This situation should not happen in `prod` - we will run migrations before re-enabling access token.

But in a general case I think we need a more expressive ingest API that allows to:
- Control whether to return an error if input data has fields that are not present in the schema, or to ignore them
- Assert that schema is exactly the same as what writer expects (e.g. `ODF-If-Schema-Hash-Matches: f16..c77`) aborting ingest if it isn't

PR #1475, opened 2025-12-04
MoleculeEncryptionMetadata

s373r:


src/sku/molecule/domain/src/snapshots/molecule_dataset_snapshots.rs
```diff
@@ -164,6 +164,7 @@ impl MoleculeDatasetSnapshots {
         DatasetSnapshots::versioned_file(
             alias,
             vec![
+                // TODO: NOT NULL cols?
```
zaychenko-sergei: That would make the change breaking, we only support adding optional columns for automatic schema migration

src/adapter/graphql/src/queries/molecule/v2/molecule_scalars_v2.rs
```diff
@@ -41,7 +43,80 @@ pub type MoleculeTag = String;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// TODO: scalar?
-pub type EncryptionMetadata = String;
+// TODO: break into scalar (w/o serde) & domain struct (w/ serde) -->
+// TODO: clarify valid values
+#[derive(Clone, Debug, InputObject, serde::Deserialize, serde::Serialize, SimpleObject)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[graphql(input_name = "MoleculeEncryptionMetadataInput")]
```
zaychenko-sergei: Does this trick allow to avoid duplicating input struct? We need to use it everywhere then... in separate ticket of course.

src/adapter/graphql/src/queries/molecule/v2/molecule_scalars_v2.rs
```diff
@@ -41,7 +43,80 @@ pub type MoleculeTag = String;
...
+    pub version: u32,
+    pub record: MoleculeEncryptionMetadata,
+}
+
+impl MoleculeEncryptionMetadataRecord {
+    pub fn new(record: MoleculeEncryptionMetadata) -> Self {
+        Self { version: 0, record }
+    }
+}
+
```
zaychenko-sergei: Move it to domain level from the beginning.

src/adapter/graphql/src/queries/molecule/v2/molecule_scalars_v2.rs
```diff
@@ -41,7 +43,80 @@ pub type MoleculeTag = String;
...
+                serde_json::from_str(&raw_json_as_str).map_err(serde::de::Error::custom)?;
+
+            Ok(Self {
+                version: helper.version,
+                record: helper.record,
+            })
+        }
+    }
+}
+// <-- TODO: break into scalar (w/o serde) & domain struct (w/ serde)
```
zaychenko-sergei: Do we really need this complicated serialization? Feels like it's doable with flattening attributes without any helpers.

src/sku/molecule/domain/src/snapshots/molecule_dataset_snapshots.rs
s373r: In the case of a Molecule versioned file, yes

But in that case, it turns out I need to remove the strict schemas (`SetDataSchema`) that I added earlier

src/adapter/graphql/src/queries/molecule/v2/molecule_scalars_v2.rs
s373r: If the structure is the same, then yes, it does

Will create a ticket

s373r: Cross-linking: https://github.com/kamu-data/kamu-cli/issues/1479

src/adapter/graphql/src/queries/molecule/v2/molecule_scalars_v2.rs
s373r: > Do we really need this complicated serialization?

Yes, it is necessary for ingesting into a dataset. 

If we don't use serialization in a string containing JSON, but simply use JSON, we will get an error when inserting
(long story short: attempt to write an object (map?), although a string is expected (expected in column)):

```
2025-12-05T10:56:32.563825Z ERROR Gql::request:Gql::MoleculeDataRoomMutV2::finish_upload_file:UpdateVersionedFileUseCaseImpl::execute{id=did:odf:fed0147baf3c2591cd68bf0580dd15b830bee085f863bc09301f653b68d3a525ab383}:PushIngestDataUseCaseImpl::execute{dataset_id=did:odf:fed0147baf3c2591cd68bf0580dd15b830bee085f863bc09301f653b68d3a525ab383}:execute_ingest{target=molecule.vitafast/foo.txt data_source=Buffer(..)}:
kamu::services::ingest::push_ingest_executor_impl: Ingest iteration failed error=
Internal(InternalError { source: ArrowError(JsonError("whilst decoding field 'encryption_metadata': expected string got {\"record\": {\"accessControlConditions\": \"EM2\"\"chain\": \"EM5\"\"contractVersion\": \"EM9\"\"dataToEncryptHash\": \"EM1\"\"encryptedAt\": \"EM4\"\"encryptedBy\": \"EM3\"\"litNetwork\": \"EM7\"\"litSdkVersion\": \"EM6\"\"templateName\": \"EM8\"}\"version\": 0}"), Some("")), context: None, backtrace: Some(<disabled>) }) error_msg=Internal error
```

> Feels like it's doable with flattening attributes

Unfortunately, we can't do without a helper structure and custom logic

--

I think I need to emphasize more clearly that this is a serialization into a JSON string, not a JSON object, so that readers don't have similar questions

src/sku/molecule/domain/src/snapshots/molecule_dataset_snapshots.rs
s373r: The TODO removed

src/adapter/graphql/src/queries/molecule/v2/molecule_scalars_v2.rs
s373r: Done

Issue #1505, opened 2025-12-17
`DatasetEntryServiceImpl::resolve_dataset_handles_by_refs()`: correct alias resolution in multi-tenant mode


s373r:
### Expectations

- MT: The absence of an account name in an alias should result in the use of the current subject name:
  - Check `resolve_account_ids_by_maybe_names()`
  - Check `group_aliases_by_resolved_owner_account_name()`
- MT: This GQL request should not result in errors:
```gql
{ datasets { byRef(datasetRef: "data-room-activity") { id } } }
```

PR #1529, opened 2025-12-23
`DatasetEntryServiceImpl::resolve_dataset_handles_by_refs()`: correct alias resolution in multi-tenant mode

s373r:
## Description

Closes: #1505 

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/domain/datasets/services/tests/tests/services/test_dataset_entry_service.rs
```diff
@@ -1,320 +0,0 @@
-// Copyright Kamu Data, Inc. and contributors. All rights reserved.
```
s373r, line 1: renamed

src/domain/datasets/services/tests/tests/services/test_dataset_entry_service_impl.rs
```diff
@@ -0,0 +1,766 @@
...
+                assert_eq!(
+                    indoc::indoc!(
+                        r#"
+                        resolved_handles:
+                        - dataset-1: dataset-1
+                        - dataset-2: dataset-2
+                        - dataset-3: dataset-3
+                        - dataset-4: dataset-4
+                        - dataset-5-not-created: dataset-5-not-created
+                        - dataset-6-not-created: dataset-6-not-created
```
s373r, line 431: Comment for readers: 

The tests highlighted an interesting situation: in the case of resolving handles, we assume that if a handle present, then everything is ready (or have already resolved), but in the case of non-existent handles, this gives an interesting result

sergiimk: How would a non-existing handle appear in the real scenario? We should guarantee that a `DatasetHandle` can only be created for a dataset that exists and is accessible during the current transaction.

sergiimk (commented):
This is probably OK for now, but I'd really like us to soon refactor the core identity types to completely remove the ambiguity of `Option<AccountName>` from lower system levels.

This ties into two changes:
- Having dataset handle contain `(dataset_id, dataset_name, owner_account_id, owner_account_name)` quad (requires denormalizing DB a little)
- Having all code always operate in multi-tenant mode so that account IDs and names are never optional

This will leave optionality of account name only at the human input level - in `DatasetRef`.

src/domain/datasets/services/tests/tests/services/test_dataset_entry_service_impl.rs
zaychenko-sergei: Invalid handle could appear if, i.e., dataset is deleted or renamed, in the middle of another operation that resolved the handle before the handle-destructing operation started.

PR #1561, opened 2026-02-04
Molecule APIv2: global activity async ingesting of dataset rows

s373r:
## Description

Closes: https://github.com/kamu-data/planning-internal/issues/131

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    
### DIFF
--- a/src/sku/molecule/services/src/services/implementation/molecule_async_global_activity_writer.rs
+++ b/src/sku/molecule/services/src/services/implementation/molecule_async_global_activity_writer.rs
@@ -74,7 +74,7 @@ impl MoleculeAsyncGlobalActivityWriter {
                     // NOTE: It's a tradeoff/compromise, but it seems better to have
                     //       a few kilobytes more data written than to cause an error
                     //       in the outbox dispatcher and stop processing the message queue.
-                    ignore_quota_check: true,
+                    skip_quota_check: true,
                 },
             )
             .await
--- a/src/adapter/graphql/src/mutations/molecule_mut/v1/molecule_mut_v1.rs
+++ b/src/adapter/graphql/src/mutations/molecule_mut/v1/molecule_mut_v1.rs
@@ -202,7 +202,7 @@ impl MoleculeProjectMut {
                     is_ingest_from_upload: false,
                     media_type: Some(file_utils::MediaType::NDJSON.to_owned()),
                     expected_head: None,
-                    ignore_quota_check: false,
+                    skip_quota_check: false,
                 },
                 None,
             )
--- a/src/adapter/http/src/data/ingest_handler.rs
+++ b/src/adapter/http/src/data/ingest_handler.rs
@@ -148,7 +148,7 @@ pub async fn dataset_ingest_handler(
                 is_ingest_from_upload,
                 media_type,
                 expected_head: None,
-                ignore_quota_check: false,
+                skip_quota_check: false,
             },
             None,
         )
--- a/src/app/cli/src/commands/ingest_command.rs
+++ b/src/app/cli/src/commands/ingest_command.rs
@@ -183,7 +183,7 @@ impl Command for IngestCommand {
                         is_ingest_from_upload: false,
                         media_type: self.get_media_type()?,
                         expected_head: None,
-                        ignore_quota_check: false,
+                        skip_quota_check: false,
                     },
                     listener.clone(),
                 )
--- a/src/domain/core/src/services/ingest/push_ingest_planner.rs
+++ b/src/domain/core/src/services/ingest/push_ingest_planner.rs
@@ -46,7 +46,7 @@ pub struct PushIngestOpts {
     /// Expected head block to prevent concurrent updates
     pub expected_head: Option<odf::Multihash>,
     /// Skip the account quota check after ingesting.
-    pub ignore_quota_check: bool,
+    pub skip_quota_check: bool,
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/domain/core/src/use_cases/push_ingest_data_use_case.rs
+++ b/src/domain/core/src/use_cases/push_ingest_data_use_case.rs
@@ -53,7 +53,7 @@ pub struct PushIngestDataUseCaseOptions {
     pub is_ingest_from_upload: bool,
     pub media_type: Option<MediaType>,
     pub expected_head: Option<odf::Multihash>,
-    pub ignore_quota_check: bool,
+    pub skip_quota_check: bool,
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/domain/datasets/services/src/use_cases/collections/update_collection_entries_use_case_impl.rs
+++ b/src/domain/datasets/services/src/use_cases/collections/update_collection_entries_use_case_impl.rs
@@ -246,7 +246,7 @@ impl UpdateCollectionEntriesUseCase for UpdateCollectionEntriesUseCaseImpl {
                         is_ingest_from_upload: false,
                         media_type: Some(MediaType::NDJSON.to_owned()),
                         expected_head: Some(expected_head_value),
-                        ignore_quota_check: false,
+                        skip_quota_check: false,
                     },
                     None,
                 )
--- a/src/domain/datasets/services/src/use_cases/versioned_files/update_versioned_file_use_case_impl.rs
+++ b/src/domain/datasets/services/src/use_cases/versioned_files/update_versioned_file_use_case_impl.rs
@@ -170,7 +170,7 @@ impl UpdateVersionedFileUseCase for UpdateVersionedFileUseCaseImpl {
                     is_ingest_from_upload: false,
                     media_type: Some(MediaType::NDJSON.to_owned()),
                     expected_head,
-                    ignore_quota_check: false,
+                    skip_quota_check: false,
                 },
                 None,
             )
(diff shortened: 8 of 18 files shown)


src/sku/molecule/services/src/use_cases/activities/molecule_view_project_activities_use_case_impl.rs
```diff
@@ -371,34 +371,40 @@ impl MoleculeViewProjectActivitiesUseCaseImpl {
...
+                    schema_name = hit.schema_name
+                ))),
+            })
+            .collect::<Result<Vec<_>, InternalError>>()?;
+
+        let total_count = if let Some(total_hits) = search_results.total_hits {
+            usize::try_from(total_hits).unwrap()
+        } else {
+            // NOTE: fallback
+            list.len()
```
zaychenko-sergei, line 404: Unless you are doing vector/hybrid search, you can be sure the `total_count` is initialized. For filtering or full-text matches, you can safely assert the total hits are available.

src/sku/molecule/services/src/use_cases/activities/molecule_view_global_activities_use_case_impl.rs
```diff
@@ -273,34 +270,40 @@ impl MoleculeViewGlobalActivitiesUseCaseImpl {
...
+                ))),
+            })
+            .collect::<Result<Vec<_>, InternalError>>()?;
+
+        let total_count = if let Some(total_hits) = search_results.total_hits {
+            usize::try_from(total_hits).unwrap()
+        } else {
+            // NOTE: fallback
+            list.len()
+        };
```
zaychenko-sergei, line 304: Same point

src/sku/molecule/services/src/use_cases/activities/molecule_append_global_data_room_activity_use_case_impl.rs
```diff
@@ -0,0 +1,91 @@
...
+    ) -> Result<(), MoleculeAppendDataRoomActivityError> {
+        // Check dataset access before sending the message
+        let _ = self
+            .global_data_room_activities_service
+            .writer(&molecule_subject.account_name, true) // TODO: try to create once as start-up job?
+            .await
+            .map_err(MoleculeDatasetErrorExt::adapt::<MoleculeAppendDataRoomActivityError>)?;
+
+        self.ensure_within_quota(molecule_subject, &activity_record)
+            .await?;
```
zaychenko-sergei, line 69: Hmm, I'm not sure. Let's say we are near the limit before the entry operation started. We still have enough quota to write a data room operation, but don't have enough quota for global activity, and this gives us inconsistent state.

From the user's point of view, writing to data room and replicating the activity record to global actrivity is a part of the same logical transaction, so we either have a quota to do it as a whole, or we block the whole operation.

Possible solutions:
 - ignore quotas in global activity writes (we will exceed quota a little bit possibly, but at least we would avoid getting an inconsistent state)
 - make quotas checked at a higher level to include all changes we make to all datasets at once, and do that before any write operation starts  (complete solution, but costlier)

zaychenko-sergei: BTW, by the time the outbox handler does it's job, the quota might be broken already with a different write operation :-)

src/sku/molecule/services/src/services/implementation/molecule_async_global_activity_writer.rs
```diff
@@ -9,100 +9,73 @@
...
         };
 
+        // Gain write access to global activities dataset
+        let global_data_room_activities_writer = self
+            .global_data_room_activities_service
+            .writer(molecule_subject_account_name, true) // TODO: try to create once as start-up job?
+            .await
+            .int_err()?;
+
+        // TODO: several attempts
```
zaychenko-sergei: Yes, we need to do retries, but if N retries fail, don't consider this as a success.
This would fail the outbox channel on this receiver, and would be reattempted on server restart.

src/sku/molecule/domain/src/entities/molecule_data_room_activity.rs
```diff
@@ -184,6 +184,22 @@ pub struct MoleculeDataRoomActivityPayloadRecord {
...
+            + self.access_level.len()
+            + self.content_type.as_ref().map_or(0, |s| s.0.len())
+            + self.description.as_ref().map_or(0, String::len)
+            + self.categories.iter().fold(0, |acc, s| acc + s.len())
+            + self.tags.iter().fold(0, |acc, s| acc + s.len());
+
+        stack_size + heap_size
+    }
+}
+
```
zaychenko-sergei, line 210: Probably Parquet would size it differently anyway. 
I think we should be clear this is a rough estimate, not a precise calculation, and emphasize that in the name of the method.

src/sku/molecule/domain/src/entities/molecule_data_room_activity.rs
```diff
@@ -45,7 +45,7 @@ pub struct MoleculeDataRoomActivity {
 
     pub version: u32,
 
-    // NOTE: This should be odf::AccountID, but kept as String for safety.
+    // NOTE: This should be odf::AccountID but kept as String for safety.
```
zaychenko-sergei, line 48: I think the note is wrong, change_by is usually a wallet address in Molecule case.

zaychenko-sergei (commented):
Good work, but raised  a few questions.

src/sku/molecule/domain/src/entities/molecule_data_room_activity.rs
s373r: > usually a wallet address in Molecule case.

Yes, `odf::AccountID` (via `DidPkh`) supports eth addresses

But if there is a Solana address there, we won't be able to use it

src/sku/molecule/domain/src/entities/molecule_data_room_activity.rs
s373r: > Probably Parquet would size it differently anyway.

Definitely. But we don't mention Parquet size in the method name -- since this is the size of the payload, I thought earlier that this would be sufficient.

However, I'll think about what name would be best for this method, since it has raised questions.

src/sku/molecule/services/src/use_cases/activities/molecule_view_project_activities_use_case_impl.rs
s373r: I understand—my change is made solely from a typing perspective, since `total_count` is an optional value.

Q: I think we can leave it as it is?

(waits for reply)

src/sku/molecule/services/src/use_cases/activities/molecule_view_global_activities_use_case_impl.rs
s373r: https://github.com/kamu-data/kamu-cli/pull/1561#discussion_r2774746477

src/sku/molecule/services/src/use_cases/activities/molecule_append_global_data_room_activity_use_case_impl.rs
s373r: > BTW, by the time the outbox handler does it's job, the quota might be broken already with a different write operation :-)

Yes, this is a problematic point of implementation.

> ignore quotas in global activity writes (we will exceed quota a little bit possibly, but at least we would avoid getting an inconsistent state)

It seems that we need to proceed as follows in this situation: if we have already reached this point, then allow the entry regardless of the quota values.

Yes, it's a tradeoff/compromise, but it seems better to have a few kilobytes more data written than to cause an error in the outbox dispatcher and stop processing the message queue.

s373r: > make quotas checked at a higher level to include all changes

I like this idea, it allows for a more atomic decision (fail fast), but at this stage, I'm afraid of getting bogged down in details -- it seems to me that in this situation, we'd be better off opting for faster iteration in delivery.

src/sku/molecule/services/src/services/implementation/molecule_async_global_activity_writer.rs
s373r: We discussed it in parallel on Slack and came to the general conclusion that we don't need it in this situation

The comment will be updated to record the context of this decision and guidance if the implementation changes.

s373r: Done -- a comment added

src/sku/molecule/domain/src/entities/molecule_data_room_activity.rs
s373r: Done -- `roughly_estimated_size_in_bytes()` name is used

src/sku/molecule/services/src/use_cases/activities/molecule_append_global_data_room_activity_use_case_impl.rs
s373r: Done -- introduced (and used) `skip_quota_check` flag to bypass quota check

src/sku/molecule/services/src/use_cases/activities/molecule_view_project_activities_use_case_impl.rs
zaychenko-sergei: A super correct approach would be to have separate response structures, where listing & full-text would return non-optional value, and vector & hybrid - an optional one. Okay, let's skip. I'm not a fan of this default.

zaychenko-sergei (approved):

PR #1563, opened 2026-02-05
Backport: DataFusion `to_table()` UDTF

s373r:


src/infra/core/src/services/query_service_impl.rs
```diff
@@ -343,6 +344,36 @@ impl QueryServiceImpl {
...
+            dataset_registry: self.dataset_registry.clone(),
+            dataset_action_authorizer: self.dataset_action_authorizer.clone(),
+        });
+        let query_dataset_data_use_case = Arc::new(QueryDatasetDataUseCaseImpl::new(
+            self_as_arc,
+            self.dataset_registry.clone(),
+            rebac_dataset_registry_facade,
+        ));
+
+        kamu_datafusion_udf::ToTableUdtf::register(&ctx, query_dataset_data_use_case);
```
sergiimk: Looks like there's a lot of dependency inversions going on here:
- Most our services AFAIK assume that authorization was done by the upper Use Case layer and don't check permissions themselves
  - Here we have a service that historically was checking permissions, because we don't always know which datasets some SQL query touches
- We also have a Service registering a UDTF (which is an SQL-level concept) by passing a Use Case (which is a layer above services) ... so we are **jumping across three layers** in a wrong direction
  - I think this works only because it's all in `kamu` core module, but would be impossible to split into different crates

**Some thoughts:**
I'm not clear why UDTF needs a use case - is it also authN related?

To separate auth from query service I see two options:

1) Prepare / execute:
- Upper layer calls `QueryService::plan_query(sql)` and gets a list of datasets it touches
- Checks user has access to all datasets
- Calls `QueryService::execute_query(plan)`

2) Auth callback:
- `QueryService` injects an associated `QueryAuthCallback` interface
- It is implemented in upper levels using `RebacDatasetRegistryFacade`
- The callback is used during query execution
- It allows us to break the cyclical dependency by not referring to Auth or Use Case components directly from the service level

s373r: > I'm not clear why UDTF needs a use case - is it also authN related?

For the most part, I was simply trying to adapt a previous solution that required changes after various optimizations and refactorings that occurred after the initial implementation, which was not backported in time.

But yes, I completely agree with the comment about layers.

> To separate auth from query service I see two options:
>    Prepare / execute:

We have previously discussed various reasons for choosing this particular implementation path. Let me remind you of the context: https://github.com/kamu-data/kamu-cli/pull/1437#discussion_r2492120628

Personally, I am eagerly awaiting the transition to a remote catalog approach, but this is a separate major task that may not affect the current implementation due to the existing interface (sync not async), which imposes additional complications.

> Auth callback:

I chose this approach (or a similar one) because it allows us to obtain a ready-made solution, and a relatively simple one at that.

--

Fixed

sergiimk: Thanks for the refactoring. Indeed there's so many things tied to this that my brain fails to retain all the context.

I still don't like `QueryService` knowing about Rebac layer and using `use_case::helpers` from a layer that is above it, but I think this connection will be easy to sever later if need.

I'm happy for this to go in as-is.

s373r: > I still don't like `QueryService` knowing about Rebac layer and using `use_case::helpers` from a layer that is above it

Yes, for sure, it's a half measure.

But this will make it easier for us to improve the code in the future.

PR #1622, opened 2026-04-30
EIP-712 Message Signing Endpoint

s373r:
Related to: https://github.com/kamu-data/planning-internal/issues/134

:warning: A rough draft, but it seems to work


.cargo/config.toml
```diff
@@ -10,3 +10,11 @@ feature-unification = true
...
 # See: https://www.reddit.com/r/rust/comments/1qzvkwf/workspace_feature_permutations_hell/
 feature-unification = "workspace"
+
+[alias]
+# Reduces recompilations due to invalidated cache between `cargo build` vs. `cargo clippy`
+# by separating `target` directories
+# See: https://github.com/rust-lang/cargo/issues/9280
+#      https://github.com/rust-lang/cargo/issues/15422
+#      (no solution, but there is a description of the root cause)
+clippy-target-dir = "clippy --target-dir target/clippy"
```
sergiimk, line 20: Ugh, I'm worried this will increase the already huge disk space usage. Both issues you linked are closed, so this is not really an acknowledged upstream bug and other people seem to be finding workarounds.

I don't have time to dive into this now, but would suggest removing this from SKU and opening a separate PR on master where we could discuss why the compilation flags between `build` and `clippy` differ.

src/adapter/http/src/data/sign/eip712/sign_eip712_handler.rs
```diff
@@ -0,0 +1,155 @@
...
+        let ed25519_private_key = secret_key
+            .get_decrypted_private_key(encryption_key)
+            .int_err()?;
+        // TODO: SEC: Molecule: Phase 3: PK zeroing after usage
+
+        let request_hash = typed_data.signing_hash_with_eip191_prefix()?;
+
+        // TODO: Molecule: Phase 3: add verification key to response?
+        // let _verification_key =
+        // odf::metadata::DidKey::new_ed25519(&private_key.verifying_key());
```
sergiimk, line 99: Agree, it would be nice to include `verificationMethod` for the top-level signature just like you already do for attestation.

sergiimk (commented):
Very nice. Thanks for leaving informative TODOs 👍 💯

.cargo/config.toml
s373r: Don't worry: it's an alias; if you don't use it, nothing changes :)

src/adapter/http/src/data/sign/eip712/sign_eip712_handler.rs
s373r: It will be done

.cargo/config.toml
s373r: > I don't have time to dive into this now, but would suggest removing this from SKU and opening a separate PR on master where we could discuss why the compilation flags between build and clippy differ.

Long story short: this is a well-known problem that is beyond our code

PR #1625, opened 2026-05-03
GQL collection::entries support maxDepth

sergiimk:
Implements `maxDepth` support.

I faced a hard choice here on how to represent presence of "directories" after reaching the depth limit.

Options were:
- Make `entries()` return an enum `CollectionEntry` and `CollectionPrefix` - this seems like a nice choice but would break the existing API
- Return entry with a path but `ref == null` - `CollectionEntry` objects that have `ref` field as required so this would also be a (somewhat) breaking change and will force everyone to check for `null` everywhere, even if they don't use `maxDepth`
- After reaching max depth return only the first entry - this is what I went with. See tests for clarity.

### DIFF
--- a/src/adapter/graphql/tests/tests/test_gql_dataset_adapter_collection.rs
+++ b/src/adapter/graphql/tests/tests/test_gql_dataset_adapter_collection.rs
@@ -449,7 +449,7 @@ async fn test_collection_path_prefix_and_max_depth() {
                 {
                     "add": {
                         "entry": {
-                            "path": "/a",
+                            "path": "/0a",
                             "ref": linked,
                         }
                     }
@@ -457,7 +457,7 @@ async fn test_collection_path_prefix_and_max_depth() {
                 {
                     "add": {
                         "entry": {
-                            "path": "/1/b",
+                            "path": "/0b/1a",
                             "ref": linked,
                         }
                     }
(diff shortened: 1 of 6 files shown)


src/adapter/graphql/tests/tests/test_gql_dataset_adapter_collection.rs
```diff
@@ -449,94 +449,160 @@ async fn test_collection_path_prefix_and_max_depth() {
...
+        harness.list_entries_ext(&did, None, Some(0)).await,
+        as_entries(&[
+            "/0a",
+            "/0b/1a",
+            // "/0b/1b/2a",
+            // "/0b/1b/2b",
+            // "/0b/1c/2a",
+            // "/0b/1c/2b",
         ])
     );
```
s373r, line 567: It's a bit surprising to see this query results, given that the value for `maxDeath` is `0`.

In my opinion, we should rely on the behavior of well-known utilities such as `find` and `tree` (and their clones).

In this case, level `1` refers only to nested folders/files (similar to the output of the `ls` command).

Next, I will provide examples of command results on a file structure similar to yours, but relative:
```sh
# find

➜  find . -maxdepth 1
.
./0b
./0a

➜  find . -maxdepth 2
.
./0b
./0b/1b
./0b/1c
./0b/1a
./0a

➜  find . -maxdepth 3
.
./0b
./0b/1b
./0b/1b/2a
./0b/1b/2b
./0b/1c
./0b/1c/2a
./0b/1c/2b
./0b/1a
./0a

# tree

➜  tree -L 1
.
├── 0a
└── 0b

2 directories, 1 file

➜  tree -L 2
.
├── 0a
└── 0b
    ├── 1a
    ├── 1b
    └── 1c

4 directories, 2 files

➜  tree -L 3
.
├── 0a
└── 0b
    ├── 1a
    ├── 1b
    │   ├── 2a
    │   └── 2b
    └── 1c
        ├── 2a
        └── 2b

4 directories, 6 files
```

However, when the `maxDepth` value is `0`, the utilities behave differently:
```sh
➜  find . -maxdepth 0 # return nothing but zero exit code
.

➜  dirs tree -L 0 # error
tree: Invalid level, must be greater than 0.
```

s373r (changes requested):

src/adapter/graphql/tests/tests/test_gql_dataset_adapter_collection.rs
sergiimk: Updated to have "depth == 1" as top level. Depth 0 will return just a single (first) entry.

PR #1642, opened 2026-05-19
Molecule: Phase 3

s373r:
Related to: https://github.com/kamu-data/planning-internal/issues/133


src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
```diff
@@ -138,6 +161,10 @@ impl CreateAccountUseCase for CreateAccountUseCaseImpl {
     ) -> Result<Account, CreateAccountError> {
         self.save_account(account, password).await?;
 
+        // Generate a new private key.
+        let (private_key, _) = odf::AccountID::new_generated_ed25519();
+        self.save_private_key(&account.id, private_key).await?;
```
s373r: Comment for readers:
- Important change

Cargo.lock
```diff
@@ -156,6 +156,12 @@ dependencies = [
  "alloy-rpc-client",
  "alloy-rpc-types",
  "alloy-serde",
+ "alloy-signer",
+ "alloy-signer-aws",
+ "alloy-signer-gcp",
+ "alloy-signer-ledger",
+ "alloy-signer-local",
+ "alloy-signer-turnkey",
```
s373r: Comment for readers:
- It looks like this "no go" thing, so we'll have to split the `alloy` into individual sub-crates

Cargo.lock
```diff
@@ -480,7 +487,7 @@ dependencies = [
  "lru 0.16.4",
  "parking_lot",
  "pin-project",
- "reqwest",
+ "reqwest 0.12.28",
```
s373r: Comment for readers:
- A duplicate is produced by `alloy-signer-gcp`

s373r (commented):
Added comments for readers

.cargo/config.toml
```diff
@@ -10,3 +10,11 @@ feature-unification = true
...
 # See: https://www.reddit.com/r/rust/comments/1qzvkwf/workspace_feature_permutations_hell/
 feature-unification = "workspace"
+
+[alias]
+# Reduces recompilations due to invalidated cache between `cargo build` vs. `cargo clippy`
+# by separating `target` directories
+# See: https://github.com/rust-lang/cargo/issues/9280
+#      https://github.com/rust-lang/cargo/issues/15422
+#      (no solution, but there is a description of the root cause)
+clippy-target-dir = "clippy --target-dir target/clippy"
```
sergiimk: Where target dir is located is currently a personal choice. In `DEVELOPERS.md` we already recommend this:
> Create a file `/home/me/code/kamu/.cargo/config.toml` that contains:
> ```toml
> [build]
> target-dir = "/home/me/code/kamu/.target"
> ```

So perhaps we could also recommend this alias as a speed up.

Otherwise it may be confusing that my builds are going into a shared directory while I also have a project-local `target/clippy`.

So I'd suggest:
* removing this from the branch
* adding it to your local `/kamu/.cargo/config.toml`
* making a PR on master to recommend this speedup in `DEVELOPERS.md`

wdyt?

Cargo.toml
```diff
@@ -137,158 +140,163 @@ members = [
     "src/e2e/app/cli/mysql",
     "src/e2e/app/cli/sqlite",
 ]
+exclude = [
+    "src/sandbox"
```
sergiimk: I'm concerned that this will become a messy dumping ground of code that doesn't compile.

I think your code in that project is valuable though, as it tests the signing flow using only alloy, without any kamu stuff. If we ever suspect a but - it lets us quickly isolate the issue.

So perhaps it makes sense to keep (and maintain!) it as an integration test in `crypto-eip712-utils` crate?

.cargo/config.toml
sergiimk: I keep confusing that this this is not a clippy configuration setting, but an alias to run clippy with a different target dir...

This makes it more complicated as we cannot recommend this alias as both an "optional speedup" and have it used by our `make clippy` target :(

sergiimk: We could do something like this in the Makefile:
```sh
CLIPPY_CMD := $(shell cargo -Z unstable-options config get alias.clippy-target-dir 2>/dev/null \
    && echo "cargo clippy-target-dir" \
    || echo "cargo clippy")

.PHONY: clippy
clippy:
    $(CLIPPY_CMD)
```
Ugly but should do the trick?

src/adapter/http/src/data/verify_handler.rs
```diff
@@ -135,9 +135,15 @@ pub async fn verify_handler(
...
 async fn verify(catalog: Catalog, request: VerifyRequest) -> Result<VerifyResponse, ApiError> {
+    use kamu_signing::common::ProofType;
+
     // 1. Validate request
     match request.proof.r#type {
-        query::ProofType::Ed25519Signature2020 => {}
+        ProofType::Ed25519Signature2020 => {}
+        ProofType::EcdsaSecp256k1Signature2019 => {
+            // TODO: Molecule: Phase 3: implement
+            return Err(ApiError::not_implemented_without_reason());
```
sergiimk, line 145: IMO: It's OK to leave it unimplemented for now, so mb just remove the comment mentioning `Molecule` so it doesn't get in when we back-integrate this to `master`

src/app/cli/src/services/config/models.rs
```diff
@@ -559,33 +559,49 @@ pub struct AwsSecretDatabasePasswordPolicyConfig {
...
-    /// - prepends a multibase prefix
+    /// - Converts default base64 encoding to base64url and removes padding
+    /// - Prepends a multibase prefix
     #[config(combine(replace))]
     pub private_key: Option<odf::metadata::PrivateKey>,
+
+    /// Secp256k1 private key used to sign EIP-712 typed data.
+    ///
+    /// ```sh
+    /// cast wallet new
```
sergiimk: A person deploying kamu will likely not have `cast` installed (installing it on nixos is still a pain). So how about suggesting standard linux tools:
```markdown
To generate use:
\```sh
od -vN 32 -An -tx1 /dev/urandom | tr -d ' \n'
\```
or
\```sh
openssl rand -hex 32
\```
```

src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
sergiimk: Ugh, I think we botched the epic of saving private keys pretty badly. We are not saving private keys for predefined AND GitHub accounts.

How I'm reading your updated code is:
- A service (`PredefinedAccountsRegistrator` or `OAuthGithub` currently) creates an `Account` struct
- It generates a DID to use as `Account::id`
- It **throws away** the associated private key
- The `Account` is passed to `CreateAccountUseCase`
- It saves it with provided DID
- And generates a new, **completely unrelated** private key to save in the DB

The whole point of private keys is to be linked to DID (public key) to validate ownership of DID.

What I think needs to happen:
- Remove this generation - saving unrelated private keys is worse than not having any
- Create a ticket
- On mainline branch we should:
  - Nuke the heck out `impl Into<Account> for AccountConfig`
  - Extend `AccountConfig` to carry optional private key
  - Have `CreateAccountUseCase` accept `AccountConfig`
  - The **only place generating** and saving the keys should be `CreateAccountUseCaseImpl`
  - The only place that supplies `AccountConfig:id` should be the `Web3WalletAuthenticationProvider`

CC @zaychenko-sergei

sergiimk: Also, let's rename the method to `maybe_save_private_key` to highlight that it may not save anything based on configuration.

src/app/cli/src/services/config/models.rs
```diff
@@ -559,33 +559,49 @@ pub struct AwsSecretDatabasePasswordPolicyConfig {
...
+    #[config(combine(replace))]
+    pub secp256k1_private_key: Option<kamu_signing::utils::Secp256k1Signer>,
 }
 
 impl IdentityConfig {
-    pub fn to_infra_cfg(&self) -> Option<kamu_adapter_http::data::query_types::IdentityConfig> {
-        self.private_key
-            .clone()
-            .map(|private_key| kamu_adapter_http::data::query_types::IdentityConfig { private_key })
+    pub fn to_infra_cfg(&self) -> Option<kamu_signing::entities::IdentityConfig> {
```
sergiimk: Since this config is identical to the "infra" struct and that I specifically designed `setty` to be used not only on application but also on library level, how about we add `setty` derives to the infra struct and embed it directly here?

Same deduplication can happen in `kamu-node`.

sergiimk: This is something we should do for many configs here, I just did not have time for such refactoring yet.

src/adapter/http/src/data/sign/eip712/sign_eip712_types.rs
```diff
@@ -0,0 +1,63 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Request
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug, serde::Deserialize, serde::Serialize, utoipa::IntoParams)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[into_params(parameter_in = Query)]
+pub struct SignEip712QueryParams {
+    /// DID of the managed key to sign with
+    pub key: odf::metadata::DidOdf,
```
sergiimk, line 26: I'm confused as to how this works for account-level signing.

When you want to sign something with a key of the dataset you own - you have `did:odf:...` key and all is fine.

But our design was supposed to support signing with account key too and those will have `did:key:...` prefix that will not parse here.

sergiimk: I was wrong, our accounts always had `did:odf:` prefix ... although I strongly believed they were `did:key:`

src/sku/molecule/domain/src/entities/molecule_project.rs
```diff
@@ -7,12 +7,38 @@
...
+
 use chrono::{DateTime, Utc};
 use internal_error::{InternalError, ResultIntoInternal};
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// TODO: revisit after IPNFT-less projects changes.
+#[nutype::nutype(
+    sanitize(lowercase),
+    validate(len_char_min = 2),
```
sergiimk: Are you sure? Where is this restriction coming from? Even NYSE has single-character ticker symbols.

src/sku/molecule/services/src/use_cases/projects/molecule_create_project_use_case_impl.rs
```diff
@@ -111,14 +105,15 @@ impl MoleculeCreateProjectUseCase for MoleculeCreateProjectUseCaseImpl {
...
-            format!("{}.{lowercase_ipnft_symbol}", molecule_account.account_name)
+            format!("{}.{symbol}", molecule_account.account_name)
                 .parse()
                 .int_err()?;
 
         let project_email = format!("support+{project_account_name}@kamu.dev")
             .parse()
             .unwrap();
 
+        // TODO: Molecule: Phase 3: looks we are ready
```
sergiimk: What does this mean?

.cargo/config.toml
s373r: Add more context:
- The main issue I was/am trying to address was the time-consuming process of re-checking code that had already been verified (or, simply, running `clippy` twice in a row). I expect the second run to produce immediate results, but that doesn't happen
- I noticed that the issue doesn't appear right away, but only after a few builds and `clippy` passes

I've been looking into this for some time and found that the re-check is caused by `event-sourcing` crate. 
```shell
# Found via this:
CARGO_LOG=cargo::core::compiler::fingerprint=trace cargo clippy-target-dir --workspace --all-targets --locked -- -D warnings
```

But this requires further investigation into why build artifacts are considered modified or something like that (the root cause of re-check).

> This makes it more complicated as we cannot recommend this alias as both an "optional speedup" and have it used by our make clippy target :(

Yes, at the moment, I'm not entirely sure that this will lead to an speedup. As it turns out, rebuilds happen with this alias as well, as I now know.

> removing this from the branch

Therefore, I have no objections to this.

> making a PR on master to recommend this speedup in DEVELOPERS.md

If my investigation yields results, I'll definitely do that, but for now, it isn’t completed. I'll continue it, though, using my local setup.

src/sku/molecule/domain/src/entities/molecule_project.rs
s373r: At the implementation time , I was sure, but now, I realize it isn't correct -- we need to do at least `1`

src/sku/molecule/services/src/use_cases/projects/molecule_create_project_use_case_impl.rs
s373r: ... ready to process the following line TODO:
```
// TODO: Remove tolerance to accounts that already exist after we have account
```

We started discussing the topic, but the conversation stalled -- I suggest we can continue here

Cargo.toml
s373r: > So perhaps it makes sense to keep (and maintain!) it as an integration test in crypto-eip712-utils crate?

It's already done :)

> as it tests the signing flow using only alloy, without any kamu stuff

`crypto-eip712-utils` is essentially a lightweight wrapper that relies solely on `internal-error`. 

I think, so as not to make anyone confused, I'll just delete this sandbox crate (all things were "ported" to other workspace crates)

src/sku/molecule/domain/src/entities/molecule_project.rs
s373r: Fixed

Cargo.toml
s373r: `sandbox-eip-712` deleted -- Done

src/adapter/http/src/data/verify_handler.rs
s373r: Makes sense -- Done

.cargo/config.toml
s373r: Deleted -- Done

src/app/cli/src/services/config/models.rs
s373r: > not have cast installed (installing it on nixos is still a pain)

Sounds sad :( 

There will be a variety of options in the comments, something for everyone

--

Done

src/app/cli/src/services/config/models.rs
s373r: > Since this config is identical to the "infra" struct

They are not identical. The infra struct has `Option<_>` fields, the config not. 

--

But I'll take another look -- maybe I'm missing something.

s373r: Made minor structural changes to better align with `setty`

Done

src/domain/accounts/services/src/use_cases/create_account_use_case_impl.rs
s373r: > Remove this generation

Done

> Create a ticket

Done, https://github.com/kamu-data/kamu-cli/issues/1647

> let's rename the method 

Done

Cargo.lock
s373r: Unwanted deps have been removed

Done

Cargo.lock
s373r: Not actual after https://github.com/kamu-data/kamu-cli/pull/1642#discussion_r3290166623

src/sku/molecule/services/src/use_cases/projects/molecule_create_project_use_case_impl.rs
sergiimk: I see. Re-reading the whole code now I con't see the point of allowing to reuse existing accounts - seem dangerous no matter how I look at it.

I can only think of that maybe it was done for eventual consistency, but whole creation process is under one transaction... If everything went well and client didn't get a response - it's probably OK to tell them "already exists" and make them re-read the project list before retrying.

I suggest we remove this `account_by_name` lookup and call `create_account_use_case` and let it fail. 

We will have to handle the name collision error though, as currently it will be raised as an internal error.

s373r: > I can only think of that maybe it was done for eventual consistency, but whole creation process is under one transaction...

I seem to recall discussions about this -- perhaps there was a moment when the creation involved several transactions

> I suggest we remove this account_by_name lookup and call create_account_use_case and let it fail.

Done

> We will have to handle the name collision error though, as currently it will be raised as an internal error.

New errors have added & propagated 

Done

resources/schema.gql
```diff
@@ -962,8 +962,14 @@ type CreateDatasetResultSuccess implements CreateDatasetResult & CreateDatasetFr
 	message: String!
 }
 
-type CreateProjectErrorConflict implements CreateProjectResult {
-	project: MoleculeProjectV2!
+type CreateProjectAccountConflictError implements CreateProjectResult {
```
sergiimk, line 965: As I mentioned in Slack - I'd prefer to avoid having this late API-breaking change and instead reused the existing conflict error with a different message.

sergiimk (changes requested):

resources/schema.gql
s373r: We discussed it in Slack and agreed that we need a new error (`CreateProjectAccountConflictError`), but we shouldn't change the name of the old one (`CreateProjectErrorConflict`)

Done

PR #1649, opened 2026-05-28
Transaction abort observability

### LINKED ISSUE #341: Need better visibility in case of client-side connection drops
In https://github.com/kamu-data/planning-internal/issues/139 we had a really interesting incident where S3 PUT request hanged indefinitely due to a bug in SDK, but the request tokio task was seemingly aborted by the HTTP stack when it saw the client TCP connection to reset after the client-side timeout.

Currently such abort is very difficult to spot:
- the spans were closed without error
- the DB transaction was neither committed nor rolled back

This ticket is to make sure we:
- Emit some log from DB transaction wrapper if it's dropped without commit/rollback
- Emit a warning from HTTP layer about the task aborting

sergiimk:
Closes: kamu-data/kamu-node#341

In this PR:
- I add two drop guards:
  - One around DB transaction state that outputs `Transaction DROPPED` warning if transaction was never committed or rolled bach
  - One around HTTP request future that outputs `HTTP disconnect` if axum drops the request future before response is produced (e.g. when client disconnects)
- To debug all this I introduce `/system/probe` endpoint that can simulate opening a DB transaction, sleeping, panicking, and returning errors
- I also realized that `trace_id` that `root_span!` macro adds into the span to make it appear in our logs gets added as a free-form attribute in OTEL spans - this causes `Duplicate tag` warnings in Tempo as it conflicts with `trace_id` that is already a core part of the OTEL span data structure. So I clean up this attribute before exporting it to OTEL. Not sure it will fix our missing traces problem, but it will reduce the set of possible problems for sure.

Let's say I hit the API server with this:
```
xh GET "http://127.0.0.1:8080/system/probe?sleep=1s&dbTransaction=true"
```

Example normal request:
image

Example request when I Ctrl+C on the client side:
image

Please have a look - I'll test this under OTEL integration with `kamu-node` tomorrow before merging.


zaychenko-sergei (approved):

src/utils/database-common/src/transactions/db_transaction.rs
```diff
@@ -33,6 +34,14 @@ impl TransactionRef {
         }
     }
 
+    /// Acquires a connection from the connection pool and starts the
+    /// transaction if it's not already started. You probably should not call
+    /// this - it's only occasionally useful for testing transactions without
+    /// downcasting to specific DB type.
+    pub async fn begin(&self) -> Result<bool, InternalError> {
+        Arc::clone(&self.inner).begin().await
```
s373r: Just out of curiosity: 

Q: why do we need a clone (rc increment) here?

s373r (commented):
Cool!

src/utils/database-common/src/transactions/db_transaction.rs
sergiimk: In a trait you can accept `self` as `begin(self: Arc<Self>)` and Rust considers it object-safe, but it does not consider `begin(self: &Arc<Self>)` object-safe for some reason.

In the end I need an `Arc` to wrap raw `TransactionState` into `TransactionRefT`:
```rust
async fn begin(self: Arc<Self>) -> Result<bool, InternalError> {
        let tx = TransactionRefT { inner: self };
        tx.begin().await
    }
```

but I'll check again if I can achieve this without extra cloning.

s373r: > re begin(self: Arc<Self>)

Maybe it's because the `Trait` is a value type, but in our code it's a ref type, and we cannot consume `self.inner` by that reason

In other words, dereferencing takes place, so it appears as though we are operating on a value (`self.inner`), but in reality, it is a reference -- and w/ a clone we have a owned value

I think it's OK to leave as is

Issue #996, opened 2024-12-18
Login, OAuth: login error if system (local) username matches GitHub username 


s373r:
### Describe the bug

- If a multi-tenant workspace is created, a system user is also created (`password` provider).
- When trying to log in, it tries to create a new account, but for the `oauth_github` provider
- Subsequently, it creates problems in the authorization process: 
  - `Credentials are already used by an existing account`

![image](https://github.com/user-attachments/assets/46c9b625-f661-4cb6-b615-a29407ec5a6b)

### Steps To Reproduce

- `kamu-cli` part:
```shell
kamu init --multi-tenant

KAMU_AUTH_GITHUB_CLIENT_ID=361a3b4fda86d0234d2f \
KAMU_AUTH_GITHUB_CLIENT_SECRET=465849325236ed49253993744069e1bec6808554 \
kamu \
  system api-server \
  --address 0.0.0.0 \
  --http-port 8080 \
  --get-token
```
- `kamu-web-ui` part:
  - Login via GitHub

### Current Behavior

`Credentials are already used by an existing account`

### Expected Behavior

No errors

### kamu system info

```sh
latest (v0.213.0)
```

### Anything else?

- **Solution:** need to update the indexes in the `account` table
- Proposed refactoring: https://github.com/kamu-data/kamu-cli/pull/1674#discussion_r3751576693

---

UPD (2026-08-11):

⚠️ Latest internal conversation re the topic: https://kamu-data.slack.com/archives/C02FVH5C04E/p1786385007035239?thread_ts=1786050617.361799&cid=C02FVH5C04E

PR #1674, opened 2026-08-04
Harden account registration: provisioning + DID secrets

### LINKED ISSUE #1647: `CreateAccountUseCaseImpl`: consolidate account provisioning
### Expectations
 
- [x] Nuke the heck out `impl Into<Account>` for `AccountConfig`
- [x] Extend `AccountConfig` to carry optional private key
- [x] Have `CreateAccountUseCase` accept `AccountConfig`
- [x] The only place generating and saving the keys should be `CreateAccountUseCaseImpl`
- [ ] The only place that supplies `AccountConfig:id` should be the `Web3WalletAuthenticationProvider`

Based on https://github.com/kamu-data/kamu-cli/pull/1642#discussion_r3276686518

### LINKED ISSUE #1672: Harden account registration: provisioning + DID secrets

s373r:
Closes #1647
Closes #1672

---
- [x] `PredefinedAccountsRegistrator`: cover w/ tests
- [x] Validate all IDs to ensure they are derived from DID secrets
---

### Added:
- `MySqlDatasetEntryRepository` has been implemented:
  - To fix a previously masked error w/ dataset searching in E2E tests
### Changed
- Revised approaches for handling predefined accounts (mostly `PredefinedAccountsRegistrator`) (#1674):
  - Now, every predefined password account has its own unique key, even for single-tenant workspaces
  - Stricter validation of predefined accounts
- New account IDs are derived from DID secret keys, not from names (#1674):
  - `DEFAULT_ACCOUNT_ID` has been removed; now we only use `TEST_ACCOUNT_ID` for testing, 
    which does not exist in the production code paths (affects new single-tenant workspaces)
- More strict typing in some places (using `Url` and `Email` instead of `String`) (#1674)
- The uniqueness of the DID secret for accounts and datasets is guaranteed (#1674)
### Fixed
- Fixed the tests failing on Fedora (MQTT, ODFEngine) (#1674):
  - MQTT: `localhost` replaced w/ direct container host
  - ODFEngine: improved handling of the initial connection: due to the nature of proxying, 
               there is a very brief initial window during which the connection may be reset


migrations/postgres/20260806220000_did_secret_keys_unique_entity.sql
```diff
@@ -0,0 +1,17 @@
+/* ------------------------------ */
+
+-- Deduplicate existing rows (keep the oldest by ctid).
+DELETE
+FROM did_secret_keys t
+    USING (SELECT ctid,
+                  ROW_NUMBER() OVER (PARTITION BY entity_type, entity_id ORDER BY ctid) AS rn
+           FROM did_secret_keys) dup
+WHERE t.ctid = dup.ctid
+  AND dup.rn > 1;
```
s373r, line 10: A comment for readers:

This part of the migration may change following investigation/validation re derivability of account IDs from DID secrets

Please do not close this thread until then

migrations/sqlite/20260806220000_did_secret_keys_unique_entity.sql
```diff
@@ -0,0 +1,15 @@
+/* ------------------------------ */
+
+-- Deduplicate existing rows (keep the oldest by rowid).
+DELETE
+FROM did_secret_keys
+WHERE rowid NOT IN (SELECT MIN(rowid)
+                    FROM did_secret_keys
+                    GROUP BY entity_type, entity_id);
```
s373r, line 8: A comment for readers:

This part of the migration may change following investigation/validation re derivability of account IDs from DID secrets

Please do not close this thread until then

migrations/postgres/20260806220000_did_secret_keys_unique_entity.sql
s373r: A more surgical approach may be chosen, without cleaning using migrations

migrations/sqlite/20260806220000_did_secret_keys_unique_entity.sql
s373r: A more surgical approach may be chosen, without cleaning using migrations

src/domain/accounts/domain/src/services/authentication_provider.rs
s373r, line 33: A comment for readers:

I spent a long time trying to replace `ProviderLoginResponse` with `PredefinedAccountsConfig`, but it was a total fiasco.

`PredefinedAccountsConfig` already contains too many fields (keys, ReBAC) to be usable -- no matter how hard I tried, I couldn't get it to work :(

scripts/s3-list-dataset-aliases.sh
```diff
@@ -0,0 +1,28 @@
...
+    --bucket "$BUCKET" \
+    --query "Contents[?ends_with(Key, 'alias')].Key" \
+    --output text \
+    --no-cli-pager \
+  | tr '\t' '\n'
+)
+
+for key in $alias_files; do
+  # NOTE: It's slow, yeah, but works
+  aws s3 cp "s3://$BUCKET/$key" -
```
s373r, line 26: A comment for readers:

I forgot what format the aliases were stored in, so I double-checked them in different envs to make sure they all matched the account names.

s373r: But this script isn't that important; I'm willing to delete it without any problem by request

scripts/test-workspace-crates-one-by-one.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Usage: ./test-workspace-crates-one-by-one.sh [--start-from crate]
```
s373r, line 5: A comment for readers:

I like this script a lot better already; I finally got around to making it -- it helps me stay focused on crate tests when making major changes

src/app/cli/src/commands/system_generate_token_command.rs
```diff
@@ -34,9 +35,25 @@ pub struct GenerateTokenCommand {
...
+
+            let account_name = odf::AccountName::from_str(login).int_err()?;
+            let account = self
+                .account_service
+                .get_account_by_name(&account_name)
+                .await
+                .int_err()?;
+
+            account.id
         } else {
```
s373r, line 57: A noticeable change: this applies to the local workspace, but now you won't be able to generate tokens for the environment (which is a good thing)

src/app/cli/src/services/accounts/models.rs
```diff
@@ -79,45 +73,30 @@ impl CurrentAccountIndication {
...
-                        Ok(subject)
-                    }
-                }
+            WorkspaceStatus::Created(_) => {
+                let maybe_account = account_service.account_by_name(&self.account_name).await?;
+                let Some(account) = maybe_account else {
+                    return Err(ToCurrentAccountSubjectError::NotRegisteredAccount {
+                        account: self.account_name.clone(),
+                    });
+                };
```
s373r, line 95: A noticeable change: it's clear that we no longer mask the account's behavior using hardcoded values for single-tenant environments.

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
```diff
@@ -86,15 +110,28 @@ pub enum AccountPropertyName {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// TODO: Embed the `AccountConfig::validate()` call
+//       into setty validation mechanism (for whole struct).
```
s373r: A comment for readers: can be treated as a feature request ^_^

src/infra/core/tests/tests/ingest/test_fetch.rs
```diff
@@ -732,7 +732,7 @@ async fn test_fetch_mqtt_empty() {
     let broker = crate::MqttBroker::new().await;
 
     let fetch_step = odf::metadata::FetchStep::Mqtt(odf::metadata::FetchStepMqtt {
-        host: "localhost".to_string(),
+        host: broker.address.clone(),
```
s373r, line 735: A comment for readers: Fedora-related fixes

src/utils/format-utils/src/format_collection.rs
```diff
@@ -11,15 +11,24 @@ use std::fmt::Display;
...
+        write!(res, "{first}").unwrap();
+        for item in iter {
+            res.push_str(", ");
+            write!(res, "{item}").unwrap();
+        }
+    }
+
+    res.push(']');
+    res
 }
```
s373r, line 32: A comment for readers: a more performant approach (in theory)

src/utils/internal-error/src/lib.rs
```diff
@@ -153,3 +151,34 @@ where
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub type ErrorIndex = usize;
+
+#[derive(thiserror::Error, Debug)]
+#[error("Batch operation partially failed: {failed}/{total} items")]
+pub struct BatchError<E>
```
s373r, line 159: A comment for readers: a naive attempt to create a generalized BatchError

migrations/postgres/20260806220000_did_secret_keys_unique_entity.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- NOTE: Leave the previous index as is:
+--       idx_auth_did_secret_keys (entity_type, entity_id);
+
+CREATE UNIQUE INDEX idx_auth_did_secret_keys_unique_entity_id
+    ON did_secret_keys (entity_id);
```
sergiimk, line 16: Hm, I thought we were missing an index altogether, but If we already have the old index as:

```
CREATE INDEX idx_auth_did_secret_keys
    ON did_secret_keys (entity_type, entity_id);
```

would it make more sense to make it `UNIQUE`?

After all we're not trying to enforce that secrets should be distinct - instead we want to make sure that every entity has only one associated secret.

src/odf/metadata/src/auth/account_id.rs
sergiimk, line 53: should we add `#[cfg(feature = 'testing')]` on top of this method to ensure it's never used in production code?

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
```diff
@@ -187,14 +225,31 @@ impl AccountConfig {
...
-        if let Some(id) = &self.id {
-            id.clone()
-        } else {
-            odf::AccountID::new_seeded_ed25519(self.account_name.as_bytes())
+    /// Resolves account ID from `id` and/or `private_key`.
+    /// Returns `None` when neither is set.
+    ///
+    /// NOTE: It is assumed that we call `validate()` first before using the
+    ///       values.
+    pub fn resolve_account_id(&self) -> Option<odf::AccountID> {
```
sergiimk: Minor: "resolve" sounds like some external call to me - how about naming this `get_account_id_from_config_or_private_key`?

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
sergiimk: `setty` already supports validation macros like [`#[config(validate(...))]`](https://github.com/kamu-data/setty/blob/master/examples/validate/src/main.rs#L11) that delegate logic to the [`validator`](https://crates.io/crates/validator/) crate.

Have you tried something like `#[config(validate(custom(function = "...")))]`?

src/odf/metadata/src/testing/shortcuts.rs
```diff
@@ -41,8 +41,9 @@ pub fn alias(account: &impl AsRef<str>, dataset_name: &impl AsRef<str>) -> Datas
 }
 
 pub fn account_id(name: &impl AsRef<str>) -> AccountID {
```
sergiimk, line 43: I kinda dislike this function. As a free-standing function it's hard to find - you need to know about its existence rather than discovering it from `AccountID`.

We already have a similar `Account::test(name)` method, so maybe for consistency this could also be moved into a method under `AccountID`?

src/utils/internal-error/src/lib.rs
sergiimk: Note: The need for this may go away if account config validation uses `validator` crate (via `setty`) as it comes with its own batch error type.

src/infra/accounts/sqlite/src/repos/sqlite_account_repository.rs
```diff
@@ -404,6 +413,38 @@ impl AccountRepository for SqliteAccountRepository {
         Ok(maybe_account_row.map(|account_row| account_row.id))
     }
 
+    async fn find_account_ids_by_unique_fields(
```
sergiimk, line 416: How about `find_account_ids_by_one_of_unique_fields` to reflect that only one of the fields need to match, not all three?

src/domain/accounts/domain/src/repos/did_secret_key_repository.rs
```diff
@@ -37,15 +36,32 @@ pub trait DidSecretKeyRepository: Send + Sync {
...
+    #[error(transparent)]
+    Duplicate(#[from] DidSecretKeyDuplicateError),
+
     #[error(transparent)]
     Internal(#[from] InternalError),
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("Entity's DID secret key already exists: '{entity:?}'")]
+pub struct DidSecretKeyDuplicateError {
```
sergiimk, line 50: As I mentioned in the comment on migration - don't think we have a goal of deduplicating secret keys - our goal is to ensure uniqueness of the entity identity.

src/adapter/oauth/src/oauth_github.rs
sergiimk: Wasn't one of the main goals of this refactoring to eliminate places where a key pair is generated and private key is discarded and make sure we save valid private keys for accounts created via GitHub login?

src/app/cli/src/services/accounts/models.rs
sergiimk: Do I understand correctly that in a single-tenant workspace `kamu` will be created as a "real" predefined account? Can you point me to where this creation happens?

src/domain/accounts/domain/src/entities/account.rs
```diff
@@ -49,6 +54,12 @@ pub enum AccountProvider {
     Web3Wallet,
 }
 
+impl AccountProvider {
+    pub fn is_password(provider: &str) -> bool {
```
sergiimk, line 64: To me `AccountProvider` enum doesn't make sense. Provider mechanism is meant to be as a plug-in extension point, so there shouldn't be an enum listing all implementations because some of them can be not owned by us. This is why `Account::provider` is a string and not an enum after all.

We can leave this out of scope, but I just want to communicate my concerns.

src/domain/accounts/domain/src/services/account_identity_generator.rs
```diff
@@ -0,0 +1,19 @@
...
+// by the Apache License, Version 2.0.
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub trait AccountIdentityGenerator: Sync + Send {
+    fn generate_ed25519(
+        &self,
+        account_name: &odf::AccountName,
+    ) -> (odf::metadata::SigningKey, odf::AccountID);
+}
```
sergiimk: This service interface makes me uneasy.

The API itself `(name) -> key_pair` raises many questions when reading the code in production context.

My understanding is that this is needed only to simplify test setup. If so, should we instead add a `seed_dids_from_names_in_tests` parameter to `AccountServiceImpl` that can be specified in test harnesses DI setup?

At least it will make it extra clear what's happening when reading the service code.

src/domain/accounts/domain/src/services/authentication_provider.rs
sergiimk: Suprised to see `AccountID` remaining a mandatory argument here, as we discussed that most providers are not authoritative sources of DIDs

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
```diff
@@ -43,10 +43,17 @@ pub trait CreateAccountUseCase: Send + Sync {
 
 #[derive(bon::Builder, Default)]
 pub struct CreateAccountUseCaseOptions {
+    pub quiet: bool,
```
sergiimk: This parameter needs a docstring - supressing a outbox message seems like a very big deal (a potential source of state inconsistency) and I'm confused why this is ever needed.

src/domain/accounts/services/src/services/login_password_auth_provider.rs
```diff
@@ -81,14 +84,13 @@ impl AuthenticationProvider for LoginPasswordAuthProvider {
             .int_err()?;
 
         Ok(ProviderLoginResponse {
-            // For password-based accounts
-            account_id: odf::AccountID::new_seeded_ed25519(account_name.as_bytes()),
-            account_name,
-            email: account.email.clone(),
-            display_name: password_login_credentials.login.clone(),
+            account_id: account.id,
```
sergiimk: To me it feels like we've coupled two separate systems:
- Password-based authentication provider - that should only know about logins and passwords
- Account/identity service - that manages data of accounts from multiple providers

It's strange to see the circular dependency that is going on here to fetch data from account services ... that goes straight back into account service.

If we would separate the two cleanly the response here would only include the login as `provider_identity_key` and nothing else.

src/domain/accounts/domain/src/services/account_service.rs
```diff
@@ -66,6 +69,20 @@ pub trait AccountService: Sync + Send {
         account_name: &odf::AccountName,
     ) -> Result<Option<odf::AccountID>, InternalError>;
 
+    async fn find_account_id_by_provider_identity_key(
+        &self,
+        provider_identity_key: &str,
+    ) -> Result<Option<odf::AccountID>, FindAccountIdByProviderIdentityKeyError>;
```
sergiimk, line 75: Should this method take `(provider_id, provider_identity_key)`?

If password-based auth provider uses `login` as `provider_identity_key` I don't understand how we guarantee that those keys don't conflict across multiple providers.

src/domain/accounts/services/src/services/login_password_auth_provider.rs
```diff
@@ -81,14 +84,13 @@ impl AuthenticationProvider for LoginPasswordAuthProvider {
...
+            account_name: account.account_name,
+            email: account.email,
+            display_name: account.display_name,
             account_type: account.account_type,
-            avatar_url: account.avatar_url.clone(),
-            provider_identity_key: password_login_credentials.login.to_ascii_lowercase(),
+            avatar_url: account.avatar_url,
+            provider_identity_key: account.provider_identity_key,
```
sergiimk, line 92: why `lowercase` is gone here? is it safe?

src/app/cli/src/services/accounts/models.rs
s373r: > Do I understand correctly that in a single-tenant workspace kamu will be created as a "real" predefined account?

Yes, the account (for new workspaces) will have its own unique ID and key

>  Can you point me to where this creation happens?

Sure!

Main actor here are:
- `PredefinedAccountsConfig::single_tenant()` -- Previously, the ID was preset; now it is set to `None`
- `PredefinedAccountsRegistrator::process_account()`: cannot find the account and create a one
- `CreateAccountUseCaseImpl::execute()` + `CreateAccountUseCaseImpl::resolve_account_key_and_id()`: Since the ID wasn't set (nor key), we generate a new pair

src/utils/internal-error/src/lib.rs
s373r: Q: In this case, yes, but maybe we need to save it for later?

migrations/postgres/20260806220000_did_secret_keys_unique_entity.sql
s373r: > would it make more sense to make it UNIQUE?

That's what I did at first, yes

> After all we're not trying to enforce that secrets should be distinct - instead we want to make sure that every entity has only one associated secret

But after one of the self-review rounds, I realized that this approach could leave room for error (at the database level) if the same `entity_id` were saved with a different `entity_type`.

Since `entity_type` is primarily a convenience feature for searching for entities, I suggest leaving it as is to ensure that `entity_id` can have only one secret

Q: WDYT?

src/odf/metadata/src/auth/account_id.rs
s373r: I've been thinking about this because it's one of the things I've wanted to fix for a long time (step by step, PR by PR) so that such methods can only be used in tests (with validation at compile time, ofc).

`<a dubious thing>`
So far, based on my observations, `cargo/rustc` can add features during the build process when it builds a set of crates and detects that the same crate is being built with different features.
`</a dubious thing>`

I've practically removed this method (for accounts) from all prod code paths; there's only one place left:

```rust
// CurrentAccountIndication

    pub async fn to_current_account_subject(
        &self,
        workspace_status: WorkspaceStatus,
        account_service: Arc<dyn AccountService>,
    ) -> Result<CurrentAccountSubject, ToCurrentAccountSubjectError> {
        match workspace_status {
            WorkspaceStatus::NoWorkspace | WorkspaceStatus::AboutToBeCreated(_) => {
                // NOTE: At this stage, real accounts do not exist yet.
                let dummy_subject = CurrentAccountSubject::logged(
                    odf::AccountID::new_seeded_ed25519(self.account_name.as_bytes()), // <---
                    self.account_name.clone(),
                );

                Ok(dummy_subject)
            }

        // <CUT>
```

I think I can update this place as well, to get rid of the method entirely -- will do

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
s373r: Will rename

src/odf/metadata/src/testing/shortcuts.rs
s373r: Let me think about that, since this isn't new code

src/infra/accounts/sqlite/src/repos/sqlite_account_repository.rs
s373r: Yes, it was hard for me to come up with a name here; you're right, your suggestion is better

~~I'll also add (at least in the description) which fields are considered unique in general~~
UPD: Never mind: unique fields are arguments

src/domain/accounts/domain/src/repos/did_secret_key_repository.rs
s373r: I'd like to clarify, since from my perspective these are two inextricably linked goals (or rather, one follows from the other)

Currently, at the infra layer, we return a `DidSecretKeyDuplicateError` if a duplicate is found (i.e., we process a unique index) -- our common practice.

Q: Are you suggesting that we return an `InternalError` in this situation?

Or am I misunderstanding your line of reasoning?

src/adapter/oauth/src/oauth_github.rs
s373r: My bad

Update this place

src/domain/accounts/domain/src/entities/account.rs
s373r: Will add as a TODO your concerns

As far as I can recall, a long time ago, I added this enum as a list of all possible providers that are generally allowed. 

And, of course, there was also a practical benefit: to verify compatibility with a specific provider, we didn't need to import an adapter crate (infra) -- all we needed was `kamu-accounts` (domain).

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
s373r: > I'm confused why this is ever needed.

The following was added to this PR: https://github.com/kamu-data/kamu-cli/pull/1528

> This parameter needs a docstring

Will add

s373r: СС @zaychenko-sergei

src/domain/accounts/services/src/services/login_password_auth_provider.rs
s373r: > To me it feels like we've coupled two separate systems:

Yes, I think that's exactly it -- that's why even small changes here create some tension

> It's strange to see the circular dependency that is going on here to fetch data from account services ... that goes straight back into account service.

I think this is because we use the authorization data to create an account later on, in cases where no password is required (oauth-github, web3)...  But I need to figure out a way to resolve this

> If we would separate the two cleanly the response here would only include the login as `provider_identity_key` and nothing else.

Will do

src/domain/accounts/services/src/services/login_password_auth_provider.rs
s373r: > why lowercase is gone here? is it safe?

I think yes -- we're just populating the data from an account that's (!) already been saved

src/domain/accounts/domain/src/services/account_service.rs
s373r: > Should this method take (provider_id, provider_identity_key)?

That's a good idea. I think this will help fix a small buy long-standing issue: 
- https://github.com/kamu-data/kamu-cli/issues/996

But provided that we remove the uniqueness constraint on `provider_identity_key` only and also add `provider_id` to the unique index (`provider_id`, `provider_identity_key`)

But now I'm starting to have my doubts... I'll think some more about how to handle this situation.

Q: WDYT?

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
s373r: Hmm, you're right -- these checks can also be expressed as checks on a specific field, rather than on the entire structure

Will do

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
s373r: Done

src/infra/accounts/sqlite/src/repos/sqlite_account_repository.rs
s373r: Done

src/domain/accounts/domain/src/entities/account.rs
s373r: Done -- Added

src/domain/accounts/services/src/services/login_password_auth_provider.rs
s373r: > If we would separate the two cleanly the response here would only include the login as provider_identity_key and nothing else.

To create an account, we need more information (email, etc.)

For now, I'll try to adapt the existing code -- let's see how it turns out

migrations/postgres/20260806220000_did_secret_keys_unique_entity.sql
sergiimk: Ok I see your logic now and agree!

src/odf/metadata/src/auth/account_id.rs
sergiimk: > rgo/rustc can add features during the build process when it builds a set of crates and detects that the same crate is being built with different features

Our `.cargo/config.toml` enables `feature-unification = "workspace"` to unify features across all workspace crates and reduce build permutations between e.g. `cargo build` and `cargo build -p kamu-cli`

But AFAIK unification only looks at features in `[dependencies]`. So If we make sure that `features = ["testing"]` is used ONLY in `[dev-dependencies]` - then testing code will not be available and attempt to use e.g. seeded ID generation from a production code will not compile.

src/domain/accounts/domain/src/repos/did_secret_key_repository.rs
sergiimk: I retract this comment after what you said about why you chose the specific UNIQUE index structure.

src/adapter/oauth/src/oauth_github.rs
s373r: Done

Now both `OAuthGithub` and `Web3WalletAuthenticationProvider` will have their own secrets upon initial authorization

In the case of `Web3WalletAuthenticationProvider`, the `account_id` generated by `Web3WalletAuthenticationProvider` will be used

src/domain/accounts/domain/src/services/authentication_provider.rs
s373r: Done -- fixed

src/domain/accounts/services/src/services/login_password_auth_provider.rs
s373r: I'll add to my answer:

Since this is a password account, the data is populated solely for typing purposes (as required by the structure) -- no account will be created based on this data, since it has already been created

src/domain/accounts/domain/src/entities/account.rs
sergiimk: > I added this enum as a list of all possible providers that are generally allowed

An more extensible approach to this would be:
- Let supported auth providers be registered in `dill::Catalog`
- Use `#[dill::meta(...)]` builder metadata to attach IDs to each provider
- Validate string values in config against the list of registered providers

But yea, let's leave this for later

src/domain/accounts/domain/src/services/account_service.rs
sergiimk: Relying on `provider_identity_key` to be unique without prefixing it with provider ID seems like a design flaw.

Unique index on `(provider_id, provider_identity_key)` makes a lot more sense to me but let's check with @zaychenko-sergei.

src/domain/accounts/services/src/services/login_password_auth_provider.rs
sergiimk: So the provider currently piggy-backs on `accounts` table for:
- email storage
- uniqueness checks

and maybe that's OK.

But perhaps I'd still avoid returning any other information - no DID, no display name or avatar - this would underline the separation between provider and the account system saying that all these other things are controlled by latter.

src/utils/internal-error/src/lib.rs
sergiimk: Tbh I'm not 100% convinced by the error design and would need to think about it more.

Personally I was trying to design something more like `BatchResult<S, E>` and `BatchResultWithInput<I, S, E>` that would be usable in conjunction with `enum BatchMode { AbortOnFirstError, TryAllRollbackOnErrors, TryAllCommitSuccessful }`.

generalizing batch operations is tough...

src/odf/metadata/src/auth/account_id.rs
s373r: > So If we make sure that features = ["testing"] is used ONLY in [dev-dependencies]

Yes, I double-checked that last time, but it seems that wasn't enough

(screenshots showing that the project can still be built -- even though, based on our expectations, it shouldn't be)

image
image
image
image

s373r: > should we add #[cfg(feature = 'testing')] on top of this method to ensure it's never used in production code?

Done -- added... but see above

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
s373r: To clarify: my request was regarding top-level validation or (https://github.com/Keats/validator#struct-level-validation)

```rs
#[setty::derive(setty::Config, Clone)]
#[config(validate(custom(function = "validate_account_config")))] // <-- this guy
pub struct AccountConfig {
```

In this case, I can also express what I want by fields validators

s373r: No luck -- probably wrong usage
```
   Compiling kamu-accounts v0.265.0 (/home/s373r/src/work/kamu-data/kamu-cli/src/domain/accounts/domain)
error: cannot find attribute `config` in this scope
   --> src/domain/accounts/domain/src/entities/predefined_accounts_config.rs:120:3
    |
120 | #[config(validate(custom(function = "validate_account_config_private_key")))]
    |   ^^^^^^
    |
help: `config` is an attribute that can be used by the derive macros `Config` and `Default`, you might be missing a `derive` attribute
    |
121 + #[derive(Config, Default)]
122 | pub struct AccountConfig {
```

sergiimk: Ah yeah I did not support the top-level validation, but it should be easy to do.

Let's keep your existing validation approach, leave a TODO, and deal with this later. Thanks for trying tho!

s373r: > In this case, I can also express what I want by fields validators

No, it didn't work -- I can't validate two fields at the same time using mechanisms like these:
```rs
// TODO: setty re-export ValidationError
use setty::__internal::validator::ValidationError;

#[setty::derive(setty::Config, Clone)]
pub struct AccountConfig {
    // ...
    #[config(
        validate(custom(function = "validate_account_config_private_key")),
        combine(replace)
    )]
    pub private_key: Option<odf::metadata::PrivateKey>,
    // ...
    #[config(
        validate(custom(function = "validate_account_config_provider")),
        default = AccountProvider::Password.to_string())
    ]
    pub provider: String,
    // ...
}
// ...
fn validate_account_config_private_key(
    _value: &odf::metadata::PrivateKey,
) -> Result<(), ValidationError> {
    todo!()
}

fn validate_account_config_provider(_value: &String) -> Result<(), ValidationError> {
    todo!()
}
```

I managed to get it to compile, but when validating one field, I need to know the values of the others as well...

E.g. validating `private_key`, `id` is also needed

src/utils/internal-error/src/lib.rs
s373r: > if account config validation uses validator

No luck at the moment :(

Cross-linking: https://github.com/kamu-data/kamu-cli/pull/1674#discussion_r3752521211

src/domain/accounts/domain/src/services/authentication_provider.rs
s373r: Not actual anymore

src/domain/accounts/domain/src/entities/predefined_accounts_config.rs
s373r: > Let's keep your existing validation approach, leave a TODO, and deal with this later.

Done -- Added the link to the TODO

src/domain/accounts/domain/src/services/account_service.rs
s373r: Cross-linking: [internal Slack conversation ](https://kamu-data.slack.com/archives/C02FVH5C04E/p1786385007035239?thread_ts=1786050617.361799&cid=C02FVH5C04E)

In short: concerns were raised regarding the need to ensure the uniqueness of account names (for URLs and SmTP).

It is proposed that this comment be addressed at https://github.com/kamu-data/kamu-cli/issues/996

--

Propose to close this thread

src/domain/accounts/domain/src/entities/account.rs
s373r: > An more extensible approach to this would be:

I really like your suggestion. At times like this, my love for DI only grows stronger. 
Added as a refactoring suggestion as part of https://github.com/kamu-data/kamu-cli/issues/996

> But yea, let's leave this for later

Closing this thread

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
s373r: UPD: Following an [internal discussion](https://kamu-data.slack.com/archives/C02FVH5C04E/p1786386235839119?thread_ts=1786050617.361799&cid=C02FVH5C04E), it was determined that the use of `quiet` was only necessary for testing purposes. 

Thus, the only place in the prod code paths (`PredefinedAccountsRegistrator::register_unknown_account()`) was updated to avoid using the param

Done -- tests stabilized

src/utils/internal-error/src/lib.rs
s373r: In that case, I see two ways to resolve the situation:
1) Leave it as is
2) Move the error declaration closer to where it's used

Q: WDYT?

src/domain/accounts/services/src/services/login_password_auth_provider.rs
s373r: > But perhaps I'd still avoid returning any other information - no DID, no display name or avatar

From `ProviderLoginResponse` removed: `account_type` & `avatar_url` fields. `display_name` field was omitted because it is generated by `Web3WalletAuthenticationProvider` and is different from `account_name`.

But we can't remove `account_id` (DID) just yet, precisely because of `Web3WalletAuthenticationProvider`

src/odf/metadata/src/testing/shortcuts.rs
s373r: Overall, I'm willing to update this part -- no problem -- but I'd like to suggest doing it as part of a separate PR, not in this one.

If you agree, I'll create a separate issue so I don't forget

Q: WDYT?

src/domain/accounts/domain/src/services/account_identity_generator.rs
s373r: > My understanding is that this is needed only to simplify test setup. If so, should we instead add a seed_dids_from_names_in_tests parameter to AccountServiceImpl that can be specified in test harnesses DI setup?

Yes, that would be much better -- it's a great idea.

I updated code and moved a couple of `AccountIdentityGenerator` trivial tests to the `AccountID` area.

Done

src/domain/accounts/domain/src/services/authentication_provider.rs
```diff
@@ -30,12 +30,10 @@ pub trait AuthenticationProvider: Sync + Send {
 
 #[derive(Debug)]
 pub struct ProviderLoginResponse {
-    pub account_id: odf::AccountID,
+    pub account_id: Option<odf::AccountID>,
     pub account_name: odf::AccountName,
     pub email: Email,
     pub display_name: AccountDisplayName,
-    pub account_type: AccountType,
-    pub avatar_url: Option<String>,
```
sergiimk, line 38: We should restore avatar_url per our discussion on slack

src/odf/metadata/src/auth/account_id.rs
sergiimk: This is a good step and the problem is likely on the cargo's side in workspace feature unification. Let's park it for later.

src/odf/metadata/src/testing/shortcuts.rs
sergiimk: Yep, let's do it in a separate PR

src/domain/accounts/services/src/services/login_password_auth_provider.rs
sergiimk: My point was not about removing those fields from `ProviderLoginResponse` entirely but instead making them optional. If provider has authority over the DID - it should return `account_id`. If it hash an avatar - it can return `avatar_url`.

src/utils/internal-error/src/lib.rs
sergiimk: Leave as is and let's revisit after we improve `setty`

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
sergiimk: Consider removing the `quiet` argument altogether

src/domain/accounts/domain/src/services/account_identity_generator.rs
sergiimk: Let's try to replace `SeedDidsFromNamesInTests` with populating DIDs in `AccountConfig::test_config_from_subject` as per our Slack discussion

src/domain/accounts/domain/src/services/authentication_provider.rs
s373r: Done

src/domain/accounts/services/src/services/login_password_auth_provider.rs
s373r: Yes, thank you for explaining that to me (several times, sorry!)

Done

src/domain/accounts/domain/src/use_cases/create_account_use_case.rs
s373r: Done

src/domain/accounts/domain/src/services/account_identity_generator.rs
s373r: We often create accounts not only through `PredefinedAccountsRegistrator`, but also by explicitly creating accounts (especially in the `kamu-adapter-http` tests).

Consequently, the addition of `SeedDidsFromNamesInTests` has been removed wherever possible

s373r: Done

PR #1680, opened 2026-08-12
Add TAB completion for resource type and context names

zaychenko-sergei:
Adds TAB completion for the two things kamu's shell completion could not previously offer, because their valid values aren't known to clap:

- Resource type selectors for get/list/delete. The accepted set (variablesets/vs, secretsets/ss, storages/st, …) comes from the resource registry, so it can never appear in a static completion script. Today you have to consult --help or kamu context api-resources to discover that vs is legal.
- Context names for kamu context use|check|delete and for --context/-c, which come from your own context registry.


src/app/cli/src/commands/complete_command.rs
```diff
@@ -29,6 +32,9 @@ pub struct CompleteCommand {
     remote_repo_reg: Option<Arc<dyn RemoteRepositoryRegistry>>,
     remote_alias_reg: Option<Arc<dyn RemoteAliasesRegistry>>,
     config_service: Arc<ConfigService>,
+    resource_type_lookup_service: Arc<dyn ResourceTypeLookupService>,
+    resource_context_registry_service: Arc<ResourceContextRegistryService>,
+    workspace_service: Arc<WorkspaceService>,
```
sergiimk, line 43: just double check that completions don't crash outside of the workspace - this is the reason most other registries above are injected as optional

src/app/cli/src/commands/completions_command.rs
```diff
@@ -48,3 +53,43 @@ impl Command for CompletionsCommand {
...
+                "enable -n mapfile 2>/dev/null || true\nkamu() {{ printf '%s\\n' 'my prod' \
+                 'other' '*.txt'; }}\n{BASH_COMPLETIONS}\nCOMP_WORDS=(kamu context \
+                 use)\nCOMP_CWORD=2\n_kamu_\nprintf '%s\\n' \"${{COMPREPLY[@]}}\"\n"
+            );
+
+            let output = StdCommand::new("bash")
+                .arg("-c")
+                .arg(&script)
+                .output()
+                .expect("failed to run bash");
```
sergiimk: Test will always fail on windows?

src/app/cli/src/commands/completions_command.rs
```diff
@@ -18,7 +18,12 @@ _kamu_()
     if test $? -ne 0; then
         return 1
     fi
-    COMPREPLY=($( echo -n \"$_COMP_OUTPUTSTR\" ))
+    COMPREPLY=()
+    if test -n \"$_COMP_OUTPUTSTR\"; then
+        while IFS= read -r _KAMU_COMP_CANDIDATE; do
+            COMPREPLY[${#COMPREPLY[@]}]=\"$_KAMU_COMP_CANDIDATE\"
+        done <<< \"$_COMP_OUTPUTSTR\"
+    fi
```
sergiimk: I'm not sure what problem this is fixing. If you'd like to re-test completions on different platforms - go for it, but otherwise I'd rather not change it and wait for a good reason to revisit.

sergiimk (commented):
Most of this code is very hard to understand.

It either:
- Requires re-writing parsing logic to be more clear (currently it's a ton of string checks and splitting impossible to grok)
- Merging as-is after extensive testing

I'd be a lot more comfortable with a more limited version of completions that only completes context names and resource types without trying to cover additional input cases like short flags `-cblah` etc.

src/app/cli/src/commands/completions_command.rs
sergiimk: ok I see the problem described in your PR description ... but do we ever have candidate completions that contain spaces? is this a real problem or just something LLM made up?

src/app/cli/src/commands/completions_command.rs
zaychenko-sergei: Removed

src/app/cli/src/commands/completions_command.rs
zaychenko-sergei: LLM pointed on specific Clap behavior we did not know about, but I agree, it's an overkill for the value, so removed.

src/app/cli/src/commands/complete_command.rs
zaychenko-sergei: Double-checked experimentally.

sergiimk (approved):

PR #1682, opened 2026-08-13
Experiment/cargo.toml formatting

s373r:
### Checklist
- [x] Add CI step(s)
- [x] Wait for changes from the IoC branch
  - https://github.com/kamu-data/kamu-cli/pull/1685


Makefile
```diff
@@ -94,6 +100,8 @@ lint-udeps:
 lint-fix:
 	cargo clippy --workspace --all-targets --fix --allow-dirty --allow-staged --broken-code
 	cargo fmt --all
+	cargo sort -g -w
+	taplo fmt
```
sergiimk: Can we introduce a new `make fmt` target? It's a bit strange to mix clippy and code formatting together.

And if we keep this target - we probably should format code before running clippy not to rebuild twice in case of formatting changes.

Cargo.toml
```diff
@@ -1,5 +1,112 @@
 [workspace]
 members = [
+    # Adapters
```
sergiimk, line 3: Does `cargo-sort` respect these comments as group separators or you sorted these manually?

sergiimk (approved):
I like it! 👍 

The line break in empty arrays was the only thing that jumped at me:
```
default = [
]
```
but I'm ready to live with it.

Makefile
s373r: > It's a bit strange to mix clippy and code formatting together

IMHO, Clippy's rules, which specifically automaticly change the code, are perfectly suited for lint-fix

> Can we introduce a new make fmt target?

But you're right -- it's better to have a make target for that

--

Will do

Cargo.toml
s373r: > Does cargo-sort respect these comments as group separators or you sorted these manually?

That is `cargo-sort`. 

The only thing I did manually was to add the groups in the comments so that sorting wouldn't break up the grouping within Infra, e.g.:

```diff
- # Infra
- ## Outbox
+ # Infra (core)
+ # Infra - Outbox
```

Makefile
s373r: Done

PR #1684, opened 2026-08-14
Login, OAuth: login error if system (local) username matches GitHub username

### LINKED ISSUE #996: Login, OAuth: login error if system (local) username matches GitHub username

s373r:
## Description

Closes: #996

- [x] Compatibility:
    
  - [x] Network APIs: ✅
    
  - [x] Configuration: ✅
    
  - [x] Container images: ✅
- [ ] Observability:
    
  - [ ] Alerts: ✅
- [ ] Documentation:
    
- [ ] Downstream effects:
    

src/domain/accounts/services/src/services/authentication_service_impl.rs
```diff
@@ -87,15 +87,15 @@ impl AuthenticationServiceImpl {
 
     fn resolve_authentication_provider(
         &self,
-        login_method: &str,
+        login_method_lowercase: &str,
     ) -> Result<Arc<dyn AuthenticationProvider>, UnsupportedLoginMethodError> {
         match self
             .authentication_providers_by_method
-            .get(&login_method.to_lowercase())
+            .get(&login_method_lowercase.to_lowercase())
```
sergiimk: should we remove `to_lowercase()` call if it's implied that argument is already normalized?

sergiimk (approved):
Looks nice and simple 👍 

The only improvement I can think of would be to replace
* `check name exists -> create account` logic
* with `try create account -> detect collision -> add suffix -> try again`

This would exclude a very small chance of collision, but the chances of it are ridiculously low, so I'm fine with merging as is.

src/domain/accounts/services/src/services/authentication_service_impl.rs
s373r: > The only improvement I can think of would be to replace
> 
> - check name exists -> create account logic
> - with try create account -> detect collision -> add suffix -> try again

btw, yes, that's a idea -- if we can rely on the database's own guarantees, there isn't much point in performing preliminary checks.

The only thing that needs to be double-checked (which is why I chose this approach in the first place) is `pg`, which does not allow a transaction to be reused if a constraint is violated -- will recheck (ACTION 1)

> should we remove to_lowercase() call if it's implied that argument is already normalized?

eagle eyes -- sure -- will update the place (ACTION 2)

s373r: ACTION 1: Yes, my memory served me right -- we even have safeguards against that kind of behavior in one instance (`PostgresDatasetEntryRepository::save_dataset_entry()`)

Done

--

> should we remove to_lowercase() call if it's implied that argument is already normalized?
> ACTION 2

Done

sergiimk: Great point about `pg` transaction aborting on constraint error. Don't think this simple case warrants us messing with `SAVEPOINT`s so let's keep the current logic as is.

migrations/mysql/20260814190000_accounts_provider_identity_key_composite_unique.sql
```diff
@@ -0,0 +1,11 @@
+/* ------------------------------ */
+
+DROP INDEX idx_accounts_provider_identity_key ON accounts;
+CREATE UNIQUE INDEX idx_uniq_accounts_provider_provider_identity_key
+    ON accounts (provider, provider_identity_key);
+
+DROP INDEX idx_accounts_email ON accounts;
+CREATE UNIQUE INDEX idx_uniq_accounts_provider_email
+    ON accounts (provider, email);
```
sergiimk: Described why I disagree with this in Slack

sergiimk (changes requested):

migrations/mysql/20260814190000_accounts_provider_identity_key_composite_unique.sql
s373r: This part of the changes has been reverted

Done

PR #1689, opened 2026-08-20
Do not panic when completions output pipe is closed early

### LINKED ISSUE #1068: Kamu command line completions crashes
### Describe the bug

Kamu command line completions crash on my Mac

### Steps To Reproduce

0. Docker installed
1. Fresh install of Kamu with `curl -s "https://get.kamu.dev" | sh`
2. Edit `~/.profile` to include `source <(kamu completions bash)`
3. Start shell

### Current Behavior

Jonathans-iMac:Kamu jonathan$ source ~/.profile-kamu
Jonathans-iMac:Kamu jonathan$ thread 'main' panicked at library/std/src/io/stdio.rs:1123:9:
failed printing to stdout: Broken pipe (os error 32)
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: std::io::stdio::_print
   3: <kamu_cli::commands::completions_command::CompletionsCommand as kamu_cli::commands::Command>::run::{{closure}}
   4: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
   5: kamu_cli::app::run::{{closure}}::{{closure}}::{{closure}}
   6: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
   7: kamu_cli::app::run::{{closure}}
   8: tokio::runtime::park::CachedParkThread::block_on
   9: tokio::runtime::context::runtime::enter_runtime
  10: tokio::runtime::runtime::Runtime::block_on
  11: kamu_cli::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

Oh no, looks like kamu has crashed!
Please try running `kamu system diagnose` to verify your environment.
If the issue persists, help us by reporting this problem at https://github.com/kamu-data/kamu-cli/issues

### Expected Behavior

Should not crash

### kamu system info

```sh
{
  "build": {
    "appVersion": "0.222.0",
    "buildTimestamp": "2025-02-06T20:38:13.953011000Z",
    "gitDescribe": "1ea1a6e",
    "gitSha": "1ea1a6e001da7a0ea9561515bcb20deb80656cf7",
    "gitCommitDate": "2025-02-06",
    "gitBranch": "HEAD",
    "rustcSemver": "1.85.0-nightly",
    "rustcChannel": "nightly",
    "rustcHostTriple": "x86_64-apple-darwin",
    "rustcCommitSha": "7c002ff9a70cb84fd1a91bc7b4a0f988cfc36fca",
    "cargoTargetTriple": "x86_64-apple-darwin",
    "cargoFeatures": "default,ingest_evm,ingest_ftp,ingest_mqtt,query_extensions_json,rust_embed,web_ui",
    "cargoOptLevel": "3"
  },
  "workspace": {
    "version": 0,
    "rootDir": ""
  },
  "containerRuntime": {
    "info": {
      "Architecture": "x86_64",
      "BridgeNfIp6tables": true,
      "BridgeNfIptables": true,
      "CDISpecDirs": [],
      "CPUSet": true,
      "CPUShares": true,
      "CgroupDriver": "cgroupfs",
      "CgroupVersion": "2",
      "ClientInfo": {
        "Arch": "amd64",
        "BuildTime": "Fri Sep 20 11:38:18 2024",
        "Context": "desktop-linux",
        "Debug": false,
        "GitCommit": "ce12230",
        "GoVersion": "go1.22.7",
        "Os": "darwin",
        "Plugins": [
          {
            "Name": "ai",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-ai",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Ask Gordon - Docker Agent",
            "Vendor": "Docker Inc.",
            "Version": "v0.1.0"
          },
          {
            "Name": "buildx",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-buildx",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Docker Buildx",
            "Vendor": "Docker Inc.",
            "Version": "v0.18.0-desktop.2"
          },
          {
            "Name": "compose",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-compose",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Docker Compose",
            "Vendor": "Docker Inc.",
            "Version": "v2.30.3-desktop.1"
          },
          {
            "Name": "debug",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-debug",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Get a shell into any image or container",
            "Vendor": "Docker Inc.",
            "Version": "0.0.37"
          },
          {
            "Name": "desktop",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-desktop",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Docker Desktop commands (Alpha)",
            "Vendor": "Docker Inc.",
            "Version": "v0.0.15"
          },
          {
            "Name": "dev",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-dev",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Docker Dev Environments",
            "Vendor": "Docker Inc.",
            "Version": "v0.1.2"
          },
          {
            "Name": "extension",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-extension",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Manages Docker extensions",
            "Vendor": "Docker Inc.",
            "Version": "v0.2.27"
          },
          {
            "Name": "feedback",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-feedback",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Provide feedback, right in your terminal!",
            "Vendor": "Docker Inc.",
            "Version": "v1.0.5"
          },
          {
            "Name": "init",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-init",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Creates Docker-related starter files for your project",
            "Vendor": "Docker Inc.",
            "Version": "v1.4.0"
          },
          {
            "Name": "sbom",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-sbom",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "View the packaged-based Software Bill Of Materials (SBOM) for an image",
            "URL": "https://github.com/docker/sbom-cli-plugin",
            "Vendor": "Anchore Inc.",
            "Version": "0.6.0"
          },
          {
            "Name": "scout",
            "Path": "/Users/jonathan/.docker/cli-plugins/docker-scout",
            "SchemaVersion": "0.1.0",
            "ShortDescription": "Docker Scout",
            "Vendor": "Docker Inc.",
            "Version": "v1.15.0"
          }
        ],
        "Version": "27.3.1",
        "Warnings": null
      },
      "Containerd": {
        "Address": "/run/containerd/containerd.sock",
        "Namespaces": {
          "Containers": "moby",
          "Plugins": "plugins.moby"
        }
      },
      "ContainerdCommit": {
        "Expected": "472731909fa34bd7bc9c087e4c27943f9835f111",
        "ID": "472731909fa34bd7bc9c087e4c27943f9835f111"
      },
      "Containers": 1,
      "ContainersPaused": 0,
      "ContainersRunning": 1,
      "ContainersStopped": 0,
      "CpuCfsPeriod": true,
      "CpuCfsQuota": true,
      "Debug": false,
      "DefaultRuntime": "runc",
      "DockerRootDir": "/var/lib/docker",
      "Driver": "overlayfs",
      "DriverStatus": [
        [
          "driver-type",
          "io.containerd.snapshotter.v1"
        ]
      ],
      "ExperimentalBuild": false,
      "GenericResources": null,
      "HttpProxy": "http.docker.internal:3128",
      "HttpsProxy": "http.docker.internal:3128",
      "ID": "91b3df91-7978-4600-a1c5-2e0c8991dffb",
      "IPv4Forwarding": true,
      "Images": 4,
      "IndexServerAddress": "https://index.docker.io/v1/",
      "InitBinary": "docker-init",
      "InitCommit": {
        "Expected": "de40ad0",
        "ID": "de40ad0"
      },
      "Isolation": "",
      "KernelVersion": "6.10.14-linuxkit",
      "Labels": [
        "com.docker.desktop.address=unix:///Users/jonathan/Library/Containers/com.docker.docker/Data/docker-cli.sock"
      ],
      "LiveRestoreEnabled": false,
      "LoggingDriver": "json-file",
      "MemTotal": 8218161152,
      "MemoryLimit": true,
      "NCPU": 16,
      "NEventsListener": 15,
      "NFd": 61,
      "NGoroutines": 92,
      "Name": "docker-desktop",
      "NoProxy": "hubproxy.docker.internal",
      "OSType": "linux",
      "OSVersion": "",
      "OomKillDisable": false,
      "OperatingSystem": "Docker Desktop",
      "PidsLimit": true,
      "Plugins": {
        "Authorization": null,
        "Log": [
          "awslogs",
          "fluentd",
          "gcplogs",
          "gelf",
          "journald",
          "json-file",
          "local",
          "splunk",
          "syslog"
        ],
        "Network": [
          "bridge",
          "host",
          "ipvlan",
          "macvlan",
          "null",
          "overlay"
        ],
        "Volume": [
          "local"
        ]
      },
      "RegistryConfig": {
        "AllowNondistributableArtifactsCIDRs": null,
        "AllowNondistributableArtifactsHostnames": null,
        "IndexConfigs": {
          "docker.io": {
            "Mirrors": [],
            "Name": "docker.io",
            "Official": true,
            "Secure": true
          },
          "hubproxy.docker.internal:5555": {
            "Mirrors": [],
            "Name": "hubproxy.docker.internal:5555",
            "Official": false,
            "Secure": false
          }
        },
        "InsecureRegistryCIDRs": [
          "127.0.0.0/8"
        ],
        "Mirrors": null
      },
      "RuncCommit": {
        "Expected": "v1.1.13-0-g58aa920",
        "ID": "v1.1.13-0-g58aa920"
      },
      "Runtimes": {
        "io.containerd.runc.v2": {
          "path": "runc",
          "status": {
            "org.opencontainers.runtime-spec.features": "{\"ociVersionMin\":\"1.0.0\",\"ociVersionMax\":\"1.0.2-dev\",\"hooks\":[\"prestart\",\"createRuntime\",\"createContainer\",\"startContainer\",\"poststart\",\"poststop\"],\"mountOptions\":[\"acl\",\"async\",\"atime\",\"bind\",\"defaults\",\"dev\",\"diratime\",\"dirsync\",\"exec\",\"iversion\",\"lazytime\",\"loud\",\"mand\",\"noacl\",\"noatime\",\"nodev\",\"nodiratime\",\"noexec\",\"noiversion\",\"nolazytime\",\"nomand\",\"norelatime\",\"nostrictatime\",\"nosuid\",\"nosymfollow\",\"private\",\"ratime\",\"rbind\",\"rdev\",\"rdiratime\",\"relatime\",\"remount\",\"rexec\",\"rnoatime\",\"rnodev\",\"rnodiratime\",\"rnoexec\",\"rnorelatime\",\"rnostrictatime\",\"rnosuid\",\"rnosymfollow\",\"ro\",\"rprivate\",\"rrelatime\",\"rro\",\"rrw\",\"rshared\",\"rslave\",\"rstrictatime\",\"rsuid\",\"rsymfollow\",\"runbindable\",\"rw\",\"shared\",\"silent\",\"slave\",\"strictatime\",\"suid\",\"symfollow\",\"sync\",\"tmpcopyup\",\"unbindable\"],\"linux\":{\"namespaces\":[\"cgroup\",\"ipc\",\"mount\",\"network\",\"pid\",\"user\",\"uts\"],\"capabilities\":[\"CAP_CHOWN\",\"CAP_DAC_OVERRIDE\",\"CAP_DAC_READ_SEARCH\",\"CAP_FOWNER\",\"CAP_FSETID\",\"CAP_KILL\",\"CAP_SETGID\",\"CAP_SETUID\",\"CAP_SETPCAP\",\"CAP_LINUX_IMMUTABLE\",\"CAP_NET_BIND_SERVICE\",\"CAP_NET_BROADCAST\",\"CAP_NET_ADMIN\",\"CAP_NET_RAW\",\"CAP_IPC_LOCK\",\"CAP_IPC_OWNER\",\"CAP_SYS_MODULE\",\"CAP_SYS_RAWIO\",\"CAP_SYS_CHROOT\",\"CAP_SYS_PTRACE\",\"CAP_SYS_PACCT\",\"CAP_SYS_ADMIN\",\"CAP_SYS_BOOT\",\"CAP_SYS_NICE\",\"CAP_SYS_RESOURCE\",\"CAP_SYS_TIME\",\"CAP_SYS_TTY_CONFIG\",\"CAP_MKNOD\",\"CAP_LEASE\",\"CAP_AUDIT_WRITE\",\"CAP_AUDIT_CONTROL\",\"CAP_SETFCAP\",\"CAP_MAC_OVERRIDE\",\"CAP_MAC_ADMIN\",\"CAP_SYSLOG\",\"CAP_WAKE_ALARM\",\"CAP_BLOCK_SUSPEND\",\"CAP_AUDIT_READ\",\"CAP_PERFMON\",\"CAP_BPF\",\"CAP_CHECKPOINT_RESTORE\"],\"cgroup\":{\"v1\":true,\"v2\":true,\"systemd\":true,\"systemdUser\":true},\"seccomp\":{\"enabled\":true,\"actions\":[\"SCMP_ACT_ALLOW\",\"SCMP_ACT_ERRNO\",\"SCMP_ACT_KILL\",\"SCMP_ACT_KILL_PROCESS\",\"SCMP_ACT_KILL_THREAD\",\"SCMP_ACT_LOG\",\"SCMP_ACT_NOTIFY\",\"SCMP_ACT_TRACE\",\"SCMP_ACT_TRAP\"],\"operators\":[\"SCMP_CMP_EQ\",\"SCMP_CMP_GE\",\"SCMP_CMP_GT\",\"SCMP_CMP_LE\",\"SCMP_CMP_LT\",\"SCMP_CMP_MASKED_EQ\",\"SCMP_CMP_NE\"],\"archs\":[\"SCMP_ARCH_AARCH64\",\"SCMP_ARCH_ARM\",\"SCMP_ARCH_MIPS\",\"SCMP_ARCH_MIPS64\",\"SCMP_ARCH_MIPS64N32\",\"SCMP_ARCH_MIPSEL\",\"SCMP_ARCH_MIPSEL64\",\"SCMP_ARCH_MIPSEL64N32\",\"SCMP_ARCH_PPC\",\"SCMP_ARCH_PPC64\",\"SCMP_ARCH_PPC64LE\",\"SCMP_ARCH_RISCV64\",\"SCMP_ARCH_S390\",\"SCMP_ARCH_S390X\",\"SCMP_ARCH_X32\",\"SCMP_ARCH_X86\",\"SCMP_ARCH_X86_64\"]},\"apparmor\":{\"enabled\":true},\"selinux\":{\"enabled\":true}},\"annotations\":{\"io.github.seccomp.libseccomp.version\":\"2.5.4\",\"org.opencontainers.runc.checkpoint.enabled\":\"true\",\"org.opencontainers.runc.commit\":\"v1.1.13-0-g58aa920\",\"org.opencontainers.runc.version\":\"1.1.13\"}}"
          }
        },
        "runc": {
          "path": "runc",
          "status": {
            "org.opencontainers.runtime-spec.features": "{\"ociVersionMin\":\"1.0.0\",\"ociVersionMax\":\"1.0.2-dev\",\"hooks\":[\"prestart\",\"createRuntime\",\"createContainer\",\"startContainer\",\"poststart\",\"poststop\"],\"mountOptions\":[\"acl\",\"async\",\"atime\",\"bind\",\"defaults\",\"dev\",\"diratime\",\"dirsync\",\"exec\",\"iversion\",\"lazytime\",\"loud\",\"mand\",\"noacl\",\"noatime\",\"nodev\",\"nodiratime\",\"noexec\",\"noiversion\",\"nolazytime\",\"nomand\",\"norelatime\",\"nostrictatime\",\"nosuid\",\"nosymfollow\",\"private\",\"ratime\",\"rbind\",\"rdev\",\"rdiratime\",\"relatime\",\"remount\",\"rexec\",\"rnoatime\",\"rnodev\",\"rnodiratime\",\"rnoexec\",\"rnorelatime\",\"rnostrictatime\",\"rnosuid\",\"rnosymfollow\",\"ro\",\"rprivate\",\"rrelatime\",\"rro\",\"rrw\",\"rshared\",\"rslave\",\"rstrictatime\",\"rsuid\",\"rsymfollow\",\"runbindable\",\"rw\",\"shared\",\"silent\",\"slave\",\"strictatime\",\"suid\",\"symfollow\",\"sync\",\"tmpcopyup\",\"unbindable\"],\"linux\":{\"namespaces\":[\"cgroup\",\"ipc\",\"mount\",\"network\",\"pid\",\"user\",\"uts\"],\"capabilities\":[\"CAP_CHOWN\",\"CAP_DAC_OVERRIDE\",\"CAP_DAC_READ_SEARCH\",\"CAP_FOWNER\",\"CAP_FSETID\",\"CAP_KILL\",\"CAP_SETGID\",\"CAP_SETUID\",\"CAP_SETPCAP\",\"CAP_LINUX_IMMUTABLE\",\"CAP_NET_BIND_SERVICE\",\"CAP_NET_BROADCAST\",\"CAP_NET_ADMIN\",\"CAP_NET_RAW\",\"CAP_IPC_LOCK\",\"CAP_IPC_OWNER\",\"CAP_SYS_MODULE\",\"CAP_SYS_RAWIO\",\"CAP_SYS_CHROOT\",\"CAP_SYS_PTRACE\",\"CAP_SYS_PACCT\",\"CAP_SYS_ADMIN\",\"CAP_SYS_BOOT\",\"CAP_SYS_NICE\",\"CAP_SYS_RESOURCE\",\"CAP_SYS_TIME\",\"CAP_SYS_TTY_CONFIG\",\"CAP_MKNOD\",\"CAP_LEASE\",\"CAP_AUDIT_WRITE\",\"CAP_AUDIT_CONTROL\",\"CAP_SETFCAP\",\"CAP_MAC_OVERRIDE\",\"CAP_MAC_ADMIN\",\"CAP_SYSLOG\",\"CAP_WAKE_ALARM\",\"CAP_BLOCK_SUSPEND\",\"CAP_AUDIT_READ\",\"CAP_PERFMON\",\"CAP_BPF\",\"CAP_CHECKPOINT_RESTORE\"],\"cgroup\":{\"v1\":true,\"v2\":true,\"systemd\":true,\"systemdUser\":true},\"seccomp\":{\"enabled\":true,\"actions\":[\"SCMP_ACT_ALLOW\",\"SCMP_ACT_ERRNO\",\"SCMP_ACT_KILL\",\"SCMP_ACT_KILL_PROCESS\",\"SCMP_ACT_KILL_THREAD\",\"SCMP_ACT_LOG\",\"SCMP_ACT_NOTIFY\",\"SCMP_ACT_TRACE\",\"SCMP_ACT_TRAP\"],\"operators\":[\"SCMP_CMP_EQ\",\"SCMP_CMP_GE\",\"SCMP_CMP_GT\",\"SCMP_CMP_LE\",\"SCMP_CMP_LT\",\"SCMP_CMP_MASKED_EQ\",\"SCMP_CMP_NE\"],\"archs\":[\"SCMP_ARCH_AARCH64\",\"SCMP_ARCH_ARM\",\"SCMP_ARCH_MIPS\",\"SCMP_ARCH_MIPS64\",\"SCMP_ARCH_MIPS64N32\",\"SCMP_ARCH_MIPSEL\",\"SCMP_ARCH_MIPSEL64\",\"SCMP_ARCH_MIPSEL64N32\",\"SCMP_ARCH_PPC\",\"SCMP_ARCH_PPC64\",\"SCMP_ARCH_PPC64LE\",\"SCMP_ARCH_RISCV64\",\"SCMP_ARCH_S390\",\"SCMP_ARCH_S390X\",\"SCMP_ARCH_X32\",\"SCMP_ARCH_X86\",\"SCMP_ARCH_X86_64\"]},\"apparmor\":{\"enabled\":true},\"selinux\":{\"enabled\":true}},\"annotations\":{\"io.github.seccomp.libseccomp.version\":\"2.5.4\",\"org.opencontainers.runc.checkpoint.enabled\":\"true\",\"org.opencontainers.runc.commit\":\"v1.1.13-0-g58aa920\",\"org.opencontainers.runc.version\":\"1.1.13\"}}"
          }
        }
      },
      "SecurityOptions": [
        "name=seccomp,profile=unconfined",
        "name=cgroupns"
      ],
      "ServerVersion": "27.3.1",
      "SwapLimit": true,
      "Swarm": {
        "ControlAvailable": false,
        "Error": "",
        "LocalNodeState": "inactive",
        "NodeAddr": "",
        "NodeID": "",
        "RemoteManagers": null
      },
      "SystemTime": "2025-02-10T22:23:06.575610897Z",
      "Warnings": [
        "WARNING: daemon is not using the default seccomp profile"
      ]
    }
  }
}
```

### Logs

No such log exists

### Anything else?

_No response_

hdimer:
## Description

Closes: #1068

`kamu completions <shell>` panics with a full backtrace and the "Oh no, looks like kamu has crashed!" banner when the consumer closes the pipe early. This hits the invocation the CLI's own `--help` recommends (`src/app/cli/src/cli.rs:241`):

```
source <(kamu completions bash)
```

On `master`, `bash`, `zsh` and `fish` all exit `101`, from two different panic sites: `print!` for `bash` (`stdio.rs:1165`), and `clap_complete::generate()` for everything else (`clap_complete-4.6.9` `shell.rs:86`, `fish.rs:277`).

### The fix

Switched to [`pipecheck`](https://docs.rs/pipecheck/) as discussed. `std::io::stdout()` is wrapped in `pipecheck::wrap()` in `run()`, so a broken pipe terminates the process by SIGPIPE from inside the `write` call and `generate()` never gets the error it expects on. The buffer is gone.

For scale on the buffering objection: the `zsh` script is ~100 KB, so the first version really was materialising the whole thing.

Two smaller points:

- The `bash` branch moves from `print!` to `write!` into the same writer, so it goes through `pipecheck` too instead of panicking on its own.
- `write_completions()` flushes explicitly. `Stdout` is line buffered and its exit-time flush happens outside the wrapper, so without this a pipe breaking on the last buffered chunk would be swallowed and we would exit `0`.

The `write_completions(&mut impl Write)` seam is kept. It mirrors the sibling command in this same feature, `CompleteCommand::complete` (`src/app/cli/src/commands/complete_command.rs:147`), and keeps the per-shell content assertions as cheap in-process tests.

### On the dependency

- Its only dependency is `libc`, already a direct dependency of `kamu-cli`, so `Cargo.lock` gains exactly one entry.
- `cargo deny check` is clean: `advisories ok, bans ok, licenses ok, sources ok`.
- It is still young (0.2.0, MIT). The crate is written to be vendored if you would rather not take the dependency: `src/pipecheck.rs` is the entire implementation, self-contained and separately licensed for copy-paste. Happy to switch to an in-tree module instead, it is a small change either way.

### What this does not cover

`clap_complete::generate()` still panics on writer errors that are *not* `BrokenPipe` (`ENOSPC` when redirecting to a full disk, say) for every shell except `bash`. That is unchanged from `master` and out of scope here.

### Verification

Local, macOS, `aarch64`:

- All of `bash`, `zsh`, `fish` now terminate by `SIGPIPE` with empty stderr. Before: `101` plus the crash banner, confirmed by reverting just this file and rebuilding.
- Generated scripts byte-identical to `master` for `bash`, `zsh`, `fish`, `powershell` and `elvish`.
- `source <(kamu completions bash)` exits `0`, and `kamu completions zsh | head -1` exits `141`.
  The latter is a genuine mid-stream break rather than a first-write one, since the `zsh` script is larger than the pipe buffer.
- `cargo fmt --check`, `make clippy` and `cargo deny check` clean; the 3 new tests pass.

The regression test spawns the binary with an already-closed pipe read end, so the failing write is deterministic rather than dependent on scheduling between parent and child.

- [x] Compatibility:
  - [x] Network APIs: ✅
  - [x] Configuration: ✅
  - [x] Container images: ✅
- [x] Observability:
  - [x] Tracing / metrics: ✅ (n/a - removes a panic in a local, non-server command)
  - [x] Alerts: ✅ (n/a)
- [x] Documentation:
  - [x] [Public documentation](https://github.com/kamu-data/kamu-docs/): ✅ (n/a - no user-facing behaviour change beyond not crashing)
- [x] Downstream effects:

---

Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Recommendation: for ease of reading, use the following format:
 ### Fixed
 -->
 
+## Unreleased
+### Fixed
+- CLI: `kamu completions` no longer panics when the output pipe is closed early, e.g. `source <(kamu completions bash)` (#1068)
+
 ## [0.266.1] - 2026-08-19
 ### Fixed
 - Login, OAuth: login error if system (local) username matches GitHub username (#1684)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8142,6 +8142,7 @@ dependencies = [
  "observability",
  "opendatafabric",
  "paste",
+ "pipecheck",
  "pretty_assertions",
  "prettytable-rs",
  "prometheus",
@@ -11391,6 +11392,15 @@ name = "pin-utils"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
+
+[[package]]
+name = "pipecheck"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2337597d75d3d1b25e67c7e5938c0b9cffa9626d2b1b91c940dbebd6b073ccdf"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "pkcs1"
--- a/src/app/cli/Cargo.toml
+++ b/src/app/cli/Cargo.toml
@@ -245,6 +245,7 @@ hex = { version = "0.4", default-features = false, features = [] }
 indoc = "2"
 itertools = "0.14"
 libc = "0.2" # Signal names
+pipecheck = "0.2" # Unix-style broken pipe handling for stdout
 regex = "1"
 secrecy = "0.10"
 shlex = "1" # Parsing partial input for custom completions
--- a/src/app/cli/src/commands/completions_command.rs
+++ b/src/app/cli/src/commands/completions_command.rs
@@ -7,7 +7,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::io::{ErrorKind, Write};
+use std::io::Write;
 
 use clap::CommandFactory as _;
 
@@ -41,27 +41,27 @@ impl CompletionsCommand {
         let mut cli = crate::cli::Cli::command();
         let bin_name = cli.get_name().to_owned();
 
-        // Note: the script is rendered into a buffer first, as
-        // `clap_complete::generate()` panics on writer errors. The fallible
-        // `Generator::try_generate()` would avoid this, but requires replicating
-        // `set_bin_name()` + `build()` by hand
-        let mut script = Vec::new();
         match self.shell {
-            clap_complete::Shell::Bash => script.extend_from_slice(BASH_COMPLETIONS.as_bytes()),
-            _ => clap_complete::generate(self.shell, &mut cli, bin_name, &mut script),
+            clap_complete::Shell::Bash => write!(output, "{BASH_COMPLETIONS}")?,
+            _ => clap_complete::generate(self.shell, &mut cli, bin_name, output),
         }
 
-        match output.write_all(&script) {
-            // Consumers may exit before reading all output, e.g. `source <(kamu completions bash)`
-            Err(err) if err.kind() == ErrorKind::BrokenPipe => Ok(()),
-            res => Ok(res?),
-        }
+        // Every generator ends in a newline today, so `Stdout`'s line buffer is empty
+        // by now - but its exit-time flush happens outside whichever writer we were
+        // handed, so flush through that writer rather than relying on it staying so
+        output.flush()?;
+
+        Ok(())
     }
 }
 
 #[async_trait::async_trait(?Send)]
 impl Command for CompletionsCommand {
     async fn run(&self) -> Result<(), CLIError> {
-        self.write_completions(&mut std::io::stdout())
+        // `source <(kamu completions bash)`, the invocation suggested by `--help`,
+        // stops reading before the script ends. Rust ignores SIGPIPE, so the write
+        // returns `BrokenPipe`, which both `write!` and `clap_complete::generate()`
+        // turn into a panic - `pipecheck` lets the signal terminate us instead
+        self.write_completions(&mut pipecheck::wrap(std::io::stdout()))
     }
 }
--- a/src/app/cli/src/commands/completions_command.rs
+++ b/src/app/cli/src/commands/completions_command.rs
@@ -7,6 +7,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::io::{ErrorKind, Write};
+
 use clap::CommandFactory as _;
 
 use super::{CLIError, Command};
@@ -32,19 +34,34 @@ pub struct CompletionsCommand {
     shell: clap_complete::Shell,
 }
 
-#[async_trait::async_trait(?Send)]
-impl Command for CompletionsCommand {
-    async fn run(&self) -> Result<(), CLIError> {
+impl CompletionsCommand {
+    pub fn write_completions(&self, output: &mut impl Write) -> Result<(), CLIError> {
         // TODO: Remove once clap allows to programmatically complete values
         // See: https://github.com/clap-rs/clap/issues/568
         let mut cli = crate::cli::Cli::command();
         let bin_name = cli.get_name().to_owned();
+
+        // Note: the script is rendered into a buffer first, as
+        // `clap_complete::generate()` panics on writer errors. The fallible
+        // `Generator::try_generate()` would avoid this, but requires replicating
+        // `set_bin_name()` + `build()` by hand
+        let mut script = Vec::new();
         match self.shell {
-            clap_complete::Shell::Bash => print!("{BASH_COMPLETIONS}"),
-            _ => {
-                clap_complete::generate(self.shell, &mut cli, bin_name, &mut std::io::stdout());
-            }
+            clap_complete::Shell::Bash => script.extend_from_slice(BASH_COMPLETIONS.as_bytes()),
+            _ => clap_complete::generate(self.shell, &mut cli, bin_name, &mut script),
         }
-        Ok(())
+
+        match output.write_all(&script) {
+            // Consumers may exit before reading all output, e.g. `source <(kamu completions bash)`
+            Err(err) if err.kind() == ErrorKind::BrokenPipe => Ok(()),
+            res => Ok(res?),
+        }
+    }
+}
+
+#[async_trait::async_trait(?Send)]
+impl Command for CompletionsCommand {
+    async fn run(&self) -> Result<(), CLIError> {
+        self.write_completions(&mut std::io::stdout())
     }
 }
--- a/src/app/cli/tests/tests/mod.rs
+++ b/src/app/cli/tests/tests/mod.rs
@@ -8,6 +8,7 @@
 // by the Apache License, Version 2.0.
 
 mod test_access_token_registry_svc;
+mod test_completions_command;
 mod test_config;
 mod test_di_graph;
 mod test_generate_cli_markdown;
--- a/src/app/cli/tests/tests/test_completions_command.rs
+++ b/src/app/cli/tests/tests/test_completions_command.rs
@@ -1,0 +1,82 @@
+// Copyright Kamu Data, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::assert_matches;
+use std::io::{Error, ErrorKind, Write};
+use std::sync::Arc;
+
+use clap::ValueEnum as _;
+use dill::TypedBuilder;
+use kamu_cli::CLIError;
+use kamu_cli::commands::*;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+struct FailingWriter(ErrorKind);
+
+impl Write for FailingWriter {
+    fn write(&mut self, _buf: &[u8]) -> std::io::Result<usize> {
+        Err(self.0.into())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Err(self.0.into())
+    }
+}
+
+fn completions_command(shell: clap_complete::Shell) -> Arc<CompletionsCommand> {
+    CompletionsCommand::builder(shell)
+        .get(&dill::Catalog::builder().build())
+        .unwrap()
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#[test_log::test]
+fn test_writes_completions() {
+    for (shell, expected) in [
+        (clap_complete::Shell::Bash, "complete -F _kamu_ kamu"),
+        (clap_complete::Shell::Zsh, "#compdef kamu"),
+        (clap_complete::Shell::Fish, "complete -c kamu"),
+    ] {
+        let mut output = Vec::new();
+        completions_command(shell)
+            .write_completions(&mut output)
+            .unwrap();
+
+        let script = String::from_utf8(output).unwrap();
+        assert!(script.contains(expected), "{shell}: {script}");
+    }
+}
+
+#[test_log::test]
+fn test_broken_pipe_is_ignored() {
+    // Note: every shell other than Bash is rendered by `clap_complete::generate()`,
+    // which panics on writer errors, so this also pins the decision to render the
+    // script into a buffer first
+    for shell in clap_complete::Shell::value_variants() {
+        completions_command(*shell)
+            .write_completions(&mut FailingWriter(ErrorKind::BrokenPipe))
+            .unwrap();
+    }
+}
+
+#[test_log::test]
+fn test_other_write_errors_are_propagated() {
+    let res = completions_command(clap_complete::Shell::Bash)
+        .write_completions(&mut FailingWriter(ErrorKind::PermissionDenied));
+
+    assert_matches!(
+        res,
+        Err(CLIError::Failure { source, .. })
+            if source.downcast_ref::<Error>().map(Error::kind) == Some(ErrorKind::PermissionDenied)
+    );
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/app/cli/tests/tests/test_completions_command.rs
+++ b/src/app/cli/tests/tests/test_completions_command.rs
@@ -56,19 +56,13 @@ fn test_writes_completions() {
 }
 
 #[test_log::test]
-fn test_broken_pipe_is_ignored() {
-    // Note: every shell other than Bash is rendered by `clap_complete::generate()`,
-    // which panics on writer errors, so this also pins the decision to render the
-    // script into a buffer first
-    for shell in clap_complete::Shell::value_variants() {
-        completions_command(*shell)
-            .write_completions(&mut FailingWriter(ErrorKind::BrokenPipe))
-            .unwrap();
-    }
-}
-
-#[test_log::test]
-fn test_other_write_errors_are_propagated() {
+fn test_write_errors_are_propagated() {
+    // Note: only Bash is covered here. Every other shell is rendered by
+    // `clap_complete::generate()`, which panics rather than returning writer
+    // errors, and its fallible `try_generate()` is no escape either - `Fish`
+    // still panics internally (`clap_complete-4.6.9` `fish.rs:277`). Broken
+    // pipes are handled a level up, by the `pipecheck` writer `run()` passes in
+    // - see the test below
     let res = completions_command(clap_complete::Shell::Bash)
         .write_completions(&mut FailingWriter(ErrorKind::PermissionDenied));
 
@@ -77,6 +71,47 @@ fn test_other_write_errors_are_propagated() {
         Err(CLIError::Failure { source, .. })
             if source.downcast_ref::<Error>().map(Error::kind) == Some(ErrorKind::PermissionDenied)
     );
+}
+
+// A broken pipe takes the whole process down, so unlike the cases above it can
+// only be observed from the outside
+#[cfg(unix)]
+#[test_log::test]
+fn test_broken_pipe_terminates_by_sigpipe() {
+    use std::os::unix::process::ExitStatusExt as _;
+
+    // Run outside any `.kamu` that happens to sit above the checkout
+    let workdir = tempfile::tempdir().unwrap();
+
+    for shell in clap_complete::Shell::value_variants() {
+        let (reader, writer) = std::io::pipe().unwrap();
+
+        // Closing the read end before spawning makes the very first write fail, so the
+        // test does not depend on the child being scheduled before or after this point
+        drop(reader);
+
+        let output = std::process::Command::new(env!("CARGO_BIN_EXE_kamu-cli"))
+            .args(["completions", &shell.to_string()])
+            .current_dir(workdir.path())
+            .stdout(writer)
+            .stderr(std::process::Stdio::piped())
+            .spawn()
+            .unwrap()
+            .wait_with_output()
+            .unwrap();
+
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        assert!(
+            !stderr.contains("panicked") && !stderr.contains("has crashed"),
+            "{shell}: {stderr}"
+        );
+        assert_eq!(
+            output.status.signal(),
+            Some(libc::SIGPIPE),
+            "{shell}: {:?}, stderr: {stderr}",
+            output.status
+        );
+    }
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////


sergiimk:
Hey @hdimer thank you for the contribution!

Given how many various commands we have that output to `stdout` I'm not sure explicitly handling the `BrokenPipe` error is the best solution. Some commads may produce a lot of output and buffering it is not an option.

I was thinking that maybe we need some `Writer` wrapper that detects `BrokenPipe` under the hood and ignores the rest of the output, but this has drawbacks too:
- Application would continue to write to the already invalid stream instead of aborting early, which can be a problem for commands like `sql` that crunch through a lot of data
- Statements like `println!` can still panic as they `unwrap()` under the hood.

After some research I see that:
- On unix Rust ignores `SIGPIPE` signal [by default](https://github.com/rust-lang/rust/issues/62569) to make it appear as a write error instead of shutting down the application (default libc behavior)
- Some applications [restore](https://github.com/BurntSushi/ripgrep/pull/586/changes) the default signal behavior
- By doing this application will get interrupted by the signal instead of seeing an error, which I believe is exactly what we want
- But there are several downsides:
  - Portability - but since we don't fully support running on Windows without WSL anyway this shouldn't be an issue
  - Spawning child processes - apparently `Command::exec` resets the signal behavior ... and we do spawn child processes often for containers :(

After even more research I found [`pipecheck`](https://docs.rs/pipecheck/latest/pipecheck/) crate that seems to provide a `Writer` wrapper that will check for underlying error and shutdown the application. So far this is the best option I see - we could start using it in commands that stream output into `stdout` and later progressively eliminate any raw `println!` statements.

What do you think?

hdimer:
Fair on the buffering, it doesn't generalize. It's fine for a few KB of completion script, but not something I'd want anywhere near `sql`.

Two things I checked before answering, because they shift the tradeoff a bit.

**The child-process objection to restoring SIGPIPE doesn't hold.** std already resets SIGPIPE to `SIG_DFL` in the child before exec, on both `spawn()` and `CommandExt::exec()` — see `do_exec` and the `posix_spawn` path in [`sys/process/unix`](https://doc.rust-lang.org/src/std/sys/process/unix/unix.rs.html), both guarded by `if !on_broken_pipe_used()`. So the containers you spawn get `SIG_DFL` today. If you restored `SIG_DFL` in `main` the ripgrep way, that guard doesn't flip (it only flips when `-Zon-broken-pipe` is passed), so children still get reset to `SIG_DFL`. And if you did use the flag, they'd inherit your `SIG_DFL`. Every path lands on the same disposition the children already see.

Sockets are fine too — std uses `MSG_NOSIGNAL`/`SO_NOSIGPIPE` on `TcpStream`, so a client dropping off the api-server can't take the process down. (`UnixStream` doesn't, rust-lang/rust#139956, but you don't use it.)

**pipecheck is really the scoped version of that same idea.** It doesn't hand you back an error — it unblocks SIGPIPE, resets the disposition to default and raises it at itself, falling back to `exit(1)`. Same "die on broken pipe" semantics as restoring the signal, just limited to the writers you opt in. It does fix this command as a side effect: `clap_complete::generate()` unwraps writer errors internally, but with pipecheck the process is already gone inside `write()`, so the buffering goes away.

So how I'd frame it:

- **pipecheck** — opt-in, aborts early, no global signal change. Doesn't help `println!`, and eliminating those is a long tail.
- **`SIG_DFL` in `main`** — one line, fixes every command and every `println!` at once, but blunt: any EPIPE anywhere is instant death.

Worth knowing that both terminate without unwinding, so no destructors run either way. If any container teardown hangs off `Drop`, that applies equally to both and isn't a reason to prefer one.

They're also not mutually exclusive: pipecheck where you stream, `SIG_DFL` if you'd rather not chase down the `println!` tail.

Either way, happy to redo this PR on pipecheck as the first adopter — it's a small command and the invocation that breaks is the one `--help` recommends. Let me know whether you want the dependency landed here or in its own PR. Only thing I'd flag is that pipecheck is on 0.2.0 and only a few months old, so if adding a dep needs sign-off on your side, better to know now than after the rewrite.

sergiimk:
Ok Mr. AI, let's redo this PR to use `pipecheck` as a more scoped solution.

hdimer:
Done, pushed (rebased on master).

Wrapping stdout in `pipecheck::wrap()` let the buffer go entirely, which was the actual objection. For scale: the `zsh` script is ~100 KB, so it really was materialising the whole thing. `kamu completions zsh | head -1` now exits `141` on a genuine mid-stream break rather than a first-write one.

Two things worth your eyes:

- `write_completions()` flushes explicitly now. `Stdout` is line buffered and its exit-time flush happens outside the wrapper, so without it a pipe breaking on the last buffered chunk would be swallowed and we would exit `0`. Anything else that adopts `pipecheck` later will need the same care.
- On the dependency: `cargo deny check` comes back `advisories ok, bans ok, licenses ok, sources ok`, and it only pulls `libc`, which you already depend on directly, so the lock gains one entry. If 0.2.0 is too young for your taste, the crate is explicitly built for vendoring - `src/pipecheck.rs` is the whole implementation, self-contained and separately MIT-licensed for copy-paste. Happy to swap the dependency for an in-tree module, it is a small change either way.

Verified by reverting just this file and rebuilding: `bash`, `zsh` and `fish` go from `101` plus the crash banner to termination by `SIGPIPE` with empty stderr, and the generated scripts are byte-identical to `master` across all five shells.

I have left the raw `println!` tail alone - that is the wider cleanup you mentioned and wants its own PR.

Also dropped the #1680 conflict note from the description: it landed on a feature branch and its final diff does not touch this file, so there is nothing to merge around.

hdimer:
Small follow-up push: the regression test now covers every `clap_complete::Shell` variant rather than three, and runs in a temp cwd so it does not depend on whatever `.kamu` sits above the checkout. Comment fixes only otherwise.

Two corrections to things I said above, both in the direction of claiming less:

**The explicit flush is a guard, not a live fix.** Every generator's output ends in a newline, so `Stdout`'s line buffer is always empty by the time we get there and the flush issues no syscall. It matters the moment a generator stops ending in `\n` or a caller passes a `BufWriter`, since the exit-time flush happens outside the wrapper and drops its error. Worth keeping, but I oversold it.

**`try_generate()` is not the alternative I implied.** `clap::Command::build` is public, so calling it is two supported lines, not the hand-rolling I suggested. What actually rules it out is that it is not fully fallible: `Fish` writes straight to the buffer and ends on `.expect()` (`clap_complete-4.6.9` `fish.rs:277`), so you would fix four shells out of five and still panic on the fifth.

**On scope, which you may want to weigh before merging.** I checked the rest of the CLI against a pre-closed pipe. `kamu config list` reproduces the exact same `stdio.rs:1165` panic as #1068. More relevant to this feature: `kamu complete`, which the bash script this PR ships invokes on every Tab press, gives

```
Error:
  0: Internal error
  1: Broken pipe (os error 32)
Help us by reporting this problem at https://github.com/kamu-data/kamu-cli/issues
```

for an ordinary Unix condition. There are ~25 raw `io::stdout()` sites and no broken-pipe handling anywhere else in `src/app/cli`.

I have deliberately not widened this PR - you asked for the scoped version and the progressive rollout, and `pipecheck` makes each follow-up a one-line change at the call site. Happy to do `complete` next as its own PR if you want the pair, since it is the one users actually hit. Worth knowing that a shared `output::stdout()` helper cannot fully encapsulate this: `pipecheck::Writer` has no `Drop`, so the flush has to stay at each call site.

sergiimk (approved):
lgtm 👍 

Will merge once CI passes.

Once merged would appreciate a PR that does the same for `complete` command.

hdimer:
Follow-up as requested: #1696 gives `complete` the same treatment.

It also converts the 11 `writeln!(...).unwrap()` sites in the completion helpers to `.int_err()?` - `pipecheck` covers `BrokenPipe` on its own, but those would still panic on any other write error, and the same file already used `.int_err()?` for three of its writes. That part is a separate commit, so it is easy to drop if you would rather keep the PR to the `pipecheck` line.

PR #1695, opened 2026-08-29
Fix SELinux detection spawning a shell built-in

### LINKED ISSUE #1692: Warning: Error detecting SELinux presence
On systems without SELinux the detection code is constantly producing warnings:

```sh
2026-08-28T23:49:47.501823Z  WARN container_runtime::runtime: Error detecting SELinux presence, error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
    at /home/sergiimk/.cargo/git/checkouts/kamu-cli-aacc64f87e73d763/a07a958/src/utils/container-runtime/src/runtime.rs:587 on tokio-rt-worker
```

Expectation: there should not be a warning if SELinux is not used.

prakrititz:
## Description

Closes: #1692

`is_selinux_present()` detected SELinux by running:

```rust
std::process::Command::new("command").arg("-v").arg("sestatus").output()
```

`std::process::Command` execs directly and does **not** go through a shell, but `command`
is a POSIX shell *built-in* — there is no `/usr/bin/command` or `/bin/command` binary on
mainstream distributions. The spawn therefore fails with `NotFound` (errno 2) on every
call, which has two effects:

1. The `Err` arm fires every time, logging `Error detecting SELinux presence` on every
   container start — the warning reported in the issue.
2. The `Ok` arm is unreachable, so `is_selinux_present()` has always returned `false`,
   **including on hosts that really do run SELinux**.

This replaces the subprocess with a filesystem probe: SELinux is enabled only when the
`selinuxfs` pseudo-filesystem is mounted at `/sys/fs/selinux`, so we check for one of its
attribute files (`/sys/fs/selinux/enforce`). The `warn!` is removed because the failure
class it reported can no longer occur — it is not being silenced. `debug!("SELinux
present: {is_present}")` is kept.

The probe checks `enforce` rather than the directory itself deliberately: `/sys/fs/selinux`
exists as an empty mount-point directory on many distributions even when SELinux is off
(verified — it is present and empty on a stock WSL2 Ubuntu image), so a bare-directory
probe would be a false positive. The attribute files only appear once `selinuxfs` is
actually mounted.

### Note for reviewers — this is a behaviour change on real SELinux hosts

The only consumer of the flag is the volume-mount suffix selection in
`ContainerRuntime::docker_run_cmd` (`:ro,Z` / `:Z` / `:ro` / none). Because detection has
always returned `false`, kamu has never emitted the `:z`/`:Z` relabel flags on any host.
With detection fixed, SELinux hosts will now get them. That is the intended correction —
it is why the flag exists — but it is the riskiest part of this change and I have no
SELinux host to validate it on, so it deserves a second pair of eyes.

Alternative considered and rejected: `sh -c "command -v sestatus"` (or spawning `sestatus`
directly and matching `ErrorKind::NotFound`). It works, but it keeps a subprocess on the
path and it detects *the SELinux userspace tooling being installed*, not *SELinux being
enabled* — a different question from the one the caller asks. The filesystem probe is also
unit-testable.

### Changelog

```
## [Unreleased]
### Fixed
- Container runtime: SELinux presence detection no longer spawns the `command` shell built-in, which is not an
  executable and thus always failed with `NotFound` - as a result SELinux was never detected and a
  `Error detecting SELinux presence` warning was logged on every container start (#1692)
```

- [x] Compatibility:
  - [x] Network APIs: ✅
  - [x] Configuration: ✅
  - [x] Container images: ✅
    
- [x] Observability:
  - [x] Tracing / metrics: ✅
    
  - [x] Alerts: ✅
- [x] Documentation:
  - [x] Changelog: ✅
  - [x] Public documentation: ✅
- [x] Downstream effects:
  - [x] kamu-node: ✅
  - [x] kamu-web-ui: ✅
  - [x] release-queue: ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ Recommendation: for ease of reading, use the following format:
 ### Fixed
 -->
 
+## [Unreleased]
+### Fixed
+- Container runtime: SELinux presence detection no longer spawns the `command` shell built-in, which is not an
+  executable and thus always failed with `NotFound` - as a result SELinux was never detected and a
+  `Error detecting SELinux presence` warning was logged on every container start (#1692)
+
 ## [0.266.1] - 2026-08-19
 ### Fixed
 - Login, OAuth: login error if system (local) username matches GitHub username (#1684)
--- a/src/utils/container-runtime/src/runtime.rs
+++ b/src/utils/container-runtime/src/runtime.rs
@@ -557,6 +557,10 @@ fn is_selinux_present() -> bool {
     false
 }
 
+/// Mount point of the `SELinux` pseudo-filesystem (`selinuxfs`)
+#[cfg(target_os = "linux")]
+const SELINUX_FS_PATH: &str = "/sys/fs/selinux";
+
 #[cfg(target_os = "linux")]
 fn is_selinux_present() -> bool {
     use once_cell::sync::OnceCell;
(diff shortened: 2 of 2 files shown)


sergiimk (approved):
lgtm 👍 

thanks for the contribution

will merge once CI tasks succeed

sergiimk (approved):

PR #1696, opened 2026-08-30
Don't fail `kamu complete` when the output pipe is closed early

hdimer:
## Description

Follow-up to #1689, as requested there. Same treatment for `kamu complete`.

Related: #1068

`complete` is the command the bash script shipped in `completions` invokes on every Tab press:

```
_COMP_OUTPUTSTR="$( kamu complete -- "${COMP_WORDS[*]}" ${COMP_CWORD} )"
```

Against a closed pipe on `master` it does not panic like `completions` did - it does something arguably worse, and asks the user to file a bug for an ordinary Unix condition:

```
Error:
  0: Internal error
  1: Broken pipe (os error 32)
Help us by reporting this problem at https://github.com/kamu-data/kamu-cli/issues
```

### The fix

`run()` wraps `std::io::stdout()` in `pipecheck::wrap()`, exactly as `completions_command.rs` now does. No dependency change: #1689 already added `pipecheck` to this crate.

`complete()` flushes explicitly before returning, for the reason given in #1689 - `Stdout` is line buffered and its exit-time flush happens outside the wrapper. It has two return points, so there are two flushes.

### The `unwrap()` conversion, which is the part worth arguing about

The completion helpers wrote through `writeln!(...).unwrap()` in 11 places, while `complete()` itself already used `writeln!(...).int_err()?` in three. Two error policies for the same operation in one file, which is also why the observed failure differed depending on which completion path ran - a panic from the helpers, an "Internal error" from `complete()`.

`pipecheck` alone fixes `BrokenPipe` on both, since the process dies inside the write. It leaves every *other* write error still panicking in the helpers. So this PR also converts those 11 sites to `.int_err()?`, making the file's policy uniform and matching what `complete()` already did.

That is more than "the same as #1689", so: **happy to drop it if you would rather this PR were just the `pipecheck` line.** It is a separate commit for exactly that reason. The conversion is mechanical, changes no output, and is covered by a test that panics without it.

Deliberately left alone: the non-write `unwrap()`s in the same helpers (`datasets.try_next()`, `glob::glob()`, `fs::read_dir()`, `entry`). Those are about missing data rather than output, and converting them is a wider behavioural change than this PR should carry.

### Verification

Local, macOS:

- All three completion shapes - subcommand, positional-via-helper, and option value, which returns early - now terminate by `SIGPIPE` with empty stderr. On `master` all three give the "Internal error / report this problem" text above.
- Output is byte-identical to `master` across five completion paths: subcommand, config key, option list, top-level command list, and path completion.
- Both new failure tests were confirmed red by reverting the source: without the conversion `test_write_errors_are_propagated` panics at the raw `unwrap()` (`complete_command.rs:104`), and without `pipecheck` the pipe test fails on "Internal error". `test_completes_config_keys` passes either way, which is the point - it pins the output the refactor must not change.
- 36 tests pass in the package, `cargo fmt --check` and `make clippy` clean.

- [x] Compatibility:
  - [x] Network APIs: ✅
  - [x] Configuration: ✅
  - [x] Container images: ✅
- [x] Observability:
  - [x] Tracing / metrics: ✅ (n/a - local, non-server command)
  - [x] Alerts: ✅ (n/a)
- [x] Documentation:
  - [x] [Public documentation](https://github.com/kamu-data/kamu-docs/): ✅ (n/a - no user-facing behaviour change beyond not failing)
- [x] Downstream effects:

---

Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Recommendation: for ease of reading, use the following format:
 ### Fixed
 - SELinux presence detection if fixed and no longer spams warnings (#1692)
 - CLI: `kamu completions` no longer panics when the output pipe is closed early, e.g. `source <(kamu completions bash)` (#1068)
-- CLI: `kamu complete` no longer reports an internal error when the output pipe is closed early (#1068)
+- CLI: `kamu complete` no longer reports an internal error when the output pipe is closed early, and propagates other output errors instead of panicking (#1068)
 
 ## [0.266.1] - 2026-08-19
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Recommendation: for ease of reading, use the following format:
 ### Fixed
 - SELinux presence detection if fixed and no longer spams warnings (#1692)
 - CLI: `kamu completions` no longer panics when the output pipe is closed early, e.g. `source <(kamu completions bash)` (#1068)
+- CLI: `kamu complete` no longer reports an internal error when the output pipe is closed early (#1068)
 
 ## [0.266.1] - 2026-08-19
 ### Fixed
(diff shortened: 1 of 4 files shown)


sergiimk (approved):
lgtm, thanks 👍

